PR #14, opened 2015-05-26
Move examples from the cloud-storage-docs-xml-api-examples repo, and update them to use Application Default Credentials.

jerjou:


waprin:
LGTM, but one small nit, for the cmd-line sample README can you add a link back to the doc page that includes from it?

cloud-storage/xml-api/serviceaccount-appengine-sample/src/main/java/com/google/api/client/sample/storage/appengine/serviceaccount/StorageSample.java
```diff
@@ -0,0 +1,90 @@
+//[START all]
```
tmatsuo: If embedded in the doc, will the license header be in the snippet?

cloud-storage/xml-api/cmdline-sample/key.json
```diff
@@ -0,0 +1,3 @@
+Please download key.json file from the Google APIs Console at https://console.developers.google.com.
```
tmatsuo: Why do you want to have this file? Can you just add it to the README?

cloud-storage/xml-api/serviceaccount-appengine-sample/src/main/java/com/google/api/client/sample/storage/appengine/serviceaccount/StorageSample.java
```diff
@@ -0,0 +1,90 @@
...
+  private static final HttpTransport HTTP_TRANSPORT = new NetHttpTransport();
+
+  /** Global instance of HTML reference to XSL style sheet. */
+  String XSL = "\n<?xml-stylesheet href=\"/xsl/listing.xsl\" type=\"text/xsl\"?>\n";
+
+  @Override
+  protected void doGet(HttpServletRequest req, HttpServletResponse resp) throws IOException {
+
+    try {
+      AppIdentityCredential credential = new AppIdentityCredential(Arrays.asList(STORAGE_SCOPE));
```
tmatsuo: I think you can also use GoogleCredential.getApplicationDefault() here. Is there a reason not using it?

cloud-storage/xml-api/cmdline-sample/src/main/java/StorageSample.java
```diff
@@ -0,0 +1,99 @@
+//[START all]
```
tmatsuo: I think we should have a coding standard and the current consensus is to use the sun's one, according to:
https://github.com/GoogleCloudPlatform/Template/wiki/style.html

What do you think?

waprin:
@tmatsuo good point I merged master so we could get the root pom in here and added the checkstyle plugin which docs are telling me defaults to sun. Need to fix some issues before it will pass on travis now though.

cloud-storage/xml-api/serviceaccount-appengine-sample/src/main/java/com/google/api/client/sample/storage/appengine/serviceaccount/StorageSample.java
jerjou: I guess so. Is the 'all' tag used for anything? Perhaps it'd be prudent to just remove the tag?

cloud-storage/xml-api/serviceaccount-appengine-sample/src/main/java/com/google/api/client/sample/storage/appengine/serviceaccount/StorageSample.java
jerjou: Yeah - I tried doing it, and it wasn't finding the method for some reason. I looked into it a bit, but didn't make much headway, so I put it off so I could at least get the cmdline sample updated.

cloud-storage/xml-api/serviceaccount-appengine-sample/src/main/java/com/google/api/client/sample/storage/appengine/serviceaccount/StorageSample.java
tmatsuo: Yeah, if nobody uses it, maybe you can just remove it.

cloud-storage/xml-api/serviceaccount-appengine-sample/src/main/java/com/google/api/client/sample/storage/appengine/serviceaccount/StorageSample.java
tmatsuo: You need a newer version of com.google.api-client. Maybe you can try [1.20.0](http://search.maven.org/#artifactdetails%7Ccom.google.api-client%7Cgoogle-api-client%7C1.20.0%7Cjar).

jerjou:
Fixed style issues in cloud-storage.

taskqueue/deferred/src/main/java/com/google/cloud/taskqueue/samples/DeferSampleServlet.java
```diff
@@ -42,15 +53,20 @@ public void run() {
...
+    /**
+     * Basic demonstration of adding a deferred task.
+     * @param request servlet request
+     * @param resp servlet response
+     */
     @Override
-    public void doGet(HttpServletRequest request, HttpServletResponse
-            resp) throws IOException {
+    public void doGet(final HttpServletRequest request, final
+    HttpServletResponse resp) throws IOException {
```
tmatsuo: How about to fold just before the second `final`?

tmatsuo:
Now I have just a nit
LGTM

To tell the truth, usable auth with Java dev app server is currently broken, so let's do that with another CL, after it is fixed.

tmatsuo:
You may want to squash before merging the PR. I prefer squashing, but it's up to you @waprin

waprin:
Hmm I don't care but I lean against a merge since a lot of those were self-contained commits and if we squash it will be one huge gigantic commit. I am not against squashing but if it's up to me I prefer not to.

tmatsuo:
@waprin Alright, don't squash then. Thanks for volunteering to fix the style!

jerjou:
Added a commit to make BigqueryUtils not final, because AsyncQuerySample was extending it. Okay to merge if travis passes?

waprin:
@tmatsuo  suggested using getApplicationDefaults in appengine/serviceaccount/StorageSample.java I still think that needs to be done? Or did I miss convo that it does'nt.

jerjou:
Six comments up - apparently usable auth doesn't work with the java dev server yet, so we'll defer it to another PR.

tmatsuo:
Yup I am happy

PR #18, opened 2015-06-25
Add Storage Transfer Service samples for java

shun-fan:
@lesv @tmatsuo

### DIFF
--- a/cloud-storage/storage-transfer/README.md
+++ b/cloud-storage/storage-transfer/README.md
@@ -1,0 +1,50 @@
+# Transfer Service sample using Java
+
+This app creates two types of transfers using the Transfer Service tool.
+
+## Prerequisites
+
+1. Set up a project on Cloud Console.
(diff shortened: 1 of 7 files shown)


tmatsuo:
Can you add tests and checkstyle configuration with google style?

The parent pom has a checkstyle configuration with oracle style. So let's not add this sub module to the parent pom for now. Just run the test and checkstyle locally and then merge after the review.

BTW the travis build is failing. I don't think it's related to your change.
{
  "code" : 404,
  "errors" : [ {
    "domain" : "global",
    "message" : "Not found: Table cloud-samples-tests:test_dataset_java.test_table_java",
    "reason" : "notFound"
  } ],
  "message" : "Not found: Table cloud-samples-tests:test_dataset_java.test_table_java"
}

jerjou:
That might be me - I'll look into it.

jerjou:
'k. bigquery test should be fixed. Apparently tables have an expiration time by default, so the test table I'd created expired and got reclaimed.

PR #19, opened 2015-06-25
Fixed indent in pom.xml

shun-fan:


jerjou:
LGTM
Do you feel like removing all the trailing whitespace in all the files, too? I missed that when I moved this all over ;)

shun-fan:
I'll take a look after I get the storage transfer sample done!

PR #20, opened 2015-07-15
Move java cloud monitoring sample into github.

jerjou:
Also - add tests.


monitoring/pom.xml
```diff
@@ -0,0 +1,76 @@
...
+  <dependencies>
+    <dependency>
+        <groupId>com.google.apis</groupId>
+        <artifactId>google-api-services-cloudmonitoring</artifactId>
+        <version>v2beta2-rev20-1.20.0</version>
+    </dependency>
+    <dependency>
+        <groupId>com.google.oauth-client</groupId>
+        <artifactId>google-oauth-client</artifactId>
+        <version>${project.oauth.version}</version>
```
waprin: I think it's better if we move this into the dependency management of the root POM to make sure we stay consistent on versions.

waprin:
Mostly LGTM, one comment.

jerjou:
Sounds good - moved the properties into the parent pom. Merging...

PR #24, opened 2015-07-23
Update to production storagetransfer service account

shun-fan:


lesv:
LGTM for this, but you should fix travis

shun-fan:
jerjou is looking at the issue. It's unrelated so I'm merging.

PR #29, opened 2015-08-22
Logging

sharbison3:
From: Sam Harbison (aka harbison@google.com)

Added some sample Cloud Logging code. I am not a Maven or Git expert, yet, so feel free to correct me if you see problems in those areas.


elibixby:
@jerjou can you PTAL when you get a sec.

jerjou:
Looks good!
I don't think the settings files are really necessary - any particular reason you included them?

Also - can you add an integration test that exercises the main function, and verifies that a valid response is returned? I've added some logs to the 'cloud-samples-tests' project that we run our tests against, so you should be able to get a response once this gets merged. On your local environment, you can verify the test passes on your own project, but for CI testing, point it at cloud-samples-tests.

sharbison3:
jerjou: Thanks, I'll get to work on this. I want to get something submitted to prove I can do it.
Settings: I'm trying to discover a reasonable way to configure and use Eclipse, git, and GitHub on my workstation--I was keeping the settings for Eclipse but for no particular reason. If you have any docs advising people how to work with GitHub Java samples, I'd be glad to see it. (What's 'maven' for "run this program" for example?)

sharbison3:
PTAL, thanks.

jerjou:
(yeah - that's the question, huh :) I've been using: "mvn exec:java -Dexec.mainClass=Classname -Dexec.args='arg1 arg2 arg3''")

jerjou:
Docs: we're still figuring that out, so no docs yet. :-/

My preference is to keep what's in the repo to a minimum, to avoid confusion, and it looks like maven can successfully compile and run without the settings files.  Can you remove them from the PR?

sharbison3:
Removed settings. PTAL.

jerjou:
Could you also remove .project, .gitignore, and .classpath? If you think the .gitignore is useful, you can move that into the .gitignore at the root of the repo.

sharbison3:
PTAL. I'll leave adding the Eclipse files to .gitignore for another time.

jerjou:
LGTM
Thanks!

PR #41, opened 2015-11-25
Update to conform to appengine java runtime

shun-fan:


shun-fan:
ludo@google.com

ludoch:
LGTM Thanks!

PR #43, opened 2015-12-08
Add an App Identity example.

tswast:
This moves the first example from  https://cloud.google.com/appengine/docs/java/appidentity/ to GitHub.

I also add a simple unit test to at least make sure this code builds &
runs.  Since this sample just wraps:

    ApiProxy.getCurrentEnvironment()
        .getAttributes()
        .get("com.google.appengine.runtime.default_version_hostname"))

there isn't really much more we can test in this example.

I will move the other examples from that page over in later PRs.

### DIFF
--- a/appengine/appidentity/README.md
+++ b/appengine/appidentity/README.md
@@ -1,0 +1,12 @@
+# App Identity sample for Google App Engine
+This sample demonstrates how to use the App Identity APIs on Google App Engine
+
+## Setup
+1. Update the <application> tag in src/main/webapp/WEB-INF/appengine-web.xml with your project name
+1. Update the <version> tag in src/main/webapp/WEB-INF/appengine-web.xml with your version name
+
+## Running locally
+    $ mvn appengine:devserver
+
+## Deploying
+    $ mvn appengine:update
--- a/appengine/appidentity/README.md
+++ b/appengine/appidentity/README.md
@@ -2,26 +2,29 @@
 This sample demonstrates how to use the App Identity APIs on Google App Engine
 
 ## Running locally
-    $ mvn appengine:devserver
+This example uses the
+[Maven gcloud plugin](https://cloud.google.com/appengine/docs/java/managed-vms/maven).
+To run this sample locally:
+
+    $ mvn gcloud:run
 
 ## Deploying
 In the following command, replace YOUR-PROJECT-ID with your
 [Google Cloud Project ID](https://developers.google.com/console/help/new/#projectnumber)
 and YOUR-VERSION with a suitable version identifier.
 
-    $ mvn appengine:update -Dappengine.appId=YOUR-PROJECT-ID -Dappengine.version=YOUR-VERSION
+    $ mvn gcloud:deploy -Dversion=YOUR-VERSION -Dgcloud_project=YOUR-PROJECT-ID
 
 ## Setup
 To save your project settings so that you don't need to enter the
-`-Dappengine.appId=YOUR-CLOUD-PROJECT-ID` or
-`-Dappengine.version=YOUR-VERSION-NAME` parameters, you can make the following
-changes:
+`-Dgcloud_project=YOUR-CLOUD-PROJECT-ID` or `-Dversion=YOUR-VERSION-NAME`
+parameters, you can make the following changes:
 
 1. Update the <application> tag in src/main/webapp/WEB-INF/appengine-web.xml with your project name
 1. Update the <version> tag in src/main/webapp/WEB-INF/appengine-web.xml with your version name
 
 You will now be able to run
 
-    $ mvn appengine:update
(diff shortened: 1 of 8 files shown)


tswast:
@shun-fan @lesv Please take a look.

appengine/appidentity/README.md
```diff
@@ -0,0 +1,12 @@
+# App Identity sample for Google App Engine
+This sample demonstrates how to use the App Identity APIs on Google App Engine
+
+## Setup
+1. Update the <application> tag in src/main/webapp/WEB-INF/appengine-web.xml with your project name
```
lesv: Is this necessary?

appengine/appidentity/README.md
```diff
@@ -0,0 +1,12 @@
...
+1. Update the <version> tag in src/main/webapp/WEB-INF/appengine-web.xml with your version name
```
lesv: Is this necessary?

lesv:
Tim - is ths for traditional GAE or MVM's?

tswast:
This is for traditional GAE. I used the ../helloworld directory as a template, so the README is mostly the same.

I think the appengine-web.xml changes are necessary since we don't have the gcloud config for deploying traditional GAE apps with Maven.

lesv:
You can then define them as var's in the POM, that are substituted when the appengine-web.xml is copied.  Then the user can do mvn appengine:deploy -Dproject=myproj-333 -Dversion=32

appengine/appidentity/README.md
shun-fan: I believe the appengine maven plugin throws an error if you don't have both application and version tags.

shun-fan:
I also use tags in the pom to define project and version. Is there a reason to prefer defining them at the command line?

tswast:
https://cloud.google.com/appengine/docs/java/ has folks modifying appengine-web.xml. Should we change the Hello World instructions to point to the command-line options?

Does it makes sense to have different instructions here for the samples beyond hello world? (I do find command-line options useful when I have a development, qa, and production project)

shun-fan:
Apparently you can override the version, but not the project id from the command line, if both are present in the appengine-web.xml. Perhaps we can specify both.

lesv:
Yes please.

On Tue, Dec 8, 2015 at 4:56 PM, Shun Fan <notifications@github.com> wrote:

> Apparently you can override the version, but not the project id from the
> command line, if both are present in the appengine-web.xml. Perhaps we can
> specify both.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/GoogleCloudPlatform/java-docs-samples/pull/43#issuecomment-163073018>
> .
>



-- 


*  •  **Les Vogel*
*  •  *Cloud Developer Relations
*  •  *lesv@google.com
*  •  *+1-4 <%2B1-650-338-7103>08-676-7023

tswast:
I added instructions for how to provide the application ID and version number via the command line. They differed slightly from what @lesv describes in https://github.com/GoogleCloudPlatform/java-docs-samples/pull/43#issuecomment-162723721 but I was able to deploy using these parameters. They match what parameters I see when I run:

    mvn help:describe -Dcmd=appengine:update -Ddetail

lesv:
Other than the comments to be on MVM's (email from PM) it's looking good.

tswast:
@lesv Makes sense. I've updated the sample to use Managed VMs.

appengine/appidentity/README.md
```diff
@@ -0,0 +1,30 @@
...
+To run this sample locally:
+
+    $ mvn gcloud:run
+
+## Deploying
+In the following command, replace YOUR-PROJECT-ID with your
+[Google Cloud Project ID](https://developers.google.com/console/help/new/#projectnumber)
+and YOUR-VERSION with a suitable version identifier.
+
+    $ mvn gcloud:deploy -Dversion=YOUR-VERSION -Dgcloud_project=YOUR-PROJECT-ID
```
lesv: If you don't provide a version gcloud will automatically create one for you.

lesv:
I don't think you need -Dversion, and I'm not sure how it gets used here.  I don't see it being used, and / or passed to gcloud-mvn-plugin in anyway.

On the other hand, everything else looks good, I liked learning about truth.  We should have a discussion about surefire vs junit vs. others someday.

tswast:
@lesv I removed the references to -Dversion. It shows up as a parameter when I run `mvn help:describe -Dcmd=gcloud:deploy -Ddetail` but it does seem to be ignored when I provide one. Since it's not needed, I'll just remove it.

Truth is really great! It's been pushed by the Google Java core libraries team more and more, and I agree with them that it increases the readability of tests and makes it harder to make mistakes on the order of arguments.

lesv:
LGTM

PR #48, opened 2016-01-05
Fix project parameter in App Identity README.

tswast:
The gcloud command was getting the project argument when I was passing
it in as -Dgcloud_project. After changing to -Dgcloud.gcloud_project it
now deploys to the project I specify there. I hadn't noticed this until
I had modified my gcloud config and deployments started going to the
wrong project.


tswast:
Made new PR https://github.com/GoogleCloudPlatform/java-docs-samples/pull/49 from branch instead of fork, which should fix Travis build.

PR #50, opened 2016-01-05
Add App Identity example to assert identity to Google APIs.

tswast:
This sample is copied almost exactly from
https://cloud.google.com/appengine/docs/java/appidentity/#asserting_identity_to_google_apis

I also add a servlet for the sample code so that it can be more easily
demonstrated and manually tested.

### DIFF
--- a/appengine/appidentity/pom.xml
+++ b/appengine/appidentity/pom.xml
@@ -34,11 +34,21 @@ Copyright 2015 Google Inc. All Rights Reserved.
       <version>${appengine.target.version}</version>
     </dependency>
     <dependency>
+      <groupId>com.google.guava</groupId>
+      <artifactId>guava</artifactId>
+      <version>19.0</version>
+    </dependency>
+    <dependency>
       <groupId>javax.servlet</groupId>
       <artifactId>servlet-api</artifactId>
       <version>2.5</version>
       <type>jar</type>
       <scope>provided</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.json</groupId>
+      <artifactId>json</artifactId>
+      <version>20151123</version>
     </dependency>
 
     <!-- Test Dependencies -->
--- a/appengine/appidentity/src/main/java/com/example/appengine/appidentity/UrlShortener.java
+++ b/appengine/appidentity/src/main/java/com/example/appengine/appidentity/UrlShortener.java
@@ -1,0 +1,78 @@
+/**
+ * Copyright 2016 Google Inc. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.example.appengine.appidentity;
+
+import com.google.appengine.api.appidentity.AppIdentityService;
+import com.google.appengine.api.appidentity.AppIdentityServiceFactory;
+import com.google.common.io.CharStreams;
+import org.json.JSONObject;
+import org.json.JSONTokener;
+
+import java.io.InputStream;
+import java.io.InputStreamReader;
+import java.io.OutputStreamWriter;
+import java.net.HttpURLConnection;
+import java.net.URL;
+import java.nio.charset.StandardCharsets;
+import java.util.ArrayList;
+
+@SuppressWarnings("serial")
+class UrlShortener {
+  // [START asserting_identity_to_Google_APIs]
+  public String createShortUrl(String longUrl) throws Exception {
+    try {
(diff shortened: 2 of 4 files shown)


tswast:
@shun-fan @lesv Please take a look. The file appengine/appidentity/src/main/java/com/example/appengine/appidentity/UrlShortener.java is almost exactly copied from the App Identity docs page. I made one slight change to include the error response in the exception it throws to make it easier to see why it wasn't working at first in my manual tests (it doesn't have the right scope when running locally, and it needs to have the Url Shortener API enabled when deployed to a cloud project)

Unfortunately, this function was not written to be testable, so rather than mock a ton out, I did not write any tests yet for this. I think it will make sense to write an integration test for this, maybe in a later PR once I talk to @broady about the test framework he's building.

appengine/appidentity/pom.xml
```diff
@@ -34,12 +34,22 @@ Copyright 2015 Google Inc. All Rights Reserved.
...
+      <version>19.0</version>
+    </dependency>
+    <dependency>
       <groupId>javax.servlet</groupId>
       <artifactId>servlet-api</artifactId>
       <version>2.5</version>
       <type>jar</type>
       <scope>provided</scope>
     </dependency>
+    <dependency>
```
lesv: Ok for now, but I think we should probably move our samples to GSON or something else.

appengine/appidentity/src/main/java/com/example/appengine/appidentity/UrlShortener.java
```diff
@@ -0,0 +1,78 @@
...
+      // The token asserts the identity reported by appIdentity.getServiceAccountName()
+      JSONObject request = new JSONObject();
+      request.put("longUrl", longUrl);
+
+      URL url = new URL("https://www.googleapis.com/urlshortener/v1/url?pp=1");
+      HttpURLConnection connection = (HttpURLConnection) url.openConnection();
+      connection.setDoOutput(true);
+      connection.setRequestMethod("POST");
+      connection.addRequestProperty("Content-Type", "application/json");
+      connection.addRequestProperty("Authorization", "Bearer  " + accessToken.getAccessToken());
```
lesv: Is there a reason you have two spaces after "Bearer"?

appengine/appidentity/src/main/java/com/example/appengine/appidentity/UrlShortener.java
```diff
@@ -0,0 +1,78 @@
...
+            InputStreamReader r = new InputStreamReader(s, StandardCharsets.UTF_8)) {
+          throw new RuntimeException(String.format(
+              "got error (%d) response %s from %s",
+              connection.getResponseCode(),
+              CharStreams.toString(r),
+              connection.toString()));
+        }
+      }
+    } catch (Exception e) {
+      // Error handling elided.
```
lesv: If it doesn't exist, shouldn't we just mention it in a comment rather than putting in a try / catch block?

lesv:
PTAL

appengine/appidentity/src/main/java/com/example/appengine/appidentity/UrlShortener.java
tswast: No reason. This was copied from https://cloud.google.com/appengine/docs/java/appidentity/#asserting_identity_to_google_apis and it was there. I removed the extra space and it still appears to work fine.

appengine/appidentity/src/main/java/com/example/appengine/appidentity/UrlShortener.java
tswast: Done. I moved this comment to a new JavaDoc for this function, partially to explain the not-so-nice "throws Exception"

tswast:
@lesv Thanks for the review. Could you take a look again?

appengine/appidentity/src/main/java/com/example/appengine/appidentity/UrlShortenerServlet.java
```diff
@@ -0,0 +1,73 @@
...
+    shortener = new UrlShortener();
+  }
+
+  @Override
+  public void doGet(HttpServletRequest req, HttpServletResponse resp) throws IOException {
+    PrintWriter w = resp.getWriter();
+    w.println("<!DOCTYPE html>");
+    w.println("<form method='post'>");
+    w.println("<label for='longUrl'>URL:</label>");
+    w.println("<input id='longUrl' name='longUrl' type='text'>");
```
lesv: It's been a while since I did this, but shouldn't it be `\>`?

appengine/appidentity/src/main/java/com/example/appengine/appidentity/UrlShortenerServlet.java
```diff
@@ -0,0 +1,73 @@
...
+    w.println("<input type='submit' value='Shorten'>");
```
lesv: ditto

lesv:
One nit, I'm OK if you disagree, then LGTM

appengine/appidentity/src/main/java/com/example/appengine/appidentity/UrlShortenerServlet.java
tswast: That's for xhtml, not for HTML5. Google Style guide recommends the HTML5 style https://google.github.io/styleguide/htmlcssguide.xml?showone=Document_Type#Document_Type

PR #51, opened 2016-01-06
Initial import of MVM/SparkJava/gcloud-java example

ajkannan:
This example shows how to use [SparkJava](http://sparkjava.com/) from a Managed VM.  The sample is a webpage that interfaces with Cloud Datastore using [gcloud-java](https://github.com/GoogleCloudPlatform/gcloud-java).


ajkannan:
Travis build using JDK7 is failing because this sample uses Java 8 lambdas.  Switching this example to use Java 7 will require downgrading the SparkJava version from version 2 to 1.  Do you think modifying the sample to use SparkJava v1 instead v2 is the way to go @ludoch @lesv?

ludoch:
No, the point of this sample is to use Java8 and the latest sparkjava.

ajkannan:
Do we know if/when we can up the Java version to Java 8 for Travis CI?  If that will take a while, then I see two alternatives (provided that we keep using the latest sparkjava):
1) Not run this app's tests via Travis CI now.  We can keep the test code around and remove sparkjava from the list of modules, so that we have access to the code for manual testing.  We can also create an issue to integrate this app with Travis.
2) Wait on placing this app in this repository. 

I'm in favor option 1, unless we can update the Java version in CI to 8 soon.  What are other people's opinions?

ludoch:
https://blog.travis-ci.com/2013-11-26-test-your-java-libraries-on-java-8/

ajkannan:
Added a commit that uses jdk8 to run CI.

ludoch:
LGTM

PR #54, opened 2016-01-13
Add Google AppEngine Managed-vms samples

shun-fan:


appengine/static-files/src/main/webapp/stylesheets/styles.css
```diff
@@ -0,0 +1,4 @@
+body {
+  font-family: Verdana, Helvetica, sans-serif;
+  background-color: #CCCCFF;
+}
```
lesv: NewLine

lesv:
LGTM

lesv:
Will merge when Travis finishes.

PR #59, opened 2016-01-22
Fix style errors in the App Identity samples.

tswast:
I checked for style errors using the following command:

    mvn checkstyle:checkstyle \
        -Dcheckstyle.config.location=google_checks.xml \
        -Dcheckstyle.consoleOutput=true


lesv:
LGTM - presuming that travis hasn't been setup correctly (it's red).  
I'm not totally convinced that writer is better than w, but that's the style.  We should install the checkstyle maven plugin.

tswast:
Yeah, I'm not convinced either that w is better, either. I'll follow up in a later PR with adding the plugin.

Looks like I did break the build (missed a rename for writer). I'll fix that and make sure travis passes before merging.

PR #60, opened 2016-01-23
Checkstyle is now enforced across most modules.

tswast:
Also, many modules were not being built by Travis, since they were not included in the root pom.xml. By changing the Travis config to use `mvn verify` instead of `mvn test`, we ensure that the checkstyle plugin will run, since [by default](https://maven.apache.org/plugins/maven-checkstyle-plugin/check-mojo.html) it is bound to the verify [phase](https://maven.apache.org/guides/introduction/introduction-to-the-lifecycle.html#Lifecycle_Reference).


lesv:
LGTM -- We may wish to tweak the checkstyle rules for DevRel and Java8

PR #62, opened 2016-01-25
Reorganize mvm samples to mvm folder

shun-fan:


lesv:
Rename appears OK, but Travis is very unhappy.  You might wish to fix the tests.

lesv:
LGTM

PR #64, opened 2016-01-28
Extra servlet in Helloworld

lesv:
1. remove it.


lesv:
@shun-fan @tswast Please review PR -- even if I merge.

tswast:
LGTM.

For the removed mapping. Does the `@WebServlet` annotation in https://github.com/GoogleCloudPlatform/java-docs-samples/blob/master/managed_vms/helloworld/src/main/java/com/example/managedvms/helloworld/HelloServlet.java#L29 set up the mapping now that we don't have any mappings in web.xml? It seems there are many ways of doing servlet mappings. (I know of this, Guice, and web.xml now)

I find the security rules a bit confusing. I guess HelloServlet will respond to /profile/* and /admin/* requests, and this is just to demonstrate how to use those security settings?

Didn't we want to force HTTPS for all paths, also, with <transport-guarantee>CONFIDENTIAL</transport-guarantee> https://cloud.google.com/appengine/docs/java/config/webxml#Secure_URLs

lesv:
Sigh - I grabbed the wrong code in haste.  Thanks.

PR #66, opened 2016-01-28
Fix helloWorld

lesv:
1. make sure it really runs on MVM
2. Fix read Me to use jetty:run
3. get rid of appengine-web.xml use app.yam
4. improved security constraint.

I need this to stage some docs, likely to get replaced by Shun later
today.


tswast:
LGTM. Test failures appear to be with Spark sample, not hello world.

PR #69, opened 2016-01-28
Update to java8

shun-fan:
Update to using app.yaml
Update to gcloud-java 0.1.3
Remove websockets for now


managed_vms/analytics/src/main/appengine/app.yaml
```diff
@@ -0,0 +1,11 @@
+runtime: java
+vm: true
+runtime_config:
```
lesv: We don't appear to be documenting this.

lesv:
LGTM

PR #70, opened 2016-01-28
Remove web.xml

lesv:
https is important, but we can’t get it working locally — I thought it
worked earlier.  So we are striking it from HelloWorld.


tswast:
LGTM.

PR #71, opened 2016-01-29
Update pom.xml

halgrimur:
Add region tag.


lesv:
LGTM

lesv:
@shun-fan @tmatsuo @tswast Note - something's wrong w/ travis

PR #73, opened 2016-01-29
Add comments to cloudstorage

shun-fan:

### DIFF
--- a/managed_vms/cloudsql/src/main/java/com/example/managedvms/cloudsql/CloudSqlServlet.java
+++ b/managed_vms/cloudsql/src/main/java/com/example/managedvms/cloudsql/CloudSqlServlet.java
@@ -16,6 +16,8 @@
 
 package com.example.managedvms.cloudsql;
 
+import org.jasypt.util.text.BasicTextEncryptor;
+
 import java.io.IOException;
 import java.io.PrintWriter;
 import java.sql.Connection;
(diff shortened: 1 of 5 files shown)


managed_vms/cloudsql/src/main/java/com/example/managedvms/cloudsql/CloudSqlServlet.java
```diff
@@ -40,12 +42,18 @@
   @Override
   public void doGet(HttpServletRequest req, HttpServletResponse resp) throws IOException,
       ServletException {
+    // encrypt user ip using PBEWithMD5AndDES
+    BasicTextEncryptor encryptor = new BasicTextEncryptor();
+    encryptor.setPassword(Double.toString(10000*Math.random()));
```
tswast: Maybe add a disclaimer comment that Math.random is not a secure way to generate encryption keys / passwords.

managed_vms/datastore/src/main/java/com/example/managedvms/datastore/DatastoreServlet.java
```diff
@@ -44,17 +46,25 @@
   @Override
   public void doGet(HttpServletRequest req, HttpServletResponse resp) throws IOException,
       ServletException {
+    // encrypt user ip using PBEWithMD5AndDES
+    BasicTextEncryptor encryptor = new BasicTextEncryptor();
+    encryptor.setPassword(Double.toString(10000*Math.random()));
```
tswast: Ditto. Not very secure to use Math.random. We should either use a secure random number generator or at least add a "don't do this" disclaimer. http://stackoverflow.com/a/11052736/101923

PR #74, opened 2016-01-29
Add comments to cloudstorage

shun-fan:

### DIFF
--- a/managed_vms/cloudsql/src/main/java/com/example/managedvms/cloudsql/CloudSqlServlet.java
+++ b/managed_vms/cloudsql/src/main/java/com/example/managedvms/cloudsql/CloudSqlServlet.java
@@ -16,6 +16,8 @@
 
 package com.example.managedvms.cloudsql;
 
+import org.jasypt.util.text.BasicTextEncryptor;
+
 import java.io.IOException;
 import java.io.PrintWriter;
 import java.sql.Connection;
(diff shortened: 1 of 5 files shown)


managed_vms/cloudsql/src/main/java/com/example/managedvms/cloudsql/CloudSqlServlet.java
```diff
@@ -40,27 +42,33 @@
   @Override
   public void doGet(HttpServletRequest req, HttpServletResponse resp) throws IOException,
       ServletException {
+    // encrypt user ip using PBEWithMD5AndDES
+    BasicTextEncryptor encryptor = new BasicTextEncryptor();
+    encryptor.setPassword(Double.toString(10000*Math.random()));
```
tswast: We should use SecureRandom or at least add a "don't do this" disclaimer. http://stackoverflow.com/a/11052736/101923

Also, 10000 possible passwords isn't all that many. :-(

managed_vms/datastore/src/main/java/com/example/managedvms/datastore/DatastoreServlet.java
```diff
@@ -44,17 +46,25 @@
   @Override
   public void doGet(HttpServletRequest req, HttpServletResponse resp) throws IOException,
       ServletException {
+    // encrypt user ip using PBEWithMD5AndDES
+    BasicTextEncryptor encryptor = new BasicTextEncryptor();
+    encryptor.setPassword(Double.toString(10000*Math.random()));
```
tswast: I'm confused. If we're getting a new password every time, we're basically just storing a random number, not an IP address that we could recover.

PR #75, opened 2016-01-29
Add comments to cloudstorage

shun-fan:


lesv:
LGTM - I was suggesting first two octets (first 16 bits of binary - but this is good)

PR #76, opened 2016-02-01
Https

shun-fan:
Add https handlers to remaining samples


lesv:
LGTM

lesv:
Please look at why travis failed.

shun-fan:
It's an issue with sparkjava demo from Ajay on Ludo's team.

ajkannan:
It looks like a flaky test because of a concurrency issue.  Since Travis does two builds for each PR when merging one upstream branch into another, there could be interleavings in interactions with Datastore that result in tests failing.  I will create a PR to fix this.

PR #78, opened 2016-02-03
Adding Tags for docs

lesv:


lesv:
@shun-fan @tswast PTAL
LGTM

PR #83, opened 2016-02-05
Fix issue with truncating recorded ip addresses

shun-fan:


lesv:
LGTM - better - I realized what everyone was doing after I approved it.

PR #87, opened 2016-02-09
Fix issue with truncating recorded ip addresses in disk

shun-fan:


lesv:
LGTM

lesv:
LGTM - Travis looks like SparkDemo again.

PR #90, opened 2016-02-10
Datastore Concepts snippets and Getting Started app

ajkannan:
/cc @pcostell @aozarov

### DIFF
--- a/datastore/README.md
+++ b/datastore/README.md
@@ -1,0 +1,14 @@
+## Datastore Samples
+
+This directory contains sample code used in Google Cloud Datastore documentation. Included here is a sample command line application, `TaskList`, that interacts with Datastore to manage a to-do list.
+
+## Run the `TaskList` sample application.
+
+1. Ensure that you have:
+  * Created a Google Developers Console project with the Datastore API enabled. Follow [these instructions](https://cloud.google.com/docs/authentication#preparation) to get your project set up. 
+  * Installed the Google Cloud SDK and run the following commands in command line: `gcloud auth login` and `gcloud config set project [YOUR PROJECT ID]`.
+  * Installed [Maven](https://maven.apache.org/) and Java 7 (or above).
+
+2. Compile the program by typing `mvn clean compile` in command line.
+
+3. Run the program by typing `mvn exec:java` in command line.
--- a/datastore/README.md
+++ b/datastore/README.md
@@ -11,4 +11,4 @@ This directory contains sample code used in Google Cloud Datastore documentation
 
 2. Compile the program by typing `mvn clean compile` in command line.
 
-3. Run the program by typing `mvn exec:java` in command line.
+3. Run the program by typing `mvn exec:java` in command line. In addition to listing tasks via this command line interface, you can view tasks you create in the [Google Cloud Developer's Console](https://console.cloud.google.com/).
--- a/datastore/pom.xml
+++ b/datastore/pom.xml
@@ -1,0 +1,65 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+  <groupId>com.google.datastore.snippets</groupId>
+  <artifactId>datastore-snippets</artifactId>
+  <version>1.0</version>
+  <packaging>jar</packaging>
+  <name>Google Cloud Datastore Snippets</name>
+  <description>
+    Example snippets for Datastore concepts and getting started documentation.
+  </description>
+  <dependencies>
+    <dependency>
+      <groupId>com.google.gcloud</groupId>
+      <artifactId>gcloud-java-datastore</artifactId>
+      <version>0.1.3</version>
+    </dependency>
+    <dependency>
+      <groupId>junit</groupId>
+      <artifactId>junit</artifactId>
+      <version>4.12</version>
+    </dependency>
+  </dependencies>
+  <build>
+    <plugins>
+<!-- // [START maven]-->
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-compiler-plugin</artifactId>
+        <version>2.5.1</version>
+        <configuration>
+          <source>1.7</source>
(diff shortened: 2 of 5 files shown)


datastore/src/main/java/com/google/datastore/snippets/Concepts.java
```diff
@@ -0,0 +1,1058 @@
...
+        .set("priority", 4)
+        .set("description", "Learn Cloud Datastore")
+        .build();
+    Entity task2 = Entity.builder(taskKey2)
+        .set("type", "Personal")
+        .set("done", false)
+        .set("priority", 5)
+        .set("description", "Integrate Cloud Datastore")
+        .build();
+    // [START batch_upsert]
```
pcostell: Sorry this changed recently, but can you include creation of entities + keys in the snippet too?

datastore/src/main/java/com/google/datastore/snippets/Concepts.java
```diff
@@ -0,0 +1,1058 @@
...
+    }
+    // [END cursor_paging]
+    return nextPageCursor;
+  }
+
+  @Test
+  public void testEventualConsistentQuery() {
+    setUpQueryTests();
+    // [START eventual_consistent_query]
+    Query<Entity> query = Query.entityQueryBuilder()
```
pcostell: If you can't set the read options on the query, you should just make this a comment that says:

    // Read consistency cannot be specified in gcloud-java.

datastore/src/main/java/com/google/datastore/snippets/Concepts.java
```diff
@@ -0,0 +1,1058 @@
...
+
+  @Test
+  public void testTransactionalUpdate() {
+    List<Key> keys = setUpTransferTests();
+    transferFunds(keys.get(0), keys.get(1), 10);
+    assertSuccessfulTransfer(keys.get(0), keys.get(1));
+  }
+
+  private void transferFunds(Key fromKey, Key toKey, long amount) {
+    // [START transactional_update]
```
pcostell: I would move the snippet to include the transferFunds method definition.

pcostell:
Looks great, thanks Ajay!

ajkannan:
Incorporated Patrick's feedback.

Also /cc @lesv

datastore/src/main/java/com/google/datastore/snippets/Concepts.java
aozarov: @pcostell is it an option in v1beta2? and should we support it for v1beta3?

pcostell: Yes, it is supported in both via `RunQueryRequest.read_options`:

    runQueryRequestBuilder = RunQueryRequest.newBuilder()
    runQueryRequestBuilder.getReadOptionsBuilder().setReadConsistency(ReadConsistency.EVENTUAL);

aozarov: Ok, just to make sure the "Yes" is also for "we should support it"?
Also, I assume ReadConsistency (eventual, consistent) only applies for get/lookup operations?

datastore/src/main/java/com/google/datastore/snippets/Concepts.java
pcostell: Ah sorry --

These should be automatically allocated ids. We should show how to extract the keys from the batch afterwards.

pcostell: Since v1beta2 doesn't support upsert with auto id allocation, can we make this a batch insert instead of upsert? Sorry I didn't catch this, these snippets were designed for v1beta3.

ajkannan: Got it, updated in the latest commit.

datastore/src/main/java/com/google/datastore/snippets/TaskList.java
```diff
@@ -0,0 +1,232 @@
...
+        System.out.println("task added");
+        break;
+      case "done":
+        assertArgsLength(args, 2);
+        long id = Long.parseLong(args[1]);
+        try {
+          markDone(id);
+          System.out.println("task marked done");
+        } catch (DatastoreException e) {
+          System.out.printf("did not find a Task entity with ID %d\n", id);
```
aozarov: Nit, you can use %n instead of \n

ajkannan:
Pushed a commit to incorporate @aozarov feedback.  Ready for another round of review.

datastore/README.md
```diff
@@ -0,0 +1,14 @@
+## Datastore Samples
+
+This directory contains sample code used in Google Cloud Datastore documentation. Included here is a sample command line application, `TaskList`, that interacts with Datastore to manage a to-do list.
```
lesv: App Engine Managed VMs or Standard?

datastore/pom.xml
```diff
@@ -0,0 +1,65 @@
...
+  </dependencies>
+  <build>
+    <plugins>
+<!-- // [START maven]-->
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-compiler-plugin</artifactId>
+        <version>2.5.1</version>
+        <configuration>
+          <source>1.7</source>
```
lesv: Should be 1.8 now.  (Unless it's for Standard)

datastore/README.md
lesv: Ah MVM / any Java.

datastore/README.md
```diff
@@ -0,0 +1,14 @@
...
+
+## Run the `TaskList` sample application.
+
+1. Ensure that you have:
+  * Created a Google Developers Console project with the Datastore API enabled. Follow [these instructions](https://cloud.google.com/docs/authentication#preparation) to get your project set up. 
+  * Installed the Google Cloud SDK and run the following commands in command line: `gcloud auth login` and `gcloud config set project [YOUR PROJECT ID]`.
+  * Installed [Maven](https://maven.apache.org/) and Java 7 (or above).
```
lesv: Emulator?

lesv: Any comments about viewing tasks in cloud console?

lesv:
LGTM - your choice to update readme / pom for nits.

datastore/README.md
ajkannan: gcloud-java downloads the emulator for you if you haven't installed it, so I left it out here.  I'll add a sentence about viewing tasks in cloud console.

aozarov:
LGTM. Nice job!

lesv:
Totally agree - very nice job!

PR #91, opened 2016-02-11
Fix checkstyle in twilio and sendgrid samples.

### LINKED ISSUE #61: Several modules fail the Google Style Checks.
In https://github.com/GoogleCloudPlatform/java-docs-samples/pull/60 I found that the following modules fail the style checks:

- [x] managed_vms/async-rest
- [x] managed_vms/twilio
- [x] managed_vms/sendgrid
- [x] managed_vms/sparkjava
- [x] unittests

It looks like most of these failures were due to using ALL_CAPS for local final variables. These should use lowerCamelCase. Final local variables are considered not constants according to the style guide. https://google.github.io/styleguide/javaguide.html#s5.2.4-constant-names

tswast:
Fixes #61.

### DIFF
--- a/managed_vms/sendgrid/pom.xml
+++ b/managed_vms/sendgrid/pom.xml
@@ -48,9 +48,6 @@
           <target>1.8</target>
         </configuration>
       </plugin>
-      <!--
-           The sendgrid sample currently (2016-01-22) fails the Google Style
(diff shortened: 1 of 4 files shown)


lesv:
LGTM - again - I think we might wish to do our own checkstyle as the all caps for important things improves the teaching.

Issue #85, opened 2016-02-09
Build failing on Jenkins due to sparkdemo


tswast:
This might be the same flakey failure we were seeing in Travis.

```
-------------------------------------------------------
 T E S T S
-------------------------------------------------------
Running com.google.appengine.sparkdemo.UserControllerTest
Feb 08, 2016 1:00:25 AM com.google.api.services.datastore.client.BaseDatastoreFactory makeClient
WARNING: Not using any credentials
Feb 08, 2016 1:00:28 AM com.google.api.services.datastore.client.BaseDatastoreFactory makeClient
WARNING: Not using any credentials
[Thread-1] INFO org.eclipse.jetty.util.log - Logging initialized @3035ms
[Thread-1] INFO spark.webserver.JettySparkServer - == Spark has ignited ...
[Thread-1] INFO spark.webserver.JettySparkServer - >> Listening on 0.0.0.0:8080
[Thread-1] INFO org.eclipse.jetty.server.Server - jetty-9.3.2.v20150730
[Thread-1] INFO org.eclipse.jetty.server.ServerConnector - Started ServerConnector@42c69b0e{HTTP/1.1,[http/1.1]}{0.0.0.0:8080}
[Thread-1] INFO org.eclipse.jetty.server.Server - Started @3198ms
[main] INFO spark.webserver.JettySparkServer - >>> Spark shutting down ...
[main] INFO org.eclipse.jetty.server.ServerConnector - Stopped ServerConnector@42c69b0e{HTTP/1.1,[http/1.1]}{0.0.0.0:8080}
[main] INFO spark.webserver.JettySparkServer - done
Tests run: 5, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 5.345 sec <<< FAILURE!
Running com.google.appengine.sparkdemo.UserServiceTest
Feb 08, 2016 1:00:33 AM com.google.api.services.datastore.client.BaseDatastoreFactory makeClient
WARNING: Not using any credentials
Feb 08, 2016 1:00:34 AM com.google.api.services.datastore.client.BaseDatastoreFactory makeClient
WARNING: Not using any credentials
Tests run: 4, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 5.119 sec

Results :

Failed tests:   testDeleteUser(com.google.appengine.sparkdemo.UserControllerTest): expected:<0> but was:<1>
```

lesv:
[...]

PR #96, opened 2016-02-20
Fix consistency issue with sparkjava demo tests.

ajkannan:
Consistency issues were likely involved with #85.

/cc @aozarov mind taking a look to check that consistency bugs are now fixed?

### DIFF
--- a/managed_vms/sparkjava/src/main/java/com/google/appengine/sparkdemo/UserService.java
+++ b/managed_vms/sparkjava/src/main/java/com/google/appengine/sparkdemo/UserService.java
@@ -22,8 +22,10 @@ import com.google.gcloud.datastore.Datastore;
 import com.google.gcloud.datastore.Entity;
 import com.google.gcloud.datastore.Key;
 import com.google.gcloud.datastore.KeyFactory;
+import com.google.gcloud.datastore.PathElement;
 import com.google.gcloud.datastore.Query;
 import com.google.gcloud.datastore.QueryResults;
+import com.google.gcloud.datastore.StructuredQuery.PropertyFilter;
 
 import java.util.ArrayList;
 import java.util.List;
@@ -42,7 +44,8 @@ public class UserService {
    */
   public UserService(Datastore datastore, String kind) {
     this.datastore = datastore;
-    this.keyFactory = datastore.newKeyFactory().kind(kind);
+    this.keyFactory =
+        datastore.newKeyFactory().ancestors(PathElement.of("SparkJavaDemo", "default")).kind(kind);
     this.kind = kind;
   }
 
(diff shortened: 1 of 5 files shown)


managed_vms/sparkjava/src/main/java/com/google/appengine/sparkdemo/UserService.java
```diff
@@ -83,7 +89,9 @@ public User createUser(String name, String email) {
    */
   public String deleteUser(String id) {
     Key key = keyFactory.newKey(id);
-    datastore.delete(key);
+    if (datastore.get(key) != null) {
```
aozarov: why not just delete?

aozarov:
Change Looks good.

managed_vms/sparkjava/src/main/java/com/google/appengine/sparkdemo/UserService.java
ajkannan: Fixed.

tswast:
I agree that it does seem most likely that this flakiness is due to eventual consistency issues.

I don't think this is a good recommended practice (to make all user entities share a parent entity). This introduces a global lock on all user reads/writes. This is definitely not good for datastore. One would be better off using a SQL database than writing Datastore code like this. (We are limited to about 1 write to the whole users table per second with this change. https://cloud.google.com/datastore/docs/concepts/structuring_for_strong_consistency -- last paragraph "no more than 1 write per second (the supported limit for entity groups)")

There a couple ways around the eventual consistency.

1. In a standard runtime App Engine app, the simplest would be to allow injecting in a local Datastore service (which has options for turning off eventual consistency). See the guide for unit testing App Engine services. https://cloud.google.com/appengine/docs/java/tools/localunittesting It looks like a similar thing is possible by starting a local development server and connecting to that when using the gcloud APIs. https://cloud.google.com/datastore/docs/tools/devserver There is a `--consistency=` option, which we can set to 1.0.
2. Remove uses of `getAllUsers()` in the tests. For example: the failure in #85 for `testDeleteUser` could have been avoided completely by doing a `datastore.get` instead of a `query` to verify that the user is deleted. If we do want to use `getAllUsers` in a test, we should not be looking for a specific number of results, only verify that the command succeeded.

Option 2 is probably showing users the best example of how to use and test Datastore, and manage eventual consistency. Option 1 is okay if easy to do, but it does completely ignore eventual consistency, which may be bad for users who try to follow this sample.

waprin:
Tim's recommendations make sense for the simpler tests that you do expect to be non-flaky.But instead of replacing the tests with non-flaky ones, consider just adding the non-flaky ones and marking the flaky one as such to be only run in certain contexts and possibly to be retried a few times. We are being strongly encouraged to run good e2e tests, so some flaky tests are inevitable, so sooner or later you will probably need a broader strategy to accommodate flakiness.

ajkannan:
Since it's a relatively simple test, I pushed another commit using option 2 as Tim suggested.

tswast:
LGTM. Thank you!

PR #97, opened 2016-02-20
Set source/target to jdk 1.7 and use latest gcloud-java release features

### LINKED ISSUE #95: Can't compile datastore sample with jdk 1.7
To successfully compile from the datastore/ folder, in java-docs-samples/datastore/pom.xml, I needed to change the source and target to use 1.7 (they are set up for 1.8).

ajkannan:
Fixes #95 and also makes minor updates to use the latest gcloud-java features.

/cc @aozarov

### DIFF
--- a/datastore/src/main/java/com/google/datastore/snippets/Concepts.java
+++ b/datastore/src/main/java/com/google/datastore/snippets/Concepts.java
@@ -911,12 +911,12 @@ public class Concepts {
     setUpQueryTests();
     // [START property_run_query]
     Query<Key> query = Query.keyQueryBuilder().kind("__property__").build();
-    QueryResults<Key> results = datastore.run(query);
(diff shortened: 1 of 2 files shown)


datastore/src/main/java/com/google/datastore/snippets/Concepts.java
```diff
@@ -916,7 +915,7 @@ public void testPropertyRunQuery() {
     Map<String, Collection<String>> propertiesByKind = new HashMap<>();
     while (results.hasNext()) {
       Key property = results.next();
```
aozarov: s/results/keys/ and even more important s/property/key/ ?

ajkannan: Fixed.

shun-fan:
LGTM

PR #98, opened 2016-03-01
Remove redundant checkstyle plugin definitions.

tswast:
The submodules do not need checkstyle plugin defined if they define the
parent.

@lesv @shun-fan We should be able to simplify the getting-started-java poms in a similar way if we define the parent in the submodule poms.


lesv:
Agree -- but it would be nice if Travis was happy.  It looks like a few things are missing.

lesv:
Basically looks good, please figure out the CI and either ping me to re-review or consider this an L G T M.

tswast:
The datastore sample was failing because it uses JUnit in the main classes and the parent pom limits JUnit to the test scope. Should be fixed now, but I'll let Travis run and see if there's anything else I missed.

PR #99, opened 2016-03-01
Remove cloudsql for now

shun-fan:
Remove cloud sql as it's currently not supported for MVMs


tswast:
Remove it from the root pom.xml as well.

lesv:
LGTM once it's gone from the pom.xml, please take a look at travis as well.

tswast:
LGTM. Looks like Travis is happy now that it's gone from the root pom.

lesv:
LGTM

lesv:
Though it would be nice to see if we can make this work somehow.  I suspect we can, so please keep a copy.

lesv:
Actually, we know it works as bookshelf works.

PR #100, opened 2016-03-03
Fix error deploying hello world.

tswast:
The `appengine.version` overrides the version tag for the instance
rather than the SDK version. Changing to `appengine.sdk.version` fixes
the errors we encounter when deploying.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -9,7 +9,7 @@
     <packaging>pom</packaging>
 
     <properties>
-        <appengine.version>1.9.19</appengine.version>
+        <appengine.sdk.version>1.9.19</appengine.sdk.version>
         <appengine.app.version>1</appengine.app.version>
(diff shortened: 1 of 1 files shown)


tswast:
CC @ludoch

pom.xml
```diff
@@ -9,7 +9,7 @@
     <packaging>pom</packaging>
 
     <properties>
-        <appengine.version>1.9.19</appengine.version>
+        <appengine.sdk.version>1.9.19</appengine.sdk.version>
```
ludoch: Latest GAE version is 1.9.32. Please upgrade.

lesv:
LGTM

pom.xml
tswast: Thanks, will upgrade in a follow-on PR.

PR #101, opened 2016-03-03
Fix dockerfile for extending runtime

shun-fan:


tswast:
LGTM.

Sorry about that! I should have updated this when I renamed it in the pom.

shun-fan:
NP, it's tucked away and the connection is not easy to see. good thing the
automated tests caught it!

PR #103, opened 2016-03-03
Update App Engine SDK version.

tswast:
We don't have many samples that use the appengine-maven-plugin anymore, but those that do now should all inherit from the root pom.xml.

CC @ludoch @shun-fan


tswast:
Wait to merge. I think I missed a couple samples.

tswast:
Okay. Should be ready.

ludoch:
LGTM.

lesv:
LGTM

PR #104, opened 2016-03-04
Fix checkstyle errors in BigQuery sample tests.

tswast:
I will start running checkstyle on the test directories soon for #61. We'll ignore some rules (Javadocs don't make as much sense for test classes, for example), but I want to make sure the majority of formatting and unused import rules are followed.

I'm working my way through the samples, but the BigQuery samples needed a lot of work, so I'm sending these so the changes can be reviewed separately.


lesv:
LGTM

tswast:
I'm looking into Travis failure. It looks the failure is caused by this change.

tswast:
I uploaded a fix. I forgot to update the constants.json file when I renamed the constants to follow the [no all-caps acronym rule](https://google.github.io/styleguide/javaguide.html#s5.3-camel-case).

lesv:
LGTM - presuming things pass.

PR #108, opened 2016-03-15
Checkstyle fixes in Cloud Storage JSON API tests.

tswast:
Getting ready to use shared java-repo-tools configuration.

I also convert the assertions to use [truth](http://google.github.io/truth/).

Since I find it more readable to get a list of object names first
instead of manually looping through, I also update the sample to use
Java 8 lambdas to extract these object names. The sample is used here:
https://cloud.google.com/storage/docs/json_api/v1/json-api-java-samples
which does not indicate the required Java version.

Depends on https://github.com/GoogleCloudPlatform/java-docs-samples/pull/107

### DIFF
--- a/storage/json-api/pom.xml
+++ b/storage/json-api/pom.xml
@@ -15,6 +15,15 @@
   <build>
     <plugins>
       <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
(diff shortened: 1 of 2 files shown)


tswast:
Closing because I told it to merge into the wrong branch.

PR #109, opened 2016-03-15
Checkstyle fixes in Cloud Storage JSON API tests.

tswast:
Getting ready to use shared java-repo-tools configuration.

I also convert the assertions to use [truth](http://google.github.io/truth/).

Since I find it more readable to get a list of object names first
instead of manually looping through, I also update the sample to use
Java 8 lambdas to extract these object names. The sample is used here:
https://cloud.google.com/storage/docs/json_api/v1/json-api-java-samples
which does not indicate the required Java version.

Depends on https://github.com/GoogleCloudPlatform/java-docs-samples/pull/107

### DIFF
--- a/storage/json-api/pom.xml
+++ b/storage/json-api/pom.xml
@@ -12,6 +12,12 @@
   <artifactId>storage-json-sample</artifactId>
   <version>1</version>
   <name>Sample accessing the Google Cloud Storage JSON API using OAuth 2.0.</name>
+
+  <properties>
+      <maven.compiler.target>1.8</maven.compiler.target>
+      <maven.compiler.source>1.8</maven.compiler.source>
+  </properties>
+
   <build>
     <plugins>
       <plugin>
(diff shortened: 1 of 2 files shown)


storage/json-api/pom.xml
```diff
@@ -15,6 +15,15 @@
   <build>
     <plugins>
       <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-compiler-plugin</artifactId>
+        <version>2.5.1</version>
+        <configuration>
+          <source>1.8</source>
```
lesv: You can specify this w/ the shortcut maven.compiler.source & maven.compiler.target  -- it's one less thing we need to track.

lesv:
Just use `<properties>` and set `maven.compiler.source` & `maven.compiler.target`

storage/json-api/pom.xml
tswast: Fixed in latest commit push. Thanks for the suggestion.

lesv:
LGTM

PR #112, opened 2016-03-15
Update gcloud-java version to 0.1.5 and clean up cloudstorage sample

ajkannan:
Updates the gcloud-java version in the remaining two gcloud-java samples.  Also cleans up the managed_vms/cloudstorage sample and docs.

I tested the cloudstorage sample manually both locally and in deployment.


lesv:
LGTM

PR #114, opened 2016-03-16
Checkstyle fix for unittests samples.

tswast:
This requires an update to the documentation at https://cloud.google.com/appengine/docs/java/tools/localunittesting. I will ping a tech writer to restage after this merges.


lesv:
LGTM

PR #115, opened 2016-03-16
Apply shared checkstyles to all samples.

tswast:
We use git submodules, so to check out all the code for
`java-docs-samples`, we need to clone with the `--recursive` flag.

```
git clone --recursive git@github.com:GoogleCloudPlatform/java-docs-samples.git
```

To check out the directory in an already-cloned repo, run:

```
git submodule update --init --recursive
```

Depends on https://github.com/GoogleCloudPlatform/java-docs-samples/pull/114.


tswast:
This should complete the `unittests` item in https://github.com/GoogleCloudPlatform/java-docs-samples/issues/61.

tswast:
Build will likely fail due to https://github.com/GoogleCloudPlatform/java-docs-samples/issues/118. I will send a PR to fix that first.

.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "java-repo-tools"]
```
lesv: I'm sure this is different than the last for a reason?

lesv:
LGTM

Issue #118, opened 2016-03-16
Clean up Monitoring V3 sample.


tswast:
Some minor edits needed for javadocs, also using Truth for some of the assertions would really help in the tests. See PR https://github.com/GoogleCloudPlatform/java-docs-samples/pull/116.

- [x] checkstyle needs to pass on the test classes. Fixed by https://github.com/GoogleCloudPlatform/java-docs-samples/pull/122.
- [x] use Truth for assertions in tests. Fixed by https://github.com/GoogleCloudPlatform/java-docs-samples/pull/122.
- [x] enable tests by default, so that they run in Travis. (Depends on whitelisting the `cloud-samples-tests` project.)
- [x] adjust line breaks to follow "the rectangle rule".
- [x] `<p>` tags in recommended line in javadocs.
- [x] make license headers regular comment, not javadocs.

waprin:
Will circle back to this as soon as get through launch craziness.

tswast:
Some test classes are currently failing checkstyle checks.

```
[INFO] --- maven-checkstyle-plugin:2.17:check (default) @ cloud-monitoring-v3-samples ---
[INFO] Starting audit...
/home/travis/build/GoogleCloudPlatform/java-docs-samples/monitoring/v3/src/test/java/CreateCustomMetricTest.java:19: error: 'org.apache.commons.lang3.RandomStringUtils' should be separated from previous import group.
/home/travis/build/GoogleCloudPlatform/java-docs-samples/monitoring/v3/src/test/java/CreateCustomMetricTest.java:53: error: At-clause should have a non-empty description.
/home/travis/build/GoogleCloudPlatform/java-docs-samples/monitoring/v3/src/test/java/CreateCustomMetricTest.java:66: error: At-clause should have a non-empty description.
/home/travis/build/GoogleCloudPlatform/java-docs-samples/monitoring/v3/src/test/java/ListResourcesTest.java:19: error: 'junit.framework.Assert' should be separated from previous import group.
Audit done.
```

https://travis-ci.org/GoogleCloudPlatform/java-docs-samples/builds/116512174

I had planned to enable this check (applying checkstyles to test classes) in PR https://github.com/GoogleCloudPlatform/java-docs-samples/pull/115.

PR #119, opened 2016-03-16
Checkstyle fixes for async-rest sample.

tswast:
Tested manually by deploying to http://async-rest-test-dot-java-docs-samples-swast-1.appspot.com/.

When doing so, I noticed that the `mvn gcloud:deploy` command would not
run, so I updated the POM to the latest version of the gcloud maven
plugin, too.

Depends on #115.
Fixes https://github.com/GoogleCloudPlatform/java-docs-samples/issues/61.


tswast:
Depends on https://github.com/GoogleCloudPlatform/java-docs-samples/pull/115.

tswast:
Fixes https://github.com/GoogleCloudPlatform/java-docs-samples/issues/61.

tswast:
Build is failing due to https://github.com/GoogleCloudPlatform/java-docs-samples/issues/118. New monitoring sample is failing checkstyle.

java-repo-tools
```diff
@@ -0,0 +1 @@
+Subproject commit 5c9a3169e1ee8fd471c0fd7501b50e1b2bc16773
```
lesv: We should document this somewhere - like our internal page.

lesv:
LGTM

java-repo-tools
tswast: Sounds good. I can do that. I also think while we're in the business of auto-updating Maven plugins with Jenkins, we can consider auto-updating submodule commit numbers, too.

PR #120, opened 2016-03-16
Revert "Apply shared checkstyles to all samples."

tswast:
Reverts GoogleCloudPlatform/java-docs-samples#115

Requiring `--recursive` to clone is no good.


tswast:
Empty revert?

PR #121, opened 2016-03-16
Revert "Checkstyle fixes for async-rest sample."

tswast:
Reverts GoogleCloudPlatform/java-docs-samples#119

Requiring `--recursive` to clone is no good. We reference a regular clone in many places in the docs.


lesv:
LGTM

PR #122, opened 2016-03-17
Fix checkstyle in monitoring v3 sample.

tswast:
In this case rolling back (the git submodule which started applying checkstyle to test classes) was the correct fix for the build breakage, but I've also fixed the checkstyle issues.

CC @lesv

### DIFF
--- a/monitoring/v3/src/test/java/ListResourcesTest.java
+++ b/monitoring/v3/src/test/java/ListResourcesTest.java
@@ -1,12 +1,12 @@
-/**
+/*
  * Copyright (c) 2015 Google Inc.
- * <p/>
+ *
  * Licensed under the Apache License, Version 2.0 (the "License"); you may not
  * use this file except in compliance with the License. You may obtain a copy of
  * the License at
- * <p/>
+ *
  * http://www.apache.org/licenses/LICENSE-2.0
- * <p/>
+ *
  * Unless required by applicable law or agreed to in writing, software
  * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
  * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
@@ -14,9 +14,10 @@
  * the License.
  */
 
+import static com.google.common.truth.Truth.assertThat;
 
 import com.google.api.services.monitoring.v3.Monitoring;
-import junit.framework.Assert;
+
 import org.junit.Before;
 import org.junit.Test;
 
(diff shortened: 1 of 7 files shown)


monitoring/v3/src/test/java/ListResourcesTest.java
```diff
@@ -34,7 +35,7 @@
   /**
    * The project ID of the project created for the integration tests.
    */
-  public static final String TEST_PROJECT_ID = "cloud-monitoring-dev";
+  public static final String TEST_PROJECT_ID = "cloud-samples-tests";
```
waprin: This project won't have the API enabled. It currently requires a whitelist but should no longer require a whitelist by tomorrow. Maybe it doesn't matter because we're skipping tests. In the Python case I just added the main test project as an owner to my whitelisted project. Idk

waprin:
LGTM. The tests won't pass but they're being skipped by default anyway ,and the tests wouldn't pass anyway unless I add the main project as an owner to the whitelisted project (whicH I can do if you do want tests to pass in the meantime).

monitoring/v3/src/test/java/ListResourcesTest.java
tswast: Thanks for the heads up.

I was able to enable the API from my corp account, but it still might require some whitelisting. I get the following error when I run it:

```
$ GOOGLE_APPLICATION_CREDENTIALS=~/src/java-docs-samples/service-account.json mvn verify -DskipTests=false
...
Running ListResourcesTest
Tests run: 3, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 2.133 sec - in ListResourcesTest
Running CreateCustomMetricTest
Tests run: 1, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 0.234 sec <<< FAILURE! - in CreateCustomMetricTest
testValueRead(CreateCustomMetricTest)  Time elapsed: 0.234 sec  <<< ERROR!
com.google.api.client.googleapis.json.GoogleJsonResponseException: 
400 Bad Request
{
  "code" : 400,
  "errors" : [ {
    "domain" : "global",
    "message" : "The project 'cloud-samples-tests' is not in the requisite Stackdriver service tier.",
    "reason" : "badRequest"
  } ],
  "message" : "The project 'cloud-samples-tests' is not in the requisite Stackdriver service tier.",
  "status" : "INVALID_ARGUMENT"
}
...
```

I sent a request to the Cloud Monitoring team. I'll mail another PR to enable the tests when it does get whitelisted.

PR #123, opened 2016-03-17
Issue 188

gregw:
A demo of using the pure jetty image with a clustered session manager


gregw:
Ooops README not ready.... will close and prepare again later

managed_vms/jetty9-gcloud-sessions/src/main/jetty-base/etc/jetty-gcloud-sessions.xml
```diff
@@ -0,0 +1,33 @@
...
+  <!-- ============================================================================================== -->
+  <!-- GCloud configuration.                                                                          -->
+  <!-- This replaces the file in jetty releases < 9.3.8 with a fix to extract the worker name from    -->
+  <!-- the GAE_MODULE_INSTANCE env variable.  This can be removed from jetty-9.3.8 onwards.           -->
+  <!-- ============================================================================================== -->
+  <New id="gconf" class="org.eclipse.jetty.gcloud.session.GCloudConfiguration">
+    <!-- To contact remote gclouddatastore set the following properties in start.ini -->
+    <!-- Either set jetty.gcloudSession.projectId or use system property/env var DATASTORE_DATASET-->
+    <Set name="projectId"><Property name="jetty.gcloudSession.projectId"/></Set>
+    <Set name="p12File"><Property name="jetty.gcloudSession.p12File"/></Set>
```
lesv: Ancient history.

managed_vms/jetty9-gcloud-sessions/src/main/jetty-base/start.d/gcloud-sessions-config.ini
```diff
@@ -0,0 +1,18 @@
+## The gcloud projectId
+## Set this property to connect to remote gcloud datastore.
+## Or, set the DATASTORE_DATASET System property/env variable instead.
+jetty.gcloudSession.projectId=jetty9-work
+
+## The p12 file associated with the project.
+## Set this property to connect to remote gcloud datastore
+jetty.gcloudSession.p12File=etc/jetty9-work-fe328934f681.p12
```
lesv: definitely should be a generic name and p12 is wrong format.

managed_vms/jetty9-gcloud-sessions/src/main/jetty-base/start.d/gcloud-sessions-config.ini
```diff
@@ -0,0 +1,18 @@
...
+
+## The serviceAccount for the Datastore.
+## Set this property to connect to to remote gcloud datastore
+jetty.gcloudSession.serviceAccount=gw-test-service-account@jetty9-work.iam.gserviceaccount.com
```
lesv: really good idea not to check these in

gregw: Doh!  will cancel and regenerate new private ones.

managed_vms/jetty9-gcloud-sessions/src/main/jetty-base/etc/jetty-gcloud-sessions.xml
gregw: @lesv Note sure I'm following you here as to what is ancient history?
The `jetty-gcloud-sessions.xml` is the current way we configure the gcloud session manager in the current jetty release.  The change in this PR was to add the Env lookup for `GAE_MODULE_INSTANCE`

So are you saying that the ancient history is the form of the gcloud data store configuration that jetty is currently using?    I think @janbartel has been having issues navigating exactly which are current APIs to use and what exactly is current best practise.  So if you have any pointers for "Usable Auth" and the JSON file we should be using to configure, then we will look to update the session manager to use them.

However, without such an update, does the ancient history of this approach negate the value of this example?  Ie should we complete this example with the current session manager or wait until we can update the session manager to a more recent version?

Perhaps we should do the example with the JDBC session manager instead?

lesv: All Google services are moving to [Usable Auth](https://github.com/GoogleCloudPlatform/gcloud-java#authentication) for 2LO (computer<==>computer).  (It automatically does auth in AppEngine, GCP (MVMs, GCE, GKE, inside or outside of Docker containers, etc), and locally.)

Datastore currently uses this, and the way we'll be doing SQL will also require something like this. (as well as MySQL username/pw).

Understandable that you don't support it yet, but it's the direction we are going and we are trying to eliminate the need for files, json or p12 except when you are running outside the environment in which case you can set an environment variable to point to your json credentials.

For the main thrust of your question -- I support going forward with what you have, but we do need to get it to move towards the new APIs / auth. ( @ludo call)

PR #132, opened 2016-03-22
Update README.md

waprin:
Just a README change so I think we can merge it. 

This recommends the new gcloud ADC command for local environments.


tswast:
LGTM

PR #133, opened 2016-03-22
Add appengine versions of managed vm samples

shun-fan:


appengine/analytics/pom.xml
```diff
@@ -0,0 +1,56 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
tswast: Can remove xml stuff and have just `<project>`, add license header.

tswast:
Please add these as modules to the root `pom.xml` https://github.com/GoogleCloudPlatform/java-docs-samples/blob/master/pom.xml#L46 so that they get built and tested by Travis. I don't see this in the file list.

appengine/analytics/src/main/java/com/example/appengine/analytics/AnalyticsServlet.java
```diff
@@ -0,0 +1,65 @@
...
+
+// [START example]
+@SuppressWarnings("serial")
+public class AnalyticsServlet extends HttpServlet {
+
+  @Override
+  public void doGet(HttpServletRequest req, HttpServletResponse resp) throws IOException,
+      ServletException {
+    String trackingId = System.getenv("GA_TRACKING_ID");
+//    HttpClient client = HttpClientBuilder.create().build();
```
tswast: Delete commented line or explain why it's here.

appengine/analytics/src/main/webapp/WEB-INF/appengine-web.xml
```diff
@@ -0,0 +1,24 @@
...
+  distributed under the License is distributed on an "AS IS" BASIS,
+  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  See the License for the specific language governing permissions and
+  limitations under the License.
+-->
+<!-- [END_EXCLUDE] -->
+<appengine-web-app xmlns="http://appengine.google.com/ns/1.0">
+  <application>YOUR-PROJECT-ID</application>
+  <version>YOUR-VERSION-ID</version>
+  <threadsafe>true</threadsafe>
```
lesv: We should consider adding `<vm>false</vm>` that is done w/ a variable `<vm>$flex</vm>` and in the pom.xml add
`<properties><flex>false</flex></properties>` so that we could just deploy to Flex w/ `mvn -Dflex=true gcloud:deploy`

appengine/memcache/pom.xml
```diff
@@ -0,0 +1,60 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
lesv: Same comment as Tim earlier.

appengine/sendgrid/pom.xml
```diff
@@ -0,0 +1,54 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
lesv: ditto

appengine/analytics/src/main/webapp/WEB-INF/appengine-web.xml
```diff
@@ -0,0 +1,24 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- [START_EXCLUDE] -->
```
lesv: Do you know this works?  I haven't managed to get it to kick in properly, it really needs a [START app] ... [END app]

shun-fan:
@lesv @tswast ptal

appengine/analytics/pom.xml
```diff
@@ -0,0 +1,69 @@
...
+      <type>jar</type>
+      <scope>provided</scope>
+    </dependency>
+  </dependencies>
+  <build>
+    <!-- for hot reload of the web application -->
+    <outputDirectory>${project.build.directory}/${project.build.finalName}/WEB-INF/classes</outputDirectory>
+    <plugins>
+      <!-- Parent POM defines ${appengine.sdk.version} (updates frequently). -->
+      <plugin>
```
lesv: I'm Ok w/ this, but would rather you used `<maven.compiler.source>1.7</maven.compiler.source>` and 
   `<maven.compiler.target>1.7/maven.compiler.target>` in `properties` and not specify the plugin.

appengine/analytics/pom.xml
```diff
@@ -0,0 +1,69 @@
...
+        <groupId>com.google.appengine</groupId>
+        <artifactId>appengine-maven-plugin</artifactId>
+        <version>${appengine.sdk.version}</version>
+      </plugin>
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
```
lesv: I'm Ok w/ this, but would rather you used `<maven.compiler.source>1.7</maven.compiler.source>` and 
   `<maven.compiler.target>1.7/maven.compiler.target>` in `properties` and not specify the plugin.

lesv:
We should fix the decrease in coverage.  And I'd still like more configurability and simplification of the pom's and appengine-web.xml so we can run these on both Standard and Flex-compat.  But LGTM for now -- Thank you.

PR #135, opened 2016-03-23
Add a new example for using Customer-Supplied Encryption Keys.

BrandonY:


storage/json-api/src/main/java/CustomerSuppliedEncryptionKeysSamples.java
```diff
@@ -0,0 +1,315 @@
...
+
+    // Now set the CSEK headers
+    final HttpHeaders httpHeaders = new HttpHeaders();
+    httpHeaders.set("x-goog-encryption-algorithm", "AES256");
+    httpHeaders.set("x-goog-encryption-key", base64CSEKey);
+    httpHeaders.set("x-goog-encryption-key-sha256", base64CSEKeyHash);
+
+    // Since our request includes our private key as a header, it is a good idea to instruct caches
+    // and proxies not to store this request.
+    httpHeaders.setCacheControl("no-store");
```
hegemonic: After extensive discussion, we decided to update the docs so that they no longer recommend using the `Cache-Control` header, so this should be removed.

storage/json-api/src/main/java/CustomerSuppliedEncryptionKeysSamples.java
```diff
@@ -0,0 +1,315 @@
...
+
+    // Now set the CSEK headers
+    final HttpHeaders httpHeaders = new HttpHeaders();
+    httpHeaders.set("x-goog-encryption-algorithm", "AES256");
+    httpHeaders.set("x-goog-encryption-key", base64CSEKey);
+    httpHeaders.set("x-goog-encryption-key-sha256", base64CSEKeyHash);
+    
+    // Since our request includes our private key as a header, it is a good idea to instruct caches
+    // and proxies not to store this request.
+    httpHeaders.setCacheControl("no-store");
```
hegemonic: As noted above, this header should be removed.

storage/json-api/src/main/java/CustomerSuppliedEncryptionKeysSamples.java
```diff
@@ -0,0 +1,315 @@
...
+    httpHeaders.set("x-goog-copy-source-encryption-key-sha256", originalBase64KeyHash);
+
+    // Specify the new CSEK that we would like to apply.
+    httpHeaders.set("x-goog-encryption-algorithm", "AES256");
+    httpHeaders.set("x-goog-encryption-key", newBase64Key);
+    httpHeaders.set("x-goog-encryption-key-sha256", newBase64KeyHash);
+    
+    // Since our request includes our private key as a header, it is a good idea to instruct caches
+    // and proxies not to store this request.
+    httpHeaders.setCacheControl("no-store");
```
hegemonic: As noted above, this header should be removed.

storage/json-api/src/main/resources/client_secrets.json
```diff
@@ -0,0 +1 @@
+{"installed":{"client_id":"122681785480-35mkukq7o3hh55i2uuftv0hp168hauet.apps.googleusercontent.com","project_id":"gcs-code-samples","auth_uri":"https://accounts.google.com/o/oauth2/auth","token_uri":"https://accounts.google.com/o/oauth2/token","auth_provider_x509_cert_url":"https://www.googleapis.com/oauth2/v1/certs","client_secret":"hf-15vXoQQY8OonIK7qdrR0L","redirect_uris":["urn:ietf:wg:oauth:2.0:oob","http://localhost"]}}
```
Capstan: Is it common practice to ship a secrets file?

lesv: No - please invalidate your secrets in pantheon as well.

BrandonY: I'm confused by this. Applications ship client IDs client secrets. For example, our own command-line client, gsutil, ships with its own client secret. Why should a sample app not be the same?

lesv: I was in error, but I do think users should be creating their own client_secrets.json file.  They aren't your app, it's their app. (Unless you are expecting them to just compile and run your app always.)

lesv: Note - it's your call - your interest is in teaching user supplied keys and while we tend to always ask the user to create their own, I can see your argument for providing one.

Though I'd suggest adding a newline to the file if you decide to keep it.

BrandonY: I have switched to using GoogleCredential.getApplicationDefault(). It adds a dependency on gcloud, but I expect that almost anyone using this example will in practice also want to be using gcloud.

storage/json-api/src/main/java/CustomerSuppliedEncryptionKeysSamples.java
BrandonY: Done.

lesv:
@tswast Tim - Could you help them get this through Travis?

Thanks for switching to Usable Auth. I really should know better than to question folks working on crypto about security - It's what happens when I review too quickly.  

I'm happy with the code, it's ok to merge once Travis passes.

tswast:
Closing this and using a PR https://github.com/GoogleCloudPlatform/java-docs-samples/pull/137 from a branch to make Travis happy.

PR #136, opened 2016-03-23
Update service account keys to use java-docs-samples service account.

tswast:
This now uses a key for the java-docs-samples service account, to better track where these keys are used and test locally by downloading a key for this account and running:

```
GOOGLE_APPLICATION_CREDENTIALS=~/src/java-docs-samples/service-account.json mvn clean verify
```

I'm making this PR because I did not know which service account we were using before, so it was impossible to figure out how to add permissions to it to let it access other projects. Now it should be very clear that `java-docs-samples` service account on the `cloud-samples-tests` project is what we use.


lesv:
LGTM - we should have a keys conversation offline.

PR #147, opened 2016-04-02
Cleanup error handling/docs in TaskList

ajkannan:
/cc @aozarov

### DIFF
--- a/datastore/src/main/java/com/google/datastore/snippets/TaskList.java
+++ b/datastore/src/main/java/com/google/datastore/snippets/TaskList.java
@@ -17,7 +17,6 @@
 package com.google.datastore.snippets;
 
 import com.google.gcloud.datastore.Datastore;
-import com.google.gcloud.datastore.DatastoreException;
 import com.google.gcloud.datastore.DatastoreOptions;
 import com.google.gcloud.datastore.DateTime;
 import com.google.gcloud.datastore.Entity;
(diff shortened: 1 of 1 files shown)


datastore/src/main/java/com/google/datastore/snippets/TaskList.java
```diff
@@ -51,7 +50,8 @@
    * Adds a task entity to the Datastore.
    *
    * @param description The task description
-   * @return The {@link Key} of the entity.
+   * @return The {@link Key} of the entity
+   * @throws DatastoreException if the Datastore put fails
```
aozarov: technically it could also throw `DatastoreException` if Id allocation failed.

datastore/src/main/java/com/google/datastore/snippets/TaskList.java
```diff
@@ -70,14 +70,18 @@ Key addTask(String description) {
    * Marks a task entity as done.
    *
    * @param id The ID of the task entity as given by {@link Key#id()}
-   * @throws DatastoreException if the task does not exist
+   * @return true if the task was found, false if not
+   * @throws DatastoreException if the transaction commit fails
```
aozarov: For similar reason to the above (there is also get) if you want to be picky you can say "if Datastore transaction fails"

aozarov:
Looks good. Few nits. Feel free to merge without them.

ajkannan:
Fixed comments

PR #148, opened 2016-04-02
Update gcloud-java version in samples

ajkannan:
/cc @shun-fan 

Also, do you think a README should be added to managed_vms/datastore detailing how to run it and what it does?


tswast:
Thanks! This fixes https://github.com/GoogleCloudPlatform/java-docs-samples/issues/150.

PR #149, opened 2016-04-03
Fix jar filename in Dockerfile

shauvik:
I had to make this change for the sample to work. 
Seems like the projectname was updated and didn't reflect here.


tswast:
Looks Good To Me. Thanks!

I am in the process of getting some deployment tests running in our Jenkins instance, so hopefully we will be notified when we make changes like this rename that break deployments.

PR #152, opened 2016-04-04
Improve READMEs for App Engine samples.

tswast:
These are linked to from https://cloud.google.com/java/, so we should
have a more welcoming landing page than the existing list of folders.

CC @lesv @shun-fan

### DIFF
--- a/appengine/README.md
+++ b/appengine/README.md
@@ -1,0 +1,101 @@
+# Google App Engine Standard Environment Java Samples
+
+This is a repository that contains Java code samples for [Google App Engine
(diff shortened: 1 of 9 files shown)


shun-fan:
LGTM

appengine/README.md
```diff
@@ -0,0 +1,101 @@
+# Google App Engine Standard Environment Java Samples
+
```
lesv: You can put a table of contents at the top to help guide the process.  see github.com/googlecloudplatform/cloud-bigtable-examples/

PR #153, opened 2016-04-04
Change Hello World compiler to Java 7.

tswast:
This sample is linked to from https://cloud.google.com/java/ as "App Engine flexible environment for Java 7" https://cloud.google.com/appengine/docs/flexible/java/hello-world. We should make sure it really is compatible with Java 7 by making sure we use that compiler instead of Java 8.


lesv:
LGTM

PR #157, opened 2016-04-08
Add cloudsql gen1 sample for app engine

shun-fan:


appengine/cloudsql/README.md
```diff
@@ -0,0 +1,12 @@
+# Cloud SQL sample for Google App Enginge
+This sample demonstrates how to use [Cloud SQL](https://cloud.google.com/sql/) on Google App Engine
+## Setup
+Before you can run or deploy the sample, you will need to create a [Cloud SQL instance](https://cloud.google.com/sql/docs/create-instance)
```
lesv: Always good to spell it out!

appengine/cloudsql/README.md
```diff
@@ -0,0 +1,12 @@
...
+1. Create a new user and database for the application. The easiest way to do this is via the [Google Developers Console](https://console.developers.google.com/project/_/sql/instances/example-instance2/access-control/users). Alternatively, you can use MySQL tools such as the command line client or workbench.
```
lesv: Can't I connect to Cloud SQL v2 instances when I'm local?  If so, why not tell them how?

appengine/cloudsql/pom.xml
```diff
@@ -0,0 +1,67 @@
...
+
+  <parent>
+    <artifactId>doc-samples</artifactId>
+    <groupId>com.google.cloud</groupId>
+    <version>1.0.0</version>
+    <relativePath>../..</relativePath>
+  </parent>
+
+  <dependencies>
+    <dependency>
```
lesv: We might wish to have this in dependencies and move them up here.

lesv:
I'm happy, but it would be nice if the README had more content and the content I pointed out.  LGTM after that.

PR #159, opened 2016-04-12
Update gcloud-java dependency and imports

ajkannan:
/cc @aozarov

### DIFF
--- a/datastore/src/main/java/com/google/datastore/snippets/Concepts.java
+++ b/datastore/src/main/java/com/google/datastore/snippets/Concepts.java
@@ -23,34 +23,33 @@ import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertNotNull;
 import static org.junit.Assert.assertNull;
 
+import com.google.cloud.datastore.Cursor;
+import com.google.cloud.datastore.Datastore;
+import com.google.cloud.datastore.DatastoreException;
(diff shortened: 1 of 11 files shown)


datastore/src/main/java/com/google/datastore/snippets/Concepts.java
```diff
@@ -114,12 +109,7 @@ public static void beforeClass() throws IOException, InterruptedException {
    */
   @Before
   public void setUp() {
-    datastore = DatastoreOptions.builder()
-        .projectId(PROJECT_ID)
-        .namespace("ghijklmnop")
-        .host("http://localhost:" + PORT)
-        .build()
-        .service();
+    datastore = HELPER.options().toBuilder().namespace("ghijklmnop").build().service();
```
aozarov: @ajkannan I think it would be nice if we overloaded `options` in the helper with a `defaultNamespace`.
Also, we should probably document the namespace setter in the Builder and getter in the options to suggests
that this is the "default namespace".

aozarov:
Looks good to me.

shun-fan:
+1

datastore/src/main/java/com/google/datastore/snippets/Concepts.java
ajkannan: Sure, I'll make an issue in our repo to do that.

PR #160, opened 2016-04-12
Add a more complex example as well.

lesv:
@tswast


appengine/urlfetch/src/main/java/com/example/appengine/UrlFetchServlet.java
```diff
@@ -1,12 +1,12 @@
 /**
```
tswast: If you use `/*` (non-Javadoc for license header) you don't need the `<p>` tags.

tswast:
LGTM, minus `<p>` tag in license header nit.

lesv:
LGTM

PR #162, opened 2016-04-13
Migrate sample from compute-getting-started-java repo.

tswast:
Only made minimal changes to follow style guide and update links.

PTAL @jsimonweb @lesv


lesv:
LGTM - sad that every time we add things, coverage regresses. :(

PR #163, opened 2016-04-15
Add compute sendgrid example

shun-fan:
Add a compute engine sendgrid sample, with readme

### DIFF
--- a/compute/sendgrid/src/main/java/com/example/compute/sendgrid/SendEmailServlet.java
+++ b/compute/sendgrid/src/main/java/com/example/compute/sendgrid/SendEmailServlet.java
@@ -22,7 +22,7 @@ import com.sendgrid.SendGridException;
 // [START example]
 public class SendEmailServlet {
   final static String SENDGRID_API_KEY = "YOUR-SENDGRID-API-KEY";
-  final static String SENDGRID_SENDER = "YOUR-SENDGRID-SENDER";
+  final static String SENDGRID_SENDER = "YOUR-SENDGRID-FROM-EMAIL";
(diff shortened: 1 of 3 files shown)


shun-fan:
@tswast  fyi since les is out today

compute/sendgrid/src/main/java/com/example/compute/sendgrid/SendEmailServlet.java
```diff
@@ -0,0 +1,46 @@
...
+
+package com.example.compute.sendgrid;
+
+import com.sendgrid.SendGrid;
+import com.sendgrid.SendGridException;
+
+// [START example]
+public class SendEmailServlet {
+  final static String SENDGRID_API_KEY = "YOUR-SENDGRID-API-KEY";
+  final static String SENDGRID_SENDER = "YOUR-SENDGRID-SENDER";
```
phunl: Is the value for sendgrid-sender an email address? Maybe something like "SENDER-EMAIL" to be consistent with "DESTINATION-EMAIL"?

lesv: +1

lesv:
LGTM w/ change of YOUR-SENDGRID-SENDER to YOUR-SENDGRID-FROM-EMAIL or YOUR-SENDGRID-SENDER-EMAIL.

For the record, I would prefer we figure out how to do these w/o editing, so that mvn appengine:deploy -DAPIKEY=blah -DSENDER_EMAIL=me@example.org

PR #171, opened 2016-04-21
Add App Engine Datastore raw APIs sample.

tswast:
These samples are pulled from the [projection query](https://cloud.google.com/appengine/docs/java/datastore/projectionqueries) and [structuring for strong consistency](https://cloud.google.com/appengine/docs/java/datastore/structuring_for_strong_consistency) pages.

I will add additional code to this sample as I convert the other pages.

@lesv PTAL.

### DIFF
--- a/appengine/datastore/pom.xml
+++ b/appengine/datastore/pom.xml
@@ -1,0 +1,114 @@
+<!--
+Copyright 2016 Google Inc. All Rights Reserved.
+
+ Licensed under the Apache License, Version 2.0 (the "License");
+ you may not use this file except in compliance with the License.
+ You may obtain a copy of the License at
+
+       http://www.apache.org/licenses/LICENSE-2.0
+
+ Unless required by applicable law or agreed to in writing, software
+ distributed under the License is distributed on an "AS IS" BASIS,
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ See the License for the specific language governing permissions and
+ limitations under the License.
+-->
+<project>
+  <modelVersion>4.0.0</modelVersion>
+  <packaging>war</packaging>
+  <version>1.0-SNAPSHOT</version>
+  <groupId>com.example.appengine</groupId>
+  <artifactId>appengine-datastore</artifactId>
+  <parent>
+    <groupId>com.google.cloud</groupId>
+    <artifactId>doc-samples</artifactId>
+    <version>1.0.0</version>
+    <relativePath>../..</relativePath>
+  </parent>
+  <dependencies>
+    <dependency>
+      <groupId>com.google.appengine</groupId>
+      <artifactId>appengine-api-1.0-sdk</artifactId>
+      <version>${appengine.sdk.version}</version>
+    </dependency>
+    <dependency>
+      <groupId>com.google.auto.value</groupId>
+      <artifactId>auto-value</artifactId>
+      <version>1.2</version>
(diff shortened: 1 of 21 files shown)


appengine/datastore/src/main/java/com/example/appengine/GuestbookStrong.java
```diff
@@ -0,0 +1,82 @@
...
+ * APIs. See the
+ * <a href="https://cloud.google.com/appengine/docs/java/datastore/">documentation</a>
+ * for more information.
+ */
+class GuestbookStrong extends AbstractGuestbook {
+  private final String guestbookName;
+
+  GuestbookStrong(
+      String guestbookName,
+      DatastoreService datastore,
```
tmatsuo: You have `DatastoreService` and `UserService` as parameters to the constructor. Is it because you wanted them to be injectable?  Generally it's an easy and readable DI mechanism and great for simple samples like ours, but actually it is not needed for App Engine services. The test helper will replace the stubs for you, so you can just create the service object from the factory in the user land code. I may miss some other reasons though.

appengine/datastore/pom.xml
```diff
@@ -0,0 +1,123 @@
...
+      <artifactId>truth</artifactId>
+      <version>0.28</version>
+      <scope>test</scope>
+    </dependency>
+  </dependencies>
+  <build>
+    <!-- for hot reload of the web application -->
+    <outputDirectory>${project.build.directory}/${project.build.finalName}/WEB-INF/classes</outputDirectory>
+    <plugins>
+      <plugin>
```
lesv: Not required, defined in parent.

lesv:
Consider Takashi's comments and mine.  Ok to merge after that.

appengine/datastore/src/main/java/com/example/appengine/GuestbookStrong.java
tswast: Yes, it's to be injectable. I also have DatastoreService as a parameter for some functions so that when I include the code the reader will see that the variable `datastore` is a `DatastoreService` type.

It wasn't clear to me if the DatastoreService created by the factory would connect to the same or a different in-memory datastore if it is called more than once. With injection, it's clear that we are definitely connecting to the same datastore in the test and in the code we are trying to test.

Plus, this is the way we always did it internally at Google. (Actually, we injected into the Servlet classes, as well, using Guice to do URL routing, but personally I find that super confusing and inappropriate for a sample)

tmatsuo: I'm pretty sure that the services will use the same database, but it allows you to re-use the service object, and I think that's a fair reason. I just wanted to point out that with the App Engine test helper, you don't have to have DI container solely for testing purpose, at least for the services provided by App Engine.

tmatsuo:
LGTM

appengine/datastore/pom.xml
tswast: Removed Maven plugin in latest commit (since it's defined in parent).

appengine/datastore/src/main/java/com/example/appengine/GuestbookStrong.java
tswast: Good point. I just did a test with no injection and it does seem to use the same database.

To get the sharing with less boilerplate, I moved the factory call to the AbstractGuestbook class. That way the constructor doesn't need to get these passed in, but we still have object sharing.

PR #173, opened 2016-04-21
Added few more sample code for Search API.

tmatsuo:
* Moved some checkstyle modules to java-repo-tools
* Moved some lines in .gitignore to the top level


tmatsuo:
@tswast @lesv I forgot to add some more code for the Search API. Also moved common configurations to the top level. Please take a look.

tmatsuo:
Also added code on https://cloud.google.com/appengine/docs/java/search/options

appengine/search/src/main/java/com/example/appengine/search/DeleteServlet.java
```diff
@@ -23,13 +23,13 @@
 import com.google.appengine.api.search.GetResponse;
 // [END delete_import]
 
-//CHECKSTYLE:OFF
+// CHECKSTYLE:OFF
```
lesv: Why did you remove checkstyle here?

lesv: Ah - I guess it's so that it stays in the right place w/ your comments.

lesv:
LGTM

PR #175, opened 2016-04-21
Update SearchOptionServlet.java

tmatsuo:
Corrected the position of an end region tag.

### DIFF
--- a/appengine/search/src/main/java/com/example/appengine/search/SearchOptionServlet.java
+++ b/appengine/search/src/main/java/com/example/appengine/search/SearchOptionServlet.java
@@ -95,8 +95,8 @@ public class SearchOptionServlet extends HttpServlet {
     } catch (SearchException e) {
       // handle exception...
     }
-    return null;
     // [END search_with_options]
+    return null;
   }
 
   @Override


tmatsuo:
@lesv Merging. Please put comments if you have any concerns

tswast:
LGTM.

appengine/search/src/main/java/com/example/appengine/search/SearchOptionServlet.java
```diff
@@ -95,8 +95,8 @@ private Index getIndex() {
     } catch (SearchException e) {
       // handle exception...
```
lesv: For the record, we really shouldn't be doing this.  We should actually handle the exception or propagate it.

tmatsuo: The comment says "handle exception", and it is deferred to users how to handle it :)

lesv: Yes - that is the point, we need to show them how to handle it properly.

PR #176, opened 2016-04-21
Update memcache sample with code from docs pages

shun-fan:
Adding servlets for the different memcache samples here
https://cloud.google.com/appengine/docs/java/memcache/

### DIFF
--- a/appengine/memcache/src/main/java/com/example/appengine/memcache/MemcacheAsyncCacheServlet.java
+++ b/appengine/memcache/src/main/java/com/example/appengine/memcache/MemcacheAsyncCacheServlet.java
@@ -1,0 +1,75 @@
+/**
+ * Copyright 2016 Google Inc. All Rights Reserved.
+ *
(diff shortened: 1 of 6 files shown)


appengine/memcache/src/main/java/com/example/appengine/memcache/MemcacheAsyncCacheServlet.java
```diff
@@ -0,0 +1,75 @@
...
+
+  @Override
+  public void doGet(HttpServletRequest req, HttpServletResponse resp) throws IOException,
+      ServletException {
+    String path = req.getRequestURI();
+    if (path.startsWith("/favicon.ico")) {
+      return; // ignore the request for favicon.ico
+    }
+
+    // [START example]
```
tmatsuo: The region tag name `example` is OK, but I'd prefer more explicit one

tmatsuo:
Can you add tests?

lesv:
LGTM - please add some testing

PR #177, opened 2016-04-21
Add sample for App Engine  Datastore setDistinct projection queries.

tswast:
Also update checkstyle to allow 1-letter variable names, since it
doesn't make sense to give long names to something like "arbitrary
column A that doesn't actually mean anything".


tswast:
Build is green. I'll merge so I can get this in the docs today. @lesv PTAL.

tmatsuo:
LGTM

lesv:
LGTM

PR #178, opened 2016-04-21
Add region tags for projection query sample.

tswast:
The START_EXCLUDE wasn't working when I tried to embed the function. I'm reverting to region tags to just include the same code as used in the existing sample.


tmatsuo:
LGTM

PR #181, opened 2016-04-22
Add App Engine raw Datastore API Entity samples.

tswast:
Pulls out all samples from https://cloud.google.com/appengine/docs/java/datastore/entities.

@laurieaw @lesv PTAL

### DIFF
--- a/appengine/datastore/README.md
+++ b/appengine/datastore/README.md
@@ -6,14 +6,41 @@ from [Google App Engine standard environment][ae-docs].
 [java-datastore]: https://cloud.google.com/appengine/docs/java/datastore/
 [ae-docs]: https://cloud.google.com/appengine/docs/java/
 
+
+## Running locally
+
+This example uses the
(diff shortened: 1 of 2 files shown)


appengine/datastore/README.md
```diff
@@ -6,14 +6,41 @@ from [Google App Engine standard environment][ae-docs].
...
+and SOME-VERSION with a valid version number.
+
+    $ mvn appengine:update -Dappengine.appId=YOUR-PROJECT-ID -Dappengine.version=SOME-VERSION
+
+
 ## Setup
+
+To save your project settings so that you don't need to enter the
+ parameters, you can:
+
```
lesv: Can't we just tell them to run it w/ mvn command line args instead of editing below.

lesv:
LGTM

appengine/datastore/README.md
tswast: Yes. I fixed in the latest commit by adding a "Deploying" section like is in the other sections.

PR #182, opened 2016-04-25
Added samples showcasing XMPP

meteatamel:
Added a number of Servlet based examples for XMPP

### DIFF
--- a/appengine/xmpp/README.md
+++ b/appengine/xmpp/README.md
@@ -1,0 +1,20 @@
+# Google App Engine Standard Environment XMPP Java API Overview
+
+This sample demonstrates how to use XMPP Java API on Google App Engine.
+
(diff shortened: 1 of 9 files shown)


lesv:
LGTM

lesv:
@tswast Could you figure out why Travis is unhappy?

tswast:
@lesv Travis passed for the merge commit. https://travis-ci.org/GoogleCloudPlatform/java-docs-samples/builds/125630754

Right now Travis is failing on the `openssl` command to decrypt the service keys. I can send a PR to ignore that error and see how much further Travis can get. Right now it's not even verifying that the sample builds.

tswast:
P.S. We missed adding this to the root pom.xml as a module, so right now it Travis is not verifying the build or check style.

PR #183, opened 2016-04-25
Update connection string url in cloudsql sample

shun-fan:


lesv:
You might wish to update the readme a bit to explain what is going on here, and what they need to do, LGTM after that.

lesv:
LGTM - @shun-fan Could you update the README to mention the revised string.

PR #184, opened 2016-04-25
Fix checkstyle errors in XMPP sample.

tswast:
Add XMPP sample to Travis build, added in https://github.com/GoogleCloudPlatform/java-docs-samples/pull/182.


tswast:
@lesv PTAL
@meteatamel FYI

meteatamel:
Thanks for fixing the checkstyle warnings. For future reference, where can I see these warnings myself?

tswast:
@metea Run `mvn clean verify` instead of `mvn clean test`. Checkstyles are done in the "verify" build phase, which happens after the tests pass.

lesv:
Argh - I should have noticed at least one of those -- Thanks.

LGTM

PR #185, opened 2016-04-25
Allow Travis to continue build if decryption fails. [DO NOT MERGE]

tswast:
Decryption of the service key can fail if Travis is building from a fork instead of a branch.

Note: I'm doing this from a fork to see how far Travis gets with this change. It won't complete all the tests, as some make calls to services that require the service account credentials, but hopefully it at least gets past the build phase.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,12 +15,16 @@
 language: java
 jdk:
   - oraclejdk8
-env:
-  - GOOGLE_APPLICATION_CREDENTIALS=$TRAVIS_BUILD_DIR/service-account.json GCLOUD_PROJECT=cloud-samples-tests
 before_install:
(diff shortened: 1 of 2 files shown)


tswast:
@lesv PTAL.

I'm actually not sure why Travis is passing. It should have failed some test that required credentials. Looking at the logs https://travis-ci.org/GoogleCloudPlatform/java-docs-samples/builds/125663818 it's running `mvn install -DskipTests=true -Dmaven.javadoc.skip=true -B -V` which is *not* what we have defined for the script.

tswast:
Looks like the `mvn install` command is getting run by Travis by default.
http://stackoverflow.com/a/32475349/101923

I'll try disabling that, since we basically do that as our test.

tswast:
Okay. Looks like it's working from Pull Request now. Builds, runs checkstyle, and skips tests. I'll close this PR and open another one from a branch to ensure that works as well.

lesv:
LGTM  --  Perhaps the script should notice that they are building from a fork and not a branch and warn, then continue w/ the checkstyle.

Issue #117, opened 2016-03-16
Make sure Travis passes for non-branch pull requests


tswast:
I did pull out integration tests already for the cloud-vision repository (though more work is needed https://github.com/GoogleCloudPlatform/cloud-vision/issues/50). It means that Travis can at least run the unit tests for pull requests from other repos, skipping the integration tests.
- [x] Ignore decryption failure.
- [x] Run build without the tests. `mvn verify -DskipTests=true
- [ ] Pull out integration tests (that need service account) and run all but those.

tswast:
Example of failed build from external PR: https://travis-ci.org/GoogleCloudPlatform/java-docs-samples/builds/117868157

lesv:
Obsolete.

PR #187, opened 2016-04-25
Allow Travis to continue build if decryption fails.

tswast:
Decryption of the service key can fail if Travis is building from a fork instead of a branch.

This replaces https://github.com/GoogleCloudPlatform/java-docs-samples/pull/185.

This addresses the first two items in https://github.com/GoogleCloudPlatform/java-docs-samples/issues/117 (ignore decryption error and skip tests in that case).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,12 +15,16 @@
 language: java
 jdk:
   - oraclejdk8
-env:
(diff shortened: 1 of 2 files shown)


lesv:
LGTM

PR #188, opened 2016-04-25
Use key name in Datastore snippet.

tswast:
Writers are using this snippet in a slightly different way, so it needs to include the key name instead of using the auto ID.


tswast:
@halgrimur This is for the "Creating Entities" page.

lesv:
LGTM

PR #192, opened 2016-04-26
2 minor tag additions for doc

meteatamel:


lesv:
LGTM

lesv:
It's easier for us if you merge from a branch rather than another :master -- travis doesn't run this way.

PR #193, opened 2016-04-26
Channel Example - Added tags and channel presence

saturnism:
Added tags for docs, and added a channel presence example.


lesv:
Fix the checkstyle, and once things pass, ok to merge.

PR #198, opened 2016-04-26
Exclude generated code files from code coverage.

tswast:
This was causing the Coveralls plugin to fail with "Source not found".

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -35,3 +35,11 @@ service-account.json
 .project
 .classpath
 .settings
+
+# Vim
+[._]*.s[a-w][a-z]
+[._]s[a-w][a-z]
+Session.vim
+.netrwhist
+*~
+tags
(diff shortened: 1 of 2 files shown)


tswast:
@lesv One more change to coverage plugin.

I was able to manually run the plugin. Hopefully this means we'll get a coverage report after Travis finishes.

```
mvn clean cobertura:cobertura coveralls:report -DrepoToken=OUR-TOKEN
```

I'm thinking we should reconsider how we handle protocol buffers, since this will be a pain to maintain, and I'm afraid we might exclude more than we mean to.

tswast:
Ugh. Coveralls had another error when running on Travis.

```
[ERROR] Failed to execute goal org.eluder.coveralls:coveralls-maven-plugin:4.1.0:report (default-cli) on project doc-samples: Build error: Line number 52 is greater than the source file appengine/datastore/src/main/java/com/example/appengine/Greeting.java size -> [Help 1]
```

I'll look into this.

tswast:
"My guess is that you have multi-module project containing duplicate package/class names"
https://github.com/trautonen/coveralls-maven-plugin/issues/73#issuecomment-112551049

We probably do this. I think I use the `com.example.appengine` in multiple modules. I'll see about pulling some of them into unique names.

tswast:
I'll close this for now. Hopefully codecov works better. https://github.com/GoogleCloudPlatform/java-docs-samples/pull/199

PR #199, opened 2016-04-26
Use codecov for coverage.

tswast:
Hopefully this works more reliably than coveralls.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,7 +26,7 @@ before_install:
 install: true
 script: ./travis.sh
 after_success:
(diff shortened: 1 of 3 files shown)


tswast:
It appears to have worked: http://codecov.io/github/GoogleCloudPlatform/java-docs-samples/commit/016bb08b0f93d2a70e9b817bb622c4cc12114bc8

I'm not sure if we'll get notification here or not, but at least it ran without failing.

lesv:
LGTM - was that was discussed earlier?

tswast:
@lesv Yes, it's what @jonparrott recommended.

PR #200, opened 2016-04-26
minor nits

lesv:


lesv:
@saturnism FYI

saturnism:
Ah ha! Great catch! I totally forgot I changed the project is to test. :)
thanks Les!

PR #202, opened 2016-04-27
Add Datastore query examples.

tswast:
From Query concepts and restrictions pages.


lesv:
LGTM - I'd probably define a static String for all the property names, but it reads better this way.

tswast:
@lesv Agreed about static string for property names. Definitely recommended for production. Too easy to make a typo without the compiler catching it otherwise.

For tests / samples, I think it does read better as strings. Makes the sample / test more self-contained. We want to prevent having to look for where these strings are defined, especially when we are only including a snippet in the docs.

PR #179, opened 2016-04-22
Requests

Annie29:
Moved code from https://cloud.google.com/appengine/docs/java/requests to Github

PR #205, opened 2016-04-27
Moving code from TasksQueue Push

Annie29:
@tswast, @lesv 
Somehow this also includes the changes from requests, which I thought I'd added in https://github.com/GoogleCloudPlatform/java-docs-samples/pull/179  .  (Could this be because I created the branch for tasks before merging the requests changes?)

I'll need to change the pom.xml file to add requests as well, it seems.


lesv:
You have conflicts that need to be resolved, and you are pushing from a fork, which means that our testing doesn't run.  If you could push from a branch it would be helpful.

appengine/requests/README.md
```diff
@@ -0,0 +1,43 @@
...
+and examine the logs to see the actual messages.
+
+## Deploying
+In the following command, replace YOUR-PROJECT-ID with your
+[Google Cloud Project ID](https://developers.google.com/console/help/new/#projectnumber) 
+and SOME-VERSION with a valid version number.
+
+    $ mvn appengine:update -Dappengine.appId=YOUR-PROJECT-ID -Dappengine.version=SOME-VERSION
+
+## Setup
```
lesv: This part appears to be redundant.

appengine/requests/pom.xml
```diff
@@ -0,0 +1,112 @@
...
+      <scope>test</scope>
+    </dependency>
+
+  </dependencies>
+  <build>
+    <!-- for hot reload of the web application -->
+    <outputDirectory>${project.build.directory}/${project.build.finalName}/WEB-INF/classes</outputDirectory>
+    <plugins>
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
```
lesv: Compiler plugin isn't required.

appengine/requests/src/test/java/com/example/appengine/requests/LoggingServletTest.java
```diff
@@ -0,0 +1,101 @@
...
+    // We expect three log messages to be created
+    // with the following messages.
+    assertThat(out).contains("An informational message.");
+    assertThat(out).contains("A warning message.");
+    assertThat(out).contains("An error message.");
+
+    // We expect three log messages to be created
+    // with the following severities.
+    // Since there's no guarantee of case, use lowercase
+    String lcOut = out.toLowerCase();
```
lesv: Not sure this will find what you are looking for.  "An informational message." contains "info".  "A warning message." contains "warning", etc.

appengine/taskqueue-push/README.md
```diff
@@ -0,0 +1,39 @@
...
+
+Go to the site `localhost:8080` to add elements to the queue.  They will appear in the log as the result of the Enqueue servlet transmitting the data to the Worker servlet.
+
+## Deploying
+In the following command, replace YOUR-PROJECT-ID with your
+[Google Cloud Project ID](https://support.google.com/cloud/answer/6158840).
+
+    $ mvn appengine:update -Dappengine.appId=YOUR-PROJECT-ID -Dappengine.version=SOME-VERSION
+
+## Setup
```
lesv: Appears redundant.

lesv:
A few things to do.

appengine/requests/README.md
Annie29: Okay, @lesv, now that I see the whole file, I'm just echoing the form @tswast  uses.  

- Running locally (obvious)

- Deploying (obvious, but this is the form used where the user specifies the project id and version on the command line

- Setup, which is deploying, but putting the parameters in a config uses.

Annie29:
So, I think I've got branches working, so replace this pull request with [number 209](https://github.com/GoogleCloudPlatform/java-docs-samples/pull/209).

PR #211, opened 2016-04-28
Replaced queue.xml with queue.yaml and updated README.md

jsimonweb:


tswast:
LGTM. YAML configs work on App Engine standard environment now?

lesv:
As Tim said - if you tried it, and it works, then Ok to merge.

lesv:
Actually, I know they work, as only Java used xml.

lesv:
Thanks

PR #214, opened 2016-04-29
Add Datastore indexes samples.

tswast:
Samples are moved from:
https://cloud.google.com/appengine/docs/java/datastore/indexes

Note: I add a new script `test-devserver.sh` to the testing config. This
script runs the `mvn appengine:devserver` plugin, waits for it to start,
then verifies that it gets a non-error response from the `/` path. I use
this to verify that the `datastore-indexes.xml` files are correct (by
disabling autoGenerate, the local devserver throws an error when a query
is used without the correct index defined, just as production does).

We should probably add any "hello world" or other projects to this
check, as well. Anywhere we say to run `mvn appengine:devserver`, it
would be good to test that it is correct.

@shun-fan @tmatsuo FYI for testing-related script.

### DIFF
--- a/appengine/datastore/indexes-exploding/pom.xml
+++ b/appengine/datastore/indexes-exploding/pom.xml
@@ -1,0 +1,91 @@
+<!--
+Copyright 2016 Google Inc. All Rights Reserved.
+
+ Licensed under the Apache License, Version 2.0 (the "License");
+ you may not use this file except in compliance with the License.
+ You may obtain a copy of the License at
+
+       http://www.apache.org/licenses/LICENSE-2.0
+
+ Unless required by applicable law or agreed to in writing, software
+ distributed under the License is distributed on an "AS IS" BASIS,
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ See the License for the specific language governing permissions and
+ limitations under the License.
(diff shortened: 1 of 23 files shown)


tswast:
I'm looking into the Travis failure.

tswast:
There were a few problems:

- The `expect` command was not installed. Added via `apt` plugin for Travis.
- The timeout was too short. It takes much longer to run the build than it does on my workstation, so the expect command was timing out when run on Travis.
- One of the indexes was actually wrong! I think I copied one over for the `indexes-perfect` samples, but didn't test it with the actual queries. The test caught it!

appengine/datastore/indexes-exploding/pom.xml
```diff
@@ -0,0 +1,91 @@
...
+ See the License for the specific language governing permissions and
+ limitations under the License.
+-->
+<project>
+  <modelVersion>4.0.0</modelVersion>
+  <packaging>war</packaging>
+  <version>1.0-SNAPSHOT</version>
+  <groupId>com.example.appengine</groupId>
+  <artifactId>appengine-datastore-indexes-exploding</artifactId>
+  <parent>
```
lesv: Some whitespace for the future.

lesv:
LGTM

PR #217, opened 2016-04-29
Datastore Samples: Change comment format for XML license header.

tswast:
Makes it easier to filter with regex, since we can't put a region tag before
the  definition.


tswast:
@lesv FYI.

Annoying that we have to do this. I bet it's a bug START_EXCLUDE doesn't work with whole files. I can't imagine any situation where it's intended to show up.

PR #219, opened 2016-04-29
Add App Engine Cloud Datastore metadata queries samples.

tswast:
From https://cloud.google.com/appengine/docs/java/datastore/metadataqueries.

@shun-fan @lesv PTAL.


lesv:
LGTM

PR #220, opened 2016-04-29
Add App Engine Datastore stats sample.

tswast:
Stats queries aren't supported on the local server for unit testing, so
no tests. :-(

@shun-fan @lesv PTAL


lesv:
Once you get travis to pass, Ok to merge.

PR #223, opened 2016-04-29
Add AE Datastore transactions samples.

tswast:
From:
https://cloud.google.com/appengine/docs/java/datastore/transactions

@lesv @shun-fan PTAL.


lesv:
LGTM

PR #227, opened 2016-05-05
Fix broken appengine-gcs-client dependency.

tswast:
See: https://travis-ci.org/GoogleCloudPlatform/java-docs-samples/builds/127925475


tswast:
Looks like it's going through. I think there was a bad version cached, based on the error message.

tswast:
@lesv FYI

lesv:
Sigh - missed that. :(  Old Maven stuff.

lesv:
Thanks for fixing.

lesv:
How did it ever pass?

PR #232, opened 2016-05-09
Update MailjetServlet.java

carieshmarie:
Add region tags for documentation

### DIFF
--- a/managed_vms/mailjet/src/main/java/com/example/managedvms/mailjet/MailjetServlet.java
+++ b/managed_vms/mailjet/src/main/java/com/example/managedvms/mailjet/MailjetServlet.java
@@ -14,6 +14,7 @@
  * limitations under the License.
  */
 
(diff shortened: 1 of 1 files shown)


managed_vms/mailjet/src/main/java/com/example/managedvms/mailjet/MailjetServlet.java
```diff
@@ -67,3 +70,4 @@ public void doPost(HttpServletRequest req, HttpServletResponse resp) throws IOEx
     }
   }
 }
+// [END mailjet_imports]
```
tswast: Looks like you meant to say `// [END app]` here.

carieshmarie: Whoops! Corrected.

PR #235, opened 2016-05-11
Add compute mailjet sample

shun-fan:
Update maven gcloud plugin version for managed vm samples
Update master pom

### DIFF
--- a/compute/mailjet/README.md
+++ b/compute/mailjet/README.md
@@ -1,0 +1,44 @@
+# Java Mailjet Email Sample for Google Compute Engine
+
+This sample demonstrates how to use [Mailjet](https://www.mailjet.com/) on
+[Google Compute Engine](https://cloud.google.com/compute/)
+
+See the [sample application documentaion][sample-docs] for more detailed
+instructions.
+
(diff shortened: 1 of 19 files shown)


compute/mailjet/README.md
```diff
@@ -0,0 +1,44 @@
...
+## Running on Compute Engine
+
+To run the sample, you will need to do the following:
+
+1. [Create a Mailjet Account](https://app.mailjet.com/signup).
+1. Create a compute instance on the Google Cloud Platform Developer's Console
+1. SSH into the instance you created
+1. Update packages and install required packages
+
+    `sudo apt-get update && sudo apt-get install git-core openjdk-8-jdk maven`
```
lesv: Should we give them a script that does this?  So they can create their instance and set it up?

lesv:
After you get travis happy, I'm happy.  Consider the script, but ok w/o.

shun-fan:
Do we need a script for such a small amount of instructions?

lesv:
I like to as it keeps them from ssh to the instance.  It also encourages them to think doing lots of stuff.

PR #236, opened 2016-05-11
An initial commit for Firebase Event Proxy

misterwilliam:
I know I still need to write integration tests. 

Here were Les' original comments and my responses:
* Apache License should be on all files.

   * Done

* Some comments are helpful.

   * Added some comments.

* Vertical whitespace is also helpful - makes things a bit more readable.

   * Added a whitespace.

* Your up status isn't dynamic it will be coming from something other than the jetty servlet (Flex it's nginx, standard is something else)

   * In my opinion this sample doesn't really require a status. I just useful for detecting when the app is really having problems.

* HTTPS isn't required.

   * I only noticed the usage of https for https://gae-fb-proxy.firebaseio.com/. I believe this encrypts traffic between the sample app and Firebase, which is a desirable behavior.

* You do nothing to solve the issue for the user in your catches, it's ok in what I read, but we generally frown on it. 

   * ACK.


lesv:
I'm going to merge this.

A couple of things need to happen at some point (soon).
1. In your pom, please reference the parent pom in ../../pom.xml
2. Add your app to the listed apps for testing and codestyle. (in the parent pom.xml)

misterwilliam:
I've created a new PR to address those issues:
https://github.com/GoogleCloudPlatform/java-docs-samples/pull/239

Do you automatically get notified of these pull requests, or should I
continue to notify you?

lesv:
I do get notified, but if you add @lesv then there is a tag that brings it to my attention.

I'm behind about 3500 emails (mostly buganizer / git) due to CDCP.  I hope to go through them while I'm on vacation. (or I might just start clean)  I do see anything from github that mentions Java in the subject.

PR #242, opened 2016-05-17
Fix "surprising queries" samples.

tswast:
We were setting the filter twice (overwriting the first) rather than
using a composite filter.  This was causing the query results to be even
more surprising than we intending.

PTAL @halgrimur @shun-fan

### DIFF
--- a/appengine/datastore/src/test/java/com/example/appengine/QueriesTest.java
+++ b/appengine/datastore/src/test/java/com/example/appengine/QueriesTest.java
@@ -44,7 +44,6 @@ import org.junit.runners.JUnit4;
 
 import java.io.PrintWriter;
 import java.io.StringWriter;
-import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
 
@@ -685,9 +684,7 @@ public class QueriesTest {
   public void queryRestrictions_surprisingMultipleValuesAllMustMatch_returnsNoEntities()
       throws Exception {
     Entity a = new Entity("Widget", "a");
-    ArrayList<Long> xs = new ArrayList<>();
-    xs.add(1L);
-    xs.add(2L);
+    List<Long> xs = Arrays.asList(1L, 2L);
     a.setProperty("x", xs);
     datastore.put(a);
 
@@ -701,6 +698,8 @@ public class QueriesTest {
     // [END surprising_behavior_example_1]
 
     // Entity "a" will not match because no individual value matches all filters.
+    // See the documentation for more details:
+    // https://cloud.google.com/appengine/docs/java/datastore/query-restrictions#properties_with_multiple_values_can_behave_in_surprising_ways
     List<Entity> results = datastore.prepare(q).asList(FetchOptions.Builder.withDefaults());
     assertThat(results).named("query results").isEmpty();
   }
@@ -729,6 +728,9 @@ public class QueriesTest {
                     new FilterPredicate("x", FilterOperator.EQUAL, 2)));
     // [END surprising_behavior_example_2]
 
+    // Only "a" and "e" have both 1 and 2 in the "x" array-valued property.
+    // See the documentation for more details:
+    // https://cloud.google.com/appengine/docs/java/datastore/query-restrictions#properties_with_multiple_values_can_behave_in_surprising_ways
     List<Entity> results = datastore.prepare(q).asList(FetchOptions.Builder.withDefaults());
     assertThat(getKeys(results)).named("query result keys").containsExactly(a.getKey(), e.getKey());
   }
@@ -752,6 +754,9 @@ public class QueriesTest {
     Query q = new Query("Widget").setFilter(new FilterPredicate("x", FilterOperator.NOT_EQUAL, 1));
     // [END surprising_behavior_example_3]
 
+    // The query matches any entity that has a some value other than 1. Only
+    // entity "e" is not matched.  See the documentation for more details:
+    // https://cloud.google.com/appengine/docs/java/datastore/query-restrictions#properties_with_multiple_values_can_behave_in_surprising_ways
     List<Entity> results = datastore.prepare(q).asList(FetchOptions.Builder.withDefaults());
     assertThat(getKeys(results))
         .named("query result keys")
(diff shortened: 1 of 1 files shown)


lesv:
Argh - Good catch! 
We probably should ask Arie to review the whole thing? @aozarov

appengine/datastore/src/test/java/com/example/appengine/QueriesTest.java
```diff
@@ -694,15 +694,15 @@ public void queryRestrictions_surprisingMultipleValuesAllMustMatch_returnsNoEnti
     // [START surprising_behavior_example_1]
```
aozarov: Fine as is,  though a shorter alternative could be:
`List<Long> xs = Arrays.asList(1L, 2L);`

appengine/datastore/src/test/java/com/example/appengine/QueriesTest.java
```diff
@@ -716,21 +716,21 @@ public void queryRestrictions_surprisingMultipleValuesEquals_returnsMatchedEntit
...
+                    new FilterPredicate("x", FilterOperator.EQUAL, 2)));
     // [END surprising_behavior_example_2]
 
     List<Entity> results = datastore.prepare(q).asList(FetchOptions.Builder.withDefaults());
-    assertThat(getKeys(results)).named("query result keys").contains(a.getKey());
-
-    // Note: When run in the test server, this matches "c" as expected and does
-    // not match "d" as expected.  For some reason it does *not* match "b".
-    // The behavior of queries on repeated values is definitely surprising.
+    assertThat(getKeys(results)).named("query result keys").containsExactly(a.getKey(), e.getKey());
```
aozarov: Any reason you are not using the Datastore `Query.setKeysOnly` instead?

appengine/datastore/src/test/java/com/example/appengine/QueriesTest.java
```diff
@@ -770,17 +770,14 @@ public void queryRestrictions_surprisingMultipleValuesTwoNotEquals_returnsMatche
     // [START surprising_behavior_example_4]
     Query q =
         new Query("Widget")
-            .setFilter(new FilterPredicate("x", FilterOperator.NOT_EQUAL, 1))
-            .setFilter(new FilterPredicate("x", FilterOperator.NOT_EQUAL, 2));
+            .setFilter(
+                CompositeFilterOperator.and(
```
aozarov: That is a strange one that seems to ignore the comment in the Datastore queries [doc](https://cloud.google.com/appengine/docs/java/datastore/queries) saying that "A query can have no more than one NOT_EQUAL filter, and a query that has one cannot have any other inequality filters". @pcostell suggests that this is because the AE Datastore **client** breaks it
to separate queries and then join the combined results.

aozarov:
Looks good to me.

aozarov:
BTW, maybe adding this [reference](https://cloud.google.com/datastore/docs/concepts/queries#properties_with_array_values_can_behave_in_surprising_ways) in the test would help with understanding why these queries behave that way.

appengine/datastore/src/test/java/com/example/appengine/QueriesTest.java
tswast: I think the results will still be a list of Entities in that case. I want to do an equality check on the keys rather than the entity.

appengine/datastore/src/test/java/com/example/appengine/QueriesTest.java
tswast: Done.

appengine/datastore/src/test/java/com/example/appengine/QueriesTest.java
tswast: I added a link to the docs.
https://cloud.google.com/appengine/docs/java/datastore/query-restrictions#properties_with_multiple_values_can_behave_in_surprising_ways

That link describes that the query acts like
```
x < 1 OR (x > 1 AND x < 2) OR x > 2
```

tswast:
+1 on adding links to the documentation. We should probably get into the habit of doing this for any code we include in the docs.

appengine/datastore/src/test/java/com/example/appengine/QueriesTest.java
aozarov: Why/how a key-only query return you Entities and not Keys?

tswast: PreparedQuery always returns Entity. https://cloud.google.com/appengine/docs/java/javadoc/com/google/appengine/api/datastore/PreparedQuery#asList-com.google.appengine.api.datastore.FetchOptions-

I don't see any method in the PreparedQuery class that returns a iterable of Key.

aozarov: Yes, but if you set the `setKeysOnly` on the query the returned Entities will only contain Keys (and could be compared as is). I think in `gcloud-java-datastore` we did a better job and return a `List<Key>` rather than a `List<Entity>` which only holds the Keys.

tswast: Correct. I still need the `getKeys()` method. It's running against a local datastore and only getting like 3 entities, so adding `setKeysOnly()` doesn't really do much for me in this case.

aozarov: Actually because you are not using `setKeysOnly` couldn't you have done:
`assertThat(results).named("query result").containsExactly(a, e);` ?

Because this is part of user interfacing samples (I assume that even the tests are meant to be used as reference) I would highly recommend using the ideal code for accomplishing what is needed (regardless if
that actually does not make any difference in small scale or for testing). Feel free to ignore if tests are not
meant for user consumption or if you think otherwise.

tswast: Ah, you're right. `.equals()` only checks the key, so that will have the same effect. I'll send a follow-up PR for this change. I do want the tests to be as easy to understand as possible.

https://cloud.google.com/appengine/docs/java/javadoc/com/google/appengine/api/datastore/Entity#equals-java.lang.Object-

PR #243, opened 2016-05-18
ae/datastore: Remove unnecessary `getKeys()` calls.

tswast:
Also, update most queries to be `setKeysOnly()`. This is a good practice
when we don't care about the property values (more efficient queries).
And it also demonstrates that the equality matching for `Entity` objects
only checks the key values, not the property values.

In response to discussion here: https://github.com/GoogleCloudPlatform/java-docs-samples/pull/242#discussion_r63552558

@aozarov PTAL.


aozarov:
LGTM

PR #244, opened 2016-05-23
Updated firebase-event-proxy sample code to latest Firebase SDK

misterwilliam:
@lesv  Got another PR for you to look at. Thanks!

### DIFF
--- a/appengine/firebase-event-proxy/gae-firebase-event-proxy/src/main/java/com/example/GaeFirebaseEventProxy/FirebaseEventProxy.java
+++ b/appengine/firebase-event-proxy/gae-firebase-event-proxy/src/main/java/com/example/GaeFirebaseEventProxy/FirebaseEventProxy.java
@@ -18,6 +18,7 @@ package com.example.gaefirebaseeventproxy;
 
 import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.ObjectMapper;
+import com.google.appengine.api.utils.SystemProperty;
 import com.google.firebase.FirebaseApp;
 import com.google.firebase.FirebaseOptions;
 import com.google.firebase.database.DataSnapshot;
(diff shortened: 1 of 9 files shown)


appengine/firebase-event-proxy/gae-firebase-event-proxy/src/main/java/com/example/GaeFirebaseEventProxy/FirebaseEventProxy.java
```diff
@@ -73,7 +70,7 @@ public void onDataChange(DataSnapshot snapshot) {
         if (snapshot.exists()) {
           try {
             // Convert value to JSON using Jackson
-            String json = new ObjectMapper().writeValueAsString(snapshot.getValue());
+            String json = new ObjectMapper().writeValueAsString(snapshot.getValue(false));
 
             // Replace the URL with the url of your own listener app.
             URL dest = new URL("http://gae-firebase-listener-python.appspot.com/log");
```
lesv: Ok for now, but it's not a good thing to ask users to change lines deep in the middle of the code.  My preferred place for GAE is to set it in the pom, so it can be overridden in the mvn command line -Dpyurl=... you could then update a properties file from mvn, using substitution ${pyurl} and read it in the code from the properties file.  If not that, then just set it as a constant early in the source file.

lesv:
Ok to merge once travis passes.

appengine/firebase-event-proxy/gae-firebase-event-proxy/src/main/java/com/example/GaeFirebaseEventProxy/FirebaseEventProxy.java
misterwilliam: I agree that what you suggest is better style. However this code will be used in an article as sample code and I think it is better to use code that directly calls Firebase API directly even though that hurts encapsulation. I think directly calling the API makes it easier for customers to understand the Firebase API because there are fewer things for them to be thinking about.

An early draft of the article is here:
https://docs.google.com/document/d/1H_W9C8cO_laKYElzbBBgIv9ECMBA3LRJqEebqfquWAI/edit#

misterwilliam:
@lesv It looks ready. Thanks!

lesv:
Sorry for the delay - forgot you don't have merge privileges.

PR #257, opened 2016-06-17
Use CMD instead of ENTRYPOINT

aslo:
Using the ENTRYPOINT directive here would override the settings from the base docker image, breaking the configuration of the cloud debugger. Instead, CMD is the proper directive to use here to specify runtime arguments.


aslo:
Closing because git-author on original patch was set incorrectly

PR #258, opened 2016-06-20
 Used CMD instead of ENTRYPOINT to set default runtime args

aslo:
Using the ENTRYPOINT directive here would override the settings from the base docker image, breaking the configuration of the cloud debugger. Instead, CMD is the proper directive to use here to specify runtime arguments.


lesv:
LGTM

lesv:
@saturnism FYI

aslo:
@lesv 👍

PR #260, opened 2016-06-28
update java repo tools

tswast:
Adds [versions plugin](http://www.mojohaus.org/versions-maven-plugin/) so that we can automatically update dependencies.

@lesv @shun-fan PTAL


lesv:
LGTM

PR #261, opened 2016-06-28
App Engine standard env uses Java Servlets 2.5.

tswast:
According to:
https://cloud.google.com/appengine/docs/java/runtime#Java_Introduction

"App Engine uses the Java Servlet 2.5 standard for web applications."

This updates these samples to use version 2.5. Annotations are not
supported. I see that the SendGrid servlet already has a web.xml
defined, so the annotation were redundant, anyway (and not functional).


lesv:
LGTM

PR #267, opened 2016-07-06
Speech api changes

puneith:
@tswast Please review. This PR contains the changes due to Speech API changes, namely Sync, Async and Streaming.


tswast:
@lesv PTAL as well.

speech/grpc/src/main/java/com/google/cloud/speech/grpc/demos/AsyncRecognizeClient.java
```diff
@@ -0,0 +1,248 @@
...
+// The results are received via the google.longrunning.Operations interface.
+//
+// Uses a service account for OAuth2 authentication, which you may obtain at
+// https://console.developers.google.com
+// API Manager > Google Cloud Speech API > Enable
+// API Manager > Credentials > Create credentials > Service account key > New service account.
+//
+// Then set environment variable GOOGLE_APPLICATION_CREDENTIALS to the full path of that file.
+
+package com.google.cloud.speech.grpc.demos;
```
tswast: For new samples we have been using `com.example` package instead of `com.google`. This is to make it more clear to the user that they can change the package to match their own project.

speech/grpc/src/main/java/com/google/cloud/speech/grpc/demos/AsyncRecognizeClient.java
```diff
@@ -0,0 +1,248 @@
...
+  public AsyncRecognizeClient(String host, int port, URI input, int samplingRate)
+      throws IOException {
+    this.host = host;
+    this.port = port;
+    this.input = input;
+    this.samplingRate = samplingRate;
+
+    GoogleCredentials creds = GoogleCredentials.getApplicationDefault();
+    creds = creds.createScoped(OAUTH2_SCOPES);
+    channel = NettyChannelBuilder.forAddress(host, port)
```
tswast: Ideally the channel would be injected into the constructor so that we could use an InProcess transport for local unit tests. https://github.com/grpc/grpc-java#transport

I understand that we aren't provided with a fake implementation, so this would be a lot of work to actually make such a unit test, though.

I'd at least pull out a static function `createChannel()` that returns the channel so we can at least simulate dependency injection by calling that function rather than passing it into the constructor, reducing the work we do in the constructor itself.

speech/grpc/src/main/java/com/google/cloud/speech/grpc/demos/AsyncRecognizeClient.java
```diff
@@ -0,0 +1,248 @@
...
+        .negotiationType(NegotiationType.TLS)
+        .intercept(new ClientAuthInterceptor(creds, Executors.newSingleThreadExecutor()))
+        .build();
+    stub = SpeechGrpc.newBlockingStub(channel);
+    statusStub = OperationsGrpc.newBlockingStub(channel);
+
+    logger.info("Created stub for " + host + ":" + port);
+  }
+
+  private RecognitionAudio createRecognitionAudio() throws IOException {
```
tswast: This function seems unnecessary. I'd rather see the RecognitionAudioFactory call whereever this is used.

speech/grpc/src/main/java/com/google/cloud/speech/grpc/demos/AsyncRecognizeClient.java
```diff
@@ -0,0 +1,248 @@
...
+        .setConfig(config)
+        .setAudio(audio)
+        .build();
+
+    Operation operation;
+    Operation status;
+    try {
+      operation = stub.asyncRecognize(request);
+
+      //Print the long running operation handle
```
tswast: [nit] Space before Print after `//`.

speech/grpc/src/main/java/com/google/cloud/speech/grpc/demos/AsyncRecognizeClient.java
```diff
@@ -0,0 +1,248 @@
+/*
```
tswast: Please run google-java-format against this file: https://github.com/google/google-java-format

I'm finding it slightly difficult to read.

speech/grpc/src/main/java/com/google/cloud/speech/grpc/demos/AsyncRecognizeClient.java
```diff
@@ -0,0 +1,248 @@
...
+  }
+
+  public static void main(String[] args) throws Exception {
+
+    String audioFile = "";
+    String host = "speech.googleapis.com";
+    Integer port = 443;
+    Integer sampling = 16000;
+
+    CommandLineParser parser = new DefaultParser();
```
tswast: To keep things simpler, Java samples usually use "properties" for these. See the bookshelf sample: https://github.com/GoogleCloudPlatform/getting-started-java/blob/master/bookshelf/4-auth/pom.xml#L28

Hmm... Seems like that may be for servers, only, though... https://github.com/GoogleCloudPlatform/getting-started-java/blob/48bfbe9530b313ffec17b32edf959a38010bae3a/bookshelf/4-auth/src/main/java/com/example/managedvms/gettingstartedjava/basicactions/ListBookServlet.java#L50

lesv:
For the README.md - is the whole SERVICE ACCOUNT stuff still required?

speech/grpc/README.md
```diff
@@ -73,20 +73,35 @@ note that the audio file must be in RAW format.  You can use `sox`
 (available, e.g. via [http://sox.sourceforge.net/](http://sox.sourceforge.net/)
 or [homebrew](http://brew.sh/)) to convert audio files to raw format.
 
-### Run the non-streaming client
+### Run the sync client
```
lesv: How does it work on Windows?

speech/grpc/bin/speech-sample-async.sh
```diff
@@ -15,4 +15,4 @@
 
 SRC_DIR=$(cd "$(dirname "$0")/.."; pwd)
 java -cp ${SRC_DIR}/target/grpc-sample-1.0-jar-with-dependencies.jar \
-    com.google.cloud.speech.grpc.demos.NonStreamingRecognizeClient "$@"
+    com.google.cloud.speech.grpc.demos.AsyncRecognizeClient "$@"
```
lesv: `com.examples.cloud.speech.AsyncRecognizeClient` might be better

speech/grpc/bin/speech-sample-async.sh
```diff
@@ -15,4 +15,4 @@
 
 SRC_DIR=$(cd "$(dirname "$0")/.."; pwd)
 java -cp ${SRC_DIR}/target/grpc-sample-1.0-jar-with-dependencies.jar \
```
lesv: is SRC_DIR necessary?

`java -cp target/grpc-sample-1.0-jar-with-dependencies.jar \` perhaps?

Or even:
mvn exec:java -PAsyncRecognizeClient

speech/grpc/bin/speech-sample-streaming.sh
```diff
@@ -15,4 +15,4 @@
 
 SRC_DIR=$(cd "$(dirname "$0")/.."; pwd)
 java -cp ${SRC_DIR}/target/grpc-sample-1.0-jar-with-dependencies.jar \
-    com.google.cloud.speech.grpc.demos.RecognizeClient "$@"
+    com.google.cloud.speech.grpc.demos.StreamingRecognizeClient "$@"
```
lesv: ditto for prior comments

speech/grpc/src/main/java/com/google/cloud/speech/grpc/demos/AsyncRecognizeClient.java
```diff
@@ -0,0 +1,248 @@
...
+
+// Client that sends audio to Speech.AsyncRecognize via gRPC and returns longrunning operation.
+// The results are received via the google.longrunning.Operations interface.
+//
+// Uses a service account for OAuth2 authentication, which you may obtain at
+// https://console.developers.google.com
+// API Manager > Google Cloud Speech API > Enable
+// API Manager > Credentials > Create credentials > Service account key > New service account.
+//
+// Then set environment variable GOOGLE_APPLICATION_CREDENTIALS to the full path of that file.
```
lesv: Is this necessary?

speech/grpc/src/main/java/com/google/cloud/speech/grpc/demos/AsyncRecognizeClient.java
lesv: +1

speech/grpc/src/main/java/com/google/cloud/speech/grpc/demos/AsyncRecognizeClient.java
```diff
@@ -0,0 +1,248 @@
...
+
+  private RecognitionAudio createRecognitionAudio() throws IOException {
+    return RecognitionAudioFactory.createRecognitionAudio(this.input);
+  }
+
+  public void shutdown() throws InterruptedException {
+    channel.shutdown().awaitTermination(5, TimeUnit.SECONDS);
+  }
+
+  /** Send an async-recognize request to server. */
```
lesv: Is this comment useful?  Should function name be asyncRecognize() ?

speech/grpc/src/main/java/com/google/cloud/speech/grpc/demos/AsyncRecognizeClient.java
```diff
@@ -0,0 +1,248 @@
...
+  private static final List<String> OAUTH2_SCOPES =
+      Arrays.asList("https://www.googleapis.com/auth/cloud-platform");
+
+  private final String host;
+  private final int port;
+  private final URI input;
+  private final int samplingRate;
+
+  private final ManagedChannel channel;
+  private final SpeechGrpc.SpeechBlockingStub stub;
```
lesv: Is `stub` a term of art for speech recognition that I'm unaware of?  Why stub?

speech/grpc/src/main/java/com/google/cloud/speech/grpc/demos/AsyncRecognizeClient.java
```diff
@@ -0,0 +1,248 @@
...
+  private final OperationsGrpc.OperationsBlockingStub statusStub;
```
lesv: same Q here?

speech/grpc/src/main/java/com/google/cloud/speech/grpc/demos/AsyncRecognizeClient.java
lesv: If it's in this [list](https://en.wikipedia.org/wiki/Stub) you should explain why a `stub` is part of our code.

speech/grpc/src/main/java/com/google/cloud/speech/grpc/demos/AsyncRecognizeClient.java
```diff
@@ -0,0 +1,248 @@
...
+        .withDescription("endpoint for api, e.g. speech.googleapis.com")
+        .hasArg()
+        .withArgName("ENDPOINT")
+        .create());
+    options.addOption(OptionBuilder.withLongOpt("port")
+        .withDescription("SSL port, usually 443")
+        .hasArg()
+        .withArgName("PORT")
+        .create());
+    options.addOption(OptionBuilder.withLongOpt("sampling")
```
lesv: It seems like FILE_PATH and RATE are the only two of these most users will be interested in, why are the others here?

speech/grpc/src/main/java/com/google/cloud/speech/grpc/demos/AsyncRecognizeClient.java
lesv: He's doing a jar-with-dependencies, so he can use a properties file in resources/

speech/grpc/src/main/java/com/google/cloud/speech/grpc/demos/StreamingRecognizeClient.java
```diff
@@ -26,12 +26,12 @@
 package com.google.cloud.speech.grpc.demos;
 
 import com.google.auth.oauth2.GoogleCredentials;
-import com.google.cloud.speech.v1.AudioRequest;
-import com.google.cloud.speech.v1.InitialRecognizeRequest;
-import com.google.cloud.speech.v1.InitialRecognizeRequest.AudioEncoding;
-import com.google.cloud.speech.v1.RecognizeRequest;
-import com.google.cloud.speech.v1.RecognizeResponse;
-import com.google.cloud.speech.v1.SpeechGrpc;
+import com.google.cloud.speech.v1beta1.RecognitionConfig;
```
lesv: why are we moving from v1 to v1beta1 ?

speech/grpc/src/main/proto/google/longrunning/operations.proto
```diff
@@ -0,0 +1,144 @@
+// Copyright (c) 2015, Google Inc.
```
lesv: Does this really need to be here?  Shouldn't it be part of the gcloud or some other api repo.

If you are going to have these, then you should put them in third_party/ as they are copies of published code.  You should talk w/ @tswast about git magic for the include as well.

speech/grpc/src/main/proto/google/speech/v1beta1/cloud_speech.proto
```diff
@@ -0,0 +1,321 @@
+syntax = "proto3";
```
lesv: Shouldn't be here, and needs copyright / license.

speech/grpc/src/test/java/com/google/cloud/speech/grpc/demos/RecognitionAudioFactoryTest.java
```diff
@@ -50,7 +50,7 @@ public void verifyBytesInSizeFromGoogleStorageFile() throws IOException {
     String audioUri = "gs://cloud-samples-tests/speech/audio.raw";
 
     URI uri = URI.create(audioUri);
-    AudioRequest audio = AudioRequestFactory.createRequest(uri);
+    RecognitionAudio audio = RecognitionAudioFactory.createRecognitionAudio(uri);
 
     int numBytes = audio.getContent().toByteArray().length;
```
lesv: seems a bit light here.

lesv:
@puneith PTAL, hook your code into the testing infrastructure which will run the google style checker.

speech/grpc/src/main/java/com/google/cloud/speech/grpc/demos/AsyncRecognizeClient.java
tswast: Stub is a gRPC term (inherited from stubby, it's predecessor). Client would be a better name, probably.

speech/grpc/src/main/java/com/google/cloud/speech/grpc/demos/AsyncRecognizeClient.java
puneith: This structure already existed. I didn't mess up with the package name. Change it still?

speech/grpc/src/main/java/com/google/cloud/speech/grpc/demos/AsyncRecognizeClient.java
puneith: I agree with you on the DIP.  I will make the change. The code you see here has mostly been copied from @tswast RecogizeClient which was qualified and then changes made :)

speech/grpc/src/main/java/com/google/cloud/speech/grpc/demos/AsyncRecognizeClient.java
lesv: Sorry about pinging on the package name, it looks like whom ever reviewed the first version wasn't on the Java team. :(

speech/grpc/src/main/java/com/google/cloud/speech/grpc/demos/AsyncRecognizeClient.java
lesv: Why are we using them?  Shouldn't the name be clearer for what we expose to users?

lesv:
@puneith @tmatsuo @tswast I just had a conversation with Susan and @elibixby about this, and we are of the opinion that you really need a generated client library rather than what you have on these samples.  The protos should really be retrieved from [GoogleAPIs](https://github.com/googleapis) and not be located in the sample.  We are suggesting that you withdraw this PR / branch until the situation has been fixed.  If you'd like to proceed, you should have a GVC with me and @tmatsuo next week.

speech/grpc/src/main/java/com/google/cloud/speech/grpc/demos/AsyncRecognizeClient.java
puneith: @lesv @tswast Its common to put company name in the package even samples. This is google code so why not? https://github.com/linkedin/api-get-started/tree/master/java/src/com/linkedin/sample

speech/grpc/bin/speech-sample-async.sh
puneith: @lesv same comment as above. Its google code so why not :)

speech/grpc/src/main/java/com/google/cloud/speech/grpc/demos/AsyncRecognizeClient.java
puneith: @lesv Not necessary. But, some developers trip over our documentation. You suggest removing?

speech/grpc/src/main/java/com/google/cloud/speech/grpc/demos/StreamingRecognizeClient.java
puneith: Because the speech API had some fundamental changes in the API so v1beta1.

speech/grpc/src/main/proto/google/longrunning/operations.proto
puneith: This has been resolved separately. @lesv please confirm if still any reservations.

speech/grpc/src/main/proto/google/speech/v1beta1/cloud_speech.proto
puneith: @lesv same comment.

speech/grpc/src/test/java/com/google/cloud/speech/grpc/demos/RecognitionAudioFactoryTest.java
puneith: @lesv I don't understand the comment.

speech/grpc/bin/speech-sample-async.sh
puneith: This allows you to run the code from anywhere. Otherwise, you have to be in the correct directory to run the code. This is better!

speech/grpc/README.md
puneith: Don't know, is that a blocker?

puneith: @lesv we don't have the instructions anywhere otherwise and assuming users would have read everything before using this sample is not correct either. So we need them IMHO.

speech/grpc/bin/speech-sample-async.sh
lesv: For the user it will become their code.  It would be good not to have code they started from marked as com.google.*,  We settled on the usage from RFC 2606 after much discussion.

speech/grpc/src/main/java/com/google/cloud/speech/grpc/demos/AsyncRecognizeClient.java
lesv: Yes

speech/grpc/src/main/java/com/google/cloud/speech/grpc/demos/AsyncRecognizeClient.java
lesv: See RFC 2606

speech/grpc/src/main/proto/google/longrunning/operations.proto
lesv: This isn't supposed to be in this REPO, and if it is, it shouldn't be in this folder path.

AFAIK, the resolution is not to have this code published.

speech/grpc/src/main/proto/google/speech/v1beta1/cloud_speech.proto
lesv: This definitely shouldn't be in this REPO and it should have a LICENSE.

speech/grpc/src/test/java/com/google/cloud/speech/grpc/demos/RecognitionAudioFactoryTest.java
lesv: It isn't much of a test.

puneith: Its some test better than nothing :)

speech/grpc/README.md
lesv: Please add an issue for this.

speech/grpc/README.md
```diff
@@ -73,20 +73,35 @@ note that the audio file must be in RAW format.  You can use `sox`
 (available, e.g. via [http://sox.sourceforge.net/](http://sox.sourceforge.net/)
```
lesv: README should mention that this is BETA software w/ a link to what that means: https://cloud.google.com/terms/launch-stages

It should also mention that this samples is for Advanced Users.

speech/grpc/README.md
puneith: Issue for windows or that we don't have instructions anywhere else?

lesv:
LGTM

lesv:
There should be some additional items for catchup after merge.

PR #274, opened 2016-07-14
Auto-update dependencies. [DO NOT MERGE]

tswast:
Every one of these updates is broken due to some breaking change in an API. Just making a PR out of it so that I we can track more easily and manually update and fix these samples.


tswast:
Closing this. I just added some bugs for the samples we need to manually update. I added the packages with breaking changes to the versions ignore list.

PR #275, opened 2016-07-19
Add Cloud Natural Language API Java sample.

jerjou:
This sample shows how to use the Cloud Natural Language API to do entity
recognition. The client libraries are vendored in while we are in alpha.

### DIFF
--- a/language/README.md
+++ b/language/README.md
@@ -1,0 +1,39 @@
+# Google Cloud Natural Language API Samples
+
+These samples demonstrate the use of the [Google Cloud Natural Language API][NL-Docs].
+
+[NL-Docs]: https://cloud.google.com/language/docs/
+
+## Prerequisites
+
+### Download Maven
+
+This sample uses the [Apache Maven][maven] build system. Before getting started, be
+sure to [download][maven-download] and [install][maven-install] it. When you use
+Maven as described here, it will automatically download the needed client
+libraries.
+
+[maven]: https://maven.apache.org
+[maven-download]: https://maven.apache.org/download.cgi
+[maven-install]: https://maven.apache.org/install.html
+
+### Set Up to Authenticate With Your Project's Credentials
+
+Please follow the [Set Up Your Project](https://cloud.google.com/natural-language/docs/getting-started#set_up_your_project)
+steps in the Quickstart doc to create a project and enable the
+Cloud Natural Language API. Following those steps, make sure that you
+[Set Up a Service Account](https://cloud.google.com/natural-language/docs/common/auth#set_up_a_service_account),
+and export the following environment variable:
+
+```
+export GOOGLE_APPLICATION_CREDENTIALS=/path/to/your-project-credentials.json
+```
+
+[cloud-console]: https://console.cloud.google.com
+[language-api]: https://console.cloud.google.com/apis/api/language.googleapis.com/overview?project=_
+[adc]: https://cloud.google.com/docs/authentication#developer_workflow
+
+## Samples
+
+- [Analyze](analysis) is a command line tool to show case the features of the API.
+
--- a/language/analysis/pom.xml
+++ b/language/analysis/pom.xml
@@ -1,0 +1,93 @@
+<!--
+Copyright 2016 Google Inc. All Rights Reserved.
+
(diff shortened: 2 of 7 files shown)


language/README.md
```diff
@@ -0,0 +1,39 @@
...
+[maven]: https://maven.apache.org
+[maven-download]: https://maven.apache.org/download.cgi
+[maven-install]: https://maven.apache.org/install.html
+
+### Set Up to Authenticate With Your Project's Credentials
+
+Please follow the [Set Up Your Project](https://cloud.google.com/natural-language/docs/getting-started#set_up_your_project)
+steps in the Quickstart doc to create a project and enable the
+Cloud Natural Language API. Following those steps, make sure that you
+[Set Up a Service Account](https://cloud.google.com/natural-language/docs/common/auth#set_up_a_service_account),
```
lesv: This really isn't necessary in most instances.  But I won't block on it.

language/analysis/pom.xml
```diff
@@ -0,0 +1,99 @@
...
+      <groupId>junit</groupId>
+      <artifactId>junit</artifactId>
+      <version>4.12</version>
+    </dependency>
+    <dependency>
+      <groupId>com.google.truth</groupId>
+      <artifactId>truth</artifactId>
+      <version>0.28</version>
+    </dependency>
+    <dependency>
```
lesv: Not really required for your usage.

lesv:
@tswast Can you enable the language api's for whatever project we run tests with?
LGTM

jerjou:
Removed unnecessary test dependency.
Also added NL api to the test project.

language/README.md
jerjou: I dunno - I feel like it's the most failsafe way to do things. The other methods of getting credentials have too many caveats attached to them :-/

lesv: I'm a huge fan of gcloud for credentials and DefaultAuth.

language/analysis/pom.xml
```diff
@@ -0,0 +1,94 @@
...
+      <groupId>junit</groupId>
+      <artifactId>junit</artifactId>
+      <version>4.12</version>
+    </dependency>
+    <dependency>
+      <!-- For checking HTTP response codes. -->
+      <groupId>javax.servlet</groupId>
+      <artifactId>javax.servlet-api</artifactId>
+      <version>3.1.0</version>
```
lesv: This isn't a servlet app, so it's not required here.

lesv:
LGTM - assuming things pass.

jerjou:
Whoops. Removed the wrong dependency -_-

lesv:
LGTM once tests pass.

language/README.md
jerjou: Here my (possibly outdated) understanding:

`gcloud`
  * `auth login` which works sometimes but fails subtly because it uses user auth on the cloudsdktool project instead of a service account that's attached to the project. Also you'll have to swap this out with a service account anyway once you go to production, if you're not on GCE/GAE.
  * `auth activate-service-account` which requires downloading a service account anyway, and what happens in your production scripts if you're not on GCE/GAE? I suppose you could have your machine run the `gcloud` command as part of setup? Seems like more overhead than setting an env variable..
  * In general, requires installing `gcloud` on every machine you run the code on (unless you're on GAE/GCE)

I assume by DefaultAuth you mean the default service account on GAE/GCE? Which is fine, unless you're not on GAE/GCE, in which case we have to mention that and explain what to do in that case, so it doesn't save any explanation space (though yes - if you _are_ running on GAE/GCE, you can just skip that paragraph).

Point being - it's fine, but there's a lot of explaining and decision points around whether you're on GAE/GCE, that folks probably don't care about when they're looking at a sample like this, so it's simpler to just have one not-too-complex explanation that works in all instances :)

lesv: `gcloud beta auth application-default` is the latest hotness.

waprin: Image of Yaktocat

waprin: Unfortunately doesn't seem to work :(

waprin: Should let Vijay know.

PR #278, opened 2016-07-28
Fix mvm cloud sql sample for local run

shun-fan:


lesv:
LGTM - please wait for Travis

PR #279, opened 2016-07-28
Fix the sleep logic for streaming as per sample rate

puneith:
@jerjou @tswast Please review. Changed the sleep time to absolute 100ms because for 32KKHz the Speech API was complaining about 

message: "Audio data is being streamed too slow. Please stream audio data approximately at real time."


speech/grpc/src/main/java/com/examples/cloud/speech/StreamingRecognizeClient.java
```diff
@@ -138,8 +138,8 @@ public void onCompleted() {
                 .build();
         requestObserver.onNext(request);
         // To simulate real-time audio, sleep after sending each audio buffer.
-        // For 16000 Hz sample rate, sleep 100 milliseconds.
-        Thread.sleep(samplingRate / 160);
+        // Sleep 100ms
+        Thread.sleep(100);
```
jerjou: Approximating realtime is dependent on the `samplingRate`, so you can't hard-code 100ms unless you're also hard-coding the `samplingRate`.

The calculation here should instead be dependent on `samplingRate`, buffer size, and encoding. Since encoding is hard-coded to 2 bytes per sample, you should sleep for:

    int BYTES_PER_SAMPLE = 2;  // because LINEAR16 - ie 16 bits per sample, at 8 bits per byte
    int BYTES_PER_BUFFER = 3200;
    int SAMPLES_PER_BUFFER = BYTES_PER_BUFFER / BYTES_PER_SAMPLE; // 1600

    int samplesPerMillisecond = samplingRate / 1000;
    int millisecondsPerBuffer = SAMPLES_PER_BUFFER / samplesPerMillisecond;
    Thread.sleep(millisecondsPerBuffer);

which for 16000 Hz is `1600 / 16 = 100` (as expected), and for 32000 Hz is `1600 / 32 = 50` (which is faster than the original `32000 / 160 = 200`). I suspect there was just an error in the original calculation. (feel free to check my math as well - but at least it gives a shorter sleep duration, which is what you'd expect if you increase the sample rate while keeping the buffer size the same)

puneith:
@jerjou PTAL

jerjou:
Cool.
Can you add a test with a non-16kHz audio file?

puneith:
@jerjou PTAL

speech/grpc/src/test/java/com/examples/cloud/speech/StreamingRecognizeClientTest.java
```diff
@@ -52,11 +52,13 @@
  */
 @RunWith(JUnit4.class)
 public class StreamingRecognizeClientTest {
-  private TestAppender appender;
+  private Writer writer;
+  private WriterAppender appender;
 
   @Before
```
jerjou: Should this be a `@BeforeClass`, since it's adding an appender (rather than replacing it)? ie if this ones once per test, won't new appenders get added as many times as there are tests?

speech/grpc/src/main/java/com/examples/cloud/speech/StreamingRecognizeClient.java
```diff
@@ -76,6 +81,10 @@ public StreamingRecognizeClient(ManagedChannel channel, String file, int samplin
     this.channel = channel;
 
     speechClient = SpeechGrpc.newStub(channel);
+
+    //Send log4j logs to Console
```
jerjou: Shouldn't this be configured in a configuration file, rather than in code? eg if a user wanted to log to a file instead of stdout..

jerjou:
Travis was unhappy  because the api wasn't enabled in the test project. I enabled it and kicked off the build again (I think?).

Thanks for adding the tests! Didn't realize there weren't already tests -_-;

Anyway - looks good from my end - just a couple minor comments. Can you get someone from the java team to sign off on it?

speech/grpc/src/main/java/com/examples/cloud/speech/StreamingRecognizeClient.java
puneith: It generally is, but I thought for the purpose of sample was bit of an overkill. But I can change if folks have other thoughts.

speech/grpc/src/test/java/com/examples/cloud/speech/StreamingRecognizeClientTest.java
puneith: I have to do it for each test so that StringWriter is new. There is an After which I now think should be AfterClass where all added appenders will be removed.

speech/grpc/src/main/java/com/examples/cloud/speech/StreamingRecognizeClient.java
jerjou: I think letting the logger log to the default destination is fine for the sample. If there's a reason that messages should go to stdout, then it should print to stdout rather than logging.

Semantics are important for samples. If you're logging something, then it should be something that shows up in logs. If you're printing something, then it should be communicating with the user. Sending logs to stdout sends sort of mixed messages, IMO.

puneith:
@tswast @lesv PTAL

speech/grpc/src/test/java/com/examples/cloud/speech/StreamingRecognizeClientTest.java
jerjou: Oh I see. No - having it in an After makes sense. Sorry - didn't see the After.

speech/grpc/src/main/java/com/examples/cloud/speech/StreamingRecognizeClient.java
puneith: I think the problem here is different. Which is we are inspecting the logs to evaluate the tests, which is not a generally we should do. Its because of this reason. Now, we shouldn't do System.out.println in the Streaming sample and therefore we have to do this trick in the JUnit Tests. One way I think of fixing this is by having the recognize method return a Stream so tha JUnit inspects the Stream and now you see its getting bit involved :)

jerjou: Oh interesting...
So this script doesn't even display any output, it just logs the response... -_-; This strikes me as a bit weird, but I realize it's tangential to this CL, so I won't block on it. I do like your idea about returning a Stream, but yeah. Another time :-)

jerjou:
Oh - would you mind updating the title to the PR?

speech/grpc/src/main/java/com/examples/cloud/speech/StreamingRecognizeClient.java
lesv: Usually, we just configure this with logging.properties, you could comment.  If you are going to run this on GCE, you might wish to integrate w/ gcloud-java's [logging](https://github.com/GoogleCloudPlatform/gcloud-java/blob/master/README.md#stackdriver-logging-alpha).

lesv:
Minor comment, consider it, change if you wish, then LGTM

lesv:
LGTM - once travis passes.

PR #282, opened 2016-08-11
New Plugins

lesv:


lesv:
@tswast @shun-fan Could one of you review?

appengine/helloworld-new-plugins/README.md
```diff
@@ -0,0 +1,59 @@
...
+* `gcloud beta auth application-default login`
+
+## Maven
+### Running locally
+
+    $ mvn appengine:run
+  
+### Deploying
+
+    $ mvn appengine:deploy
```
tswast: Not a problem with your code, but it's annoying that they used the same name for the plugin. That has a lot of potential for confusion as to which plugin someone is using.

appengine/helloworld-new-plugins/build.gradle
```diff
@@ -0,0 +1,52 @@
...
+  jcenter()
+  mavenCentral()
+}
+
+apply plugin: 'java'
+apply plugin: 'war'
+apply plugin: 'com.google.cloud.tools.appengine'
+
+dependencies { // Servlet for App Engine Standard
+//  providedCompile group: 'javax.servlet', name: 'servlet-api', version:'2.5'
```
tswast: I guess they uncomment this line if they are not using `beta-quickstart`? Maybe add a comment explaining that only one of these lines should be uncommented and when.

appengine/helloworld-new-plugins/gradlew
```diff
@@ -0,0 +1,164 @@
+#!/usr/bin/env bash
+
+##############################################################################
```
tswast: Auto-generated? In that case, maybe add to `.gitignore`. If not auto-generated, add a license header.

appengine/helloworld-new-plugins/gradlew.bat
```diff
@@ -0,0 +1,90 @@
+@if "%DEBUG%" == "" @echo off
```
tswast: Ditto. License header if feasible, otherwise if this is auto-generated, consider removing and add to .gitignore

Looks like at least one person does add these to .gitignore. http://www.w3ma.com/android-studio-gitignore-avoid-useless-files/

appengine/helloworld-new-plugins/pom.xml
```diff
@@ -0,0 +1,74 @@
...
+ See the License for the specific language governing permissions and
+ limitations under the License.
+-->
+<!-- [START pom] -->
+<project>
+  <modelVersion>4.0.0</modelVersion>
+  <packaging>war</packaging>
+  <version>1.0-SNAPSHOT</version>
+  <groupId>com.example.appengine</groupId>
+  <artifactId>appengine-helloworld</artifactId>
```
tswast: Give a new name to disambiguate from the other helloworld(s) in the test output.

tswast:
@lesv A few comments pertaining to the Gradle files. LGTM once fixed.

tswast:
We should add tests to run locally using both plugins. We'll need to extend `./java-repo-tools/scripts/test-localhost.sh` slightly. Also call the new tests for the local server plugins to `./travis.sh`.

appengine/helloworld-new-plugins/gradlew
lesv: Auto-generated, but needed.  This provides the gradle magic that lets you build stuff without the correct tooling on your machine.

lesv:
I'm going to merge this, then disable spark-java testing for a bit.

PR #283, opened 2016-08-12
Stop hardcoding Netty dependency so example works on Android

louiscryan:
Using ManagedChannelBuilder allows GRPC to auto-detect the best transport for the runtime environment on which this code will run.

### DIFF
--- a/speech/grpc/src/main/java/com/examples/cloud/speech/AsyncRecognizeClient.java
+++ b/speech/grpc/src/main/java/com/examples/cloud/speech/AsyncRecognizeClient.java
@@ -88,7 +88,7 @@ public class AsyncRecognizeClient {
     GoogleCredentials creds = GoogleCredentials.getApplicationDefault();
     creds = creds.createScoped(OAUTH2_SCOPES);
     ManagedChannel channel =
-        NettyChannelBuilder.forAddress(host, port)
+        ManagedChannelBuilder.forAddress(host, port)
             .negotiationType(NegotiationType.TLS)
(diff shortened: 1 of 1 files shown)


lesv:
@louiscryan Your change failed integration tests.  Not sure why the CLA bot doesn't like you -- I know who you are.

@puneith FYI

louiscryan:
Sorry. I was being a bit lazy with my PR. I just wanted to show that they Netty dependency shouldn't be hardcoded so that the example would work happily on Android and elsewhere using GRPCs environment sensing.

I suspect this problem may be present in other sample code throughout this repo.

The compile failure is a simple import fix. Can you pick this change up and work through the other examples to see if they have the same issue?

I've definitely signed the CLA

puneith:
Can you check your email address in your commit messages.

louiscryan:
Fixed CLA and import (hopefully)

lesv:
LGTM once travis passes.

lesv:
@louiscryan Didn't compile.
@puneith Can you look at fixing for louis?

puneith:
@louiscryan I don't see .negotiationType(NegotiationType.TLS) in the API. Where is that coming from?

Also make sure to use grpc-all 0.14.1 in the pom.

lesv:
@louiscryan PING

louiscryan:
No need to call .negotiationType(TLS) it's the default.

Please update all the samples to use grpc 1.0 now that its released

tswast:
@louiscryan I've updated the sample to grpc 1.0 and merged your changes into https://github.com/GoogleCloudPlatform/java-docs-samples/pull/326. Closing this PR.

PR #286, opened 2016-08-12
Update dependencies for Speech sample.

tswast:
Also, fix shell-check warnings in scripts.

Switches to the auth library recommended in the grpc documentation.
There was an old version of Guava being included from this, which I had
to exclude from the pom.

FYI @puneith 

@lesv Please take a look.

### DIFF
--- a/speech/grpc/bin/speech-sample-async.sh
+++ b/speech/grpc/bin/speech-sample-async.sh
@@ -13,6 +13,6 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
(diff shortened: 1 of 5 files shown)


lesv:
LGTM - I hate that we need to do an exclude.  Have bugs been reported?

tswast:
@lesv I didn't report a bug. I know Dataflow has a similar exclude https://github.com/GoogleCloudPlatform/DataflowJavaSDK-examples/blob/master/pom.xml#L221 which is how I was able to track down the weird errors I was getting without this.

PR #292, opened 2016-08-15
Don't recommend running the tests at package time.

tswast:
Users probably don't want to run all the system tests to try out the BigQuery samples.


lesv:
LGTM - curious why yours passed.

PR #293, opened 2016-08-15
Disable testing for a bit

lesv:
Sparkjava has been flaky, disable testing, while I figure out why.


lesv:
@tswast FYI

lesv:
Merging - temporary.

PR #294, opened 2016-08-16
Changes property name

WalterHub:
Uses 'category' instead of 'type' as a property name. 'type' could cause confusion with data type.


lesv:
LGTM

PR #295, opened 2016-08-16
Remove output only fields

waprin:


lesv:
@tswast @waprin  BigQuery failing for some reason here?

tswast:
```com.google.api.client.googleapis.json.GoogleJsonResponseException: 503 Service Unavailable``` Sounds to me like a flake. I'll hit retry on Travis.

I'm actually refactoring the BigQuery tests right now. I'll see if I need to add any retry logic while I'm at it.

PR #297, opened 2016-08-16
Refactor BigQuery samples.

tswast:
Fixes the following issues:

- Subclassing a static utility class unncecessarily.
- Parsing constants from a JSON file for every test instead of defining
  actual constants.
- Rename Bigquery to BigQuery.
- Formatted using google-java-format
  https://github.com/google/google-java-format

@lesv PTAL


bigquery/src/main/java/com/google/cloud/bigquery/samples/AsyncQuerySample.java
```diff
@@ -41,23 +38,20 @@
    * @throws IOException IOException
    * @throws InterruptedException InterruptedException
    */
-  public static void main(final String[] args)
-      throws IOException, InterruptedException {
+  public static void main(final String[] args) throws IOException, InterruptedException {
     Scanner scanner = new Scanner(System.in);
```
lesv: Some reason these aren't command line args, with defaults?

bigquery/src/main/java/com/google/cloud/bigquery/samples/BigqueryServiceFactory.java
```diff
@@ -30,16 +30,14 @@
 import java.util.Collection;
 
 /**
- * This class creates our Service to connect to Bigquery including auth.
+ * This class creates our Service to connect to BigQuery including auth.
  */
-public final class BigqueryServiceFactory {
+public final class BigQueryServiceFactory {
```
lesv: Should the file name be changed?

bigquery/src/main/java/com/google/cloud/bigquery/samples/ExportDataCloudStorageSample.java
```diff
@@ -42,20 +42,17 @@ protected ExportDataCloudStorageSample() {
    * @throws InterruptedException Should never be thrown.
    */
   // [START main]
-  public static void main(final String[] args)
-      throws IOException, InterruptedException {
+  public static void main(final String[] args) throws IOException, InterruptedException {
     Scanner scanner = new Scanner(System.in);
```
lesv: I'm sure there is a good reason why we don't use command line args, w/ reasonable defaults.

bigquery/src/main/java/com/google/cloud/bigquery/samples/StreamingSample.java
```diff
@@ -160,18 +149,22 @@ public void remove() {
    * @throws IOException ioexception
    */
   // [START streamRow]
-  public static TableDataInsertAllResponse streamRow(final Bigquery bigquery,
+  public static TableDataInsertAllResponse streamRow(
+      final Bigquery bigquery,
```
lesv: Should it be BigQuery bigQuery ??

lesv:
LGTM - all comments are optional.

bigquery/src/main/java/com/google/cloud/bigquery/samples/BigqueryServiceFactory.java
tswast: Sigh. Yes. I'm having trouble getting git to accept that the filename is changed (I think because I'm on my Mac)

bigquery/src/main/java/com/google/cloud/bigquery/samples/AsyncQuerySample.java
tswast: That was the way it was in the existing sample, and it didn't bother me as much as some of the other things I fixed. I can look at changing this, too.

tswast:
@dpebot merge when green

lesv:
@tswast Your choice - DPEbot doesn't think this is green.

tswast:
Interesting. The Travis logs look fine to me. Maybe @jonparrott can help track down why it didn't automerge.

jonparrott:
It seems github's search indexing isn't quick enough for me to effectively filter the PRs. I'm gonna see if I can't get a fix in this afternoon.

It should work, but every now and then github will be slow about indexing so it'll fail to find the PR.

PR #299, opened 2016-08-16
Add tags for compiler properties

lesv:
1. tags for compiler properties
2. gradle comment


lesv:
LGTM

lesv:
@carieshmarie FYI

PR #300, opened 2016-08-16
Move plugin version definitions to properties.

tswast:
This a workaround for a missing feature in the codehaus versions maven
plugin where is cannot update plugins directly, but it can detect when
there are new plugins and update the related properties. https://github.com/mojohaus/versions-maven-plugin/issues/8#issuecomment-225593578

I don't actually update anything (or I tried not to, anyway). I'll get dpebot to do that.

@lesv

### DIFF
--- a/storage/xml-api/serviceaccount-appengine-sample/pom.xml
+++ b/storage/xml-api/serviceaccount-appengine-sample/pom.xml
@@ -31,8 +31,8 @@
   <packaging>war</packaging>
 
   <properties>
-    <maven.compiler.target>1.6</maven.compiler.target>
-    <maven.compiler.source>1.6</maven.compiler.source>
+    <maven.compiler.target>1.7</maven.compiler.target>
+    <maven.compiler.source>1.7</maven.compiler.source>
     <google-api-client.version>1.21.0</google-api-client.version>
     <webappDirectory>${project.build.directory}/${project.build.finalName}
     </webappDirectory>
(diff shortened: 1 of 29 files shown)


tswast:
@dpebot please merge when green

storage/xml-api/serviceaccount-appengine-sample/pom.xml
```diff
@@ -17,10 +31,15 @@
   <packaging>war</packaging>
 
   <properties>
+    <maven.compiler.target>1.6</maven.compiler.target>
+    <maven.compiler.source>1.6</maven.compiler.source>
```
lesv: 1.6 -- really?

vision/face-detection/pom.xml
```diff
@@ -29,6 +29,11 @@
     <relativePath>../..</relativePath>
   </parent>
 
+  <properties>
+    <maven.compiler.target>1.7</maven.compiler.target>
```
lesv: Probably should be 1.8

vision/label/pom.xml
```diff
@@ -29,6 +29,11 @@
     <relativePath>../..</relativePath>
   </parent>
 
+  <properties>
+    <maven.compiler.target>1.7</maven.compiler.target>
```
lesv: Probably should be 1.8

vision/landmark-detection/pom.xml
```diff
@@ -29,6 +29,11 @@
     <relativePath>../..</relativePath>
   </parent>
 
+  <properties>
+    <maven.compiler.target>1.7</maven.compiler.target>
```
lesv: Probably should be 1.8

lesv:
I'm generally happy with these changes, a few nits that can be ignored.

storage/xml-api/serviceaccount-appengine-sample/pom.xml
tswast: /shruggie That's what it was. I was a bit surprised, too. I'll update that one, too.

PR #305, opened 2016-08-18
BigQuery query samples: command line parameters.

tswast:
Switches all parameters to be supplied by properties rather than standard in.

Adds a flag to switch between legacy and standard SQL syntax.

@lesv PTAL.


lesv:
Do you want to provide an initial properties file?  Otherwise LGTM

tswast:
Oh, I guess that would be a different option. I had the sample define a default if the property is not found (for all but the projectId).

PR #306, opened 2016-08-18
Update the build.gradle

lesv:
1. More comments to build.gradle
2. add the wrapper jars


lesv:
@tswast PTAL at your leisure.

tswast:
LGTM.

Cool. I like the additional comments. (I assume `default` in the comments means "this is the default but I'm explicitly documenting it here"?)

PR #311, opened 2016-08-18
Adds system tests for storage transfer samples.

### LINKED ISSUE #113: Time passed to startTimeOfDay must be UTC
Sample code from `AwsRequester.java` inaccurately prescribes Pacific Date/Time values here:

```
/**
 * Specify times below using US Pacific Time Zone.
 */
private static final String START_DATE = "YYYY-MM-DD";
private static final String START_TIME = "HH:MM:SS";
```

This bit us when setting up a transfer job recently. Browsing the Schedule source you can see UTC should be passed:

```
/**
 * The time in UTC at which the transfer will be scheduled to start in a day. Transfers may start
 * later than this time. If not specified, transfers are scheduled to start at midnight UTC.
 * @param startTimeOfDay startTimeOfDay or {@code null} for none
 */
public Schedule setStartTimeOfDay(TimeOfDay startTimeOfDay)
```

tswast:
Also:
- Makes comments more accurate (UTC not Pacific time).
- Change to use properties instead of hard-coded constants for arguments.

@lesv 

Fixes #113


lesv:
LGTM - I hate that we are changing the same file :(

I see mention of UTC, but no where that it's set or actually tested.  (ie. your tests don't seem to request UTC in your tests).  If you don't think that's an issue – fine.

lesv:
@tswast Your travis changes need help. :(

tswast:
@lesv Travis is happy now.

As far as UTC goes: yes, it would be good to have a test that verifies this, but there don't seem to be any guarantees on when the copy / transfer job will run. From my manual testing, I scheduled it, but then it didn't seem to run right at that minute, though it did run eventually.

It's possible that it is really Pacific time and the API docs are wrong, but we did get a report in #113 that it was actually UTC, so I'm inclined to believe the API docs.

PR #313, opened 2016-08-25
Add style guidelines to CONTRIBUTING file.

tswast:
Also, add a link to google-java-format, which should help with keeping
samples within the style guidelines.

For internal bug 28395294.

@lesv PTAL


lesv:
LGTM

PR #315, opened 2016-08-29
Add gcloud-java sample for BigQuery synchronous querying.

tswast:
This loosely follows the Python sample here:

    https://github.com/GoogleCloudPlatform/python-docs-samples/blob/master/bigquery/cloud-client/sync_query.py

@lesv PTAL


lesv:
LGTM

PR #318, opened 2016-08-31
update hello world for new gradle & maven plugins

lesv:


lesv:
LGTM - I'm updating a doc CL that needs this.

PR #319, opened 2016-08-31
New plugin

lesv:
1. add license & tags to app.yaml
2. fix some versions and naming in the pom.xml
3. add tags to pom.xml
4. remove a blank line from build.gradle


lesv:
This works, I've tested.

PR #321, opened 2016-09-01
Use APT Addon for Gradle.

tswast:
@lesv FYI


lesv:
LGTM - wow - that's simpler.

PR #326, opened 2016-09-01
Stop hardcoding Netty dependency so example works on Android.

tswast:
Replaces https://github.com/GoogleCloudPlatform/java-docs-samples/pull/283.


lesv:
LGTM

PR #327, opened 2016-09-02
Revert "Auto-update dependencies."

tswast:
Reverts GoogleCloudPlatform/java-docs-samples#323


lesv:
Why did we do this?

tswast:
Because we did not want to update the jetty maven plugin version. We want to [make sure the Jetty version locally matches the one used in production](https://github.com/GoogleCloudPlatform/java-docs-samples/pull/323#issuecomment-244258847).

PR #338, opened 2016-09-14
Updated async-rest demo to latest Flex

gregw:
Also used as demo for https://github.com/GoogleCloudPlatform/jetty-runtime/issues/4


lesv:
LGTM

gregw:
@lesv thanks @ludoch I don't have merge permission.

PR #341, opened 2016-09-19
vision: Got empty response when requests are gzipped.

tswast:
This adds a workaround to the remaining samples for an issue where
requests over a certain size will "succeed" but return an empty response
from the vision API when gzipped.

The workaround is to disable gzip compression. Most images should be in
compressed formats anyway, so it's not like we are saving much bandwidth
with it.


lesv:
LGTM

PR #344, opened 2016-09-26
No model for Gradle

lesv:
1. Remove the model component for Gradle
2. update to latest maven plugin as well.


lesv:
@GoogleCloudPlatform/cloud-tools-for-java FYI
@tswast FYI
@jerjou PTAL

PR #345, opened 2016-09-28
move managed_vms -> flexible

lesv:


tswast:
Nice.

lesv:
Yea - but I need to fix the master pom.xml

tswast:
Probably will need to update the root pom.xml.

PR #348, opened 2016-09-29
Translate sample

puneith:
@lesv @tswast Please review translate sample.

### DIFF
--- a/translate/README.md
+++ b/translate/README.md
@@ -31,7 +31,7 @@ mvn clean compile assembly:single
 ```
 
 We can then run the assembled JAR file with the `java` command. The variable
-$COMMAND takes two values `detect` and `translate'.
+$COMMAND takes three values `langsupport`, `detect` and `translate`.
 
 ```
 JAR_FILE=target/translate-1.0-SNAPSHOT-jar-with-dependencies.jar
--- a/translate/README.md
+++ b/translate/README.md
@@ -31,13 +31,13 @@ mvn clean compile assembly:single
 ```
 
 We can then run the assembled JAR file with the `java` command. The variable
-$COMMAND takes two values `detect' and `translate'.
+$COMMAND takes two values `detect` and `translate'.
 
 ```
 MAIN_CLASS=com.google.cloud.translate.samples.TranslateText
 JAR_FILE=target/translate-1.0-SNAPSHOT-jar-with-dependencies.jar
 java -cp $JAR_FILE $MAIN_CLASS <detect|translate|langsupport> <text>
-<optional_source> <optional_target>
+    <optional_source> <optional_target>
 ```
 
 Example Usage:
@@ -50,7 +50,7 @@ TARGET_LANG="fr"
 
 Translate API Features:
 
- * Languages supported by the API
+ * List the languages supported by the API
    ```
    java -cp $JAR_FILE $MAIN_CLASS langsupport
    ```
--- a/translate/README.md
+++ b/translate/README.md
@@ -46,9 +46,20 @@ Example Usage:
 INPUT="A quick brown fox jumped over a lazy dog."
 SOURCE_LANG="en"
 TARGET_LANG="fr"
+```
 
+Languages supported by the API
+```
+java -cp $JAR_FILE $MAIN_CLASS langsupport
+```
+
+Detect input text language
+```
 java -cp $JAR_FILE $MAIN_CLASS detect "$INPUT"
+```
+
+Translate input text (with options)
+```
 java -cp $JAR_FILE $MAIN_CLASS translate "$INPUT"
 java -cp $JAR_FILE $MAIN_CLASS translate "$INPUT" $SOURCE_LANG $TARGET_LANG
-java -cp $JAR_FILE $MAIN_CLASS langsupport
 ```
--- a/translate/README.md
+++ b/translate/README.md
@@ -48,18 +48,18 @@ SOURCE_LANG="en"
 TARGET_LANG="fr"
 ```
 
-Languages supported by the API
-```
-java -cp $JAR_FILE $MAIN_CLASS langsupport
-```
+ * Languages supported by the API
+   ```
+   java -cp $JAR_FILE $MAIN_CLASS langsupport
+   ```
 
-Detect input text language
-```
-java -cp $JAR_FILE $MAIN_CLASS detect "$INPUT"
-```
+ * Detect input text language
+   ```
+   java -cp $JAR_FILE $MAIN_CLASS detect "$INPUT"
+  ```
 
-Translate input text (with options)
-```
-java -cp $JAR_FILE $MAIN_CLASS translate "$INPUT"
-java -cp $JAR_FILE $MAIN_CLASS translate "$INPUT" $SOURCE_LANG $TARGET_LANG
-```
+ * Translate input text (with options)
+   ```
+   java -cp $JAR_FILE $MAIN_CLASS translate "$INPUT"
+   java -cp $JAR_FILE $MAIN_CLASS translate "$INPUT" $SOURCE_LANG $TARGET_LANG
+   ```
--- a/translate/README.md
+++ b/translate/README.md
@@ -1,0 +1,49 @@
+# Google Cloud Translate Sample
+
+This sample demonstrates the use of [Google Cloud Translate
+API][Translate-Docs] for translating and detecting language text.
+
+[Translate-Docs]: https://cloud.google.com/translate/docs/
+
+## Java Version
+
+This sample requires you to have
+[Java8](https://docs.oracle.com/javase/8/docs/technotes/guides/install/install_overview.html).
+
+## Download Maven
+
+This sample uses the [Apache Maven][maven] build system. Before getting started,
+be
+sure to [download][maven-download] and [install][maven-install] it. When you use
+Maven as described here, it will automatically download the needed client
+libraries.
+
+[maven]: https://maven.apache.org
+[maven-download]: https://maven.apache.org/download.cgi
+[maven-install]: https://maven.apache.org/install.html
+
+## Run the sample
+
+To build the sample, we use Maven.
+
+```bash
+mvn clean compile assembly:single
+```
+
+We can then run the assembled JAR file with the `java` command. The variable
+$COMMAND takes two values `detect' and `translate'.
+
+```
+MAIN_CLASS=com.google.cloud.translate.samples.TranslateText
+JAR_FILE=translate-1.0-SNAPSHOT-jar-with-dependencies.jar
+java -cp $JAR_FILE $MAIN_CLASS <detect|translate> <text>
+```
+
+Example Usage:
+
+```
+INPUT="A quick brown fox jumped over a lazy dog."
+
+java -cp $JAR_FILE $MAIN_CLASS detect "$INPUT"
+java -cp $JAR_FILE $MAIN_CLASS translate "$INPUT"
+```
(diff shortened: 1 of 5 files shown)


translate/README.md
```diff
@@ -0,0 +1,67 @@
...
+## Run the sample
+
+To build the sample, we use Maven.
+
+```bash
+mvn clean compile assembly:single
+```
+
+We can then run the assembled JAR file with the `java` command. The variable
+$COMMAND takes two values `detect' and `translate'.
```
tswast: Nit. I think wrong kind of end-quote was used. ' instead of `.

translate/README.md
```diff
@@ -0,0 +1,67 @@
...
+
+```
+MAIN_CLASS=com.google.cloud.translate.samples.TranslateText
```
lesv: You could set this in your manifest when you create the jar -- actually it looks like it is done.

translate/README.md
```diff
@@ -0,0 +1,67 @@
...
+JAR_FILE=target/translate-1.0-SNAPSHOT-jar-with-dependencies.jar
+java -cp $JAR_FILE $MAIN_CLASS <detect|translate|langsupport> <text>
+<optional_source> <optional_target>
```
tswast: indent this line 4 spaces to show it is a continuation of the previous line.

translate/README.md
```diff
@@ -0,0 +1,67 @@
...
+
+```
+INPUT="A quick brown fox jumped over a lazy dog."
+SOURCE_LANG="en"
+TARGET_LANG="fr"
+```
+
+Translate API Features:
+
+ * Languages supported by the API
```
tswast: Change this line to: List the languages supported by the API.

translate/pom.xml
```diff
@@ -0,0 +1,63 @@
...
+            <manifest>
+              <mainClass>com.google.cloud.translate.samples.TranslateText</mainClass>
+            </manifest>
+          </archive>
+          <descriptorRefs>
+            <descriptorRef>jar-with-dependencies</descriptorRef>
+          </descriptorRefs>
+        </configuration>
+      </plugin>
+      <plugin>
```
lesv: Compiler plugin isn't really required, you can set in `<properties>`:
    `<maven.compiler.target>1.8</maven.compiler.target>`
    `<maven.compiler.source>1.8</maven.compiler.source>`
and that is all you need.

translate/README.md
```diff
@@ -0,0 +1,67 @@
...
+mvn clean compile assembly:single
+```
+
+We can then run the assembled JAR file with the `java` command. The variable
+$COMMAND takes two values `detect' and `translate'.
+
+```
+MAIN_CLASS=com.google.cloud.translate.samples.TranslateText
+JAR_FILE=target/translate-1.0-SNAPSHOT-jar-with-dependencies.jar
+java -cp $JAR_FILE $MAIN_CLASS <detect|translate|langsupport> <text>
```
lesv: `java -jar $JAR_FILE <detect|translate|langsupport> <text>`

translate/pom.xml
```diff
@@ -0,0 +1,63 @@
+<?xml version="1.0"?>
+<project xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd" xmlns="http://maven.apache.org/POM/4.0.0"
```
tswast: Add a license header after the `<?xml>` line and before the `<project>` line. 

Optionally, you can also delete the `<?xml>` line and make the `<project>` line just `<project>`. See: https://github.com/GoogleCloudPlatform/java-docs-samples/blob/master/appengine/helloworld-new-plugins/pom.xml

translate/src/main/java/com/google/cloud/translate/samples/TranslateText.java
```diff
@@ -0,0 +1,114 @@
...
+import com.google.cloud.translate.Translate.TranslateOption;
+import com.google.cloud.translate.Translation;
+import com.google.cloud.translate.testing.RemoteTranslateHelper;
+import com.google.common.collect.ImmutableList;
+
+import java.io.PrintStream;
+import java.util.List;
+
+public class TranslateText {
+  private static final Translate TRANSLATE = RemoteTranslateHelper.create().options().service();
```
tswast: Are you sure this is the class we need to use for a sample? The fact that this is in a package called `testing` makes me wary to suggest it for production, non-test code.

tswast: Probably want to pass the `Translate` object to each individual function here rather than creating it statically, also. That way it is more clear that developers can create the object as part of a normal lifecycle.

Consider even having each function create a Translate object, depending on how these snippets will be used, that could be the most clear.

translate/src/test/java/com/google/cloud/translate/samples/TranslateTextTest.java
```diff
@@ -0,0 +1,121 @@
...
+    // Arrange
+    ByteArrayOutputStream bout = new ByteArrayOutputStream();
+    PrintStream out = new PrintStream(bout);
+
+    // Act
+    TranslateText.detectLanguage("Mit Macht kommt große Verantwortung.", out);
+
+    // Assert
+    String got = bout.toString();
+    assertThat(got).contains("language=de, confidence=0.9293963");
```
tswast: Remove the confidence from the test.

translate/src/test/java/com/google/cloud/translate/samples/TranslateTextTest.java
```diff
@@ -0,0 +1,121 @@
...
+    // Arrange
+    ByteArrayOutputStream bout = new ByteArrayOutputStream();
+    PrintStream out = new PrintStream(bout);
+
+    // Act
+    TranslateText.detectLanguage("With power comes great responsibility.", out);
+
+    // Assert
+    String got = bout.toString();
+    assertThat(got).contains("language=en, confidence=0.79742646");
```
tswast: Remove the confidence from the test. That's way too precise. Very fragile if the model changes at all.

translate/src/test/java/com/google/cloud/translate/samples/TranslateTextTest.java
```diff
@@ -0,0 +1,121 @@
...
+    // Arrange
+    ByteArrayOutputStream bout = new ByteArrayOutputStream();
+    PrintStream out = new PrintStream(bout);
+
+    // Act
+    TranslateText.translateTextWithOptions("Mit Macht kommt große Verantwortung.", "de", "es", out);
+
+    // Assert
+    String got = bout.toString();
+    assertThat(got).contains("Con el poder viene una gran responsabilidad.");
```
tswast: I'm wary about this one, too. Since I'm sure there is more than one possible translation. Maybe change it to look for a few important words (like responsabilidad)

translate/README.md
puneith: done

translate/README.md
puneith: done

translate/README.md
puneith: done

translate/src/test/java/com/google/cloud/translate/samples/TranslateTextTest.java
puneith: Agree, but its still better to have the notion of confidence captured so that if confidence drops to say 0.2 then we catch that. Let me make appropriate changes.

translate/src/test/java/com/google/cloud/translate/samples/TranslateTextTest.java
puneith: Same.

translate/src/test/java/com/google/cloud/translate/samples/TranslateTextTest.java
puneith: Currently, we only get one translation. If this changes it will be good to know why and make modifications to the test appropriately. But, better to know than qualify the test even if accidentally a bad translation is returns which contains  responsabilidad as only relevant word.

translate/src/test/java/com/google/cloud/translate/samples/TranslateTextTest.java
tswast: Alright, maybe just the first few digits, then? This is too precise. I don't care if it goes from 0.797 to 0.798, but maybe we do care if it goes from 0.79 to 0.80? (Probably we don't, but we do care if it changes to below 0.5?)

lesv: Wow, your comparing to a string?  That's not a great idea. (not my first choice of wording).

Confidence will change with time, why not claim it will be within a range of .7 .. .9  This stuff tends to jump around as the model changes.

puneith: Fixed it.

translate/src/test/java/com/google/cloud/translate/samples/TranslateTextTest.java
puneith: Fixed

translate/pom.xml
puneith: done

translate/README.md
puneith: done

translate/README.md
puneith: yes

translate/src/test/java/com/google/cloud/translate/samples/TranslateTextTest.java
```diff
@@ -0,0 +1,134 @@
...
+    // Assert
+    String got = bout.toString();
+    assertThat(got).contains("language=de");
+
+    // Assert
+    Double confidence = Double.parseDouble(
+        got.split("confidence=")[1].split("}")[0]
+        );
+    assertThat(confidence).isLessThan(1.0);
+    assertThat(confidence).isGreaterThan(0.9);
```
lesv: `assertThat(confidence).isWithin(.05).of(.95)` might be more readable.

lesv:
LGTM from me, but I'd like the minor change in your use of Truth.  You should consider Tim's ask

translate/src/test/java/com/google/cloud/translate/samples/TranslateTextTest.java
puneith: done

translate/src/main/java/com/google/cloud/translate/samples/TranslateText.java
puneith: done ptal

translate/src/main/java/com/google/cloud/translate/samples/TranslateText.java
```diff
@@ -0,0 +1,143 @@
...
+      out.printf("Name: %s, Code: %s\n", language.name(), language.code());
+    }
+  }
+
+  /**
+   * Create Google Translate API Service.
+   *
+   * @return Google Translate Service
+   */
+  public static Translate createTranslateService() {
```
tswast: Are the default options not sufficient that this function is needed? Please file a bug against google-cloud-java if the defaults aren't working well for you.

puneith: They do, but this is to demonstrate that we can set them and how. Obviously, developers will still have to look at the API documentation, but gives them a peek into the options.

puneith:
@lesv whats the minor change for truth to reconsider? @tswast

tswast:
Removing the dependence on exact string comparison for the confidences was my concern, which you have addressed.

translate/src/test/java/com/google/cloud/translate/samples/TranslateTextTest.java
lesv: This was what I was asking for. (truth)

PR #357, opened 2016-10-10
Make each speech sample independent

jerjou:


jerjou:
@dpebot would you mind merging once Travis says it's okay?

jerjou:
@dpebot? can you please merge when Travis is green?

tswast:
Looks like the Travis failure is for Vision API. Go ahead and merge since the failure isn't related to this change.

```
 FaceDetectAppIT.detectFaces_badImage_throwsException:64 Not true that IOException message (<"image-annotator::bad image data.: image processing error!">) contains <"malformed request">
```

The error message for bad image data has changed. (Used to be malformed request)

lesv:
This sample uses an old version of boringSSL.   Should not have been LGTM'd

lesv:
Well, that's too strong -- it would be good to upgrade it, however.

jerjou:
Hm... the pom lists 1.1.33.Fork23, which [seems to be](http://search.maven.org/#search%7Cga%7C1%7Cboringssl) the latest, no?

jerjou:
plus, isn't that dpebot's job? :)

lesv:
That's Ok, but all the stuff around it is wrong.  See https://github.com/grpc/grpc-java/blob/master/SECURITY.md

DPE Bot doesn't change things that don't come from a property.

PR #360, opened 2016-10-11
Fix Vision tests.

tswast:
They were brittle, since they were looking for a specific error message,
which changed in the service.

@jerjou @lesv PTAL.


tswast:
Ugh. Fix one and more appear. Filed https://github.com/GoogleCloudPlatform/java-docs-samples/issues/361 for failing Cloud Monitoring tests.

PR #368, opened 2016-10-12
Remove java-repo-tools subtree.

tswast:
Now uses the published shared-configuration parent. There is still a
java-repo-tools submodule, but it is only used for the test-localhost.sh
script, so should not be needed for building & running samples.

@lesv PTAL


tswast:
This the first step. As I figure out to bundle more resources into the shared-configuration package, hopefully we can also get rid of the submodules, too.

tswast:
Submodules are a lot less weird than subtrees. The reason I didn't do a git clone is that submodules are basically the same as a git clone, but they have an extra metadata file to say which commit we are at.

That way we can make updates to java-repo-tools and then update the commit number in this repo when we are ready (in case there were breaking changes). [Submodules are a lot easier to do this updating than subtrees](http://stackoverflow.com/questions/8191299/update-a-submodule-to-the-latest-commit)

Yes, I agree the scripts and config files should all be bundled in Maven central, but hopefully this is a step in the right direction until I can figure that bit out.

java-repo-tools
```diff
@@ -0,0 +1 @@
+Subproject commit 74755a28f01aedf57992f473eb2d0201b1e04f2a
```
tswast: @lesv This is the metadata that ties the submodule to the specific commit. Doing a git clone followed by a checkout of this commit number should be exactly the same. (This differs from subtrees, which don't require the extra clone, but then the histories were merged, which was pretty awful)

lesv: Yep - saw that.  I just prefer things that are a bit less "magical".

PR #370, opened 2016-10-19
Add appengine-api-1.0-sdk dependency to taskqueue sample.

tswast:
This sample was breaking in the latest release due to this missing
dependency.

@lesv PTAL

### DIFF
--- a/appengine/taskqueue-push/pom.xml
+++ b/appengine/taskqueue-push/pom.xml
@@ -45,6 +45,11 @@ Copyright 2016 Google Inc. All Rights Reserved.
       <artifactId>json</artifactId>
       <version>20160810</version>
     </dependency>
+    <dependency>
+      <groupId>com.google.appengine</groupId>
+      <artifactId>appengine-api-1.0-sdk</artifactId>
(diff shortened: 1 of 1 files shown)


lesv:
How did it ever work?

tswast:
I think one of the other dependencies must have also depended on the App Engine API SDK transitively.

Actually, probably the [appengine-maven-plugin](https://github.com/GoogleCloudPlatform/java-docs-samples/blob/85d464027f6f8b0e7ba04abc9c72261b5c2284d2/appengine/taskqueue-push/pom.xml#L32). That probably belongs in the plugins section, not dependencies.

lesv:
Yes - agree - could you fix, or I will when i do my mass pass next week.

tswast:
I fixed this in https://github.com/GoogleCloudPlatform/java-docs-samples/pull/371. That's not to say that there aren't other samples that depend on it, but these are the only two that blew up when updating the App Engine SDK version.

lesv:
I meant your suggestion of moving the appengine-maven-plugin to the plugin section.

PR #371, opened 2016-10-19
Add appengine-api-1.0-sdk dep to users sample.

tswast:
dpebot found another one when I tried to re-run the dependency updater this morning.

The users sample and the taskqueue-push sample were depending on the
appengine-maven-plugin, which used to depend transitively on the
appengine-api-1.0-sdk package, but it does not in the latest version.

These samples should not have been using the plugin as a dependency,
anyway.

@lesv PTAL


tswast:
Or @jerjou

PR #373, opened 2016-10-19
Add test to Firebase TicTacToe sample

jerjou:
Also add example test for testing things that incorporate a urlfetch. (I hit a lot of dead-ends before I found the documentation for how to do this...)


jerjou:
@dpebot merge when green, plz

jerjou:
FYI added commit to fix style errors.

PR #374, opened 2016-10-19
More firebase tests

jerjou:
Also, enforces logged-in state from web.xml instead of in the servlet, for brevity.

FYI diffbase on #373


appengine/firebase-tictactoe/src/main/webapp/WEB-INF/web.xml
```diff
@@ -21,6 +21,15 @@ http://java.sun.com/xml/ns/javaee/web-app_2_5.xsd" version="2.5">
...
     </welcome-file-list>
+    <security-constraint>
+      <web-resource-collection>
+        <web-resource-name>entire-app</web-resource-name>
+        <url-pattern>/*</url-pattern>
+      </web-resource-collection>
+      <auth-constraint>
+        <role-name>*</role-name>
+      </auth-constraint>
+    </security-constraint>
```
lesv: You can specify `https` only in your `appengine-web.xml`.  I don't think this will require being logged in, if that's what your trying for.

jerjou: I'm just trying to require sign-in, as per [this article](https://cloud.google.com/appengine/docs/java/config/webxml#Security_and_Authentication)

lesv: Ah, but you are in standard.  That article suggests that you block subpaths /auth/* Not /*

jerjou: Right - the tictactoe app requires every path to be authenticated, so I modified it to block all paths.

PR #376, opened 2016-10-20
Address changes for analyzeSyntax.

gguuss:
Will need an update to pom.xml when the updated client library is published.


language/analysis/src/test/java/com/google/cloud/language/samples/AnalyzeIT.java
```diff
@@ -59,10 +59,10 @@
...
   @Test public void analyzeSentiment_returnPositive() throws Exception {
     // Act
-    Sentiment sentiment = 
+    Sentiment sentiment =
         analyzeApp.analyzeSentiment(
             "Tom Cruise is one of the finest actors in hollywood and a great star!");
-    
+
     // Assert
     assertThat((double)sentiment.getMagnitude()).isGreaterThan(0.0);
```
puneith: @gguuss v1 has score now and no polarity so the test will need to change.

language/analysis/src/main/java/com/google/cloud/language/samples/Analyze.java
```diff
@@ -195,15 +197,13 @@ public Sentiment analyzeSentiment(String text) throws IOException {
    * Gets {@link Token}s from the string {@code text}.
    */
   public List<Token> analyzeSyntax(String text) throws IOException {
-    AnnotateTextRequest request =
-        new AnnotateTextRequest()
+    AnalyzeSyntaxRequest request =
```
puneith: @gguuss The analyzeSentiment is still returning Mag and Polarity. It should return score and mag.

language/analysis/src/main/java/com/google/cloud/language/samples/Analyze.java
```diff
@@ -136,7 +138,7 @@ public static void printSyntax(PrintStream out, List<Token> tokens) {
   /**
    * Connects to the Natural Language API using Application Default Credentials.
    */
-  public static CloudNaturalLanguageAPI getLanguageService() 
+  public static CloudNaturalLanguageAPI getLanguageService()
     throws IOException, GeneralSecurityException {
     GoogleCredential credential =
         GoogleCredential.getApplicationDefault().createScoped(CloudNaturalLanguageAPIScopes.all());
```
puneith: @gguuss You need to make changes to printSyntax() to add more tokens returned in v1.

language/analysis/src/main/java/com/google/cloud/language/samples/Analyze.java
```diff
@@ -136,7 +138,7 @@ public static void printSyntax(PrintStream out, List<Token> tokens) {
   /**
```
puneith: @gguuss The analyzeSentiment is still returning Mag and Polarity. It should return score and mag.

gguuss: Done, showing score and hiding polarity.

language/analysis/src/main/java/com/google/cloud/language/samples/Analyze.java
gguuss: Done. I didn't see additional tokens in syntax, I currently have lemma, text, part of speech, dependency edge. I did, however, add "mentions" to entities, which was what I think you meant.

PR #379, opened 2016-10-24
Double quotes for html attributes.

jerjou:


lesv:
NOTE - it's ok to merge.  @jerjou Could you look at why Entities is failing travis.
@tswast Any suggestions you could give Jerjou would be helpful.

jerjou:
#380 should address the `entities` failure.

tswast:
Apparently the natural-language API changed it's mind.

"John was seriously injured in an accident" now has positive polarity. We probably need to update the text. I'm not sure it's worth filing a bug, since I think sentiment is supposed to use multiple sentences to be accurate.
https://github.com/GoogleCloudPlatform/java-docs-samples/blob/master/language/analysis/src/test/java/com/google/cloud/language/samples/AnalyzeIT.java#L79

PR #385, opened 2016-10-25
update endpoints-frameworks-v2 beta versions

tangiel:


lesv:
Happy to merge, these changes should happen automatically every night.  If they aren't it's a problem with the DPE Bot.

lesv:
I'm going to admin approve since it worked before I updated the branch.

PR #389, opened 2016-10-26
Move Datastore concepts to test directory so the tests actually run.

tswast:
@lesv ptal


tswast:
Docs CL has approval. When should we check these in?

lesv:
Tomorrow.

tswast:
Okay, just as well. I want to verify that the tests are actually running before merging anyway. I think I might have to add the JUnit runwith annotation and/or rename the class with a Test suffix

tswast:
My thoughts last night were correct. Moving the Concepts to the `test/` directory made the plugin not complain that the directory was missing, but it still didn't actually run them. Adding the Test suffix and the RunWith annotation they now run.

I'll update the docs CL to match this change, as well.

@lesv

PR #391, opened 2016-10-28
Fix up sendgrid README

jerjou:


jerjou:
But at least the markdown isn't broken, now! :no_mouth:

PR #396, opened 2016-11-03
Update expected error message.

jerjou:


jerjou:
I imagine they just changed the error message to make it more human-readable. I wouldn't be surprised if the other test in this file (that looks for `NOT_FOUND`) breaks in the next couple days for similar reasons..

PR #401, opened 2016-11-06
Use getters for DatastoreOptions and the Datastore service

glaforge:
`defaultInstance()` and `service()` seem to be deprecated in favor of `getDefaultInstance()` and `getService()`


lesv:
Yep - recent change.

PR #402, opened 2016-11-07
Removes Datastore geospatial sample.

tswast:
In reference to internal bug 32697836.

@lesv PTAL


tswast:
Because the API is in a closed alpha and they are not accepting whitelist applications. The tech writers removed the documentation today because they were getting too much feedback asking to join the whitelist.

I suppose we could keep the sample around, but it's not referenced anywhere in the docs anymore.

PR #408, opened 2016-11-15
update /flexible/* to env-flex

lesv:
1. app.yaml -> env:flex
2. remove any compat samples
3. Make sure we use new appengine & jetty plugins
4. Update artivactId’s and groupId’s to flexible where appropriate

TODO:
[] flexible/async-rest - shouldn’t be using Docker
[] flexible/endpoints - does appengine:deploy work?
[] flexible/* - See if we really want to use EnvVar’s to pass values,
perhaps we should modify to use web.xml context vars.
[] flexible/memcache — NEED WORK


lesv:
@jerjou @tswast FYI

PR #413, opened 2016-11-15
Update pom.xml

puneith:
@lesv


lesv:
@puneith @tswast @jerjou This is breaking the build - we need to get the project WL'd for Translate

I'm going to Admin Approve.

PR #414, opened 2016-11-15
added function without and with model

puneith:
@lesv @tswast @sgreenberg @gguuss Made changes to the sample so that it works for people without WL as well.

### DIFF
--- a/translate/src/main/java/com/google/cloud/translate/samples/TranslateText.java
+++ b/translate/src/main/java/com/google/cloud/translate/samples/TranslateText.java
@@ -61,6 +61,34 @@ public class TranslateText {
 
   /**
    * Translate the source text from source to target language.
+   * Make sure that your project is whitelisted. 
+   *
+   * @param sourceText source text to be translated
+   * @param sourceLang source language of the text
+   * @param targetLang target language of translated text
+   * @param out print stream
+   */
+  public static void translateTextWithOptionsAndModel(
+      String sourceText,
+      String sourceLang,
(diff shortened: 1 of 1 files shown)


translate/src/main/java/com/google/cloud/translate/samples/TranslateText.java
```diff
@@ -61,13 +61,14 @@ public static void translateText(String sourceText, PrintStream out) {
...
    * Translate the source text from source to target language.
+   * Make sure that your project is whitelisted. 
    *
    * @param sourceText source text to be translated
    * @param sourceLang source language of the text
    * @param targetLang target language of translated text
    * @param out print stream
    */
-  public static void translateTextWithOptions(
+  public static void translateTextWithOptionsAndModel(
```
gguuss: Any reason not to just pass the model and not set it if the value is null?

puneith: I guess I was trying to clearly demarcate WL & non-WL users and so to avoid confusion? Sorry did quite last minute so may not have a very good explanation.

tswast: I'm okay with duplicating if we are documenting each separately in the docs.

lesv:
Don't you need to update the README for this?

puneith:
@lesv The documentation will talk about the premium model feature and getting whitelisted. We have sample code which people can refer to.

PR #431, opened 2016-12-02
Update Dockerfile

Eng-Fouad:
Updating the war file named endpoints-1.0-SNAPSHOT.war to flexible-endpoints-1.0-SNAPSHOT.war, otherwise the docker building process fails.


Eng-Fouad:
I signed the CLA.

PR #432, opened 2016-12-02
Add GCS examples for Natural Language

gguuss:
Adds Google Cloud Storage (GCS) examples and tests to be at parity with other samples for inclusion in snippets.


language/analysis/README.md
```diff
@@ -30,12 +30,13 @@ mvn clean compile assembly:single
 ```
 
 We can then run the assembled JAR file with the `java` command. The variable $COMMAND takes
-three values `entities`, `sentiment` or `syntax`.
+six values `entities-text`, `entities-file`, `sentiment-text`, `sentiment-file`,
+`syntax-text`, or `syntax-file`.
```
jerjou: Oi. Perhaps would be cleaner to just detect a gcs uri? As in:

```java
if (command.equals("entities")) {
  if (text.startsWith("gs://")) {
    app.analyzeEntitiesFile(text);
  } else {
    app.analyzeEntitiesText(text);
  }
} // ... etc
```
inside `Analyze.java`

language/analysis/src/main/java/com/google/cloud/language/samples/Analyze.java
```diff
@@ -164,19 +170,42 @@ public Analyze(LanguageServiceClient languageApi) {
   }
 
   /**
+   * Gets {@link Entity}s from the string representing the GCS {@code path}.
```
jerjou: Perhaps:

    * Gets {@link Entity}s from the contents of the object at the given GCS {@code path}

Otherwise it sounds like it's looking for entities in the path string itself. (here and below)

language/analysis/src/test/java/com/google/cloud/language/samples/AnalyzeIT.java
```diff
@@ -59,32 +59,75 @@
...
+
+  @Test public void analyzeSentiment_returnNeutralFile() throws Exception {
+      // Act
+      Sentiment sentiment =
+          analyzeApp.analyzeSentimentFile("gs://cloud-samples-tests/natural-language/"
+                  + "sentiment/bladerunner-neutral.txt");
+
+      // Assert
+      assertThat((double)sentiment.getMagnitude()).isGreaterThan(1.0);
+      assertThat((double)sentiment.getScore()).isWithin(0.0);
```
jerjou: Seems a bit brittle, no? Perhaps `isWithin(0.1)`?

language/analysis/README.md
jmdobry: That would moves us toward writing library code in the snippets, and the other 6 languages have separate snippets for raw strings vs gcs files, so Java would need to do the same to fit into the docs where the other snippets live.

gguuss: I think he just means in the command parser, I am in favor of simplifying the args parser for a lighter-weight usage string.

gguuss: Done.

language/analysis/src/test/java/com/google/cloud/language/samples/AnalyzeIT.java
gguuss: Ack, changed to 0.1

language/analysis/README.md
jerjou: Yup - yay!

gguuss:
Sorry for merging a little aggressively and good catch - I'll fix the style for shell variables in a separate PR if that works.  Should we also be making `$JAR_FILE` and `$MAIN_CLASS` local variables as well?

lesv:
I think that's more inline with the style guide.  (ie. YES)

gguuss:
SGTM.

PR #436, opened 2016-12-06
update echo endpoint to support upcoming tutorial

tangiel:


lesv:
Travis is unhappy.   CheckStyle complained.

lesv:
ping

lesv:
@tangiel ping

tangiel:
@lesv apologies, fixed now.

appengine/endpoints-frameworks-v2/backend/src/main/java/com/example/echo/Echo.java
```diff
@@ -32,20 +38,80 @@
         ownerDomain = "echo.example.com",
         ownerName = "echo.example.com",
         packagePath = ""
-      )
+      ),
+    // [START_EXCLUDE]
+    issuers = {
+      @ApiIssuer(
+        name = "firebase",
+        issuer = "https://securetoken.google.com/tangd-customer",
```
lesv: Is this a general thing?  I can't find it documented anywhere.

appengine/endpoints-frameworks-v2/backend/src/main/java/com/example/echo/Echo.java
```diff
@@ -59,19 +125,48 @@ public Message echo(Message message) {
...
+   *
+   * Note that httpMethod is not required here. Without httpMethod, this will default to GET due
+   * to the API method name. httpMethod is added here for example purposes.
+   */
+  // [START firebase_auth]
+  @ApiMethod(
+      path = "firebase_user",
+      httpMethod = ApiMethod.HttpMethod.GET,
+      authenticators = {EspAuthenticator.class},
+      issuerAudiences = {@ApiIssuerAudience(name = "firebase", audiences = {"tangd-customer"})}
```
lesv: same question - is `tangd-customer` something that should be modified by the user?  If so, a comment would be good.

appengine/endpoints-frameworks-v2/backend/src/main/java/com/example/echo/Echo.java
tangiel: I replaced with YOUR-PROJECT-ID to match the rest of the sample.

appengine/endpoints-frameworks-v2/backend/src/main/java/com/example/echo/Echo.java
tangiel: Replaced with YOUR-PROJECT-ID.

lesv:
You might wish to update the README to mention all the places a user needs to edit.  (Hint - it's not just the pom.xml anymore.

I presume you don't want this merged until the docs are ready?

tangiel:
I updated the README to replace the YOUR-PROJECT-ID. As for the client ID stuff, it will be better to link to the docs when they're published for that, as it's much more involved. However, I don't mind merging this now so we can start using the code fragments.

Issue #252, opened 2016-06-02
Where are AutoValue_ImageText and AutoValue_Word defined?


junyongyou:
I want to use Text detection app, but it seems AutoValue_ImageText and AutoValue_Word are not defined. How can I fix this? Thanks a lot.

tswast:
Those classes are generated by the [AutoValue](https://github.com/google/auto/tree/master/value) plugin. If you are using Maven it should just work.

tswast:
Were you able to figure it out? Which IDE are you using?

junyongyou:
Sorry not to respond earlier. I actually did not figure it out. I have installed Maven, and it works. I am using Intellij IDEA, but the two classes cannot be generated. Is there something I need to config further? Thanks a lot for your help. 
BTW, I can use other Vision modules, such as labelling, face detection.

tswast:
Sorry about this (someone else reported privately to me the same issue). I am not seeing an IntelliJ plugin for AutoValue listed here https://github.com/google/auto/blob/master/value/userguide/index.md I only see instructions for Gradle and Maven. I'll update the sample to remove AutoValue references.

jaickerag:
Hi,
Its Working for me using this settings 

![image](https://cloud.githubusercontent.com/assets/22210123/18902149/d4e44854-851e-11e6-8056-81a1576f4792.png)

scurrent:
I seem to be having the same problem as junyoungyou. 
Running Eclipse, using Maven, and the sample Vision project won't compile because it cannot find: 
AutoValue_ImageText   or AutoValue_Word

tswast:
There is a work-around for using AutoValue with IntelliJ if you enable a setting to delegate to Maven for annotation processing.

![image](https://cloud.githubusercontent.com/assets/247555/20986765/d1990698-bc7e-11e6-8b6b-192506faaa97.png)

That said, I have removed references to AutoValue from the sample code in https://github.com/GoogleCloudPlatform/java-docs-samples/pull/439. After that is merged into master, you should not encounter this problem anymore.

PR #439, opened 2016-12-07
Remove dependency on AutoValue.

tswast:
AutoValue saves us from having to write boilerplate, but it only works
with Maven or Gradle. Those users using Eclipse and IntelliJ to compile
have been having difficulty.

Fixes https://github.com/GoogleCloudPlatform/java-docs-samples/issues/252

@lesv PTAL


tswast:
Yeah, I'm not happy about it either. 🙁  It will make the sample easier to run, though.

PR #442, opened 2016-12-08
Update Endpoints samples:

sepehre:
1. Change default service names to the new cloud.goog format.
2. Change app.yaml beta_settings to endpoints_api_service.

### DIFF
--- a/endpoints/getting-started/src/main/appengine/openapi.yaml
+++ b/endpoints/getting-started/src/main/appengine/openapi.yaml
@@ -3,7 +3,7 @@ info:
   description: "A simple Google Cloud Endpoints API example."
   title: "Endpoints Example"
   version: "1.0.0"
-host: "YOUR-PROJECT-ID.appspot.com"
+host: "echo-api.endpoints.YOUR-PROJECT-ID.cloud.goog"
 basePath: "/"
 consumes:
 - "application/json"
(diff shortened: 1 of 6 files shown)


sepehre:
Please do not merge until documentation updated have been coordinated.

endpoints/getting-started/src/main/appengine/openapi.yaml
```diff
@@ -3,7 +3,7 @@ info:
   description: "A simple Google Cloud Endpoints API example."
   title: "Endpoints Example"
   version: "1.0.0"
-host: "YOUR-PROJECT-ID.appspot.com"
+host: "echo-api.endpoints.YOUR-PROJECT-ID.cloud.goog"
```
lesv: Can you use https with this?

sepehre: This field is not a DNS name that the application will be called at (not yet at least). It is the Endpoints service name.

Eng-Fouad: I wonder why you chose `cloud.goog` :)

lesv:
You've got some conflicting changes in `endpoints/getting-started/src/main/appengine/openapi.yaml`.

Also, any chance you could update to the new maven plugin?  https://cloud.google.com/appengine/docs/java/tools/using-maven  We'll have to do it early Jan.  The commands become appengine:run & appengine:deploy

sepehre:
Let's abandon this PR in favor of https://github.com/GoogleCloudPlatform/java-docs-samples/pull/447

If you don't mind, let's please do the maven plugin change separately

lesv:
Sg.

PR #446, opened 2016-12-09
Use latest (beta) versions of cloud client libraries.

tswast:
Note this wasn't automatically done because dpebot currently skips beta
releases. We can look at how to add an exception for the
com.google.cloud libraries if they are going to stay beta a while.

### DIFF
--- a/language/cloud-client/pom.xml
+++ b/language/cloud-client/pom.xml
@@ -37,7 +37,7 @@
     <dependency>
       <groupId>com.google.cloud</groupId>
       <artifactId>google-cloud-language</artifactId>
-      <version>0.7.0</version>
+      <version>0.8.0</version>
     </dependency>
 
     <!-- Test dependencies -->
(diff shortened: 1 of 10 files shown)


language/cloud-client/pom.xml
```diff
@@ -37,7 +37,7 @@
     <dependency>
       <groupId>com.google.cloud</groupId>
       <artifactId>google-cloud-language</artifactId>
-      <version>0.7.0</version>
+      <version>0.8.0</version>
```
lesv: -beta ??

lesv: Sigh

lesv:
Could you update getting-started-java ??

tswast:
Sure. I can do that now.

lesv:
Actually - I'll have a PR later today for that - I'll do it.

tswast:
Okay.

PR #448, opened 2016-12-13
enforce java 7 on Endpoints sample

tangiel:
Currently, mvn appengine:update fails because it compiles Java 8
bytecode. This forces the app code to Java 7.

### DIFF
--- a/appengine/endpoints-frameworks-v2/backend/pom.xml
+++ b/appengine/endpoints-frameworks-v2/backend/pom.xml
@@ -154,6 +154,15 @@
                     </jvmFlags -->
                 </configuration>
             </plugin>
+            <plugin>
+              <groupId>org.apache.maven.plugins</groupId>
+              <artifactId>maven-compiler-plugin</artifactId>
+              <version>3.6.0</version>
+              <configuration>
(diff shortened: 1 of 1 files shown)


appengine/endpoints-frameworks-v2/backend/pom.xml
```diff
@@ -154,6 +154,15 @@
                     </jvmFlags -->
                 </configuration>
             </plugin>
+            <plugin>
```
lesv: We usually do this without adding the plugin.

```xml
    <properties>
        <maven.compiler.target>1.8</maven.compiler.target>
        <maven.compiler.source>1.8</maven.compiler.source>
    </properties>
```

Note - apparently we need to fix a bunch of files.

lesv:
Please use the suggested method. if this is urgent, I can approve & merge as is.

appengine/endpoints-frameworks-v2/backend/pom.xml
tangiel: Done.

lesv:
Thanks.

PR #458, opened 2016-12-16
update framework sample dependency versions

tangiel:


tangiel:
@lesv could you take a quick peek?

tswast:
Yes. I think it was skipped due to the `-beta`. I'll remove that filter from dpebot and see what happens. Maybe it's fine to use beta versions for most packages that use them?

lesv:
I think the problem comes in when we switch between numeric sorting and alphabetic sorting.  Some number 1.0.0-beta3, followed by 1.0.0; which Maven seems to believe is older than the beta3 qualifier for some reason, in some cases.

PR #486, opened 2017-01-12
Add Vision API quickstart.

jmdobry:


lesv:
LGTM

lesv:
Though - you might wish to add it to the parent pom.xml so that it gets built and tested.

PR #496, opened 2017-01-31
Kms samples

jerjou:
Working on tests, but figured I'd send this out first to start the review.

### DIFF
--- a/kms/README.md
+++ b/kms/README.md
@@ -1,0 +1,30 @@
+# Cloud Key Management Service
+
+Google [Cloud Key Management Service](https://cloud.google.com/kms/) is a
+cloud-hosted key management service that lets you manage encryption for your
+cloud services the same way you do on-premise. You can generate, use, rotate and
+destroy AES256 encryption keys. These sample Java applications demonstrate
+how to access the KMS API using the Google Java API Client Libraries.
+
+## Quickstart
+
+Install [Maven](http://maven.apache.org/).
+
+Build your project with:
+
+    mvn clean compile assembly:single
+
+You can run the quickstart with:
+
+    java -cp target/kms-samples-1.0.0-jar-with-dependencies.jar \
+        com.example.Quickstart [your-project-id]
+
+and can see the available snippet commands with:
+
+    java -cp target/kms-samples-1.0.0-jar-with-dependencies.jar \
+        com.example.Snippets
+
+For example:
+
+    java -cp target/kms-samples-1.0.0-jar-with-dependencies.jar \
+        com.example.Snippets createKeyRing -p <your-project-id> myFirstKeyRing
--- a/kms/src/main/java/com/example/Snippets.java
+++ b/kms/src/main/java/com/example/Snippets.java
@@ -1,0 +1,467 @@
+/*
+ * Copyright (c) 2017 Google Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
+ * in compliance with the License. You may obtain a copy of the License at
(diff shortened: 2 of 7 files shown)


kms/src/main/java/com/example/Snippets.java
```diff
@@ -0,0 +1,448 @@
...
+    CloudKMS kms = createAuthorizedClient();
+
+    // The resource name of the cryptoKey version
+    String cryptoKeyVersion = String.format(
+        "projects/%s/locations/%s/keyRings/%s/cryptoKeys/%s/cryptoKeyVersions/%s",
+        projectId, location, ringId, keyId, version);
+
+    DestroyCryptoKeyVersionRequest destroyRequest = new DestroyCryptoKeyVersionRequest();
+
+    CryptoKeyVersion destroyed = kms.projects().locations().keyRings().cryptoKeys()
```
lesv: This is probably what our style guide says to do, but it seems like 
```
kms.projects()
    .locations()
    .keyRings()
    .cryptoKeys()
    .cryptoKeyVersions()
    .destroy(cryptoKeyVersion, destroyRequest)
    .execute();
```
is more appropriate.

kms/src/main/java/com/example/Snippets.java
```diff
@@ -0,0 +1,448 @@
...
+
+    // Get the current IAM policy and add the new account to it.
+    Policy iamPolicy = getCryptoKeyPolicy(projectId, ringId, keyId);
+
+    List<Binding> bindings = iamPolicy.getBindings();
+    // Filter out the given member
+    for (Binding b : bindings) {
+      if (role.equals(b.getRole()) && b.getMembers().contains(member)) {
+        b.getMembers().remove(member);
+        /*
```
lesv: Huh?

kms/README.md
```diff
@@ -0,0 +1,25 @@
...
+
+You can run the quickstart with:
+
+    java -cp target/kms-samples-1.0.0-jar-with-dependencies.jar \
+        com.example.Quickstart [your-project-id]
+
+and can see the available snippet commands with:
+
+    java -cp target/kms-samples-1.0.0-jar-with-dependencies.jar \
+        com.example.Snippets
```
lesv: Do you want to mention any of the command line arguments here?

jerjou: After you run this, it'll show you the available arguments, which you can run, which will in turn show the available arguments for that subcommand.

I suppose having an actual example invocation wouldn't hurt..

kms/src/main/java/com/example/Snippets.java
jerjou: Ah. Right.
Since `.remove` can be optionally implemented, I wasn't sure if the `List` returned by `getMembers()` had it, so this was the contingency plan. Turns out it does, so this works, but I forgot I had the contingency here :-) Removed.

kms/src/main/java/com/example/Snippets.java
jerjou: I kind of like having most of the path in the same line, because it's just boilerplate to get to the main thing you care about (which is on the next line). That way the thing you care about is distinct 
from the noise.

jerjou:
Updated - renamed a bunch of the command-line arguments to be more verbose, to be consistent with the naming from the API (and normalized some other inconsistent naming I had).

Having trouble with ADC while running tests 😫 It's mentioning a mysterious project id... sigh...

lesv:
Once you figure out your mystery, you might wish to add the sample to the parent pom.xml to get tested.

PR #497, opened 2017-02-02
Add Speech API quickstart sample.

jmdobry:


speech/cloud-client/pom.xml
```diff
@@ -0,0 +1,59 @@
+<!--
+  Copyright 2017 Google Inc. All Rights Reserved.
```
lesv: No need for "All Rights Reserved."

speech/cloud-client/README.md
```diff
@@ -0,0 +1,28 @@
...
+Install [Maven](http://maven.apache.org/).
+
+Build your project with:
+
+  mvn clean package -DskipTests
+
+You can then run a given `ClassName` via:
+
+  mvn exec:java -Dexec.mainClass=com.example.speech.ClassName \
+      -DpropertyName=propertyValue \
```
lesv: I don't see the need for any arguments in your sample.

jmdobry: Done.

speech/cloud-client/pom.xml
jmdobry: Done.

speech/cloud-client/pom.xml
```diff
@@ -0,0 +1,59 @@
+<!--
```
jerjou: Need to add this to the top-level pom.xml in order for the tests to run

jmdobry: Really? The test definitely ran during the [travis build](https://travis-ci.org/GoogleCloudPlatform/java-docs-samples/builds/197720320), or are you refereing to something else?

jerjou: Huh.
...okay, yeah - it's a quirk of how we run travis builds. ie for PRs, we don't run the full suite of tests - we see what files have changed, then cd into that directory before running tests. So for the PR, it'll run the right tests, but when running the full suite, it won't know about these tests without being in the parent pom.

PR #498, opened 2017-02-02
Translate use nmt

gguuss:
This renames the translate classpath to something other than Google and also changes the default model to NMT so we can highlight the latest features.


gguuss:
FYI @lesv, a really tiny change for Translate.

PR #500, opened 2017-02-03
Add getting started hello world sample for Endpoints gRPC.

jeffmendoza:
cc @lesv

Hello World code is from here: https://github.com/grpc/grpc-java/tree/master/examples

I added the cli lib to the client and the interceptor class to insert the api key. Also, I separated out the api/client/server into different directories.

I added the endpoints integration including the api_config.yaml, Dockerfile, and container-engine.yaml


jeffmendoza:
cc @gguuss In case you're interested in how the gradle config came out.

endpoints/getting-started-grpc/README.md
```diff
@@ -0,0 +1,130 @@
+# Endpoints Getting Started with gRPC & Java Quickstart
+
```
jeffmendoza: Added some links. I wanted to keep this terse, as the Quickstart on c.g.c will have all the details. I'll link to that when it is published.

endpoints/getting-started-grpc/README.md
```diff
@@ -0,0 +1,130 @@
...
+    ```
+    gcloud service-management deploy out.pb api_config.yaml
+    # The Config ID should be printed out, looks like: 2017-02-01r0, remember this
+
+    # set your project to make commands easier
+    GCLOUD_PROJECT=<Your Project ID>
+
+    # Print out your Config ID again, in case you missed it
+    gcloud service-management configs list --service hellogrpc.endpoints.${GCLOUD_PROJECT}.cloud.goog
```
jeffmendoza: This is correct. Endpoints needs a globally unique name for the service. We were previously using <project>.appspot.com, but that was causing confusion, as it's not necessarily using AppEnine, and is actually setting up DNS to wherever your API is running. <project>.cloud.goog is a new namespace available, so all the endpoints docs are using <apiname>.endpoints.<project>.cloud.goog

jeffmendoza:
@lesv Comments addressed. Let me know if there is a boilerplate you want me to use at the top of the README. I'm mostly just trying to paste all the bash commands needed to make the sample work, so the writers can use them for the Quickstart.

I forgot to mention that Gradle doesn't need to be installed the wrapper takes care of that (./gradlew)

PR #502, opened 2017-02-06
adds KMS region tags

bshaffer:
Adds region tags for all relevant samples/functions. 

However, the [quickstart for datastore](https://cloud-dot-devsite.googleplex.com/datastore/docs/reference/libraries#client-libraries-usage-java) looks a little bit different (just the name of the function, and the fact it's in a dedicated file). Can we get this refactored for Java? I can try to do it but it might take a minute.


bshaffer:
I get the following error when I try to run the EncryptFileCommand:

```sh
# java -cp target/kms-samples-1.0.0-jar-with-dependencies.jar com.example.CryptFileCommands
Error: Main method not found in class com.example.CryptFileCommands, please define the main method as:
   public static void main(String[] args)
```

bshaffer:
nm, I was able to figure it out and fix the corresponding index error! PTAL.

bshaffer:
I got ambitious and added `Quickstart.java`. I compiled and it worked. I tried to match it to the other quickstarts, but LMK if I missed something!

bshaffer:
@lesv no because the README is accurate - I updated the code to match :D. I can update it if you'd like, however. 

Follow-up question, are the [Quickstart instructions](https://cloud.google.com/kms/docs/reference/libraries#client-libraries-install-java) accurate? I was able to run without modifying `pom.xml`, so can this and/or the gradle instructions be removed?

lesv:
QS Instructions doesn't even come close to matching what's in the pom.xml with the exclude of guava and bringing in a good version.  We should fix.

jerjou:
I originally had it so that the Quickstart was in a separate file, but then when I went to test it, it seemed weird to me that you had a quickstart to list keyrings, without any ability to create keyrings. Which is why I put it in with the rest in Snippets. Seemed to me that it can be included in the docs from the Snippets just as easily as from the Quickstart, and it didn't seem to make that much sense as a standalone file... no?

jerjou:
It also seems weird to me that we're cluttering up our source code with region tags just to surround a function, but I seem to have lost that battle.

Added a test for Quickstart - LGTM. Thanks for catching that command index error! :)

PR #515, opened 2017-02-14
bump endpoints-framework version to 2.0.1

tangiel:


lesv:
We have a bot that does this every evening.

PR #518, opened 2017-02-16
Change Maven size

lesv:


lesv:
LGTM - push so Travis will always pick this up.

PR #523, opened 2017-02-17
Get full IT working again

lesv:
mvn verify
* Spanner Works
1. remove comments from pom.xml
2. explicitly pass the SystemPropertyValues in local pom — not sure how
it worked earlier.
3. Fix a comment in the QuickStart sample
4. Restore stdOut in both IT samples

### DIFF
--- a/storage/storage-transfer/README.md
+++ b/storage/storage-transfer/README.md
@@ -54,18 +54,21 @@ Creating a one-time transfer from Amazon S3 to Google Cloud Storage.
   1. Set the `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY` environment variables.
 1. In AwsRequester.java, fill in the user-provided constants.
 1. Compile the package with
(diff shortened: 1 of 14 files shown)


lesv:
Still not passing, but close and I have to merge to get the updated memory settings.

storage/storage-transfer/README.md
```diff
@@ -54,18 +54,21 @@ Creating a one-time transfer from Amazon S3 to Google Cloud Storage.
   1. Set the `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY` environment variables.
 1. In AwsRequester.java, fill in the user-provided constants.
 1. Compile the package with
-   ```
-   mvn compile
-   ```
+
```
jerjou: This, unfortunately, resets the numbering. We should either use bulletted lists, or indent the heck out of this block in hopes that it'll format correctly :-)

PR #525, opened 2017-02-17
Give Maven a bit more memory

lesv:


lesv:
Will first watch the Before, then merge and watch the after.

PR #527, opened 2017-02-18
bump versions and switch to uberjars

tangiel:


lesv:
I don't seem to be able to update your branch -- we just updated the build environment yesterday, so you'll need to do that before Travis will pass.

tangiel:
I rebased on the latest.

lesv:
Looks like an error in the updated script.  I'm going to merge -- and if there are any issues - will fix over the weekend.

PR #529, opened 2017-02-20
Remove SUDO (ie. make a Container again)

lesv:


lesv:
Merged to make sure SUDO is back on.

PR #531, opened 2017-02-21
revert endpoints-framework-all dependency

tangiel:
This clashes with the GetSwaggerDoc task, since the tool doesn't depend
on the -all dependency, so we revert it for now.


frankyn:
Hi Daniel,

I have two quesitons:

1. How does `-all` dependency clash with the GetSwaggerDoc task?

1. I tried running the GetSwaggerDoc task and SLF4J had an incompatibility issue:
```
SLF4J: The requested version 1.6.99 by your slf4j binding is not compatible with [1.5.5, 1.5.6]
SLF4J: See http://www.slf4j.org/codes.html#version_mismatch for further details.
```

PR #538, opened 2017-03-01
Add jenkins scripts & deployment test for sparkjava

jerjou:


jerjou:
Addresses #538

jerjou:
Updated to make the jenkins script more robust (deleting the test version as you mentioned (which requires telling the deploy not to start serving all traffic -_-), and also doing so if there was an error), and incorporate another jenkins test that was in the jenkins shell command.

PTAL

jenkins.sh
```diff
@@ -0,0 +1,27 @@
+#!/bin/bash
```
lesv: copyright?

jenkins.sh
```diff
@@ -0,0 +1,27 @@
...
+
+set -x
+
+delete_app_version() {
+  yes | gcloud --project=$GOOGLE_PROJECT_ID \
```
lesv: https://google.github.io/styleguide/shell.xml
Quote your variables and prefer "${GOOGLE_PROJECT_ID}"

jenkins.sh
```diff
@@ -0,0 +1,27 @@
...
+  delete_app_version
+
+  exit $errcode
+}
+trap handle_error ERR
+
+# Find all jenkins.sh's and run them.
+find . -mindepth 2 -maxdepth 5 -name jenkins.sh -type f | while read path; do
+  (
+  pushd "${path%jenkins.sh}"
```
lesv: % not matched.

jerjou: Doesn't have to be: http://wiki.bash-hackers.org/syntax/pe#substring_removal

jenkins.sh
jerjou: Oh right. Not used to copyrighting my shell scripts ^_^

jerjou:
Fixed, and added `shellcheck` for style checking.

appengine/helloworld/jenkins.sh
```diff
@@ -0,0 +1,25 @@
...
+
+set -xe
+
+mvn clean appengine:update \
+  -Dappengine.additionalParams="--service_account_json_key_file=${GOOGLE_APPLICATION_CREDENTIALS}" \
+  -Dappengine.appId="${GOOGLE_PROJECT_ID}" \
+  -Dappengine.version="${GOOGLE_VERSION_ID}" \
+  -DskipTests=true
+
+curl -f "http://${GOOGLE_VERSION_ID}-dot-${GOOGLE_PROJECT_ID}.appspot.com/"
```
lesv: Do you check any of the results?

jerjou: I just moved this wholesale from the jenkins script, so don't know much about it. Presumably it's just to make sure it returns a 200, and it looks like curl returns a non-zero exit code if it doesn't.

tswast: Correct. The [`curl -f` is checking for a non-successful response](https://curl.haxx.se/docs/manpage.html#-f).

Obviously we might want to check for more than just the error code, but this actually catches quite a lot of possible errors.

lesv:
Based on Tim's comments, I'm going to merge.  Travis is currently unhappy w/ Spanner again, I'm going to fix on a separate thread.

jerjou:
Yup - I would have merged yesterday, but I wanted to coordinate this change with a change to the jenkins config (ie how to run jenkins tests). Will make that change now.

PR #540, opened 2017-03-03
Fix jenkins script.

jerjou:
* Fail when one of the tests fails
* Deploy to different app versions
* Clean up app versions in the background after each test

### DIFF
--- a/jenkins.sh
+++ b/jenkins.sh
@@ -14,12 +14,16 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-set -x
+set -xe
 shopt -s globstar
 
+app_version=""
+
 delete_app_version() {
-  yes | gcloud --project="$GOOGLE_PROJECT_ID" \
-    app versions delete "$GOOGLE_VERSION_ID"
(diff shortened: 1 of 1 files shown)


jenkins.sh
```diff
@@ -36,10 +40,18 @@ shellcheck ./**/*.sh
...
-  pushd "${path%jenkins.sh}"
-  /bin/bash ./jenkins.sh
+  pushd "$dir"
+  # Need different app versions because flex can't deploy over an existing
+  # version
+  GOOGLE_VERSION_ID="$app_version" /bin/bash ./jenkins.sh
+  echo "Return code: $?"
+
+  # Clean up the app version in the background
+  delete_app_version "$app_version" &
```
tswast: You might want to [nohup](http://stackoverflow.com/questions/15595374/whats-the-difference-between-nohup-and-ampersand) this, actually. That way if the script terminates, it'll finish the clean-up script.

jerjou: Good thinking - added.

lesv:
Ok to merge w/o Travis -- I had hoped to fix it yesterday, will today.

PR #544, opened 2017-03-05
remove "secure: always"

bmenasha:
the "secure" directive it does nothing in Flexible and it's presence can imply support for this feature.


bmenasha:
trying to close in favor of another

PR #545, opened 2017-03-05
remove "secure: always" from Flexible samples.

bmenasha:
The "secure" directive is ignored in Flexible.


lesv:
Expected failure.

bmenasha:
>  there something else that should be there, or is Flex now always secure?

There is no replacement for this feature in the Flexible platform see internal issue b/28618963

Thanks

PR #548, opened 2017-03-06
Updates vision to the latest library

gguuss:


vision/cloud-client/src/test/java/com/example/vision/DetectIT.java
```diff
@@ -227,4 +224,78 @@ public void testPropertiesGcs() throws Exception {
...
+  @Test
+  public void testCropHintsGcs() throws Exception {
+    // Act
+    String[] args = {"crop", "gs://cloud-samples-tests/vision/wakeupcat.jpg"};
+    Detect.argsHelper(args, out);
+
+    // Assert
+    String got = bout.toString();
+    assertThat(got).contains("vertices {");
+    assertThat(got).contains("x: 599");
```
lesv: Will this always be true -- including possible changes?

lesv:
LGTM

vision/cloud-client/src/test/java/com/example/vision/DetectIT.java
gguuss: It's entirely possible that the API result could change over time for these. My expectation however is that the crop hints from this image should match the bounds of the image because it is already framed and cropped.

gguuss:
Thanks Les, I'll keep an eye out for any brittleness in the tests, they're passing locally for me.

PR #563, opened 2017-03-15
Update Endpoints Migration sample

frankyn:
* Updated plugin/dependency versions
* Removed discovery doc from documentation. A discovery document doesn't need to be generated manually in the new version of endpoints frameworks.
* Removed iOS from samples because it's not yet tested.
* Moved directory from discovery/ to migration-example/
* Tested sample using Android


lesv:
Argh - things are conflicting.

frankyn:
It's on my radar. I'll take care of it after my next meeting. I've been slow in fixing this.

lesv:
LGTM once circle passes.

PR #567, opened 2017-03-17
Monitoring VTK snippets

waprin:
Adds VTK Monitoring snippets

### DIFF
--- a/monitoring/cloud-client/src/main/java/com/example/monitoring/Snippets.java
+++ b/monitoring/cloud-client/src/main/java/com/example/monitoring/Snippets.java
@@ -1,0 +1,592 @@
+/*
+ * Copyright 2016 Google Inc. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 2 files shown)


monitoring/cloud-client/src/main/java/com/example/monitoring/Snippets.java
```diff
@@ -0,0 +1,592 @@
...
+  }
+
+  /**
+   * Demonstrates listing time series and aggregating them.
+   *
+   */
+  void listTimeSeriesAggregrate() throws IOException {
+    // [START monitoring_read_timeseries_align]
+    MetricServiceClient metricServiceClient = MetricServiceClient.create();
+    String projectId = System.getProperty("projectId");
```
lesv: Isn't there another way to get this?   (Most of the veneer's have it - now)

lesv:
Looks like there is more than one typo.

waprin:
No it's just that this project doesn't actually have any compute engine instances so no time series. Unfortunately have to make tests less useful to make them more general..

PR #569, opened 2017-03-18
Information on the 3 api clients.

cobookman:
Customer was confused around why there is 3 api clients. Can we remove the rest, src api clients from the repo? If not let's put an explanation as to how these 3 api clients differ, and recommend to always use the cloud-client if possible.


lesv:
@tswast Care to comment -- I'm going to approve the PR, but I'd like your thoughts on removing the  REST client.  (I'm ok with it.)

tswast:
Please do not remove the rest client samples.

Per agreement with the veneer team, I cannot include API client samples on documentation pages, but the cloud client does not support all features, so if I need/want to document a feature that they aren't going to get to for a few quarters, this is the workaround: make a REST sample.

PR #573, opened 2017-03-20
Setup circleci ADC

jerjou:
Also, move all tests to use the project given in the environment variable, using buckets named after the project.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -1,0 +1,16 @@
+machine:
+  java:
+    version: oraclejdk8
+  environment:
+    chromedriver_path: /usr/local/bin/chromedriver
+dependencies:
+  pre:
+    - sudo mkdir /opt/webdriver/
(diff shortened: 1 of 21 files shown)


circle.yml
```diff
@@ -0,0 +1,17 @@
+machine:
+  java:
+    version: oraclejdk8
```
lesv: openjdk8 ??

jerjou: Yeah, for some reason I thought we were using oracle. Are we using openjdk?

jerjou: Hm.. doesn't look like circle [has openjdk8](https://circleci.com/docs/1.0/build-image-precise/#java), so if we want java 8, we have to go with oracle.

jerjou:
Okay - spanner samples were failing intermittently, but I think it's because both travis and circle were running at the same time and clobbering state. I'm going to merge and then turn off travis.

PR #581, opened 2017-03-24
Testing tweaks

jerjou:
Remove chromedriver var, since it's in the path, don't print out the secrets, etc.


jerjou:
Merging for now, so I can swap out the creds. LMK if there's any concerns.

PR #584, opened 2017-03-27
Fix region tags

waprin:


waprin:
@lesv FYI ,force merging since it's just region tag comments

lesv:
LGTM

PR #585, opened 2017-03-27
Fixes the Endpoints getting-started sample.

wmwong:
The Dockerfile and openapi.yaml cannot be in the appengine directory.
These have been moved out into their own directories.

Also, the jetty user needs access to the war file. The Dockerfile has
been updated accordingly.


lesv:
@frankyn PTAL
@wmwong - A bit surprised by this, but Frank should know for sure.

frankyn:
@lesv I agree with this change.

**2 cents:**
There are issues in the [associated doc on cloudsite](https://cloud.google.com/endpoints/docs/quickstart-app-engine) that need to be updated as well. Instructions expect files to be located in `java-docs-samples/endpoints/getting-started/`, but they're located in `endpoints/getting-started/src/main/appengine/`. 

**Action-item** Paths need to be updated in the [associated doc on cloudsite](https://cloud.google.com/endpoints/docs/quickstart-app-engine) along with this PR.

lesv:
@dpebot merge when green

lesv:
Docs have been updated, waiting on review from Frank and Lawrence.

frankyn:
Hi @wmwong,

I have a question, why did you place the openapi.yaml in `src/main/openapi` instead of in the root of this sample `endpoints/getting-started`? I'm not sure if there were complications, but if not we should move this down to the root of the sample. I can commit a change for this.

PR #588, opened 2017-03-29
Add tasks samples

waprin:
PR for a branch rather then master since it's still a white-list only alpha.


lesv:
Should we publish this?

PR #601, opened 2017-04-07
Speech v1

gguuss:
Addresses changes in the new library, adds streaming example, and updates GCS uri in README to a public one.


gguuss:
@lesv /FYI I added a new streaming example and have updated to v1 which is in the process of releasing

PR #606, opened 2017-04-10
Removes deprecated async local file method

gguuss:


gguuss:
@lesv FYI, it looks like async recognize on local files is discouraged, removing from sample.

lesv:
@dpebot merge when green

PR #608, opened 2017-04-11
BQ: remove API client lib samples. Use a job for queries.

tswast:
The documentation no longer has any references to the API client library
samples, even in translations, so I'm removing these samples to prevent
confusion.

Per a conversation with BigQuery engineering team, I am updating all
samples to explicitly create a query job, rather than using the
jobs.query method. There isn't a distinction between "synchronous" and
"asynchronous".

Updating the query parameters sample is blocked by https://github.com/GoogleCloudPlatform/google-cloud-java/issues/1892


bigquery/cloud-client/src/test/java/com/example/bigquery/QuerySampleIT.java
```diff
@@ -0,0 +1,128 @@
...
+import org.junit.Before;
+import org.junit.Ignore;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.junit.runners.JUnit4;
+
+import java.io.ByteArrayOutputStream;
+import java.io.PrintStream;
+
+/** Tests for synchronous query sample. */
```
jabubake: I assume "synchronous" is not relevant here as well.

tswast: Done.

PR #617, opened 2017-04-17
Switch async GCS example to long file

danaharon:


lesv:
CLA status was good before gus did his thing, and I know Gus, so I'll manually merge.

PR #618, opened 2017-04-17
Speech client library version update.

gguuss:


gguuss:
@lesv FYI, there's an improvement in async calls in the updated version of the client library.

PR #625, opened 2017-04-22
Add a bit more to the readme

lesv:
1. Put an index up (Note that we have to GAE Standard samples at the top level)
2. Add some info about credentials.

@jabubake @tswast FYI


lesv:
I'm going to merge as it's a README

PR #630, opened 2017-04-25
Update pom.xml

piaxc:
Updating to include google client api libraries.


lesv:
@dpebot merge when green

lesv:
Ah, circle isn't going to run :(

PR #632, opened 2017-04-26
Flex app merge

jabubake:
After looking at how how we use these samples in our docs, I think our vision of having the single app must not burden users who just want to see the single servlet Flex sample demonstrating a particular functionality. It should also not burden the contributors.

Let me know if this current approach makes sense. the samples-builder under flexible/java-runtime/samples creates the bundled app on the fly from all or a subset of the samples.

The only part I'm having trouble with is ensuring this works with the auto-updates for dependencies  by dpebot.


jabubake:
Merge issues with master, will recreate pull request after i fix

PR #627, opened 2017-04-24
Adds an Endpoints sample with multiple versions.

wmwong:
This is a copy of the existing getting-started sample with extensions to support multiple versions.

The difference in the API between v1 and v2 is the use of "message" vs "msg".

The code sample shows:
- Simple input layer that receives API requests from multiple major versions (v1/v2 servlets).
- Simple translation to a backend format (v1/v2 translators).
- Simple backend processing (uppercasing the message).
- Simple translation to the response format (v1/v2 translators).

No integration tests were written (similar to the other Endpoints samples) because this sample doesn't call any services. Instead, it is meant to be a sample that is deployed to GCP.

PR #642, opened 2017-05-02
Updates the Endpoints getting-started sample.

wmwong:
Adds the same updates in the comments in
https://github.com/GoogleCloudPlatform/java-docs-samples/pull/627.

Also fixes the variables to replace in app.yaml.


lesv:
@dpebot merge when green

PR #665, opened 2017-05-16
J8 samples

lesv:


lesv:
@jabubake I doubt you'll be able to review the whole thing, and there is more to go.  The basic changes are make sure all files in `appengine-java8` have 1.3.1 maven plugin, and `<runtime>java8</runtime>` everywhere.  I found some `appengine` that seem to have MVM config, and don't really build.

PR #671, opened 2017-05-17
Fix variable substitution

lesv:
Add pretty back in to give us variable substitution.

This is really overkill, but I don’t want to fix all the substitution
methods yet.


lesv:
@jabubake FYI / PTAL

PR #672, opened 2017-05-19
web.xml and Annotations

lesv:
1. Upgrade all web.xml to 3.1
2. Add annotations
3. Hello World is old, remove this one.


lesv:
I'm going to merge so that doug can use this layout of files.

PR #676, opened 2017-05-24
[Endpoints Framework] Updating migration sample for endpoints framework plugin GA

frankyn:
Changes:
- Version changes to the sample to reflect the Endpoints Frameworks plugin GA release. 
- Added information for generating a discovery document in the README
- Set `hostname` for the generated discovery document in `pom.xml` and `build.gradle`.


frankyn:
@lesv PTAL when you have a moment.

appengine/endpoints-frameworks-v2/migration-example/build.gradle
```diff
@@ -20,10 +20,10 @@ buildscript {    // Configuration for building
   }
   dependencies {
     // App Engine Gradle plugin
-    classpath 'com.google.cloud.tools:appengine-gradle-plugin:1.3.0'
+    classpath 'com.google.cloud.tools:appengine-gradle-plugin:1.3.1'
```
lesv: Using + w/ Gradle will get the latest.

PR #682, opened 2017-05-30
Use the socketFactory everywhere

lesv:
1. No longer use the Google Connector/j


lesv:
@jabubake FYI

appengine-java8/cloudsql/src/main/webapp/WEB-INF/appengine-web.xml
```diff
@@ -16,10 +16,10 @@
   <threadsafe>true</threadsafe>
   <runtime>java8</runtime>
 
-  <use-google-connector-j>true</use-google-connector-j>
+  <!-- <use-google-connector-j>true</use-google-connector-j> NO LONGER REQUIRED -->
```
jabubake: should this be in here, if it is no longer required ?

PR #684, opened 2017-05-31
Pubsub getting started

jabubake:
Initial commit (WIP)


pubsub/cloud-client/src/main/java/com/example/pubsub/CreatePullSubscriptionExample.java
```diff
@@ -0,0 +1,62 @@
...
+    String topicId = args[0];
+
+    // Your subscription ID eg. "my-sub"
+    String subscriptionId = args[1];
+
+    TopicName topicName = TopicName.create(projectId, topicId);
+
+    // Create a new subscription
+    SubscriptionName subscriptionName = SubscriptionName.create(projectId, subscriptionId);
+    try (SubscriptionAdminClient subscriptionAdminClient = SubscriptionAdminClient.create()) {
```
lesv: Rather than throw an exception (above), shouldn't we catch them and explain why they might get one and what to do if they do?

pubsub/cloud-client/src/main/java/com/example/pubsub/PublisherExample.java
```diff
@@ -0,0 +1,74 @@
...
+public class PublisherExample {
+
+  static final int MESSAGE_COUNT = 5;
+
+  // use the default project id
+  private static final String PROJECT_ID = ServiceOptions.getDefaultProjectId();
+
+  //schedule a message to be published, messages are automatically batched
+  private static ApiFuture<String> publishMessage(Publisher publisher, String message)
+      throws Exception {
```
lesv: Rather than throwing an exception, shouldn't we catch them? and explain what's going on?

pubsub/cloud-client/src/main/java/com/example/pubsub/CreatePullSubscriptionExample.java
jabubake: Submitted https://github.com/GoogleCloudPlatform/google-cloud-java/issues/2295

lesv:
Ping

PR #689, opened 2017-06-02
gaeinfo

lesv:
Simple Servlet that displays most info


lesv:
Useful to have around.

PR #692, opened 2017-06-07
Add a comment about @WebServlet

lesv:


lesv:
Testing isn't going to run due to issues w/ circle.

lesv:
@jabubake @frankyn FYI - I'm going to merge so that doc samples can be republished.

Issue #687, opened 2017-06-01
/iap sample fails approximately 70%


DazWilkin:
## In which file did you encounter the issue?

"Cloud Identity-Aware Proxy Java Samples"
/iap

## Did you change the file? If so, how?

no changes

## Describe the issue

Test fails significantly (~70%) but I see no evident pattern.

The error is always:

`testGenerateAndVerifyIapRequestIsSuccessful(com.example.iap.BuildAndVerifyIapRequestIT): The Token's Signature resulted invalid when verified using the Algorithm: SHA256withECDSA`

I will add some logging to the sample tomorrow to see whether I can find any correlation.

I ran a simple repetition of the documented test:

    for t in {1..100}; do
      mvn -Dtest=com.example.iap.BuildAndVerifyIapRequestIT verify \
      | egrep "Tests run: [0-9], Failures: [0-9], Errors: [0-9], Skipped: [0-9], Time elapsed: [0-9]*.[0-9]* sec"; \
      sleep 60s; \
    done
```
Tests run: 2, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 2.419 sec <<< FAILURE!
Tests run: 2, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 2.276 sec
Tests run: 2, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 2.231 sec <<< FAILURE!
Tests run: 2, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 2.325 sec <<< FAILURE!
Tests run: 2, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 2.451 sec <<< FAILURE!
Tests run: 2, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 2.363 sec <<< FAILURE!
Tests run: 2, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 2.176 sec <<< FAILURE!
Tests run: 2, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 2.21 sec
Tests run: 2, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 2.262 sec <<< FAILURE!
Tests run: 2, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 2.295 sec <<< FAILURE!
Tests run: 2, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 2.275 sec
[...]

PR #694, opened 2017-06-08
Iap sample v2

jabubake:
Using jjwt instead of auth0 jwt library. Auth0 jwt library has intermittent token verification failures for IAP on OpenJDK.


jabubake:
Resolves #687

lesv:
Circle will fail at the moment - Ok to merge as Admin.

PR #699, opened 2017-06-12
BuildIapRequest: Switch aud to client ID

matthewg:
Client ID is the preferred thing to use for the aud claim.
--msachs@, IAP TLM


jabubake:
The integration test : https://github.com/GoogleCloudPlatform/java-docs-samples/blob/master/iap/src/test/java/com/example/iap/BuildAndVerifyIapRequestIT.java will need an update as well,  I'll clone your PR and make the appropriate changes. Keeping this PR open till the other is ready.

jabubake:
@matthewg Cloned here : https://github.com/GoogleCloudPlatform/java-docs-samples/compare/patch-iap-samples. PTAL.

PR #732, opened 2017-06-27
Adds missing dependency until the language library is fixed.

gguuss:
Also fixes missing parent from language pom.xml and resolves the lint issues that appear in the tests.


language/analysis/pom.xml
```diff
@@ -34,6 +41,13 @@ limitations under the License.
     </dependency>
     <!-- [END dependencies] -->
 
+    <!-- FIXME: remove after client fixes depenency issue -->
```
jabubake: Gus, can you link the tracking issue : https://github.com/GoogleCloudPlatform/google-cloud-java/issues/2192, just so we can track when to remove

PR #738, opened 2017-06-29
Postgres & CloudSQL update for flex

lesv:
1. update cloudSql to latest plugin and make the sample a bit cleaner.
2. add postgresql sample.


lesv:
I'm going to merge, but I'd rather the init method threw all the things I catch.

lesv:
Well, I can't change init() to add the other types, so I'm going to keep this.

PR #746, opened 2017-07-06
Closes image annotator client.

gguuss:


vision/cloud-client/src/main/java/com/example/vision/Detect.java
```diff
@@ -261,9 +259,10 @@ public static void detectLabels(String filePath, PrintStream out) throws IOExcep
         AnnotateImageRequest.newBuilder().addFeatures(feat).setImage(img).build();
     requests.add(request);
 
-    BatchAnnotateImagesResponse response =
-        ImageAnnotatorClient.create().batchAnnotateImages(requests);
+    ImageAnnotatorClient client = ImageAnnotatorClient.create();
```
lesv: I'm going to say Ok to this, but in the future, it might be better to use a [try w/ resources](https://docs.oracle.com/javase/tutorial/essential/exceptions/tryResourceClose.html) it will auto call close.

gguuss: Done.

gguuss:
Changed to try with resources, much cleaner.

Issue #747, opened 2017-07-10
gradle build not working for appengine-java8/endpoints-v2-backend


mpoehler:
gradle build appengine-java8/endpoints-v2-backend does not work. 

During the build the task replaceProjectId replaces the projectId in the wrong file. Is currently exploded-backend, but must be exploded-endpoints-v2-backend. 

I'll provide a pull-request in a minute.

mpoehler:
Fixed with pull request #748

PR #748, opened 2017-07-10
fix for #747, replaceProjectId wrong directory

mpoehler:
also adopted the sourceCompatibility & targetCompatibility to 1.8

### DIFF
--- a/appengine-java8/endpoints-v2-backend/build.gradle
+++ b/appengine-java8/endpoints-v2-backend/build.gradle
@@ -54,14 +54,14 @@ endpointsServer {
   hostname = "echo-api.endpoints.${projectId}.cloud.goog"
 }
 
-sourceCompatibility = 1.7  // App Engine Standard uses Java 7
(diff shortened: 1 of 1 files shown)


appengine-java8/endpoints-v2-backend/build.gradle
```diff
@@ -54,14 +54,14 @@ endpointsServer {
...
-targetCompatibility = 1.7  // App Engine Standard uses Java 7
+sourceCompatibility = 1.8
+targetCompatibility = 1.8
 
 // this replaces the ${endpoints.project.id} in appengine-web.xml and web.xml
 task replaceProjectId(type: Copy) {
     from 'src/main/webapp/WEB-INF/'
     include '*.xml'
-    into 'build/exploded-backend/WEB-INF'
+    into 'build/exploded-${archivesBaseName}/WEB-INF'
```
frankyn: Use double-quotes as the single-quotes doesn't substitute ${archivesBaseName}.

mpoehler:
Sorry! Next time I double check the PR is working!

PR #755, opened 2017-07-12
Spring Integration channel adapter code samples

joaoandremartins:
Aid on getting started with the Spring Integration GCP Pub/Sub channel
adapters.


joaoandremartins:
@jabubake @meltsufin @artembilan can you please take a look?

pom.xml
```diff
@@ -89,6 +89,7 @@
         <module>pubsub/cloud-client</module>
         <module>spanner/cloud-client</module>
         <module>speech/cloud-client</module>
+        <module>spring/integration/pubsub</module>
```
jabubake: Lets just use spring/pubsub for now

spring/integration/pubsub/pom.xml
```diff
@@ -0,0 +1,130 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
jabubake: All the files are missing the license header.

spring/integration/pubsub/src/main/java/com/example/spring/pubsub/PubsubApplication.java
```diff
@@ -0,0 +1,153 @@
+package com.example.spring.pubsub;
```
jabubake: License header missing

spring/integration/pubsub/src/main/java/com/example/spring/pubsub/PubsubApplication.java
```diff
@@ -0,0 +1,153 @@
...
+import com.google.pubsub.v1.Subscription;
+import com.google.pubsub.v1.SubscriptionName;
+import com.google.pubsub.v1.Topic;
+import com.google.pubsub.v1.TopicName;
+import java.io.IOException;
+
+import com.google.cloud.pubsub.v1.AckReplyConsumer;
+import com.google.protobuf.ByteString;
+
+import java.util.List;
```
jabubake: Please use [Google Java Formatter](https://github.com/google/google-java-format)  for the source files.

spring/integration/pubsub/src/main/java/com/example/spring/pubsub/PubsubApplication.java
```diff
@@ -0,0 +1,153 @@
...
+  private PubsubOutboundGateway messagingGateway;
+
+  @Autowired
+  private PubsubAdmin admin;
+
+  public static void main(String[] args) throws IOException {
+    SpringApplication.run(PubsubApplication.class, args);
+  }
+
+  @GetMapping("/listTopics")
```
jabubake: Might be worth moving the @RestController and mappings to a different class and keeping the Application class simple (less cognitive load)

spring/integration/pubsub/src/main/java/com/example/spring/pubsub/PubsubApplication.java
```diff
@@ -0,0 +1,153 @@
...
+  public List<String> listTopics() {
```
jabubake: public methods : please provide a Javadoc

spring/integration/pubsub/pom.xml
```diff
@@ -0,0 +1,130 @@
...
+
+  <properties>
+    <spring-cloud-gcp.version>1.0.0.BUILD-SNAPSHOT</spring-cloud-gcp.version>
+    <spring-boot.version>1.5.2.RELEASE</spring-boot.version>
+  </properties>
+
+  <dependencies>
+    <dependency>
+      <groupId>org.springframework.cloud</groupId>
+      <artifactId>spring-cloud-gcp-core</artifactId>
```
meltsufin: You shouldn't need this. It's a transitive dependency of spring-cloud-gcp-starter-pubsub.

spring/integration/pubsub/pom.xml
```diff
@@ -0,0 +1,130 @@
...
+      <version>${spring-cloud-gcp.version}</version>
+    </dependency>
+    <dependency>
+      <groupId>org.springframework.cloud</groupId>
+      <artifactId>spring-cloud-gcp-pubsub</artifactId>
```
meltsufin: Also not needed.

spring/integration/pubsub/pom.xml
```diff
@@ -0,0 +1,130 @@
...
+      <version>${spring-cloud-gcp.version}</version>
+    </dependency>
+    <dependency>
+      <groupId>org.springframework.cloud</groupId>
+      <artifactId>spring-cloud-gcp-starter-pubsub</artifactId>
+      <version>${spring-cloud-gcp.version}</version>
+    </dependency>
+    <dependency>
+      <groupId>org.springframework.cloud</groupId>
+      <artifactId>spring-integration-gcp</artifactId>
```
meltsufin: Again, I think it's a transitive dependency of spring-cloud-gcp-starter-pubsub.

spring/integration/pubsub/src/main/java/com/example/spring/pubsub/PubsubApplication.java
```diff
@@ -0,0 +1,153 @@
...
+          GcpHeaders.ACKNOWLEDGEMENT);
+      consumer.ack();
+    };
+  }
+
+  @Bean
+  @ServiceActivator(inputChannel = "pubsubOutputChannel")
+  public MessageHandler messageSender(PubsubTemplate pubsubTemplate) throws IOException {
+    PubsubMessageHandler outboundAdapter = new PubsubMessageHandler(pubsubTemplate);
+    outboundAdapter.setTopic("test");
```
meltsufin: Who creates this "topic"?

joaoandremartins: It has to have been previously created.

spring/integration/pubsub/pom.xml
joaoandremartins: This one isn't.

spring/integration/pubsub/src/main/java/com/example/spring/pubsub/PubsubApplication.java
meltsufin: Maybe it's better to provide the ability to select a topic to which to post the message then?

joaoandremartins: You can't do that dynamically.

joaoandremartins:
It sounds like this shouldn't be so much of a code sample, but probably be broken into pieces and added to a tutorial or getting started guide. I might do that, instead.

spring/integration/pubsub/pom.xml
artembilan: Why XML configs should be with the License as well?
Is that your policy?
We in Spring only care about classes (Java/Groovy/Kothlin etc.)

spring/integration/pubsub/src/main/java/com/example/spring/pubsub/PubsubApplication.java
```diff
@@ -0,0 +1,153 @@
...
+import org.springframework.web.bind.annotation.GetMapping;
+import org.springframework.web.bind.annotation.PostMapping;
+import org.springframework.web.bind.annotation.RequestParam;
+import org.springframework.web.bind.annotation.RestController;
+import org.springframework.web.servlet.view.RedirectView;
+
+@SpringBootApplication
+@IntegrationComponentScan
+@RestController
+@ComponentScan(basePackages = {"org.springframework.cloud.gcp"})
```
artembilan: We don't need `@IntegrationComponentScan` since Boot `1.5` any more.
Also it isn't clear why do you need that `@ComponentScan`.
If you don't see some beans from the Spring-GCP project, then some auto-configuration is missed.
Should be fixed there.

Remember `META-INF/spring.factories` with the entry `org.springframework.boot.autoconfigure.EnableAutoConfiguration=\` ?

spring/integration/pubsub/src/main/java/com/example/spring/pubsub/PubsubApplication.java
```diff
@@ -0,0 +1,153 @@
...
+          + ((ByteString) message.getPayload()).toStringUtf8());
+      AckReplyConsumer consumer = (AckReplyConsumer) message.getHeaders().get(
+          GcpHeaders.ACKNOWLEDGEMENT);
+      consumer.ack();
+    };
+  }
+
+  @Bean
+  @ServiceActivator(inputChannel = "pubsubInputChannel")
+  public MessageHandler receiveMessageInParallel() {
```
artembilan: It's not in parallel if you don't use `Executor` option for the `PublishSubscribeChannel`

spring/integration/pubsub/src/main/java/com/example/spring/pubsub/PubsubApplication.java
```diff
@@ -0,0 +1,153 @@
...
+  @ServiceActivator(inputChannel = "pubsubOutputChannel")
+  public MessageHandler messageSender(PubsubTemplate pubsubTemplate) throws IOException {
+    PubsubMessageHandler outboundAdapter = new PubsubMessageHandler(pubsubTemplate);
+    outboundAdapter.setTopic("test");
+    return outboundAdapter;
+  }
+
+  @Bean
+  public MessageChannel pubsubOutputChannel() {
+    return new PublishSubscribeChannel();
```
artembilan: Would be much clear to determine why do we need `PublishSubscribeChannel` here as well.
Looks like `DirectChannel` is pretty enough.

Also you might consider to not declare `@Bean` for this channel at all - and Spring Integration will take care about channel auto-creation for you by provided name in the `@ServiceActivator`.

joaoandremartins: I wanted to test delivery to all subscribers, instead of round-robin, which is what I was seeing with `DirectChannel`.

From reading `PublishSubscriberChannel` Javadoc, it seems indeed that the delivery is indeed not parallel, but single-threaded. Parallelization isn't my main focus here anyway, but rather the ability to receive the sent message in multiple subscribers at a time. Maybe I should rename the methods to `messageReceiver1()` and `messageReceiver2()` or something in those lines to keep it simpler, without going into `Executor` details..

artembilan: Really? I don't see more subscriber to this `pubsubOutputChannel`.
That is my concern to keep sample as simple as possible.

I agree about renaming subscribers to the `pubsubInputChannel` to avoid similar questions about parallelism

joaoandremartins: To your point about not needing `@Bean`, IIUC that's true for outbound adapters (and it appears we're already relying on `@ServiceActivator` to do that for us), but the inbound adapter seems to require the channel explicitly, e.g., `adapter.setOutputChannel(inputChannel);`.

artembilan: Yeah... I don't tell anything about *inbound* part here.
That is fully fine, especially that you would like to get there publish-subscribe scenario.
So, yes, definitely we have to declare `MessageChannel` bean if we need something else rather than just plain `DirectChannel`.

My concern is on this line of code **only** about this `pubsubOutputChannel` bean defintion.

joaoandremartins:
We decided to turn this into a getting started guide, instead, so I'll close this one for the time being.

PR #757, opened 2017-07-14
IAP sample update : new JWT header

jabubake:
FYI, @matthewg

Closing : https://github.com/GoogleCloudPlatform/java-docs-samples/pull/756


jabubake:
Les, agree : its because we need service account credentials to sign the JWT.

PR #759, opened 2017-07-18
Fix an issue in openapi.yaml.

liminw:
In our sample, both api_key and the auth issuer are required, the config
should use "AND" semantics.


lesv:
@frankyn FYI

PR #769, opened 2017-07-24
Add openapi-appengine.yaml file for App Engine deployments.

p-buse:
After conversation with the Endpoints team, this was the most effective solution we came up with to guide users through setting up Endpoints on App Engine. Having users uncomment code proved to be difficult. If you have another idea that could work better, please let me know. Otherwise, this will have to do.


endpoints/getting-started/openapi-appengine.yaml
```diff
@@ -0,0 +1,122 @@
+# Copyright 2015 Google Inc.
```
lesv: If your not submitting this from a Tardis, it should probably be 2017

p-buse: Ha. Fixed.

p-buse:
It's an update of a sample that was written to cover all of the Endpoints quickstart guides. Unfortunately, the host is different on App Engine, and so the Endpoints team has requested that we add a second OpenAPI configuration specifically for that.

lesv:
Ah - you committed on a fork, not a branch, so tests don't run.  I'll merge.

PR #774, opened 2017-07-28
Update to latest repo-tools

lesv:
1. check style is no longer stops us, but there are a bunch of failures
we should fix.
2. Broke out compute & flex pom.xml’s so that they exist.
3. Make more project include the parent pom.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -25,11 +25,7 @@
     <parent>
         <groupId>com.google.cloud.samples</groupId>
         <artifactId>shared-configuration</artifactId>
-        <version>1.0.4</version>
-<!--
-        <relativePath>../java-repo-tools</relativePath>
- -->
-
+        <version>1.0.5</version>
     </parent>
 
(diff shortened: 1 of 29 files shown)


pom.xml
```diff
@@ -48,58 +44,50 @@
...
-        <module>flexible/mailgun</module>
-        <module>flexible/mailjet</module>
-        <module>flexible/memcache</module>
-        <module>flexible/postgres</module>
-        <module>flexible/pubsub</module>
-        <module>flexible/sendgrid</module>
-        <module>flexible/sparkjava</module>
-        <module>flexible/static-files</module>
-        <module>flexible/twilio</module>
+        <module>flexible</module>
```
jabubake: Awesome !

lesv:
Still broken are:

* appengine-guestbook-cloud-datastore-j8
* iap-samples
* cloud-logging-samples
* vision-landmark-detection

PR #778, opened 2017-07-31
Adds support for word time offset

gguuss:


speech/cloud-client/pom.xml
```diff
@@ -34,7 +34,7 @@
   </properties>
 
   <!-- FIXME(lesv) - temp to fix an issue w/ GA Datastore -->
```
lesv: You could remove this and the commented out stuff.

speech/cloud-client/pom.xml
```diff
@@ -33,25 +33,12 @@
...
-      </dependency>
-    </dependencies>
-  </dependencyManagement>
- -->
-
   <dependencies>
     <!-- [START dependencies] -->
     <dependency>
       <groupId>com.google.cloud</groupId>
       <artifactId>google-cloud</artifactId>
```
jabubake: Why do we include `google-cloud` (all client libs) instead of `google-cloud-speech` ?

gguuss: I had previously been using google-cloud-speech but this change was requested before.  I'm happy to specify either dependency.

jabubake: @gguuss :my  comment should n't block this PR, but I think we want to include `google-cloud-speech` though. We don't recommend using `google-cloud` (brings in all the client libraries).

PR #782, opened 2017-08-01
Fixes seconds reported in word time offsets and enables maven checks

gguuss:


speech/cloud-client/pom.xml
```diff
@@ -21,10 +21,9 @@
 
   <!-- Parent defines config for testing & linting. -->
   <parent>
-    <artifactId>doc-samples</artifactId>
-    <groupId>com.google.cloud</groupId>
-    <version>1.0.0</version>
-    <relativePath>../..</relativePath>
+      <groupId>com.google.cloud.samples</groupId>
```
lesv: Why not look at the doc-samples pom?

gguuss: It seemed the checks were not being enforced for this sample (the change in Recognize.java fixes what was a broken test), AFAICT, the checks are now being enforced with this change, maybe it would be worth rolling back and seeing if that makes a difference.

PR #783, opened 2017-08-01
Update pom.xml

phanikumarmss:
The above dependency in pom.xml causing publishing error java.lang.NoClassDefFoundError: io/grpc/internal/AtomicBackoff and messages are not getting published to topic.


lesv:
Thanks

PR #792, opened 2017-08-08
Add Templating

lesv:
1. App Engine Sockets (moved from GoogleArchive)
2. J8 GAE Info now uses ThymeLeaf templating
3. Flex GAE Info for combat (WL’d only) sample
4. J8 Metadata - Only uses the Metadata server


lesv:
@halgrimur FYI
@llatif FYI

lesv:
Not sure why circle isn't testing.  Once this is approved, I'll work on that. :(

appengine-java8/gaeinfo/pom.xml
```diff
@@ -62,6 +62,12 @@ Copyright 2017 Google Inc.
       <version>2.8.1</version>
     </dependency>
 
+    <dependency>
+      <groupId>org.thymeleaf</groupId>
+      <artifactId>thymeleaf</artifactId>
+      <version>3.0.7.RELEASE</version>
+    </dependency>
+
```
jabubake: Yay !

PR #796, opened 2017-08-08
breaking change

lesv:
Try switching to circleci v2


lesv:
This breaks builds, but it actually runs things - a nice change.  I'd rather break builds for a bit so we fix the build.

PR #802, opened 2017-08-09
upgrade Guava

lesv:
Fix dip testing

### DIFF
--- a/dlp/src/test/java/com/example/dlp/InspectIT.java
+++ b/dlp/src/test/java/com/example/dlp/InspectIT.java
@@ -44,7 +44,7 @@ public class InspectIT {
   public void setUp() {
     bout = new ByteArrayOutputStream();
     out = new PrintStream(bout);
-    System.setOut(out);
+    System.setOut(out); // TODO(b/64541432) DLP currently doesn't support GOOGLE DEFAULT AUTH
     assertNotNull(System.getenv("GOOGLE_APPLICATION_CREDENTIALS"));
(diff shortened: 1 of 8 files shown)


lesv:
@jabubake FYI

dlp/src/test/java/com/example/dlp/InspectIT.java
```diff
@@ -44,7 +44,7 @@
   public void setUp() {
     bout = new ByteArrayOutputStream();
     out = new PrintStream(bout);
-    System.setOut(out);
+    System.setOut(out); // TODO(b/64541432) DLP currently doesn't support GOOGLE DEFAULT AUTH
```
jabubake: Les, can we refer to the [Github issue](https://github.com/GoogleCloudPlatform/google-cloud-java/issues/2231) instead : so users can view details ?

vision/cloud-client/pom.xml
```diff
@@ -50,7 +50,7 @@
     <dependency>
       <groupId>com.google.guava</groupId>
       <artifactId>guava</artifactId>
-      <version>20.0</version>
+      <version>23.0</version>
```
jabubake: Is there a reason we went for 23.0 here and 22.0 in guestbook on J8 ?

lesv:
Guava needs to be set manually.  I think I used 23 where it worked.  In the case of DLP, it seemed like where the issue needed to go wasn't on Github.

PR #808, opened 2017-08-15
IAP samples update

jabubake:
- switching to nimbus library for ease of JWK loading
- adding constraints for iat, alg

@matthewg

### DIFF
--- a/iap/src/main/java/com/example/iap/VerifyIapRequestHeader.java
+++ b/iap/src/main/java/com/example/iap/VerifyIapRequestHeader.java
@@ -13,6 +13,7 @@
  */
 
 package com.example.iap;
+// [START verify_iap_request]
 
 import com.google.api.client.http.HttpRequest;
 import com.google.common.base.Preconditions;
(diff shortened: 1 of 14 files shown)


iap/src/main/java/com/example/iap/VerifyIapRequestHeader.java
```diff
@@ -46,127 +37,93 @@
...
-        PublicKey publicKey = kf.generatePublic(keySpec);
-        if (publicKey instanceof ECPublicKey) {
-          return (ECPublicKey) publicKey;
-        }
-      } catch (InvalidKeySpecException | NoSuchAlgorithmException var7) {
-        throw new IOException("Unexpected exception reading data", var7);
-      }
-    }
-    return null;
+  boolean verifyJwt(String jwtToken, String expectedAudience) throws Exception {
```
lesv: Shouldn't there also be an expires time that should be verified?

lesv:
P.S. Circle was unhappy -- as long as you got tested, I'm Ok.

iap/src/main/java/com/example/iap/VerifyIapRequestHeader.java
jabubake: Good catch, added

jabubake:
`mvn clean verify`  did work for me locally.

PR #809, opened 2017-08-15
Don't check for location

lesv:
Vision is moving back and forth between knowing the image and thinking
it’s in Ancient Rome.  Let’s not test such things.


jabubake:
LGTM

PR #811, opened 2017-08-16
updating Mailjet code samples for Send API v3.1

adamyanliev:
Mailjet Send API v3.1 is being released, so updating the code samples to match the new Send API v3.1 configuration. Reference can be found [here](https://dev.mailjet.com/beta/).

### DIFF
--- a/compute/mailjet/src/main/java/com/example/compute/mailjet/MailjetSender.java
+++ b/compute/mailjet/src/main/java/com/example/compute/mailjet/MailjetSender.java
@@ -21,8 +21,9 @@ import com.mailjet.client.MailjetClient;
 import com.mailjet.client.MailjetRequest;
 import com.mailjet.client.MailjetResponse;
 import com.mailjet.client.errors.MailjetException;
+import com.mailjet.client.ClientOptions;
 import com.mailjet.client.errors.MailjetSocketTimeoutException;
-import com.mailjet.client.resource.Email;
+import com.mailjet.client.resource.Emailv31;
 // [END mailjet_imports]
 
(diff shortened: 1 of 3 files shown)


lesv:
Thanks for the contribution.

Shouldn't you also update the pom.xml's for this.

adamyanliev:
@lesv checked the pom.xml files, they are up-to-date.

compute/mailjet/src/main/java/com/example/compute/mailjet/MailjetSender.java
```diff
@@ -34,23 +35,27 @@
...
-    MailjetRequest email = new MailjetRequest(Email.resource)
-        .property(Email.FROMEMAIL, sender)
-        .property(Email.FROMNAME, "pandora")
-        .property(Email.SUBJECT, "Your email flight plan!")
-        .property(Email.TEXTPART,
-            "Dear passenger, welcome to Mailjet! May the delivery force be with you!")
-        .property(Email.HTMLPART,
-            "<h3>Dear passenger, welcome to Mailjet!</h3><br/>May the delivery force be with you!")
-        .property(Email.RECIPIENTS, new JSONArray().put(new JSONObject().put("Email", recipient)));
+    MailjetRequest email = new MailjetRequest(Emailv31.resource)
```
lesv: We don't use tabs, but do use two spaces, not 4.  If you could fix that, then I'll LGTM

adamyanliev:
@lesv thanks for the explanation. I updated the spacing, should be fine now.

lesv:
Thank you again for your contribution.

PR #818, opened 2017-08-21
updating Mailjet java version in pom.xml files

adamyanliev:
Mailjet java version 4.1.1 released today, updating pom.xml files to match


lesv:
Thanks for your contribution, we have a bot that does this, but it's blocked on a couple of things at the moment.

PR #821, opened 2017-08-22
KMS: Add set-primary-version to KMS snippets

White0ut:
This fixes failing integration tests. I've executed the tests on my
personal project to confirm. Important changes are:

* Each test execution creates a CryptoKey.
	* CryptoKeys are free and creating a new CryptoKey fixes the caching
          problem of creating a key version, setting it as primary, then
          relying on the new primary in a subsequent call.
* Destruction tests now destroy keys that they created rather than the primary
  key set for the execution.
* Added the setPrimaryVersion snippet w/ tests.


lesv:
Merging before tests can run due to it being on a fork, I'll be giving you permission to commit to a branch shortly.

PR #826, opened 2017-08-24
Stackdriver logging samples (JUL, Logback, Client library)

jabubake:
- Also updated List Logs snippets to use `google-cloud-java`


logging/cloud-client/src/main/java/com/example/logging/QuickstartSample.java
```diff
@@ -40,11 +47,12 @@ public static void main(String... args) throws Exception {
     String text = "Hello, world!";
 
     LogEntry entry = LogEntry.newBuilder(StringPayload.of(text))
+        .setSeverity(Severity.ERROR)
         .setLogName(logName)
         .setResource(MonitoredResource.newBuilder("global").build())
         .build();
 
-    // Writes the log entry
+    // Writes the log entry asynchronously
```
lesv: Out of curiosity, do we do anything to improve the odds that logs get written even if the app crashes? Can we? Big try / catch in main?

logging/jul/src/main/java/com/example/logging/jul/Quickstart.java
```diff
@@ -0,0 +1,31 @@
+/*
+  Copyright 2017, Google, Inc.
```
lesv: Don't think the "," is nesc.

logging/logback/src/main/java/com/example/logging/logback/Quickstart.java
```diff
@@ -0,0 +1,31 @@
+/*
+  Copyright 2017, Google, Inc.
```
lesv: comma

logging/cloud-client/src/main/java/com/example/logging/QuickstartSample.java
jabubake: we do offer a logging.flush() but would n't help with the try / catch in main. 
I get your point though, so will look into it separately as an issue on `google-cloud-java`.

logging/jul/src/main/java/com/example/logging/jul/Quickstart.java
jabubake: Done. Fixed missing licenses too.

logging/logback/src/main/java/com/example/logging/logback/Quickstart.java
jabubake: Done.

PR #829, opened 2017-08-25
patching dlp test

jabubake:


jabubake:
DLP succeeds, logging is broken because of the missing file (added in the other PR). 
Forcing merge here.

PR #850, opened 2017-09-12
[Endpoints] Update Endpoints + Guice sample

frankyn:
Review of sample provided by @tangiel. 

- Moved all but Guice filters to `configureServlets()` from `web.xml`.
- Added additional commentary on why `/_ah/api/*` is used instead of `/*`.


frankyn:
Thank you!

PR #855, opened 2017-09-19
Add logging properties file

lesv:
1. logging.properties
2. get rid of old properties from appengine-web.xml


lesv:
This won't block the stop the build and there are other issues that need to be fixed.

PR #868, opened 2017-09-26
Java sample for doing online prediction in Cloud Machine Learning Engine.

lluunn:


lesv:
Thanks for your contribution.  Please take a look at CONTRIBUTING.md, also at some of our other samples for inspiration.

You are missing licenses at the top of your files.  Also, I can't actually figure out how this is useful to anyone.  Did you push it by accident?

lluunn:
Sorry for the missing license, I will add it soon.

Could you clarify why do you think it's not useful and what should be improved? We (Cloud Machine Learning Engine team) want to provide some sample code for users to use our service in different languages, e.g.  C# and JS: https://github.com/GoogleCloudPlatform/nodejs-docs-samples/pull/432
https://github.com/GoogleCloudPlatform/dotnet-docs-samples/pull/290

lesv:
Partially it's me, but I'd like to see some comments on what you are doing.  -- I'll mark a few things inline.

mlengine/online-prediction/src/main/java/OnlinePredictionSample.java
```diff
@@ -0,0 +1,55 @@
...
+import java.io.File;
+
+public class OnlinePredictionSample {
+  public static void main(String[] args) throws Exception {
+    HttpTransport httpTransport = GoogleNetHttpTransport.newTrustedTransport();
+    JsonFactory jsonFactory = JacksonFactory.getDefaultInstance();
+    Discovery discovery = new Discovery.Builder(httpTransport, jsonFactory, null).build();
+
+    RestDescription api = discovery.apis().getRest("ml", "v1").execute();
+    RestMethod method = api.getResources().get("projects").getMethods().get("predict");
```
lesv: Some kind of comment how you are inspecting the discovery document and getting the Prediction API.

mlengine/online-prediction/src/main/java/OnlinePredictionSample.java
```diff
@@ -0,0 +1,55 @@
...
+
+    JsonSchema param = new JsonSchema();
+    String projectId = "YOUR_PROJECT_ID";
+    String modelId = "YOUR_MODEL_ID";
```
lesv: Something about your Model here would be good.

mlengine/online-prediction/src/main/java/OnlinePredictionSample.java
```diff
@@ -0,0 +1,55 @@
...
+    System.out.println(url);
+
+    String contentType = "application/json";
+    File requestBodyFile = new File("input.txt");
+    HttpContent content = new FileContent(contentType, requestBodyFile);
+    System.out.println(content.getLength());
+
+    GoogleCredential credential = GoogleCredential.getApplicationDefault();
+    HttpRequestFactory requestFactory = httpTransport.createRequestFactory(credential);
+    HttpRequest request = requestFactory.buildRequest(method.getHttpMethod(), url, content);
```
lesv: A comment about this.

Why aren't you actually generating an APIary client so don't need to do it this way?  Seems simple enough, but there are some named things when you create it from the proto usually.

lesv:
Also, please go to github/ (internal link) and register.  Didn't realize you were a Googler, or I'd have reached out directly.

mlengine/online-prediction/src/main/java/OnlinePredictionSample.java
lluunn: Done

mlengine/online-prediction/src/main/java/OnlinePredictionSample.java
lluunn: There isn't particular reason for choosing discovery over direct API.
It just that I tried discovery first.
Which is the typical or recommended pattern?

mlengine/online-prediction/README.md
```diff
@@ -0,0 +1,9 @@
+# Cloud Machine Learning Engine - Online Prediction with Java
```
lesv: Could you look at our other README's and add content about auth.

mlengine/online-prediction/pom.xml
```diff
@@ -0,0 +1,49 @@
+<project>
```
lesv: License needed here

mlengine/online-prediction/README.md
```diff
@@ -0,0 +1,9 @@
...
+
+Modify the OnlinePredictionSample.java with your project/model/version information.
+
+Compile and run the sample:
```
lesv: Should mention that this is a part of a tutorial and provide a link.

mlengine/online-prediction/src/main/java/OnlinePredictionSample.java
lesv: In this case, I'm happy with just using discovery.  As you will need to generate a One Platform API for GA.  (I'd rather see a 1P now, but can live w/ this)

mlengine/online-prediction/README.md
lluunn: Added, PTAL

mlengine/online-prediction/pom.xml
lluunn: Done.

lluunn:
Thanks for reviewing this!

lluunn:
Thank you!

PR #869, opened 2017-09-26
Updates instructions for jar installation

gguuss:


iot/api-client/manager/README.md
```diff
@@ -10,7 +10,7 @@ PubSub topic for Cloud IoT as described in [the parent README](../README.md).
 Manually install [the provided client library](https://cloud.google.com/iot/resources/java/cloud-iot-core-library.jar)
```
lesv: If you are really going to do this, shouldn't you put it in a GCS bucket?  Also, you might wish to include the pom.xml as it has the dependencies in it.

gguuss: The intent is to fallback to the existing way we're distributing the bits. For now, the existing library is on devsite where it's whitelisted but today I realized the instructions are currently broken as the groupID didn't match what's in POM.xml.

lesv: Circle is still failing - would be good to fix it.

lesv: Actually, it's the Storage-JSON API failing, so I'm ok w/ this once you figure out the library in a way that can be tested.

PR #870, opened 2017-09-27
Update package import to appengine

nnegrey:


lesv:
On a fork, so testing doesn't run.

PR #874, opened 2017-09-28
Use published artifact for IoT

gguuss:


lesv:
I know gguuss has signed the CLA - not sure why CLABot is unhappy.

PR #882, opened 2017-10-09
Java 8 way to concatenate strings

medvedev:
Let's show people how to write clear and compact code in Java8


lesv:
Thanks for the PR, but I don't think this is the appropriate venue for this.

PR #883, opened 2017-10-09
Post-launch fixes

gguuss:
* Fixes bug in types used for patch (Java)
  * Adds test for device patch
* Updates HTTP headers to be lowercase for HTTP/2
* Adds snippets for MQTT


iot/api-client/manager/src/main/java/com/example/cloud/iot/examples/DeviceRegistryExample.java
```diff
@@ -343,8 +353,9 @@ public static Device getDevice(String deviceId, String projectId, String cloudRe
         GoogleCredential.getApplicationDefault().createScoped(CloudIotScopes.all());
     JsonFactory jsonFactory = JacksonFactory.getDefaultInstance();
     HttpRequestInitializer init = new RetryHttpInitializerWrapper(credential);
-    CloudIot service = new CloudIot(GoogleNetHttpTransport.newTrustedTransport(), jsonFactory,
-        init);
+    final CloudIot service = new CloudIot.Builder(
+        GoogleNetHttpTransport.newTrustedTransport(),jsonFactory, init)
+        .setApplicationName(APP_NAME).build();
```
lesv: try mvn checkstyle:check or checkstyle:checkstyle

Not sure if that is the right indentation.

gguuss: `mvn checkstyle:checkstyle` is passing without fail for all 3 samples. The . on the continuation looks a little funny but this is two indentions at a TS of 2 and should be consistent with the rest of the sample code.

PR #890, opened 2017-10-18
adding serving url sample

jabubake:
@andrewsg FYI


appengine-java8/images/src/main/java/com/example/appengine/images/ImagesServlet.java
```diff
@@ -108,6 +109,16 @@ public void doGet(HttpServletRequest req, HttpServletResponse resp) throws IOExc
         ByteBuffer.wrap(rotatedImage.getImageData()));
     //[END rotate]
 
+    // [START servingUrl]
+    // Create a fixed dedicated URL that points to the GCS hosted file
+    ServingUrlOptions options = ServingUrlOptions.Builder
+            .withGoogleStorageFileName("/gs/" + bucket + "/image.jpeg")
```
lesv: I always think of this as gs:// - please check this.  Otherwise LGTM

lesv: Ah - this is blob store.  It's ok.

PR #892, opened 2017-10-24
[Endpoints] Add region tags and update dependencies

frankyn:
This PR is to add region tags around the backend/ sample for use in documentation and update plugin versions. I'm not sure why they aren't getting picked up in dependency auto-update.

### DIFF
--- a/appengine-java8/endpoints-v2-backend/build.gradle
+++ b/appengine-java8/endpoints-v2-backend/build.gradle
@@ -18,8 +18,10 @@ buildscript {
   }
 
   dependencies {
-    classpath 'com.google.cloud.tools:endpoints-framework-gradle-plugin:+'
-    classpath 'com.google.cloud.tools:appengine-gradle-plugin:+'
+    // [START endpoints_plugin]
+    classpath 'com.google.cloud.tools:endpoints-framework-gradle-plugin:1.0.2'
+    // [END endpoints_plugin]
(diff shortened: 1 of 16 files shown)


appengine-java8/endpoints-v2-backend/build.gradle
```diff
@@ -39,14 +41,17 @@ def projectId = 'YOUR_PROJECT_ID'
...
 apply plugin: 'com.google.cloud.tools.endpoints-framework-server'
+// [END endpoints_plugin_apply]
 apply plugin: 'com.google.cloud.tools.appengine'
 
 dependencies {
-    // For real projects: use concrete versions here instead of the '+' to make your build consistent
-    compile 'com.google.endpoints:endpoints-framework:+'
-    compile 'com.google.endpoints:endpoints-management-control-appengine:+'
-    compile 'com.google.endpoints:endpoints-framework-auth:+'
+    compile 'com.google.endpoints:endpoints-framework:2.0.8'
```
lesv: I'm sure there is a reason this is lonely. (no tag)  Just checking.

frankyn: Yes, the dependencies below will be used in documentation for "Adding API Management" and not starting from scratch.

frankyn:
Thanks @lesv. Is there something specific I need to do for Maven versions to be updated automatically besides the SDK?

PR #894, opened 2017-10-25
Added missing argument, plus minor edits.

WalterHub:


WalterHub:
Addressing as part of a different pull request.

PR #895, opened 2017-10-25
Adds snippet for restoring a crypto key version

WalterHub:
New snippet to use at https://cloud.google.com/kms/docs/destroy-restore.

### DIFF
--- a/kms/src/test/java/com/example/SnippetsIT.java
+++ b/kms/src/test/java/com/example/SnippetsIT.java
@@ -209,12 +209,21 @@ public class SnippetsIT {
   public void restoreCryptoKeyVersion_restores() throws Exception {
     Snippets.createCryptoKeyVersion(PROJECT_ID, LOCATION_ID, KEY_RING_ID, CRYPTO_KEY_ID);
 
-    Matcher matcher = Pattern.compile(".*cryptoKeyVersions/(\\d+)\",\"state\":\"DESTROY_SCHEDULED\".*",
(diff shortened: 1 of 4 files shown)


kms/src/test/java/com/example/SnippetsIT.java
```diff
@@ -204,6 +204,24 @@ public void destroyCryptoKeyVersion_destroys() throws Exception {
         KEY_RING_ID, CRYPTO_KEY_ID, version));
   }
 
+
+  @Test
+  public void restoreCryptoKeyVersion_restores() throws Exception {
+    Snippets.createCryptoKeyVersion(PROJECT_ID, LOCATION_ID, KEY_RING_ID, CRYPTO_KEY_ID);
+
+    Matcher matcher = Pattern.compile(".*cryptoKeyVersions/(\\d+)\",\"state\":\"DESTROY_SCHEDULED\".*",
```
lesv: This might be a bit brittle.  What if it's actually done?

WalterHub: Do you mean if the destruction already occurred after the wait period?

lesv: Yes

lesv:
@kurtisvg Sponge looks unhappy - I'm expecting to see a bit more  PTAL

WalterHub:
(fyi, I'm not sure why the UI is indicating I dismissed the comment from @lesv)

kurtisvg:
@lesv Tests for this branch are still WIP. If you want to pull it, I'll take responsibility for fixing any tests it might break.

PR #902, opened 2017-10-31
Increase Kokoro testing reliability 

kurtisvg:
This PR increases the reliability of a few of the tests. Some are most likely still flaky, but I will address those as they appear. 

It also removes leftover files from older testing and moves the GitHub templates into the .github folder.


kurtisvg:
@lesv - FYI

Issue #885, opened 2017-10-11
Bug: KMS & Logging IT tests are flakey...


lesv:
kms-samples
logging-google-cloud-samples-api

We should fix this.

PR #903, opened 2017-11-01
Increase Logging Testing stability

kurtisvg:
Increase the stability of the Logging tests.

@lesv - FYI

### DIFF
--- a/logging/cloud-client/src/main/java/com/example/logging/ListLogs.java
+++ b/logging/cloud-client/src/main/java/com/example/logging/ListLogs.java
@@ -39,13 +39,16 @@ public class ListLogs {
 
       String logFilter = "logName=projects/" + options.getProjectId() + "/logs/" + logName;
 
-      // List log entries
+      // List all log entries
       Page<LogEntry> entries = logging.listLogEntries(
           EntryListOption.filter(logFilter));
-      for (LogEntry logEntry : entries.iterateAll()) {
-        System.out.println(logEntry);
-      }
-      // Use entries.getNextPage() to paginate
+      do {
+        for (LogEntry logEntry : entries.iterateAll()) {
+          System.out.println(logEntry);
+        }
+        entries = entries.getNextPage();
+      } while(entries != null);
+
     }
     // [END listlogs]
   }
--- a/logging/cloud-client/src/test/java/com/example/logging/LoggingIT.java
+++ b/logging/cloud-client/src/test/java/com/example/logging/LoggingIT.java
@@ -39,6 +39,9 @@ import org.junit.runners.JUnit4;
 @SuppressWarnings("checkstyle:abbreviationaswordinname")
 public class LoggingIT {
 
+  private final String QUICKSTART_LOG = "my-log";
+  private final String TEST_WRITE_LOG = "test-log";
+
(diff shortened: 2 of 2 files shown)


kurtisvg:
Also satisfies #885 (KMS tests should already have been addressed in #902)

logging/cloud-client/src/main/java/com/example/logging/ListLogs.java
```diff
@@ -39,13 +39,16 @@ public static void main(String... args) throws Exception {
...
-      for (LogEntry logEntry : entries.iterateAll()) {
-        System.out.println(logEntry);
-      }
-      // Use entries.getNextPage() to paginate
+      do {
+        for (LogEntry logEntry : entries.iterateAll()) {
+          System.out.println(logEntry);
+        }
+        entries = entries.getNextPage();
+      } while(entries != null);
```
lesv: There are forms of getting entries that aren't fully paged, but are infinite.  I'm presuming you've verified that this is actually paged?

logging/cloud-client/src/test/java/com/example/logging/LoggingIT.java
```diff
@@ -56,37 +59,36 @@ public void setUp() {
 
   @After
   public void tearDown() {
+    // Clean up created logs
+    deleteLog(QUICKSTART_LOG);
```
lesv: do you want to do this `@before` as well? incase something broke last time?

logging/cloud-client/src/test/java/com/example/logging/LoggingIT.java
```diff
@@ -39,6 +39,9 @@
 @SuppressWarnings("checkstyle:abbreviationaswordinname")
 public class LoggingIT {
 
+  private final String QUICKSTART_LOG = "my-log";
```
lesv: Should these go to $TEMP?

kurtisvg: This is the name of the log in StackDriver that it can later be referred by, not the location it's logging too.

logging/cloud-client/src/test/java/com/example/logging/LoggingIT.java
kurtisvg: I'm concerned that part of the reasons the tests are failing periodically is that the deletes are actually being registered after the test line is logged. By moving the cleanup to the tearDown, it should always clean up unless the build is somehow forced to stop between the file log and that this exact moment.

logging/cloud-client/src/test/java/com/example/logging/LoggingIT.java
lesv: SGTM

logging/cloud-client/src/test/java/com/example/logging/LoggingIT.java
lesv: I'm just concerned that some crash happens and it doesn't execute - so I ask Q's.  If you are happy, I'm happy. (ie you don't think that will happen)

lesv: sgtm

PR #909, opened 2017-11-06
[Endpoints] Fix local development for Guice sample.

frankyn:
There exists a bug with the DevAppServer and Guice + Basic Scaling that causes the main serving port 8080 to 404 when accessing the sample API endpoint, but works from an instance port when this isn't the correct way to access the API endpoint.

Reported here: b/68948960


frankyn:
Thanks @lesv

PR #916, opened 2017-11-14
[Endpoints] Add skeleton v2 sample

frankyn:
This is a PR to add a skeleton V2 sample.

### DIFF
--- a/appengine-java8/endpoints-v2-skeleton/pom.xml
+++ b/appengine-java8/endpoints-v2-skeleton/pom.xml
@@ -103,24 +103,6 @@
                     <hostname>YOUR_PROJECT_ID.appspot.com</hostname>
                 </configuration>
             </plugin>
-            <!--
-                 The versions-maven-plugin is used for dependency updates and
-                 is not required.
-            -->
-            <plugin>
-                <groupId>org.codehaus.mojo</groupId>
-                <artifactId>versions-maven-plugin</artifactId>
-                <version>2.3</version>
-                <executions>
-                    <execution>
-                        <phase>compile</phase>
-                        <goals>
-                            <goal>display-dependency-updates</goal>
-                            <goal>display-plugin-updates</goal>
-                        </goals>
-                    </execution>
-                </executions>
-            </plugin>
         </plugins>
     </build>
     <!-- [END pom_build] -->
--- a/appengine-java8/endpoints-v2-skeleton/pom.xml
+++ b/appengine-java8/endpoints-v2-skeleton/pom.xml
@@ -85,7 +85,7 @@
             <plugin>
                 <groupId>com.google.cloud.tools</groupId>
                 <artifactId>appengine-maven-plugin</artifactId>
-                <version>1.3.1</version>
+                <version>1.3.2</version>
                 <configuration>
                     <!-- deploy configuration -->
                 </configuration>
@@ -96,7 +96,7 @@
                 <version>1.0.2</version>
                 <configuration>
                     <!-- plugin configuration -->
-                    <hostname>YOUR-PROJECT-ID.appspot.com</hostname>
+                    <hostname>YOUR_PROJECT_ID.appspot.com</hostname>
                 </configuration>
             </plugin>
             <!--
--- a/appengine-java8/endpoints-v2-skeleton/pom.xml
+++ b/appengine-java8/endpoints-v2-skeleton/pom.xml
@@ -1,0 +1,124 @@
+<!--
+ Copyright 2017 Google Inc.
+
+ Licensed under the Apache License, Version 2.0 (the "License");
+ you may not use this file except in compliance with the License.
+ You may obtain a copy of the License at
+
+       http://www.apache.org/licenses/LICENSE-2.0
+
+ Unless required by applicable law or agreed to in writing, software
+ distributed under the License is distributed on an "AS IS" BASIS,
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ See the License for the specific language governing permissions and
+ limitations under the License.
+-->
+<project>
+    <modelVersion>4.0.0</modelVersion>
+    <packaging>war</packaging>
+    <version>1.0-SNAPSHOT</version>
+
+    <groupId>com.example.skeleton</groupId>
+    <artifactId>endpoints-j8-skeleton</artifactId>
+
+    <parent>
+        <artifactId>appengine-java8-samples</artifactId>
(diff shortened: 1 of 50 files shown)


appengine-java8/endpoints-v2-skeleton/pom.xml
```diff
@@ -0,0 +1,124 @@
...
+-->
+<project>
+    <modelVersion>4.0.0</modelVersion>
+    <packaging>war</packaging>
+    <version>1.0-SNAPSHOT</version>
+
+    <groupId>com.example.skeleton</groupId>
+    <artifactId>endpoints-j8-skeleton</artifactId>
+
+    <parent>
```
jabubake: If we are providing this as a skeleton app, maybe we should take out the parent reference here.

appengine-java8/endpoints-v2-skeleton/pom.xml
```diff
@@ -0,0 +1,124 @@
...
+                            <filtering>true</filtering>
+                            <targetPath>WEB-INF</targetPath>
+                        </resources>
+                    </webResources>
+                </configuration>
+            </plugin>
+            <plugin>
+                <groupId>com.google.cloud.tools</groupId>
+                <artifactId>appengine-maven-plugin</artifactId>
+                <version>1.3.2</version>
```
jabubake: noticed you used 1.3.3 in the gradle version

appengine-java8/endpoints-v2-skeleton/pom.xml
```diff
@@ -0,0 +1,124 @@
...
+                    <!-- deploy configuration -->
+                </configuration>
+            </plugin>
+            <plugin>
+                <groupId>com.google.cloud.tools</groupId>
+                <artifactId>endpoints-framework-maven-plugin</artifactId>
+                <version>1.0.2</version>
+                <configuration>
+                    <!-- plugin configuration -->
+                    <hostname>YOUR_PROJECT_ID.appspot.com</hostname>
```
jabubake: should this be marked as "MUST REPLACE" ?

appengine-java8/endpoints-v2-skeleton/pom.xml
```diff
@@ -0,0 +1,124 @@
...
+                </configuration>
+            </plugin>
+            <!--
+                 The versions-maven-plugin is used for dependency updates and
+                 is not required.
+            -->
+            <plugin>
```
jabubake: If this is not required and we cal this a skeleton app, maybe we can opt out of including it ? (making for a simpler `pom.xml`)

appengine-java8/endpoints-v2-skeleton/src/main/java/com/example/skeleton/MyApi.java
```diff
@@ -0,0 +1,30 @@
...
+ */
+
+package com.example.skeleton;
+
+import com.google.api.server.spi.config.Api;
+
+/**
+ * MyApi skeleton endpoints sample
+ * Add your first API methods in this class, or you may create another class.
+ * In that case, update your src/main/webapp/WEB-INF/web.xml accordingly.
```
jabubake: "In that case" comment, can  you rephrase that to, if you create another class ?

appengine-java8/endpoints-v2-skeleton/src/main/webapp/WEB-INF/appengine-web.xml
```diff
@@ -0,0 +1,30 @@
...
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ See the License for the specific language governing permissions and
+ limitations under the License.
+-->
+<!-- [START appengine_web_xml] -->
+<appengine-web-app xmlns="http://appengine.google.com/ns/1.0">
+    <runtime>java8</runtime>
+    <threadsafe>true</threadsafe>
+
+    <basic-scaling>
```
jabubake: Is this a requirement for the user in their skeleton app, if not, can we leave it out ?

jabubake:
@frankyn : also, do you know why the tests failed ?

appengine-java8/endpoints-v2-skeleton/pom.xml
frankyn: I guess I misunderstood, but I thought Dependency Auto-update used this plugin to know which dependencies to update?

appengine-java8/endpoints-v2-skeleton/src/main/webapp/WEB-INF/appengine-web.xml
frankyn: Good call, I'll remove this.

appengine-java8/endpoints-v2-skeleton/pom.xml
frankyn: I forgot to add this sample to the parent pom.xml.. The rationale for keeping it in is for dependency updates as IIRC it's dependent on the parent pom.

frankyn:
Failure is from Firestore
```
[INFO] Google Cloud Firestore Samples ..................... FAILURE [ 11.197 s]
```

appengine-java8/endpoints-v2-skeleton/pom.xml
frankyn: The maven appengine plugin is at 1.3.2

frankyn:
I updated the PR to reflect your comments @jabubake PTAL. Thank you for your review!

appengine-java8/endpoints-v2-skeleton/pom.xml
jabubake: looks like @kurtisvg is working on a related cleanup, so ignore my comment here.

appengine-java8/endpoints-v2-skeleton/pom.xml
jabubake: 👍

appengine-java8/endpoints-v2-skeleton/pom.xml
jabubake: we don't include it within each sample though:
already included here.
https://github.com/GoogleCloudPlatform/java-repo-tools/blob/master/pom.xml#L237

jabubake:
LGTM, just a minor update required in removing the versions plugin from the pom.xml

appengine-java8/endpoints-v2-skeleton/pom.xml
frankyn: ah thanks for clarifying! Will fix and then merge.

PR #926, opened 2017-11-20
Add section_id 'dependencies' around pubsub dependencies in pubsub/cloud-client/pom.xml.

dzlier-gcp:
For use in new java landing page samples section.


kurtisvg:
Overriding status check since there are issues with the DLP API at the moment.

PR #933, opened 2017-11-29
Removed TaskQueue samples using Restv1 API.

kurtisvg:
The TaskQueue REST API is being deprecated shortly, so this PR removes the samples that use that API.


kurtisvg:
I moved the remaining sample (pull) into the AppEngine folder, and then combined the push and defer samples into one taskqueue folder with it as well.

kurtisvg:
@jabubake I fixed a few mistakes with pom structure, PTAL when you have time

PR #935, opened 2017-11-29
Update samples to v1

nnegrey:


jabubake:
Kokoro tests are failing for other reasons. Merging

PR #940, opened 2017-12-05
Revert "Updated references to container-engine.yaml to kubernetes-engine.yaml"

lesv:
Reverts GoogleCloudPlatform/java-docs-samples#932


lesv:
@kurtisvg PTAL

PR #944, opened 2017-12-06
Transaction Sample update

lesv:
1. An example that shows queue add w/ transactions should actually use
the transaction.


lesv:
LGTM - due to likely unrelated failure of build.

PR #952, opened 2017-12-12
Update runtime to the latest GA version

patflynn:


patflynn:
based off of [user feedback on stack overflow.](https://stackoverflow.com/questions/47719703/how-to-create-a-new-project-using-both-app-engine-standard-java-8-and-gradle/47758688#47758688)

patflynn:
just noticed the java 8 sample directory next to this one...

Perhaps we should make java 8 the default and remove the java 7 sample?

patflynn:
[This cloud doc](https://cloud.google.com/appengine/docs/standard/java/tools/gradle#creating_a_new_project) needs updating too to ensure that the default user-flow takes people to the latest runtime.

lesv:
We probably should update the cloud docs to use one of the other samples in an appengine-8 folder.  There are ones in getting-started-java and in this repo, but hello worlds tend to be in g-s-j

patflynn:
https://b/70555407

PR #954, opened 2017-12-12
Adds tests for HTTP device sample.

gguuss:


iot/api-client/manager/resources/rsa_private.pem
```diff
@@ -0,0 +1,28 @@
+-----BEGIN PRIVATE KEY-----
```
lesv: Did you really mean to publish this?

gguuss: Yes, it's used after the device is registered to connect and test the "http device" samples.  The authorization credential is revoked during tear-down of the tests.

gguuss: At any rate, this file is needed to connect and providing a dummy / static one is probably simpler than generating them with code or shelling out to `openssl` in the tests.

iot/api-client/manager/README.md
```diff
@@ -134,3 +134,80 @@ Patch a device with RSA:
...
+
+    mvn clean compile
+
+## Running the sample
+
+The following command summarizes the sample usage:
+
+```
+    mvn exec:java \
+        -Dexec.mainClass="com.example.cloud.iot.examples.HttpExample" \
```
lesv: This seems long winded, but Ok

iot/api-client/manager/resources/rsa_private.pem
lesv: Sg - It won't surprise me for this to trigger some scanners, in which case, I'll ask you to encrypt a new one and we can decrypt when we test.

iot/api-client/manager/resources/README.md
```diff
@@ -0,0 +1,4 @@
+# Test public certificate files
+
+The public certificates in this folder are only provided for testing and should
+not be used for registering your devices.
```
lesv: Do we tell them how to create their own somewhere?

iot/api-client/manager/README.md
gguuss: Yeah, for the Manager sample, there are a few things that can are put in environment variables but for the device samples it would pretty much just be the project ID, which doesn't help much with the command length.

iot/api-client/manager/resources/README.md
gguuss: Yes, added note.

PR #969, opened 2017-12-21
Maven Dependency Cleanup - Appengine

kurtisvg:
This PR cleans up the dependency structure for the Appengine folder. It does the following:

1. Changes parents for individual projects to our `shared-configuration` so they users can move a relevant folder without breaking things.
2. Addresses any exists check-style violations 
3. Adds the `Open in Cloud Shell` button for all README.md

Additionally, it deletes the appengine/channel samples since the service has been shutdown.


lesv:
Wow - you just want me unhappy. 241 files?

appengine/remote/remote-server/src/main/java/com/example/appengine/remote/RemoteServlet.java
```diff
@@ -1,5 +1,5 @@
-/**
- * Copyright 2016 Google Inc.
+/*
+ * Copyright 2015 Google Inc.
```
lesv: In general (and I saw this earlier), we don't change the dates.  (Rolling them forward or backward).

lesv:
I presume you meant Open in Cloud Shell, not SQL.

kurtisvg:
Merging - Issue with test is because of Vision Beta.

PR #980, opened 2018-01-04
Maven Dependency Cleanup - Appengine Java8

kurtisvg:
This is the same as #969, but for Appengine Java 8.

### DIFF
--- a/appengine-java8/spanner/pom.xml
+++ b/appengine-java8/spanner/pom.xml
@@ -22,16 +22,23 @@
   <artifactId>appengine-spanner-j8</artifactId>
   <version>1.0-SNAPSHOT</version>
   <packaging>war</packaging>
+
+  <!--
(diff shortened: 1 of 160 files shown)


appengine-java8/spanner/pom.xml
```diff
@@ -22,16 +22,23 @@
...
+    <groupId>com.google.cloud.samples</groupId>
+    <artifactId>shared-configuration</artifactId>
+    <version>1.0.8</version>
   </parent>
+
+  <properties>
+    <maven.compiler.target>1.8</maven.compiler.target>
+    <maven.compiler.source>1.8</maven.compiler.source>
+
+    <failOnMissingWebXml>false</failOnMissingWebXml>
```
lesv: Shouldn't be required if we are using the correct version of the maven war plugin.

PR #985, opened 2018-01-10
Updated bigquery, cloud-tasks, and compute folders.

kurtisvg:
Updated samples in biquery, cloud-tasks, and compute folders by updating their parent projects, cleaning up checkstyle violations, and adding in Cloud Shell buttons to READMEs.


lesv:
Hadn't realized this was here.

PR #989, opened 2018-01-12
Updated CloudSQL samples.

kurtisvg:
This is an update to both CloudSQL samples for GAE Standard (Java8). 

The 'postgres' sample has been renamed to 'cloudsql-postgres' so that it is next to 'cloudsql' and easier to find. It also updates the version numbers on the socket factory dependencies. 

Both samples have been also adjusted to remove IP Logging (since it may be considered PII) and instead just log visitors instead. 

@piaxc - FYI


appengine-java8/cloudsql-postgres/src/main/java/com/example/appengine/postgresql/PostgreSqlServlet.java
```diff
@@ -1,5 +1,5 @@
-/*
- * Copyright 2016 Google Inc.
+/**
+ * Copyright 2017 Google Inc.
```
jabubake: if this file just moved, you can retain 2016 in the license

appengine-java8/cloudsql/pom.xml
```diff
@@ -79,7 +79,7 @@
     <dependency>
       <groupId>com.google.cloud.sql</groupId>
       <artifactId>mysql-socket-factory</artifactId> <!-- mysql-socket-factory-connector-j-6 if using 6.x.x -->
```
jabubake: what does that comment above mean ? Java 6 ? can you make that more explicit or remove if it is no longer required

kurtisvg:
@jabubake updated.

PR #998, opened 2018-01-18
Moving translate-pubsub

dzlier-gcp:
Moving translate-pubsub sample from [GoogleCloudPlatform/getting-started-java](https://github.com/GoogleCloudPlatform/getting-started-java) to [GoogleCloudPlatform/java-docs-samples](https://github.com/GoogleCloudPlatform/java-docs-samples).

Only changes from that project's version are updates to the pom.xml & README.md where appropriate, and fixing a typo in index.jsp (importing translate_pubsub.PubSubHome instead of translatepubsub.PubSubHome).

### DIFF
--- a/appengine-java8/translate-pubsub/pom.xml
+++ b/appengine-java8/translate-pubsub/pom.xml
@@ -1,0 +1,89 @@
+<!--
+  Copyright 2018 Google Inc.
+
+  Licensed under the Apache License, Version 2.0 (the "License");
+  you may not use this file except in compliance with the License.
+  You may obtain a copy of the License at
+
+       http://www.apache.org/licenses/LICENSE-2.0
+
+  Unless required by applicable law or agreed to in writing, software
+  distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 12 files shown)


appengine-java8/translate-pubsub/pom.xml
```diff
@@ -0,0 +1,80 @@
...
+-->
+<!-- [START project] -->
+<project>
+  <modelVersion>4.0.0</modelVersion>
+  <packaging>war</packaging>
+  <version>1.0-SNAPSHOT</version>
+  <groupId>com.example.appengine</groupId>
+  <artifactId>appengine-translate-pubsub</artifactId>
+  <!-- Parent POM defines ${appengine.sdk.version} (updates frequently). -->
+  <parent>
```
kurtisvg: Parent should be set as follows to enable checkstyle checks:
```
  <!--
  -->
  <parent>
    <version>1.0.8</version>
  </parent>
  <properties>
```

dzlier-gcp: Done

PR #1006, opened 2018-01-24
Consolidate and cleanup Kokoro test script. 

kurtisvg:
This tidies up and condenses the test script used to trigger Kokoro tests.

It removes some unused variables, directs output from tests themselves into a file, and removes the clutter from 'downloading....' for all the dependencies.

### DIFF
--- a/.kokoro/tests/run_tests.sh
+++ b/.kokoro/tests/run_tests.sh
@@ -13,7 +13,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-set -xeo pipefail
+set -eo pipefail
 shopt -s globstar
 # We spin up some subprocesses. Don't kill them on hangup
 trap '' HUP
--- a/.kokoro/tests/run_tests.sh
+++ b/.kokoro/tests/run_tests.sh
@@ -13,61 +13,35 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-set -eo pipefail
+set -xeo pipefail
 shopt -s globstar
-
-set -xe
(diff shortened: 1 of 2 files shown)


pom.xml
```diff
@@ -20,6 +20,10 @@
     <artifactId>doc-samples</artifactId>
     <packaging>pom</packaging>
 
+    <prerequisites>
```
lesv: You might remove this.  Things get weird with and w/o it.

.kokoro/tests/run_tests.sh
```diff
@@ -13,61 +13,35 @@
...
+gcloud components update --quiet
+echo "********** GCLOUD INFO ***********"
+gcloud -v
+echo "********** MAVEN INFO  ***********"
+mvn -v
+echo "********** GRADLE INFO ***********"
+gradle -v
+
+# Setup required enviormental variables
 export GOOGLE_APPLICATION_CREDENTIALS=${KOKORO_GFILE_DIR}/service-acct.json
```
lesv: You can turn echo off for this stuff

kurtisvg: Is there a way to do that? Or will ` &> /dev/null/` silence it?

lesv: https://www.gnu.org/software/bash/manual/html_node/The-Set-Builtin.html

`set ` option of some kind.

pom.xml
lesv: Yea - I think what we need to do is set the versions of the maven plugins to use.

kurtisvg:
@lesv PTAL

PR #1012, opened 2018-01-30
Add region tags

ellenevans:
One for create task and one for lease and acknowledge task


ellenevans:
Changes made

lesv:
@dpebot merge when green

PR #1013, opened 2018-01-31
Add region tags

ellenevans:


lesv:
@dpebot merge when green

PR #1018, opened 2018-02-08
Adds exponential backoff to MQTT

gguuss:


gguuss:
@lesv Yes, in HTTP I was able to reuse the retry that is in Apiary, because this device example is using MQTT and a the PAHO client, it's a little more custom / verbose.

PR #1019, opened 2018-02-08
Cloud CDN

DanSanche:
I added some samples for Cloud CDN url signing on java: https://cloud.google.com/storage/docs/access-control/create-signed-urls-program
I believe we don't wont to merge this until the feature is ready to launch
this is my first submission to a sample repo, so let me know if I'm doing anything weird


DanSanche:
updated author info, since that was blocking the build

PR #1030, opened 2018-02-14
Reduce noise in diff tests.

kurtisvg:
Changes diff tests to only output for projects tested to reduce noise, and fixes so that only leaf projects are run.


.kokoro/tests/diff_tests.sh
```diff
@@ -37,35 +37,36 @@ gcloud auth activate-service-account\
...
-    RTN=$?
+    CHANGED=$?
+    # Only test leafs to prevent testing twice
+    PARENT=$(grep "<modules>" pom.xml -c)
     set -e
 
     # Check for changes to the current folder
-    if [ "$RTN" -eq 1 ]; then
-        echo -e "\n Change detected. Running tests. \n "
+    if [ "$CHANGED" -eq 1 ] && [ "$PARENT" -eq 0 ]; then
```
lesv: "${...}"

kurtisvg:
@lesv I going to go ahead and pull it for now, but I'll talk with Dane about running shellcheck on it in the future as part of the tests.

PR #1035, opened 2018-02-20
Device sample updates

gguuss:
* Updates HTTP sample to use v1 instead of beta path
* Updates MQTT sample to use TLS v1.2 as it's the only supported version on the device bridge.


gguuss:
> Probably TLSv1.3 isn't rolled out to Java yet, but it should be soon ??

Good point!  The bridge (as opposed to "server" as it's not a full server) currently is best supported by TLS-v1.2 and MQTT.

> Shouldn't a v1 (GA'd) service be using the One Platform API's?

Great point, the HTTP Device side works with its own identity which makes it difficult to work as a 1P-API.

PR #1040, opened 2018-02-27
Post-GA nits

gguuss:
Fixes keyFactory type for ES vs EC keys in HTTP example (MQTT was done before) and adds more comprehensive examples to README.


iot/api-client/manager/README.md
```diff
@@ -8,6 +8,10 @@ This sample app demonstrates device management for Google Cloud IoT Core.
 Note that before you can run the sample, you must configure a Google Cloud
 PubSub topic for Cloud IoT as described in [the parent README](../README.md).
 
+Before running the samples, you can set the `GCLOUD_PROJECT` and
```
lesv: Shouldn't this be GOOGLE_CLOUD_PROJECT ??

iot/api-client/manager/README.md
```diff
@@ -68,6 +72,7 @@ Create an ES device:
     java -cp target/cloudiot-manager-demo-1.0-jar-with-dependencies.jar \
```
lesv: Wow that's long

gguuss: I agree, updated the manager samples to use `mvn:exec`, much cleaner.

PR #1043, opened 2018-02-28
Add Spanner BatchQuery sample.

jsimonweb:


spanner/cloud-client/src/main/java/com/example/spanner/BatchSample.java
```diff
@@ -0,0 +1,115 @@
+/*
+ * Copyright 2017 Google Inc.
```
jabubake: Update to 2018

PR #1048, opened 2018-03-02
Fixing Auto Updates

kurtisvg:


kurtisvg:
@lesv - FYI 

I've rolled back a few samples that are still having issues in particular monitoring/, bigtable/, and dataflow/spanner-io. They seem to have non-trivial updates required to the samples. I'll work with @dzlier-gcp to figure out what is going wrong.

PR #1052, opened 2018-03-08
[DO NOT MERGE] Change to managed rollout

simonz130:
We introduce a new feature where we can specify in YAML that service configuration for Endpoints is going to be managed. This means that the service will pull latest deployed service configuration automatically, rather than requiring customer to specify specific version of service configuration.

New flag:
**--rollout_strategy=managed**
Previously the only way to specify a version was using the following flag: 
**--version=SERVICE_CONFIG_ID**


simonz130:
Disregard this change - please, use my other change that is complete.

Issue #991, opened 2018-01-14
Typo in Firestore tutorial


CeH9:
Seems like in [comments](https://github.com/GoogleCloudPlatform/java-docs-samples/blob/master/firestore/src/main/java/com/example/firestore/snippets/References.java#L52-L53) should be "users", not "employees".

[here too](https://github.com/GoogleCloudPlatform/java-docs-samples/blob/master/firestore/src/main/java/com/example/firestore/snippets/References.java#L65-L66)

[Tutorial that reffers to this code](https://firebase.google.com/docs/firestore/data-model)

jabubake:
@campionfellin Tx for the fix !

PR #1057, opened 2018-03-13
Fix typo in comment

campionfellin:
This is a quick fix to Issue #991


lesv:
@dpebot merge when green

PR #1062, opened 2018-03-15
Add auth example for https://cloud.google.com/docs/authentication/production.

dzlier-gcp:


auth/README.md
```diff
@@ -0,0 +1,41 @@
...
+
+<a href="https://console.cloud.google.com/cloudshell/open?git_repo=https://github.com/GoogleCloudPlatform/java-docs-samples&page=editor&open_in_editor=auth/README.md">
+<img alt="Open in Cloud Shell" src ="http://gstatic.com/cloudssh/images/open-btn.png"></a>
+
+GCP client libraries use a strategy called Application Default Credentials (ADC) to find your 
+application's credentials. When your code uses a client library, the strategy checks for your 
+credentials in the following order: First, ADC checks to see if the environment variable 
+`GOOGLE_APPLICATION_CREDENTIALS` is set. If the variable is set, ADC uses the service account file 
+that the variable points to. If the environment variable isn't set, ADC uses the default service 
+account that Compute Engine, Kubernetes Engine, App Engine, and Cloud Functions provide, for
```
lesv: Do you wish to mention that it fetches from the Metadata server?

Also some of this is in transition.  @jonparrott PTAL

auth/src/main/java/com/google/cloud/auth/samples/AuthExample.java
```diff
@@ -0,0 +1,104 @@
...
+import com.google.common.collect.Lists;
+
+import java.io.FileInputStream;
+import java.io.IOException;
+
+/**
+ * Hit google cloud translate REST endpoint; uses google-auth library.
+ */
+public class AuthExample {
+  public static void main(String[] args) throws IOException {
```
lesv: I presume this is intentional.  Our standard tends to put main() at the end, but if you think it's better in front - ie. folks actually look at it before running, then I'm ok w/ it.

auth/README.md
dzlier-gcp: Oh, I just copied the most relevant paragraph from https://cloud.google.com/docs/authentication/production. I could also see what kind of comments are on the other repos' tutorials on the subject.

auth/src/main/java/com/google/cloud/auth/samples/AuthExample.java
dzlier-gcp: I don't have a preference, so I will move it to meet the standard.

This did make me notice that the class comment needs to be updated as well though.

auth/README.md
jonparrott: Could you just link to the docs instead? This is likely to get stale. :(

auth/src/main/java/com/google/cloud/auth/samples/AuthExample.java
```diff
@@ -0,0 +1,104 @@
...
+    Page<Bucket> buckets = storage.list();
+    for (Bucket bucket : buckets.iterateAll()) {
+      System.out.println(bucket.toString());
+    }
+  }
+  // [END auth_cloud_implicit]
+
+  // [START auth_cloud_explicit]
+  static void authExplicit(String jsonPath) throws IOException {
+    // You can specify a credential file by providing a path to GoogleCredentials.
```
jonparrott: The explicit sample should *only* show the case where it reads from a file, it shouldn't show the implicit case as well.

dzlier-gcp:
Done with feedback changes.

lesv:
Kokoro seems unhappy

PR #1063, opened 2018-03-16
Creating sample for StackDriver Alerts.

dzlier-gcp:


monitoring/v3/pom.xml
```diff
@@ -0,0 +1,50 @@
+<!--
+  Copyright 2016 Google Inc. All Rights Reserved.
```
SurferJeffAtGoogle: Update the year to 2018.

dzlier-gcp: Done

PR #1076, opened 2018-03-29
Remove basePath: "/"

navinger:
I'm updating all the openapi files in the getting-started sample in all the sample repos to remove basePath: "/"
Here's the reason from simonz130:

From the OpenAPI 2 spec:
* basePath: "If it is not included, the API is served directly under the host. The value MUST start with a leading slash (/). "
* Paths for methods: "A relative path to an individual endpoint. The field name MUST begin with a slash. The path is appended to the basePath in order to construct the full URL."

This OpenAPI getting-started sample have basePath: "/", which (per strict spec interpretation) means all the paths start with double-slashes. (e.g "//v1/shelves" rather than "/v1/shelves"). Removing basepath="/" fixes that.

### DIFF
--- a/endpoints/getting-started/openapi.yaml
+++ b/endpoints/getting-started/openapi.yaml
@@ -20,7 +20,6 @@ info:
   version: "1.0.0"
 host: "echo-api.endpoints.YOUR-PROJECT-ID.cloud.goog"
 # [END swagger]
-basePath: "/"
 consumes:
   - "application/json"
 produces:


lesv:
```
- testing endpoints/getting-started
[ERROR] src/main/java/com/example/endpoints/AuthInfoServlet.java:[1] (header) RegexpHeader: Line does not match expected header line of '^\W+\*?\W*Copyright 20\d\d Google Inc.$'.
[ERROR] src/main/java/com/example/endpoints/EchoServlet.java:[1] (header) RegexpHeader: Line does not match expected header line of '^\W+\*?\W*Copyright 20\d\d Google Inc.$'.
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-checkstyle-plugin:2.17:check (default) on project endpoints: You have 2 Checkstyle violations. -> [Help 1]
```

PR #1084, opened 2018-04-05
Updates for the Developer Portal

navinger:
Updates so the sample API can be used easily in the Developer Portal. 
-- In schemes, it has to be "http" so that you can test the echo API in the Developer Portal. 
-- In the echoMessage definition, adding type: "object" provides a better user experience in the Developer Portal.

### DIFF
--- a/endpoints/getting-started/openapi.yaml
+++ b/endpoints/getting-started/openapi.yaml
@@ -25,7 +25,9 @@ consumes:
 produces:
   - "application/json"
 schemes:
-  - "https"
+# Uncomment the next line if you configure SSL for this API.
+# - "https"
+  - "http"  
 paths:
   "/echo":
     post:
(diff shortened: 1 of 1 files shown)


endpoints/getting-started/openapi.yaml
```diff
@@ -25,7 +25,9 @@ consumes:
 produces:
   - "application/json"
 schemes:
-  - "https"
+# Uncomment the next line if you configure SSL for this API.
```
lesv: And comment out the `http` line

navinger:
I asked an engineer on the developer portal team if you really do have to comment out the http line, and his response was:
"no, multiple schemes are OK
it's not either/or"

Please merge this change to master. You can't test the sample code in the Developer Portal without this change.

lesv:
SGTM - though typically users want to lock down http when they use https.  We are encouraging folks to go http only.

lesv:
@dpebot merge when green

PR #1089, opened 2018-04-12
Report individual test statuses to kokoro.

dzlier-gcp:


lesv:
@dpebot merge when green

PR #1091, opened 2018-04-20
BigQuery: Deletes unused samples and standardizes region tags

alixhami:
Query samples moved to google-cloud-java in [PR 3180](https://github.com/GoogleCloudPlatform/google-cloud-java/pull/3180)


tswast:
Thanks! I know the "label" samples are out-of-date (well, they still work but there's a better way) now that the cloud client has support for them, but it's good to have them at least use the right region tag.

PR #1099, opened 2018-05-09
dpebot dependency updates + test fixes.

dzlier-gcp:
Also add "sponge_log" to end of failsafe error reports so they show up in fusion.

### DIFF
--- a/vision/beta/cloud-client/pom.xml
+++ b/vision/beta/cloud-client/pom.xml
@@ -40,12 +40,12 @@
     <dependency>
       <groupId>com.google.cloud</groupId>
       <artifactId>google-cloud-vision</artifactId>
-      <version>1.24.1</version>
+      <version>1.29.0</version>
     </dependency>
     <dependency>
       <groupId>com.google.cloud</groupId>
       <artifactId>google-cloud-storage</artifactId>
-      <version>1.24.1</version>
+      <version>1.29.0</version>
     </dependency>
     <!-- [END dependencies] -->
 
@@ -60,7 +60,7 @@
     <dependency>
       <groupId>com.google.truth</groupId>
       <artifactId>truth</artifactId>
-      <version>0.39</version>
+      <version>0.40</version>
       <scope>test</scope>
     </dependency>
   </dependencies>
--- a/vision/beta/cloud-client/src/test/java/com/example/vision/DetectIT.java
+++ b/vision/beta/cloud-client/src/test/java/com/example/vision/DetectIT.java
@@ -244,7 +244,7 @@ public class DetectIT {
 
     // Assert
     String got = bout.toString();
-    assertThat(got).contains("Landmark");
+    assertThat(got).contains("History");
     assertThat(got).contains("Best guess label: palace of fine arts");
   }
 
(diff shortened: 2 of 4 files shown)


dataflow/spanner-io/pom.xml
```diff
@@ -53,6 +53,7 @@
         <artifactId>maven-failsafe-plugin</artifactId>
         <version>2.19.1</version>
         <configuration>
+          <reportNameSuffix>sponge_log</reportNameSuffix>
```
lesv: Should this be in java-repo-tools or somewhere else?

kurtisvg: The parent pom is a version behind, it probably just needs to be manually updated since we released a new version

dzlier-gcp: This is the only failsafe plugin usage in the whole project, so I wasn't sure if it was necessary to move it up. But the parent pom is the version where I added sponge_log to surefire - it's just that failsafe and surefire don't use the same plugin config.

lesv: I ask questions when I review the answer can be, (and in this case appears to be), NO, this is the right spot.

dzlier-gcp:
Updating the apache beam version causes this error in the spanner tests:

```
Tests run: 4, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 19.525 sec <<< FAILURE! - in com.example.dataflow.SpannerReadIT
readDbEndToEnd(com.example.dataflow.SpannerReadIT)  Time elapsed: 5.692 sec  <<< ERROR!
org.apache.beam.sdk.Pipeline$PipelineExecutionException: com.google.cloud.spanner.SpannerException: INVALID_ARGUMENT: io.grpc.StatusRuntimeException: INVALID_ARGUMENT: Query is not root partitionable since it does not have a DistributedUnion at the root. Please run EXPLAIN for query plan details.
	at com.example.dataflow.SpannerReadIT.readDbEndToEnd(SpannerReadIT.java:140)
Caused by: com.google.cloud.spanner.SpannerException: INVALID_ARGUMENT: io.grpc.StatusRuntimeException: INVALID_ARGUMENT: Query is not root partitionable since it does not have a DistributedUnion at the root. Please run EXPLAIN for query plan details.
Caused by: java.util.concurrent.ExecutionException: io.grpc.StatusRuntimeException: INVALID_ARGUMENT: Query is not root partitionable since it does not have a DistributedUnion at the root. Please run EXPLAIN for query plan details.
Caused by: io.grpc.StatusRuntimeException: INVALID_ARGUMENT: Query is not root partitionable since it does not have a DistributedUnion at the root. Please run EXPLAIN for query plan details.
```

I can't find much information about how to ensure the code in `[SpannerReadAll.java](dataflow/spanner-io/src/main/java/com/example/dataflow/SpannerReadAll.java)` has a Distributed union at the root. When I try to mimic the Sql version of that in the console, it passes just fine. Is this a bug I should file with the spanner team, or should I revert the apache beam version?

lesv:
You might wish to make sure that beam is using `<provided>` for the spanner library, rather than including it themselves.  You might need to put in a `<excludes>` and the earlier version of the spanner client.  This doesn't mean that it's a drop in replacement, as our client library team has been know to change stuff.

dzlier-gcp:
I have still not gotten spanner tests to pass, but since Vision is working I would like to submit that part so that test starts passing, and then continue to work on the spanner part.

PR #1100, opened 2018-05-10
Add tags to CloudSQL samples

kurtisvg:
Add some anchors to allow sections to be linked to in the docs


flexible/cloudsql/src/main/resources/config.properties
```diff
@@ -11,5 +11,6 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-
+# [START properties]
```
lesv: This doesn't look like it's actually showing that much.  What are you teaching here?

PR #1101, opened 2018-05-11
Add more tags for CloudSQL

kurtisvg:
I missed a few doc tags in the last PR.

@piaxc - FYI


flexible/postgres/src/main/resources/config.properties
```diff
@@ -11,5 +11,6 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-
+# [START properties]
 sqlUrl=${sqlURL}
+# [END properties]
```
lesv: You might wish a new line here.

PR #1106, opened 2018-05-16
[DO NOT MERGE] Updating all the region tags to have a prefix of datastore

billyjacobson:
Updating all the region tags to have a prefix of datastore

### DIFF
--- a/datastore/pom.xml
+++ b/datastore/pom.xml
@@ -57,10 +57,10 @@
   </dependencies>
   <build>
     <plugins>
-<!-- // [START maven]-->
+<!-- // [START datastore_maven]-->
 
-<!-- // [END maven]-->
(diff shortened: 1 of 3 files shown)


datastore/pom.xml
```diff
@@ -57,10 +57,10 @@
   </dependencies>
   <build>
     <plugins>
-<!-- // [START maven]-->
+<!-- // [START datastore_maven]-->
```
lesv: Seems pointless.

billyjacobson: At the team summit we are fixing all the region tags so they can be imported into the new system

lesv: Ah, nope I meant the import of a blank line, not the change, I understand the change.

PR #1119, opened 2018-06-06
Add stackdriver alert v3 integration tests.

dzlier-gcp:


monitoring/v3/src/test/java/AlertIT.java
```diff
@@ -0,0 +1,107 @@
...
+  public void testBackupPolicies() throws IOException {
+    AlertSample.main(new String[]{"backup", "-j", policyFileName});
+    File backupFile = new File(policyFileName);
+    assertTrue(backupFile.exists());
+    String fileContents =
+        String.join("\n", Files.readLines(backupFile, StandardCharsets.UTF_8));
+    assertTrue(fileContents.contains("test-policy"));
+  }
+
+  // TODO: Complete restore backup test when parse/unparse issue is figured out.
```
kurtisvg: If there is a bug associated with this task, it might be useful to update the comment with the id for future use.

dzlier-gcp: Done.

PR #1120, opened 2018-06-06
Auto-update dependencies.

dzlier-gcp:
Same as #1098, but without update to apache beam version 2.4.0. Still working on that, but everything else passes fine.

### DIFF
--- a/appengine-java8/analytics/pom.xml
+++ b/appengine-java8/analytics/pom.xml
@@ -87,7 +87,7 @@
     <dependency>
       <groupId>org.mockito</groupId>
       <artifactId>mockito-core</artifactId>
-      <version>2.18.3</version>
+      <version>2.19.0</version>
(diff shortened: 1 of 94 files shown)


kurtisvg:
It looks like vision/beta/cloud-client is failing.

dzlier-gcp:
I have seen it timeout like that before, I think it might be flakey. Let me try running it again and see if it succeeds.

kurtisvg:
I rolled back the tests that were failing in PR #1098, then rebased this PR to reduce the time needed to run tests.

dzlier-gcp:
Thanks for doing that - still not sure why this test is timing out now. Will look into that today.

dzlier-gcp:
Added commit updating version of guava in newly failing tests.

dzlier-gcp:
Tests pass, just need a LGTM.

PR #1128, opened 2018-06-12
Adding Spanner STRUCT param samples

jabubake:


spanner/cloud-client/src/test/java/com/example/spanner/SpannerSampleIT.java
```diff
@@ -133,6 +133,20 @@ public void testSample() throws Exception {
     assertThat(out).contains("1 4 2017-10-05 11000");
     assertThat(out).contains("1 19 2017-11-02 15000");
     assertThat(out).contains("2 42 2017-12-23 7000");
+
+    runSample("writestructdata");
+    out = runSample("querywithstruct");
+    assertThat(out).contains("6");
```
kurtisvg: These "contains" asserts should considered switching to "equals" since if the input contains a 6 in the stack trace they may still pass.

jabubake: Ack, will confirm changes work once the library is released.

jabubake:
@kurtisvg @jsimonweb FYI : would be great if you could review + merge.

PR #1144, opened 2018-06-22
Fix auth sample to explicitly set the projectId

chingor13:
When explicitly setting credentials for a service, we should also advise the user to explicitly set their project id. `ServiceOptions` will not look for the project id in the service account json if provided (See https://github.com/GoogleCloudPlatform/google-cloud-java/issues/3386). Instead, it will still use ADC's rules for finding the projectId with may or may not be what the user intends.

Context: internal b/110528397

### DIFF
--- a/auth/src/main/java/com/google/cloud/auth/samples/AuthExample.java
+++ b/auth/src/main/java/com/google/cloud/auth/samples/AuthExample.java
@@ -96,24 +96,25 @@ public class AuthExample {
       authImplicit();
       return;
     }
-    String suite = args[0];
-    if ("explicit".equals(suite)) {
-      if (args.length >= 3) {
-        authExplicit(args[1], args[2]);
-      } else {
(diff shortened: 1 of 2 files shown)


chingor13:
cc @jabubake

jabubake:
@hostirosti @theacodes FYI

hostirosti:
Can we hold back on this change? It doesn't address the underlying issue of inconsistent behavior between credentials provided via GOOGLE_APPLICATION_CREDENTIALS and .setCredentials(...) using the same service account credentials both times.

kurtisvg:
@chingor13 @hostirosti Is this change still wanted? If not, are there any objections to closing the PR?

hostirosti:
I lost track, but I believe the underlying issue was fixed iirc. Can you check with @theacodes around this? If the project id is picked up from the credentials this change is obsolete afaik.

PR #1153, opened 2018-07-13
Add task create & lease/acknowledge sample.

dzlier-gcp:
Still working on the appengine version, but thought I would create the pull request now so reviews can start, since we were wanting this done by EOD.

I do not have a README currently because the content seems copied from the documentation at https://cloud.google.com/tasks/docs/quickstart-pull, which someone else will be working on this weekend. I will grab it from there and pull it down when it is done.

### DIFF
--- a/appengine-java8/tasks/src/main/java/com/example/task/CreateTask.java
+++ b/appengine-java8/tasks/src/main/java/com/example/task/CreateTask.java
@@ -1,0 +1,155 @@
+/*
+ * Copyright 2018 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.example.task;
+
+import com.google.cloud.tasks.v2beta2.AppEngineHttpRequest;
+import com.google.cloud.tasks.v2beta2.CloudTasksClient;
+import com.google.cloud.tasks.v2beta2.HttpMethod;
+import com.google.cloud.tasks.v2beta2.QueueName;
+import com.google.cloud.tasks.v2beta2.Task;
+import com.google.common.base.Strings;
+import com.google.protobuf.ByteString;
+import com.google.protobuf.Timestamp;
+
+import java.nio.charset.Charset;
+import java.time.Clock;
+import java.time.Instant;
+
+import org.apache.commons.cli.CommandLine;
+import org.apache.commons.cli.CommandLineParser;
+import org.apache.commons.cli.DefaultParser;
+import org.apache.commons.cli.HelpFormatter;
+import org.apache.commons.cli.Option;
+import org.apache.commons.cli.Options;
+import org.apache.commons.cli.ParseException;
+
+public class CreateTask {
+  private static String GGOGLE_CLOUD_PROJECT_KEY = "GOOGLE_CLOUD_PROJECT";
+
+  private static Option PROJECT_ID_OPTION = Option.builder("pid")
+      .longOpt("project-id")
+      .desc("The Google Cloud Project, if not set as GOOGLE_CLOUD_PROJECT env var.")
+      .hasArg()
+      .argName("project-id")
+      .type(String.class)
+      .build();
+
+  private static Option QUEUE_OPTION = Option.builder("q")
+      .required()
+      .longOpt("queue")
+      .desc("The Cloud Tasks queue.")
+      .hasArg()
+      .argName("queue")
+      .type(String.class)
+      .build();
+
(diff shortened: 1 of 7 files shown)


appengine-java8/tasks/src/main/java/com/example/task/CreateTask.java
```diff
@@ -0,0 +1,155 @@
...
+  private static Option IN_SECONDS_OPTION = Option.builder("s")
+      .longOpt("in-seconds")
+      .desc("Schedule time for the task to create.")
+      .hasArg()
+      .argName("in-seconds")
+      .type(int.class)
+      .build();
+
+  public static void main(String... args) throws Exception {
+    // [START cloud_tasks_appengine_create_task]
```
kurtisvg: Are you sure that you want the region tag to start here? It feels like most of the command line parsing isn't relevant enough that we want to link to it from elsewhere.

appengine-java8/tasks/src/main/java/com/example/task/TaskServlet.java
```diff
@@ -0,0 +1,48 @@
...
+    description = "Create Cloud Task",
+    urlPatterns = "/tasks/create"
+)
+public class TaskServlet extends HttpServlet {
+  private static Logger log = Logger.getLogger(TaskServlet.class.getName());
+
+  @Override
+  public void doPost(HttpServletRequest req, HttpServletResponse resp) throws IOException {
+    log.info("Received task request: " + req.getServletPath());
+    if (req.getParameter("payload") != null) {
```
kurtisvg: It might be a good idea to add a line of logging for when the payload is null - that way if the user forgets to set a payload the endpoint will still logged that it was triggered.

tasks/pom.xml
```diff
@@ -0,0 +1,95 @@
...
+
+  <build>
+    <sourceDirectory>src/main/java</sourceDirectory>
+    <plugins>
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-compiler-plugin</artifactId>
+        <version>3.2</version>
+      </plugin>
+      <plugin>
```
kurtisvg: Is building the jar with dependencies required for deployment? If not, can we remove this section?

tasks/src/main/java/com/example/Quickstart.java
```diff
@@ -0,0 +1,173 @@
...
+          .build());
+      System.out.println("Task created: " + newTask.getName());
+    }
+  }
+  // [END cloud_tasks_create_task]
+
+  // [START cloud_tasks_lease_and_acknowledge_task]
+  private static void pullAndAckTask(String projectId, String queueName, String location) {
+    try (CloudTasksClient client = CloudTasksClient.create()) {
+      LeaseTasksRequest.Builder reqBuilder = LeaseTasksRequest.newBuilder()
```
kurtisvg: Any reason you are passing a builder instead of just the request?

tasks/src/main/java/com/example/Quickstart.java
```diff
@@ -0,0 +1,173 @@
...
+          .setMaxTasks(1)
+          .setResponseView(Task.View.FULL);
+      LeaseTasksResponse response = client.leaseTasks(reqBuilder.build());
+      if (response.getTasksCount() == 0) {
+        System.out.println("No tasks found in queue.");
+        return;
+      }
+      Task task = response.getTasksList().get(0);
+      System.out.println("Leased task: " + task.getName());
+      client.acknowledgeTask(AcknowledgeTaskRequest
```
kurtisvg: What do you think about building the AcknowledgeTaskRequest before the function and passing it in? I think it increases readability and makes it more obvious what the function requires.

tasks/src/main/java/com/example/Quickstart.java
dzlier-gcp: I'm not, on line 148 it passes `reqBuilder.build()`

tasks/src/main/java/com/example/Quickstart.java
dzlier-gcp: Yeah that's fine.

appengine-java8/tasks/src/main/java/com/example/task/CreateTask.java
dzlier-gcp: I wasn't sure, but I also wanted to include enough context to understand what was happening. I don't know if these tags are used in the documentation, I was just copying them from the python sample. But I don't mind moving it.

appengine-java8/tasks/src/main/java/com/example/task/TaskServlet.java
dzlier-gcp: These log statements were actually just for my debugging purposes initially, but yeah I can add one for a null payload. Warning level probably, since it doesn't affect the execution?

dzlier-gcp:
Requested changes done.

appengine-java8/tasks/src/main/java/com/example/task/TaskServlet.java
kurtisvg: Typically logging is debug < info < warning < critical, so using ' debug' or 'info' is probably typical (but I don't think it makes a big difference for a smaller application).

tasks/src/main/java/com/example/Quickstart.java
kurtisvg: Yup that's what I meant - typically the builder pattern is a replacement for `new`, so it seems weird to trigger the build there.

PR #1155, opened 2018-07-16
Add maven-exec-plugin info to tasks/pom.xml.

dzlier-gcp:
Just copying the pom.xml sections from appengine/tasks/pom.xml. These plugin fields allow `exec:java` to complete without waiting for a loose thread from the tasks API. Otherwise, the `mvn exec:java` command will hang and then fail.


dzlier-gcp:
Can't seem to add Averi as a reviewer/assignee, cc'ing here. akitsch@google.com

tasks/pom.xml
```diff
@@ -75,6 +76,17 @@ Copyright 2018 Google LLC
 
   <build>
     <sourceDirectory>src/main/java</sourceDirectory>
+    <plugins>
+      <plugin>
+        <groupId>org.codehaus.mojo</groupId>
+        <artifactId>exec-maven-plugin</artifactId>
+        <version>${maven-exec-plugin.version}</version>
```
kurtisvg: Does this need to be a property? I feel like there were some issues with DPEbot's auto updating properties, but I might be remembering incorrectly.

dzlier-gcp: I'm not sure. I can change it though, I was just copying from another sample in this repo that uses this plugin.

PR #1156, opened 2018-07-16
Add Cloud Tasks Readmes

averikitsch:


appengine-java8/tasks/README.md
```diff
@@ -0,0 +1,105 @@
...
+`gcloud beta tasks queues describe $QUEUE_ID`, with the location embedded in
+the "name" value (for instance, if the name is
+"projects/my-project/locations/us-central1/queues/my-appengine-queue", then the
+location is "us-central1").
+
+```
+export LOCATION_ID=us-central1
+```
+
+Create a task, targeted at the `log_payload` endpoint, with a payload specified:
```
dzlier-gcp: Won't affect execution since the CreateTask class has it correct, but the endpoint should be 'tasks/create' here like you have below. Because it's important to be different from python.

appengine-java8/tasks/README.md
```diff
@@ -0,0 +1,105 @@
...
+```
+
+## Running the Samples
+
+Set environment variables:
+
+First, your project ID:
+
+```
+export PROJECT_ID=my-project-id
```
dzlier-gcp: We tend to prefer GOOGLE_CLOUD_PROJECT because gcloud will (sometimes) set this up for us, and then the "project" or "project-id" parameter can be omitted from some requests.

tasks/README.md
```diff
@@ -0,0 +1,64 @@
...
+From the project folder, build your project with:
+
+```
+mvn clean assembly:single
+```
+
+Optionally, you can set up your settings as environment variables:
+
+```
+export PROJECT_ID=<YOUR_PROJECT_ID>
```
dzlier-gcp: Same as above re: GOOGLE_CLOUD_PROJECT.

appengine-java8/tasks/README.md
averikitsch: Done.

appengine-java8/tasks/README.md
averikitsch: Done.

tasks/README.md
averikitsch: Done.

appengine-java8/tasks/README.md
```diff
@@ -0,0 +1,105 @@
...
+### Deploying
+
+```
+mvn appengine:deploy
+```
+
+Verify the index page is serving:
+
+```
+gcloud app browse
```
dzlier-gcp: Problem: this application doesn't have a base GET handler that would accept requests to the base path on either one, so that would need to be added. It would be pretty trivial to add a "hello world" message on get. If that is what you want to do, just let me know and I could add a commit to do that. Otherwise, we would just need to remove the "verify the index page is serving" step, since the app works without that needing to occur and it's just a sanity checking step for the user.

averikitsch: removed.

PR #1164, opened 2018-07-29
Remove deprecated references to categories and fix paths.

realjordanna:
These were removed when the API went GA. Also fix a couple typos, we refer to "info types" as InfoTypes in all formal documentation.


kurtisvg:
@dpebot merge when green

PR #1169, opened 2018-08-01
Create monitoring uptime check config samples.

dzlier-gcp:
Also organize the existing alert samples in a traditional directory, and clean up the code a bit.

The cleanup to AlertSample is just using a try() block to create the AlertPolicyClient or NotificationChannelServiceClient as needed instead of creating them for an instance of the AlertSample.


dzlier-gcp:
Crap, I had them all at 100 and then the review screen made it look like 120 was the limit.

PR #1172, opened 2018-08-03
Add IAM doc code snippets

awkoren:
Snippets for: 
- Quickstart
- Query grantable roles
- Service accounts
- Service account keys


awkoren:
1) Permissions needed for tests:

roles/iam.securityReviewer
roles/iam.serviceAccountAdmin
roles/iam.serviceAccountKeyAdmin

2) The error below seems to indicate the test cannot find the default project. Is GOOGLE_PROJECT_ID set in the test harness? 

"message" : "You don't have permission to query grantable roles for //cloudresourcemanager.googleapis.com/projects/null.",

iam/api-client/src/main/java/com/google/iam/snippets/GrantableRoles.java
```diff
@@ -0,0 +1,59 @@
+// Copyright 2018 Google Inc.
```
dzlier-gcp: Please use "LLC", as well as javadoc style comments in this and all other java files, i.e.

/** Copyright 2018 Google LLC
 *
 * Licensed...
 */

iam/api-client/pom.xml
```diff
@@ -0,0 +1,78 @@
...
+  <name>iam-snippets</name>
+
+    <!--
+    The parent pom defines common style checks and testing strategies for our samples.
+    Removing or replacing it should not affect the execution of the samples in anyway.
+  -->
+  <parent>
+    <groupId>com.google.cloud.samples</groupId>
+    <artifactId>shared-configuration</artifactId>
+    <version>1.0.9</version>
```
dzlier-gcp: Please update this to version 1.0.11.

iam/api-client/src/test/java/com/google/iam/snippets/GrantableRolesIT.java
```diff
@@ -0,0 +1,34 @@
...
+@RunWith(JUnit4.class)
+@SuppressWarnings("checkstyle:abbreviationaswordinname")
+public class GrantableRolesIT {
+
+  @Test
+  public void testQuickstart() throws Exception {
+    String project = System.getenv("GOOGLE_CLOUD_PROJECT");
+    String resource = "//cloudresourcemanager.googleapis.com/projects/" + project;
+
+    GrantableRoles.main(new String[] {resource});
```
dzlier-gcp: Can you intercept the output of the program running and then verify that the output is what you'd expect? There are several examples of doing this in other IT's in this repo, just search for "bout" and you'll find one.

iam/api-client/src/test/java/com/google/iam/snippets/QuickstartIT.java
```diff
@@ -0,0 +1,31 @@
...
+import org.junit.runner.RunWith;
+import org.junit.runners.JUnit4;
+
+@RunWith(JUnit4.class)
+@SuppressWarnings("checkstyle:abbreviationaswordinname")
+public class QuickstartIT {
+
+  @Test
+  public void testQuickstart() throws Exception {
+    Quickstart.main(new String[0]);
```
dzlier-gcp: Here as well, please verify the output matches what is expected when this runs. We don't just want to catch exceptions, we want the tests to verify that the output hasn't changed. Some calls will return error messages without actually throwing any exceptions.

iam/api-client/src/test/java/com/google/iam/snippets/ServiceAccountsIT.java
```diff
@@ -0,0 +1,50 @@
...
+import org.junit.runners.JUnit4;
+
+@RunWith(JUnit4.class)
+@SuppressWarnings("checkstyle:abbreviationaswordinname")
+public class ServiceAccountsIT {
+
+  @Test
+  public void testServiceAccounts() throws Exception {
+
+    String projectId = System.getenv("GOOGLE_CLOUD_PROJECT");
```
dzlier-gcp: Same

iam/api-client/pom.xml
awkoren: This results in an error. Also the root pom.xml and many of the other samples I looked at use 1.0.9 so I'm not sure why this change is needed?

Failure to find com.google.cloud.samples:shared-configuration:pom:1.0.11 in https://repo.maven.apache.org/maven2 was cached in the local repository, resolution will not be reattempted until the update interval of central has elapsed or updates are forced

iam/api-client/src/test/java/com/google/iam/snippets/GrantableRolesIT.java
awkoren: Done

iam/api-client/src/test/java/com/google/iam/snippets/ServiceAccountsIT.java
awkoren: Done

iam/api-client/src/main/java/com/google/iam/snippets/GrantableRoles.java
awkoren: Done -- however, this now causes the tests to fail: 

RegexpHeader: Line does not match expected header line of '^\W+\*?\W*Copyright 20\d\d Google Inc.$'.

Seems like the test needs to be updated to use LLC...

awkoren:
Addressed all feedback.
Note that changing "Google Inc" to "Google LLC" is causing the tests to fail: 
RegexpHeader: Line does not match expected header line of '^\W+\*?\W*Copyright 20\d\d Google Inc.$'.

dzlier-gcp:
Ah, I thought 1.0.11 was available, but try 1.0.10. If the tests still fail with that LLC/Inc change, then you can change them back and hopefully 1.0.11 will address it.

dzlier-gcp:
@dpebot Merge when green

awkoren:
It looks like all statuses are green and a reviewer has approved -- what else needs to happen to trigger dpebot?

PR #1177, opened 2018-08-09
Modified publish message command line arg

anguillanneuf:
Code expects arg[0] and arg[1]. README only provided arg[0].

### DIFF
--- a/pubsub/cloud-client/README.md
+++ b/pubsub/cloud-client/README.md
@@ -39,7 +39,7 @@ For more samples, see the samples in
 
 #### Publish messages
 ```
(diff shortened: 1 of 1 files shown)


pubsub/cloud-client/README.md
```diff
@@ -39,7 +39,7 @@ For more samples, see the samples in
 
 #### Publish messages
 ```
-  mvn exec:java -Dexec.mainClass=com.example.pubsub.PublisherExample -Dexec.args=my-topic
+  mvn exec:java -Dexec.mainClass=com.example.pubsub.PublisherExample -Dexec.args="my-topic, num-messages"
```
nnegrey: What do you think of just setting `num-messages` as `5`, since the comment below says that's what it does and more clearly portrays that, or you could update the comment on line 44. Either way, up to you. :)

PR #1179, opened 2018-08-13
update region tags

averikitsch:
Updating and removing region tags only

### DIFF
--- a/flexible/mailjet/src/main/appengine/app.yaml
+++ b/flexible/mailjet/src/main/appengine/app.yaml
@@ -6,7 +6,7 @@ handlers:
   script: this field is required, but ignored
 
 # [START gae_flex_mailjet_yaml]
-gae_flex_mailjet_yaml:
+env_variables:
   MAILJET_API_KEY: YOUR-MAILJET-API-KEY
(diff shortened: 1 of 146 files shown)


flexible/mailjet/src/main/appengine/app.yaml
```diff
@@ -5,8 +5,8 @@ handlers:
 - url: /.*
   script: this field is required, but ignored
 
-# [START env_variables]
-env_variables:
+# [START gae_flex_mailjet_yaml]
+gae_flex_mailjet_yaml:
```
kurtisvg: :thinking:

flexible/sendgrid/src/main/appengine/app.yaml
```diff
@@ -5,8 +5,8 @@ handlers:
 - url: /.*
   script: this field is required, but ignored
 
-# [START env_variables]
-env_variables:
+# [START gae_flex_sendgrid_yaml]
+gae_flex_sendgrid_yaml:
```
kurtisvg: Here too.

flexible/mailjet/src/main/appengine/app.yaml
averikitsch: 🤦

PR #1197, opened 2018-08-27
Remove unused dependency region tags and update used region tags

alixhami:
The goal with sample region tags is to have a globally unique tag for each sample. The "dependencies" tag is used throughout the java-docs-samples repo, which goes against best practices. This PR accomplishes the following:
+ For region tags that are not used on cloud.google.com, remove the region tag
+ For region tags that are used on cloud.google.com, add a prefix of product_java_ (ex. bigquery_java_dependencies)

Note that this PR does not include all products, only ones that I am responsible for adding to the sample tracker. Others will be performing this update for the remaining products.

### DIFF
--- a/bigquery/cloud-client/pom.xml
+++ b/bigquery/cloud-client/pom.xml
@@ -36,13 +36,13 @@
   </properties>
 
   <dependencies>
-    <!-- [START dependencies] -->
+    <!-- [START bigquery_java_dependencies] -->
     <dependency>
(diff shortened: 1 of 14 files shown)


alixhami:
I updated my approach with the unused region tags. Rather than deleting them, I gave them names that use the standard globally unique naming conventions, so that we avoid people adding the plain "dependencies" region tag in the future.

PR #1210, opened 2018-09-11
Fix DialogFlow tests and update to canonical sample format.

dzlier-gcp:
DialogFlow tests are failing, and while fixing them I ran into many issues with the tests simply parsing the `stdout` content to verify things were working as intended.

With canonical samples, we are no longer formatting the samples as commandline executables, and instead of writing to `stdout`, we are returning the requested information directly, as that is more likely to be what users are looking for anyway. So I updated the DF samples to do so, and the tests to match.


dzlier-gcp:
More context on @beccca 's response: 

The issue is with deleting the `System.out.println` lines in the classes, as these are determined by the DPEs in collaboration with the product teams to demonstrate the format of the responses returned from the various API calls in the code snippets embedded in our documentation. When thinking about the canonical format, I interpreted the decision that "we no longer have to make CLI's/executables" to mean there was no need to write to `stdout` if the code is not an executable, but that wasn't the main reason for the write lines in the first place. I will be adding those back in.

The issue I wanted to resolve in this change was making the functions actually return something other than void, for 2 reasons:

1. Unit tests that capture and regex match on stdout are brittle, prone to problems, and don't do much to verify that the call was actually successful, particularly when it's only matching on the manual string part and not on the content returned itself.
2. Functions like "list_" and "get_" that return void are functionally useless for anything other than documentation - they can't be copied into a developer's project and used as-is, because they don't do anything actually useful in practice.

dialogflow/cloud-client/src/main/java/com/example/dialogflow/DetectIntentKnowledge.java
```diff
@@ -74,10 +69,9 @@ public static void detectIntentKnowledge(
         // Build the query with the TextInput
         QueryInput queryInput = QueryInput.newBuilder().setText(textInput).build();
 
-        KnowledgeBaseName knowledgeBaseName = KnowledgeBaseName.of(projectId, knowledgeBaseId);
         QueryParameters queryParameters =
             QueryParameters.newBuilder()
-                .addKnowledgeBaseNames(knowledgeBaseName.toString())
+                .addKnowledgeBaseNames(knowledgeBaseId)
```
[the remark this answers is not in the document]
nnegrey: I see in the test you pass in a `KnowledgeBaseName`

dzlier-gcp:
Updated to return the `System.out` lines in the sample classes.

dialogflow/cloud-client/src/main/java/com/example/dialogflow/DetectIntentKnowledge.java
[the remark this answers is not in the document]
dzlier-gcp: using `knowledgeBaseName.toString()` in fact was not working, but I will try it one more time to double check.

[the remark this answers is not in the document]
dzlier-gcp: (from chat)
There's no way to directly get the knowledgeBaseId from a KnowledgeBase object. You can either get the full name ("projects/..../knowledgeBases/ID") or the display name.

You could parse the ID out from the full name of course and pass that in, but that implies that the user is also expected to pass in the ID by itself. If they're supposed to be getting it from a KnowledgeBase object themselves, then they'd just have to do the same extra steps. If they are expected to just have it alone as a config value or something, that would work, but they could also just as easily store the entire KnowledgeBaseName as config and not just the ID part...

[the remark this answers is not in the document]
dzlier-gcp: However, I didn't realize that a Document needed to exist in order to get answers from `detectIntentKnowledge`, so adding that in now along with better validation in the unit test. It's very strange that the document just needs to exist, but nothing from it needs to be supplied in the call to `SessionsClient.detectIntent`

dialogflow/cloud-client/src/main/java/com/example/dialogflow/DetectIntentKnowledge.java
```diff
@@ -104,48 +101,14 @@ public static void detectIntentKnowledge(
           System.out.format(" - Answer: '%s'\n", answer.getAnswer());
           System.out.format(" - Confidence: '%s'\n", answer.getMatchConfidence());
         }
+
+        allKnowledgeAnswers.add(Tuple.of(text, sessionsClient
```
kurtisvg: Another suggestion on the new sample guidelines is to explicitly define variables ahead of time, and avoid multiple nesting in function calls.

dialogflow/cloud-client/src/main/java/com/example/dialogflow/ContextManagement.java
```diff
@@ -17,41 +17,38 @@
...
   /**
    * Lists contexts
+   *
    * @param sessionId Identifier of the DetectIntent session.
    * @param projectId Project/Agent Id.
+   * @return List of Contexts found.
    */
-  public static void listContexts(String sessionId, String projectId) throws Exception {
+  public static List<Context> listContexts(String sessionId, String projectId) throws Exception {
+    List<Context> contexts = Lists.newArrayList();
```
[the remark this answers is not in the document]
dzlier-gcp: What's your take on the return statement being inside vs outside the tags?

dialogflow/cloud-client/src/main/java/com/example/dialogflow/DetectIntentKnowledge.java
dzlier-gcp: Updated to break out nested calls.

dialogflow/cloud-client/src/main/java/com/example/dialogflow/ContextManagement.java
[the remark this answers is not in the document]
nnegrey: I'd say leave the return outside.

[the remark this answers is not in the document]
kurtisvg: I think that sounds reasonable.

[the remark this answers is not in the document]
dzlier-gcp: From the perspective of having these regions be copy-pastable by developers, though, I would think the function tag and return statement would be useful so that they could just plop the function into their code directly. They would basically have to write a function wrapper around the code anyway regardless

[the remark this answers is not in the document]
kurtisvg: I see what you are saying, but I think the presumption is that they can copy paste the code into an existing code and get it working. Even if they copy the function as a whole they still have to create a class around it, so I don't think including the function signature really make's a significant impact.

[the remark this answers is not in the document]
dzlier-gcp: hey every keystroke counts :)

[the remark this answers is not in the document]
nnegrey: I feel we should make sure this gets decided in the new rubric. (To have a solid answer going forward)

dialogflow/cloud-client/src/test/java/com/example/dialogflow/KnowledgeBaseManagementIT.java
```diff
@@ -35,100 +47,128 @@
...
   }
 
-}
\ No newline at end of file
+  @Test
+  public void testDetectIntentKnowledge() throws Exception {
+    // Create a Knowledge Base
+    KnowledgeBase knowledgeBase =
+        KnowledgeBaseManagement.createKnowledgeBase(PROJECT_ID, KNOWLEDGE_BASE_NAME);
+    String knowledgeBaseId = knowledgeBase.getName();
```
nnegrey: Rename to `knowledgeBaseName`

dialogflow/cloud-client/src/main/java/com/example/dialogflow/DetectIntentKnowledge.java
```diff
@@ -16,52 +16,50 @@
...
-   * @param texts The texts to be processed.
+   * @param sessionId       Identifier of the DetectIntent session.
+   * @param languageCode    Language code of the query.
+   * @param texts           The texts to be processed.
+   * @return The KnowledgeAnswers found for each text.
    */
-  public static void detectIntentKnowledge(
+  public static List<Tuple<String, KnowledgeAnswers>> detectIntentKnowledge(
       String projectId,
       String knowledgeBaseId,
```
nnegrey: Rename to `knowledgeBaseName` to match the API and what is being passed in.

dialogflow/cloud-client/src/main/java/com/example/dialogflow/DetectIntentKnowledge.java
```diff
@@ -16,52 +16,50 @@
...
-   * @param sessionId Identifier of the DetectIntent session.
-   * @param languageCode Language code of the query.
-   * @param texts The texts to be processed.
+   * @param sessionId       Identifier of the DetectIntent session.
+   * @param languageCode    Language code of the query.
+   * @param texts           The texts to be processed.
+   * @return The KnowledgeAnswers found for each text.
    */
-  public static void detectIntentKnowledge(
+  public static List<Tuple<String, KnowledgeAnswers>> detectIntentKnowledge(
```
nnegrey: Curious (correct me if I'm wrong): Is the `Tuple` more of a Python thing? That is specially added as part of google.cloud.java.

Could a better return that makes navigating the answers for all of a `text` be?
`Hashmap<String, List<KnowledgeAnswers>>`

Downside being that you have to instantiate the `List` for each `text`

dialogflow/cloud-client/src/main/java/com/example/dialogflow/DocumentManagement.java
```diff
@@ -22,27 +22,23 @@
...
 
+public class DocumentManagement {
   //  [START dialogflow_list_document]
+
   /**
-   * @param projectId Project/Agent id.
    * @param knowledgeBaseId Knowledge Base id.
    */
-  public static void listDocuments(String projectId, String knowledgeBaseId) throws Exception {
+  public static List<Document> listDocuments(String knowledgeBaseId)
```
nnegrey: Rename to `knowledgeBaseName`

dialogflow/cloud-client/src/main/java/com/example/dialogflow/DocumentManagement.java
```diff
@@ -51,22 +47,25 @@ public static void listDocuments(String projectId, String knowledgeBaseId) throw
...
+   * @param displayName     display name of the Document.
+   * @param mimeType        MIME type of the Document. e.g. text/csv, text/html
+   * @param knowledgeType   Knowledge Type of the Document. e.g. FAQ, EXTRACTIVE_QA
+   * @param contentUri      Uri of the Document. e.g. gs://path/mydoc.csv, http://mypage.com/faq.html
+   * @return The created document.
    */
-  public static void createDocument(
-      String projectId,
+  public static Document createDocument(
       String knowledgeBaseId,
```
nnegrey: Rename to `knowledgeBaseName`

dialogflow/cloud-client/src/main/java/com/example/dialogflow/DocumentManagement.java
```diff
@@ -82,39 +81,39 @@ public static void createDocument(
...
   // [START dialogflow_get_document]
+
   /**
-   * @param projectId Project/Agent id.
-   * @param knowledgeBaseId Knowledge Base id.
    * @param documentId Document Id.
+   * @return The requested document.
    */
-  public static void getDocument(String projectId, String knowledgeBaseId, String documentId)
+  public static Document getDocument(String documentId)
```
nnegrey: Rename `documentName`

dialogflow/cloud-client/src/main/java/com/example/dialogflow/DocumentManagement.java
```diff
@@ -124,97 +123,23 @@ public static void getDocument(String projectId, String knowledgeBaseId, String
...
 
   // [START dialogflow_delete_document]
+
   /**
-   * @param projectId Project/Agent id.
-   * @param knowledgeBaseId Knowledge Base id.
    * @param documentId Document Id.
    */
-  public static void deleteDocument(String projectId, String knowledgeBaseId, String documentId)
+  public static void deleteDocument(String documentId)
```
nnegrey: `documentName`

dialogflow/cloud-client/src/main/java/com/example/dialogflow/KnowledgeBaseManagement.java
```diff
@@ -74,86 +78,31 @@ public static void createKnowledgeBase(String projectId, String displayName) thr
 
   /**
    * @param knowledgeBaseId Knowledge base id.
-   * @param projectId Project/agent id.
+   * @return The retrieved KnowledgeBase.
    */
-  public static void getKnowledgeBase(String projectId, String knowledgeBaseId) throws Exception {
-
+  public static KnowledgeBase getKnowledgeBase(String knowledgeBaseId) throws Exception {
```
nnegrey: `knowledgeBaseName`

dialogflow/cloud-client/src/main/java/com/example/dialogflow/KnowledgeBaseManagement.java
```diff
@@ -74,86 +78,31 @@ public static void createKnowledgeBase(String projectId, String displayName) thr
...
+
   // [START dialogflow_delete_knowledge_base]
 
   /**
    * @param knowledgeBaseId Knowledge base id.
-   * @param projectId Project/agent id.
    */
-  public static void deleteKnowledgeBase(String projectId, String knowledgeBaseId)
-      throws Exception {
+  public static void deleteKnowledgeBase(String knowledgeBaseId) throws Exception {
```
nnegrey: `knowledgeBaseName`

dialogflow/cloud-client/src/test/java/com/example/dialogflow/KnowledgeBaseManagementIT.java
```diff
@@ -35,100 +47,128 @@
...
-    documentManagement.listDocuments(PROJECT_ID,knowledgeBaseId);
-    got = bout.toString();
-    assertThat(got).contains("Display Name: " + DOCUMENT_BASE_NAME);
-    documentId = got.split("documents/")[1].split("- MIME Type")[0].trim();
+  @Test
+  public void testDocumentManagement() throws Exception {
+    // Create a Knowledge Base
+    KnowledgeBase knowledgeBase =
+        KnowledgeBaseManagement.createKnowledgeBase(PROJECT_ID, KNOWLEDGE_BASE_NAME);
+    String knowledgeBaseId = knowledgeBase.getName();
```
nnegrey: `knowledgeBaseName`

dialogflow/cloud-client/src/main/java/com/example/dialogflow/DetectIntentKnowledge.java
dzlier-gcp: Originally I had this as a Map<String, KnowledgeAnswers>, but the downside is that maps aren't guaranteed order so you wouldn't actually know what the "last" response was. An ordered list of key value pairs was the best alternative I could think of. 

But as with the other conversation, that's more an issue of how this sample is set up in the first place: passing in multiple text inputs and then running them through a loop in the method instead of having the method just accept a single text and running it multiple times in the tests/usecase. Might be worth just refactoring it that way to simplify.

nnegrey: True, but you know the order in which you sent the `texts` so you could use that to search for the answers for each `text`. 

Guess it's sort of a trade off, my only real concern is the use of the `Tuple` and how that flows with Java

dzlier-gcp: I can live with that

dzlier-gcp:
Hi all, I'm having an issue with this that I'm not sure I should report to the DialogFlow team directly as an actual bug found by the Unit Tests.

For the tests in [KnowledgeBaseManagementIT](https://github.com/GoogleCloudPlatform/java-docs-samples/pull/1210/files#diff-03a7d42fa3b59f957c05c29175020c23R146), particularly the testIntentKnowledge, it creates a KnowledgeBase based on the [Storage Docs FAQ](https://cloud.google.com/storage/docs/faq).

The test question used to be "Where is my data stored?", which is the second question in the Storage section. Half the time, it was interpreting that as "Is my data redundant?", the first question in that section. When I changed the test text to "Is my data redundant?", it always succeeded locally, but now it's having the same problem run with kokoro where it thinks the question is the other one half the time.

This seems like a potential API problem though, since the input text is an exact match for one of the questions. Should I file a bug with the team, or just make the test more forgiving?

nnegrey:
I'd start by reaching out to DF team and see if something changed on the backend, this sounds like an API bug.

dzlier-gcp:
Created bug b/115888589, but there was no automatic assignee so I'll forward it to the dialogflow team to make sure they see it.

dialogflow/cloud-client/src/main/java/com/example/dialogflow/DetectIntentKnowledge.java
```diff
@@ -104,48 +101,15 @@ public static void detectIntentKnowledge(
           System.out.format(" - Answer: '%s'\n", answer.getAnswer());
           System.out.format(" - Confidence: '%s'\n", answer.getMatchConfidence());
         }
+
+        KnowledgeAnswers answers = sessionsClient
+            .detectIntent(detectIntentRequest)
+            .getQueryResult()
```
madongfly: Why send the same RPC again instead of reusing the previously returned response?

dialogflow/cloud-client/src/test/java/com/example/dialogflow/KnowledgeBaseManagementIT.java
```diff
@@ -35,100 +47,127 @@
...
+    assertEquals(1, allAnswers.size());
+    KnowledgeAnswers knowledgeAnswers = allAnswers.get(TEXTS.get(0));
+    for (String text : TEXTS) {
+      assertEquals(1, knowledgeAnswers.getAnswersCount());
+      Answer answer = knowledgeAnswers.getAnswers(0);
+      assertEquals(text, answer.getFaqQuestion());
+      assertEquals(document.getName(), answer.getSource());
+      assertThat(answer.getAnswer()).contains("Cloud Storage");
+    }
+  }
```
madongfly: Please cleanup the resources (remove the knowledge base) after the test.

dialogflow/cloud-client/src/main/java/com/example/dialogflow/DetectIntentKnowledge.java
dzlier-gcp: Just a miss, good catch!

dialogflow/cloud-client/src/test/java/com/example/dialogflow/KnowledgeBaseManagementIT.java
dzlier-gcp: This is done in the `@After tearDown()` method above, line 77.

beccasaurus:
_ping_ ~ what's the status of this? looks like we've got LGTMs :)

dzlier-gcp:
Ah, thanks for the ping, I missed the LGTM. I will update the branch and submit.

PR #1211, opened 2018-09-12
Update Cloud tasks samples

averikitsch:


appengine-java8/tasks/README.md
```diff
@@ -79,23 +73,19 @@ Create a task, targeted at the `/tasks/create` endpoint, with a payload specifie
 
 ```
 mvn exec:java -Dexec.mainClass="com.example.task.CreateTask" \
-    -Dexec.args="--project-id $GOOGLE_CLOUD_PROJECT \
+    -Dexec.args="--project-id $PROJECT_ID \
```
dzlier-gcp: Why is this being changed? It's GOOGLE_CLOUD_PROJECT in all the samples I can think of, though I wouldn't be surprised if there were inconsistencies. Was there guidance on this that I missed?

appengine-java8/tasks/src/main/java/com/example/task/CreateTask.java
```diff
@@ -121,22 +121,36 @@ public static void main(String... args) throws Exception {
     String payload = params.getOptionValue(PAYLOAD_OPTION.getOpt(), "default payload");
 
     // [START cloud_tasks_appengine_create_task]
+    // Instantiates a client.
     try (CloudTasksClient client = CloudTasksClient.create()) {
+
+      // TODO(developer): Uncomment these lines and replace with your values.
+      // String project = "my-project-id";
```
dzlier-gcp: Vars should be `projectId` and `queueName` to match below if this is kept as is. But uncommenting these lines would be a compile error because these fields are all declared up above. In addition, they are all pulled from CLI opts so the user is presumably already supplying them if they are running the CLI, unless this comment is more aimed at docs readers than people running the sample.

averikitsch: Trying to comprise between the code snippet for the docs and running the sample with the CLI. Changing the comment to "Variables provided by the CLI."

appengine-java8/tasks/README.md
averikitsch: Trying to make it consistent between the samples; there are many inconsistencies between repos/samples. I don't think there is an official call.

averikitsch:
Pull queues are removed from the beta launch.

PR #1216, opened 2018-09-20
Fix broken sample and add test for sample

nnegrey:


video/cloud-client/src/test/java/com/example/video/DetectIT.java
```diff
@@ -51,7 +52,16 @@ public void tearDown() {
 
   @Test
   public void testLabels() throws Exception {
-    String[] args = {"labels", LABEL_FILE_LOCATION};
+    String[] args = {"labels", LABEL_GCS_LOCATION};
+    Detect.argsHelper(args);
+    String got = bout.toString();
+    // Test that the video with a cat has the whiskers label (may change).
+    assertThat(got.toUpperCase()).contains("WHISKERS");
```
kurtisvg: It might be worthwhile to update these tests to check that some labels were successfully returned, instead of a specific label (unless the docs specifically mention it should be detecting that label)

nnegrey: Like switching the functions to return the result and then check if the size of returned labels are not 0?

I know that for ML APIs, I believe we still want to validate some of the results to make sure the results are accurate. But let me go double check.

kurtisvg: Just something to consider if the labels change a lot. Not blocking PR.

nnegrey: Merging for now, will check with team on testing changes.

beccasaurus: +1 @kurtisvg *Love the idea!*

This came up in conversation only last week for testing responses from Machine Learning APIs based on models that _can and do change over time_

I love the idea of providing a list of test inputs/outputs and verifying that *at least 1* of the assertions passed

It is very much à la Data Driven Tests aka Parameterized Tests

(see: [JUnit4 Theories](https://github.com/junit-team/junit4/wiki/Theories) & other xUnit-style frameworks which implement 'Theories' for data-driven tests – [C# xUnit Theory](https://xunit.github.io/docs/getting-started-dotnet-core#write-first-theory), [Go Table-Driven-Tests](https://github.com/golang/go/wiki/TableDrivenTests), [Python parameterized package](https://github.com/wolever/parameterized), ...)

The main difference versus typical parameterized tests is that they usually exist to verify ALL SCENARIOS SUCCEED (versus *any one of many scenarios succeed*).

----

**pseudocode** _example of parameterized tests_

```php
method testMySample
  run_sample "input.png"

  assert output includes "Label: Dog"
end
```

^--- rather than 1 test that asserts "Dog" is returned as a label when analyzing an image file, provide a LIST of input/outputs to test

```php
test_cases = [
  [ "dog.png", "Label: Dog" ],
  [ "cat.png", "Label: Cat" ],
  [ "cat.png", "Label: Feline" ]
]

method data_driven_test file_name, output_assertion
    run_sample file_name

  assert output includes output_assertion
end
```

And we want to assert _at least one_ of the test cases passes

----

I love this. But I also want to be careful implementing it at this time. Because we have test-case consistency **across 7 languages**. When we start using this technique, I want to be clear about the expectations across languages. This also adds a new responsibility to the author of the canonical sample: rather than coming up with 1 test case, they will be tasked with coming up with multiple which should be stable over time – these should probably come from the **product team**.

#misc: _I'm in the design phases of a generator for code sample tests (initially only for code samples that are generated as well) and this is one of my design requirements. I am backing in table/data-driven tests with the toggle option of either verifying that ALL SUCCESS (default) or ANY ONE SUCCEEDS (available option)_

PR #1218, opened 2018-09-21
KMS changes

DanSanche:
implements changes suggested in b/113025362 and b/113025156


kms/src/main/java/com/example/Asymmetric.java
```diff
@@ -16,6 +16,7 @@
 
 package com.example;
 
+// [START kms_get_asymmetric_public]
```
kurtisvg: I don't think you should add the imports for the entire file. Several of these imports aren't even being used in the snippet that is requesting them.

One suggesting for handling this problem is the approach that's being used in #1232. This puts each snippet in it's own file, and the region tags include the imports.

Another suggestion is to add a comment detailing that the PemReader import is needed, and maybe even specify the mvn package name for the user to more easily add it to their pom.

kms/src/main/java/com/example/Asymmetric.java
```diff
@@ -114,20 +115,18 @@ public static String encryptRSA(String message, CloudKMS client, String keyPath)
...
-    byte[] ciphertext = cipher.doFinal(message.getBytes(StandardCharsets.UTF_8));
-    return Base64.getEncoder().encodeToString(ciphertext);
+    return cipher.doFinal(plaintext);
   }
   // [END kms_encrypt_rsa]
 
   // [START kms_sign_asymmetric]
   /** Create a signature for a message using a private key stored on Cloud KMS */
-  public static String signAsymmetric(String message, CloudKMS client, String keyPath)
+  public static byte[] signAsymmetric(byte[] message, CloudKMS client, String keyPath)
```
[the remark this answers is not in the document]
DanSanche: I think message is better in this context, since that's the term typically used in cryptography to describe the data being signed: https://en.wikipedia.org/wiki/Digital_signature

kms/src/main/java/com/example/Asymmetric.java
DanSanche: We had a discussion about the best way to present these imports in b/113025156, and we decided the universal import region tag would be sufficient. Is it more standard to include imports in each region tag? I could do it that way, but it would require a pretty big refactor, and new PRs for the other languages

(btw, I changed the region tag here - these should be imports for all asymmetric samples, not just get_asymmetric_public)

kurtisvg: Up until now the standard has been to not include imports in region tags. If we switch the new format, each snippet will be in it's own file so that only relevant imports will be shown. 

What page will this be on? I don't think lumping all the imports into one region tag is a good solution for multiple different snippets - it won't be clear which imports go to which snippet.

kurtisvg: The other suggestion (if you want to avoid a large refactor until the new format is decided upon) would be to update the comment for the functions using the 3rd party library with just that import. 

For example add something like `this snippet uses BouncyCastleProvider, which can be imported at org.bouncycastle.jce.provider.BouncyCastleProvider` to the function comment. This way the third party import would be visible and still only listed for the relevant functions.

DanSanche:
@kurtisvg I did some experimenting, and ended up ended up deciding it looks best with the comments. I think I'll remove the import region tags and go with this for the other languages

PR #1220, opened 2018-09-28
fixed a missing [ in the end tag

merla18:
needed for integration with product documentation


kurtisvg:
@dzlier-gcp Can you please merge this? It's a doc only change, but it is failing because language/automl isn't in the parent pom so the tests are outdated/failing.

PR #1228, opened 2018-10-11
Adds commands beta sample

gguuss:


iot/api-client/beta-features/manager/pom.xml
```diff
@@ -0,0 +1,132 @@
+<!--
+  Copyright 2018 Google Inc.
```
hongalex: Are we changing this to `Copyright 2018 Google LLC`?

PR #1231, opened 2018-10-12
Added Cloud SQL MySQL Servlet connectivity sample.

kurtisvg:
This is the canonical sample for a new series of samples highlighting connection best practices for Cloud SQL users. It's a simple web app that allows voting between two different groups ("Tabs" vs "Spaces") and records the votes into a MySQL backend.

@saturnism PTAL


cloud-sql/mysql/servlet/README.md
```diff
@@ -0,0 +1,61 @@
+# Connecting to Cloud SQL - MySQL
+
+## Before you begin
+
+1. If you haven't already, set up a Java Development Environment (including google-cloud-sdk and 
+maven utilities) by following the [java setup guide](https://cloud.google.com/java/docs/setup).
+
+1. Create a 2nd Gen Cloud SQL Instance by following these
```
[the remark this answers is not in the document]
lesv: https://cloud.google.com/resource-manager/docs/creating-managing-projects#creating_a_project

cloud-sql/mysql/servlet/README.md
```diff
@@ -0,0 +1,61 @@
...
+1. Create a service account with the 'Cloud SQL Client' permissions by following these 
+[instructions](https://cloud.google.com/sql/docs/mysql/connect-external-app#4_if_required_by_your_authentication_method_create_a_service_account).
+Download a JSON key to use to authenticate your connection. 
+
+1. Use the information noted in the previous steps:
+```bash
+export GOOGLE_APPLICATION_CREDENTIALS=/path/to/service/account/key.json
+export CLOUD_SQL_CONNECTION_NAME='<MY-PROJECT>:<INSTANCE-REGION>:<MY-DATABASE>'
+export DB_USER='my-db-user'
+export DB_PASS='my-db-pass'
```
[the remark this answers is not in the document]
kurtisvg: I've added a warning to the sections with environment variables that they aren't secure with a link to KMS as a secure solution. I would prefer to avoid using mvn / gradle to configure properties because it's permanent and forces the user to rebuild rather than restart the application. We are planning on adding instructions for GKE/GCE as well, which could mean a lot of repeated steps if they have to rebuild to change instance name.

cloud-sql/mysql/servlet/src/main/java/com/example/cloudsql/ConnectionPoolContextListener.java
```diff
@@ -0,0 +1,138 @@
...
+    // sit idle for this many milliseconds are retried if minimumIdle is exceeded.
+    config.setIdleTimeout(600000); // 10 minutes
+    // [END cloud_sql_idle_connections]
+    // [END_EXCLUDE]
+
+    // Initialize the connection pool using the configuration object.
+    DataSource pool = new HikariDataSource(config);
+    // [END cloud_sql_mysql_connection_pool]
+    return pool;
+  }
```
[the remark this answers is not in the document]
kurtisvg: I'm going to skip this for now as well - we want to expand the sample to additional run-times soon and I want to avoid any run-time specific options.

cloud-sql/mysql/servlet/src/main/java/com/example/cloudsql/ConnectionPoolContextListener.java
```diff
@@ -0,0 +1,142 @@
...
+
+  private void createTableSchema(DataSource pool) {
+    // Safely attempt to create the table schema.
+    try (Connection conn = pool.getConnection()) {
+      PreparedStatement createTableStatement = conn.prepareStatement(
+          "CREATE TABLE IF NOT EXISTS votes ( "
+              + "vote_id SERIAL NOT NULL, time_cast timestamp NOT NULL, canidate CHAR(6) NOT NULL, "
```
saturnism: what does `canidate` mean?

cloud-sql/mysql/servlet/src/main/java/com/example/cloudsql/IndexServlet.java
```diff
@@ -0,0 +1,139 @@
...
+      voteCtStmt.setString(1, "spaces");
+      ResultSet spacesResult = voteCtStmt.executeQuery();
+      spacesResult.next(); // Move to the first result
+      voteCt = spacesResult.getInt(1);
+
+    } catch (SQLException e) {
+      // If something goes wrong, the application needs to react appropriately. This might mean
+      // getting a new connection and executing the query again, or it might mean redirecting the
+      // user to a different page to let them know something went wrong.
+      throw new ServletException("Unable to successfully connect to the database. Please check the "
```
saturnism: pls propagate `e` in `ServletException`

See https://tomcat.apache.org/tomcat-5.5-doc/servletapi/javax/servlet/ServletException.html#ServletException(java.lang.String,%20java.lang.Throwable)

cloud-sql/mysql/servlet/src/main/java/com/example/cloudsql/IndexServlet.java
```diff
@@ -0,0 +1,139 @@
...
+  @Override
+  public void doPost(HttpServletRequest req, HttpServletResponse resp)
+      throws IOException {
+    // Get the team from the request and record the time of the vote.
+    String team = req.getParameter("team");
+    if (team != null) {
+      team = team.toLowerCase();
+    }
+    Timestamp now = new Timestamp(new Date().getTime());
+    if (team == null || !team.equals("tabs") && !team.equals("spaces")) {
```
saturnism: to be clear on the operation precedence, `if (team == null || ("tabs".equals(team) && !"spaces".equals(team))`

cloud-sql/mysql/servlet/src/main/java/com/example/cloudsql/IndexServlet.java
```diff
@@ -0,0 +1,139 @@
...
+      voteStmt.setString(2, team);
+
+      // Finally, execute the statement. If it fails, an error will be thrown.
+      voteStmt.execute();
+
+    } catch (SQLException e) {
+      // If something goes wrong, handle the error in this section. This might involve retrying or
+      // adjusting parameters depending on the situation.
+      // [START_EXCLUDE]
+      System.out.println("An SQL error occurred during executions: \n" + e.toString());
```
saturnism: same comments as above regarding exception handling.

cloud-sql/mysql/servlet/README.md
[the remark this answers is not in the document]
saturnism: +1; Definitely do not use Maven/Gradle configuration properties for runtime environment values. Those are not idiomatic.

cloud-sql/mysql/servlet/src/main/java/com/example/cloudsql/ConnectionPoolContextListener.java
```diff
@@ -0,0 +1,142 @@
...
+    }
+  }
+
+  @Override
+  public void contextDestroyed(ServletContextEvent event) {
+    // This function is called when the Servlet is destroyed.
+    DataSource pool = (DataSource) event.getServletContext().getAttribute("my-pool");
+    if (pool != null) {
+      try {
+        pool.unwrap(HikariDataSource.class).close();
```
saturnism: Since we know this is already an `HikariDataSource`, it should simply cast to it.

`HikariDataSource pool = (HikariDataSource) ...

cloud-sql/mysql/servlet/src/main/java/com/example/cloudsql/ConnectionPoolContextListener.java
kurtisvg: `canidate` is the subject of the vote (e.g. `tabs` or `spaces`).

cloud-sql/mysql/servlet/src/main/java/com/example/cloudsql/ConnectionPoolContextListener.java
```diff
@@ -0,0 +1,142 @@
...
+    // Safely attempt to create the table schema.
+    try (Connection conn = pool.getConnection()) {
+      PreparedStatement createTableStatement = conn.prepareStatement(
+          "CREATE TABLE IF NOT EXISTS votes ( "
+              + "vote_id SERIAL NOT NULL, time_cast timestamp NOT NULL, canidate CHAR(6) NOT NULL, "
+              + "PRIMARY KEY (vote_id) );"
+      );
+      createTableStatement.execute();
+    } catch (SQLException e) {
+      throw new Error(
```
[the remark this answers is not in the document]
kurtisvg: What is the preferred way to stop the application if this context fails to initialize correctly?

cloud-sql/mysql/servlet/src/main/webapp/index.jsp
```diff
@@ -0,0 +1,117 @@
...
+      } else {
+        leadTeam = "SPACES";
+        voteDiff = spaceVoteCt - tabVoteCt;
+      }
+    }
+%>
+<div class="section">
+    <div class="center">
+        <h4>
+            <% if(voteDiff != 0) { %>
```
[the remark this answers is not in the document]
kurtisvg: Could you provide more context on this? When should I use a taglib over the `<= ... %>`? Is `<%= ... &>` unsafe or inefficient? 

Is the advised taglib to use the JSTL core library?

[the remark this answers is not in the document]
saturnism: JSTL core library works. The taglibs are idiomatic in the JSP world, and you don't need to use `request.getAttribute()` to get the variables, since taglibs will be able to get the data directly from the attribute. It'll be cleaner, less code, and safer since it'll escape tags so you don't end up w/ script injections.

Just for example (and probably don't work as is since it's been a while since I had to write JSP :):

```
<c:choose>
<c:when test="${tabVoteCt == spaceVoteCt}">
  TABS and SPACES are tied
</c:when>
<c:when test="${tabVoteCt > spaceVoteCt}">
  TABS won
</c:when>
<c:when test="${tabVoteCt < spaceVoteCt}">
  SPACES won
</c:when>

</c:choose>
```

To iterate through the list:
```
<c:forEach items="${recentVotes}" var="vote">
  <li class="collection-item avatar">
    <c:choose>
      ...
    </c:choose>
  </li>
</c:forEach>
```

cloud-sql/mysql/servlet/src/main/java/com/example/cloudsql/ConnectionPoolContextListener.java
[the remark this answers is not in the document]
saturnism: I wouldn't use Error. Instead, throw a new `RuntimeException` w/ the cause. `throw new RuntimeException("oops", e);`

cloud-sql/mysql/servlet/src/main/java/com/example/cloudsql/ConnectionPoolContextListener.java
saturnism: is it candidate?

cloud-sql/mysql/servlet/src/main/java/com/example/cloudsql/IndexServlet.java
```diff
@@ -0,0 +1,139 @@
...
+
+      voteCtStmt.setString(1, "tabs");
+      ResultSet tabResult = voteCtStmt.executeQuery();
+      tabResult.next(); // Move to the first result
+      tabCt = tabResult.getInt(1);
+
+      voteCtStmt.setString(1, "spaces");
+      ResultSet spacesResult = voteCtStmt.executeQuery();
+      spacesResult.next(); // Move to the first result
+      voteCt = spacesResult.getInt(1);
```
saturnism: should this be `spaceCt`?

cloud-sql/mysql/servlet/src/main/webapp/index.jsp
```diff
@@ -0,0 +1,117 @@
...
+                <i class="material-icons circle blue">space_bar</i>
+                <% } %>
+                <span class="title">A vote for <b><%= v.getCandiate().toUpperCase() %></b></span>
+                <p>was cast at <%= v.getTimeCast() %>.</p>
+            </li>
+        <% } %>
+    </ul>
+</div>
+</body>
+<footer>
```
saturnism: `footer` tag should be part of the `body` tag. And if that's done, we should remove `footer` tag altogether and simply use `script` in the end.

```
<html>
  <head></head>
  <body>
    ...
    <script>...</script>
  </body>
</html>
```

cloud-sql/mysql/servlet/src/main/java/com/example/cloudsql/Vote.java
```diff
@@ -0,0 +1,47 @@
...
+ * limitations under the License.
+ */
+
+package com.example.cloudsql;
+
+import java.sql.Timestamp;
+
+public class Vote {
+
+  private String candiate;
```
saturnism: candidate?

cloud-sql/mysql/servlet/src/main/webapp/index.jsp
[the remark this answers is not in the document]
kurtisvg: Thanks for the background - I've made the adjustments.

cloud-sql/mysql/servlet/src/main/java/com/example/cloudsql/ConnectionPoolContextListener.java
kurtisvg: Sorry yes - my bad. Fixed.

cloud-sql/mysql/servlet/README.md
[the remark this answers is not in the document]
lesv: I might argue that, but I'm ok with it.

cloud-sql/mysql/servlet/src/main/java/com/example/cloudsql/IndexServlet.java
lesv: IRL you'd probably log the event, maybe add to some monitor value as well before propagating.

PR #1232, opened 2018-10-15
Propose new sample format.

kurtisvg:
Updated InspectString and InspectFile snippets to the newly proposed format.


dlp/src/main/java/dlp/snippets/InspectFile.java
```diff
@@ -0,0 +1,91 @@
...
+        System.out.println("\tQuote: " + f.getQuote());
+        System.out.println("\tInfo type: " + f.getInfoType().getName());
+        System.out.println("\tLikelihood: " + f.getLikelihood());
+      }
+    } catch (Exception e) {
+      System.out.println("Error during inspectFile: \n" + e.toString());
+    }
+  }
+}
+// [START dlp_inspect_file]
```
nnegrey: END

dlp/src/main/java/dlp.snippets/InspectFile.java
```diff
@@ -0,0 +1,91 @@
...
+        infoTypes.add(InfoType.newBuilder().setName(typeName).build());
+      }
+
+      // Set the inspect configuration for request
+      InspectConfig config = InspectConfig.newBuilder()
+          .addAllInfoTypes(infoTypes)
+          .setIncludeQuote(true)
+          .build();
+
+      // Construct request
```
lesv: comment seems unhelpful.  The rest are.  Could we say something else or additional?

dlp/src/main/java/dlp.snippets/InspectFile.java
```diff
@@ -0,0 +1,91 @@
...
+import com.google.privacy.dlp.v2.ProjectName;
+import com.google.protobuf.ByteString;
+import java.io.FileInputStream;
+import java.util.ArrayList;
+import java.util.List;
+
+public class InspectFile {
+
+  // Inspects the specified file.
+  public static void inspectFile(String projectId, String filePath, String fileType) {
```
lesv: Is this code threadsafe?  (I would suspect it is because the client is created here).  How do I get greater performance?  Should I cache the client(s)?

dlp/src/main/java/dlp.snippets/InspectFile.java
```diff
@@ -0,0 +1,91 @@
...
+
+      // Construct request
+      InspectContentRequest request = InspectContentRequest.newBuilder()
+          .setParent(project.toString())
+          .setItem(item)
+          .setInspectConfig(config)
+          .build();
+
+      // Run request and parse response
+      InspectContentResponse response = dlp.inspectContent(request);
```
lesv: How long does this take?  Should I make it async?  Do it in it's own thread, or in a separate microservice?  How does this code guide my architecture choices?

dlp/src/test/java/dlp/snippets/InspectTests.java
```diff
@@ -0,0 +1,82 @@
...
+    System.setOut(null);
+    bout.reset();
+  }
+
+  @Test
+  public void testInspectString() {
+    InspectString.inspectString(PROJECT_ID,  "I'm Gary and my email is gary@example.com");
+
+    String output = bout.toString();
+    assertThat(output, CoreMatchers.containsString("Info type: PHONE_NUMBER"));
```
dzlier-gcp: Wait, is this verifying that the DLP API found a PHONE_NUMBER in that test string? Because I don't see one. If the output is only verifying that it was _looking_ for phone numbers, then verifying that it was also looking for email addresses doesn't verify that it found one...

kurtisvg: The snippet prints out any findings out to console, so this test verifies that at least one of the findings had the infotype `PHONE_NUMBER`. 

The test was incorrect to look for an `"EMAIL_ADDRESS"` - I had changed the test string but forgot to update the following asserts.

dlp/src/main/java/dlp.snippets/InspectFile.java
kurtisvg: The snippets with longer running operations use PubSub actions for longer running jobs. Do you think that with a comparison between different snippets there is enough suggestions for architecture choices? If not, do you have suggestions on how to better convey it?

lesv: I'd be inclined to mention it in the preamble (class or method comments).

dlp/src/main/java/com/example/dlp/Inspect.java
```diff
@@ -72,165 +72,6 @@
 
 public class Inspect {
```
tswast: There's still a lot going on in this snippets class, including some command-line logic. Is this file intended to show the new sample format or just the other files?

dlp/src/main/java/dlp/snippets/InspectFile.java
```diff
@@ -0,0 +1,91 @@
...
+      // Construct the request to be sent by the client
+      InspectContentRequest request = InspectContentRequest.newBuilder()
+          .setParent(project.toString())
+          .setItem(item)
+          .setInspectConfig(config)
+          .build();
+
+      // Use the client to send the request and parse results
+      InspectContentResponse response = dlp.inspectContent(request);
+      System.out.println("Findings: " + response.getResult().getFindingsCount());
```
tswast: A little whitespace might between this and the above line might help show that line 79 is the one actually making the API request. In Python we had a pattern of doing

```
InspectContentResponse response = dlp.inspectContent(request);  // API request
```

to be 100% clear which lines are network calls. That might have been a pattern from when the client was less clear about what methods actually make requests, though.

dlp/src/main/java/com/example/dlp/Inspect.java
kurtisvg: I only ported over two of the samples into the new format - I'll move the rest over if/when the new format is finalized.

dlp/src/main/java/dlp/snippets/InspectFile.java
kurtisvg: Good point. I've added the break and split the comments to make sure it is clear.

tswast: Thanks!

dlp/src/main/java/com/example/dlp/Inspect.java
```diff
@@ -756,28 +597,7 @@ public static void main(String[] args) throws Exception {
...
-          filePath,
-          minLikelihood,
-          maxFindings,
-          infoTypesList,
-          customInfoTypesList,
-          includeQuote,
-          projectId);
-      // gcs file inspection
-    } else if (cmd.hasOption("gcs")) {
+    if (cmd.hasOption("gcs")) {
```
dzlier-gcp: I don't know if this is in scope or just because the executable already exists in this project, but are we going to continue to surface executables with new samples? Last I heard it was "no," but just wanted to confirm.

kurtisvg: No, the new format puts the nix on future CLIs

dlp/src/test/java/dlp/snippets/InspectTests.java
```diff
@@ -0,0 +1,91 @@
+/*
+ * Copyright 2018 Google LLC
```
dzlier-gcp: Should we start using 2019 now?

kurtisvg:
@dzlier-gcp Ready for another review.

kurtisvg:
@dzlier-gcp ping

dlp/src/test/java/dlp/snippets/RedactTests.java
```diff
@@ -16,6 +16,7 @@
 
 package dlp.snippets;
 
+// [START dlp_redact_image]
```
dzlier-gcp: Is this intentional? If it is there's no end tag.

kurtisvg: It was not intentional. Fixed.

PR #1235, opened 2018-10-17
Added missing region tags

navinger:
I want to update the docwidgets in the Adding API Management doc (https://cloud.google.com/endpoints/docs/frameworks/java/adding-api-management) to use the updated sample in appengine-java8/endpoints-v2-backend. But there were a few region tags missing, so I added them.


navinger:
@lesv PTAL

Issue #1239, opened 2018-10-25
Pub/Sub: java.lang.IllegalThreadStateException


anguillanneuf:
## In which file did you encounter the issue?

1. https://github.com/GoogleCloudPlatform/java-docs-samples/tree/master/pubsub/cloud-client/src/main/java/com/example/pubsub/CreatePullSubscriptionExample.java
2. https://github.com/GoogleCloudPlatform/java-docs-samples/tree/master/pubsub/cloud-client/src/main/java/com/example/pubsub/CreateTopicExample.java
3. https://github.com/GoogleCloudPlatform/java-docs-samples/tree/master/pubsub/cloud-client/src/main/java/com/example/pubsub/PublisherExample.java

## Describe the issue

Without `System.exit(0)` in main, users will see the following `java.lang.IllegalThreadStateException`. 

But adding it will cause tests to fail though. 

I opened a PR #1240 to have this information in the comments. 

```
[WARNING] thread Thread[ObjectCleanerThread,1,com.example.pubsub.PublisherExample] was interrupted but is still alive after waiting at least 15000msecs
[WARNING] thread Thread[ObjectCleanerThread,1,com.example.pubsub.PublisherExample] will linger despite being asked to die via interruption
[WARNING] thread Thread[grpc-default-worker-ELG-1-1,5,com.example.pubsub.PublisherExample] will linger despite being asked to die via interruption
[WARNING] thread Thread[grpc-default-worker-ELG-1-2,5,com.example.pubsub.PublisherExample] will linger despite being asked to die via interruption
[WARNING] thread Thread[grpc-default-worker-ELG-1-3,5,com.example.pubsub.PublisherExample] will linger despite being asked to die via interruption
[WARNING] thread Thread[grpc-shared-destroyer-0,5,com.example.pubsub.PublisherExample] will linger despite being asked to die via interruption
[WARNING] NOTE: 5 thread(s) did not finish despite being asked to  via interruption. This is not a problem with exec:java, it is a problem with the running code. Although not serious, it should be remedied.
[WARNING] Couldn't destroy threadgroup org.codehaus.mojo.exec.ExecJavaMojo$IsolatedThreadGroup[name=com.example.pubsub.PublisherExample,maxpri=10]
java.lang.IllegalThreadStateException
    at java.lang.ThreadGroup.destroy (ThreadGroup.java:778)
[...]

PR #1240, opened 2018-10-25
cleanupDeamonThreads to fix illegalThreadStateException

anguillanneuf:
I described why I wanted to add it in Issue #1239 .


kurtisvg:
@anguillanneuf - It also looks like `google-cloud-pubsub` has fallen out of date (`1.38.0` to `1.49.0`). I would suggest upgrading and see if perhaps the issue was already fixed before filing a bug against the client library.

PR #1241, opened 2018-10-26
Add OpenCensus sample code.

dzlier-gcp:
Java version of https://github.com/GoogleCloudPlatform/golang-samples/pull/646/files.

### DIFF
--- a/opencensus/pom.xml
+++ b/opencensus/pom.xml
@@ -1,0 +1,82 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+ Copyright 2018 Google LLC
+
+ Licensed under the Apache License, Version 2.0 (the "License");
+ you may not use this file except in compliance with the License.
+ You may obtain a copy of the License at
+
+       http://www.apache.org/licenses/LICENSE-2.0
+
+ Unless required by applicable law or agreed to in writing, software
+ distributed under the License is distributed on an "AS IS" BASIS,
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ See the License for the specific language governing permissions and
+ limitations under the License.
+-->
+<!-- [START opencensus_pom] -->
+<project>
+  <modelVersion>4.0.0</modelVersion>
+  <packaging>jar</packaging>
+  <groupId>com.example</groupId>
+  <artifactId>opencensus-samples</artifactId>
+  <version>1.0</version>
+
+  <!--
+    The parent pom defines common style checks and testing strategies for our samples.
+    Removing or replacing it should not affect the execution of the samples in anyway.
+  -->
+  <parent>
+    <groupId>com.google.cloud.samples</groupId>
+    <artifactId>shared-configuration</artifactId>
+    <version>1.0.10</version>
+  </parent>
+
+  <properties>
+    <maven.compiler.source>1.8</maven.compiler.source>
+    <maven.compiler.target>1.8</maven.compiler.target>
+    <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+    <opencensus.version>0.15.0</opencensus.version>
+  </properties>
+
+  <dependencies>
+    <!-- [START monitoring_opencensus_metrics_dependencies] -->
+    <dependency>
+      <groupId>io.opencensus</groupId>
+      <artifactId>opencensus-api</artifactId>
+      <version>${opencensus.version}</version>
+    </dependency>
+    <dependency>
+      <groupId>io.opencensus</groupId>
+      <artifactId>opencensus-impl</artifactId>
+      <version>${opencensus.version}</version>
+    </dependency>
+    <dependency>
+      <groupId>io.opencensus</groupId>
+      <artifactId>opencensus-exporter-stats-stackdriver</artifactId>
+      <version>${opencensus.version}</version>
+    </dependency>
+    <!-- [END monitoring_opencensus_metrics_dependencies] -->
+  </dependencies>
+
+  <build>
+    <plugins>
+      <plugin>
+        <groupId>org.codehaus.mojo</groupId>
+        <artifactId>exec-maven-plugin</artifactId>
+        <version>1.6.0</version>
+        <executions>
+          <execution>
+            <goals>
+              <goal>java</goal>
+            </goals>
+          </execution>
+        </executions>
+        <configuration>
+          <mainClass>com.example.opencensus.Quickstart</mainClass>
+        </configuration>
+      </plugin>
+    </plugins>
+  </build>
+</project>
+<!-- [END opencensus_pom] -->
--- a/opencensus/src/main/java/com/example/opencensus/Quickstart.java
+++ b/opencensus/src/main/java/com/example/opencensus/Quickstart.java
@@ -1,0 +1,88 @@
+/*
+ * Copyright 2018 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.example.opencensus;
+
+// [START monitoring_opencensus_metrics_quickstart]
+
+import com.google.common.collect.Lists;
+
+import io.opencensus.exporter.stats.stackdriver.StackdriverStatsExporter;
+import io.opencensus.stats.Aggregation;
+import io.opencensus.stats.BucketBoundaries;
+import io.opencensus.stats.Measure.MeasureLong;
+import io.opencensus.stats.Stats;
+import io.opencensus.stats.StatsRecorder;
+import io.opencensus.stats.View;
+import io.opencensus.stats.View.Name;
+import io.opencensus.stats.ViewManager;
+
+import java.io.IOException;
+import java.util.Collections;
+import java.util.Random;
+import java.util.concurrent.TimeUnit;
+
+public class Quickstart {
+  private static final int EXPORT_INTERVAL = 60;
+  private static final MeasureLong LATENCY_MS = MeasureLong.create(
+      "task_latency",
+      "The task latency in milliseconds",
+      "ms");
+  // Latency in buckets:
+  // [>=0ms, >=100ms, >=200ms, >=400ms, >=1s, >=2s, >=4s]
+  private static final BucketBoundaries LATENCY_BOUNDARIES = BucketBoundaries.create(
+      Lists.newArrayList(0d, 100d, 200d, 400d, 1000d, 2000d, 4000d));
+  private static final StatsRecorder STATS_RECORDER = Stats.getStatsRecorder();
+
+  public static void main(String[] args) throws IOException, InterruptedException {
+    // Register the view. It is imperative that this step exists,
+    // otherwise recorded metrics will be dropped and never exported.
+    View view = View.create(
+        Name.create("task_latency_distribution_4"),
+        "The distribution of the task latencies.",
+        LATENCY_MS,
+        Aggregation.Distribution.create(LATENCY_BOUNDARIES),
+        Collections.emptyList());
+
+    ViewManager viewManager = Stats.getViewManager();
+    viewManager.registerView(view);
+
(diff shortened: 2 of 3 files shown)


opencensus/src/main/java/com/example/opencensus/Quickstart.java
```diff
@@ -0,0 +1,87 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.example.opencensus;
+
+// [START monitoring_opencensus_metrics_quickstart]
```
SurferJeffAtGoogle: I don't see the corresponding closing tag.

opencensus/pom.xml
```diff
@@ -0,0 +1,82 @@
...
+  </parent>
+
+  <properties>
+    <maven.compiler.source>1.8</maven.compiler.source>
+    <maven.compiler.target>1.8</maven.compiler.target>
+    <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+  </properties>
+
+  <dependencies>
+    <!-- [START dependencies] -->
```
kurtisvg: This should probably start with `monitoring_opencensus` for the sample tracker

opencensus/src/main/java/com/example/opencensus/Quickstart.java
```diff
@@ -0,0 +1,87 @@
...
+
+public class Quickstart {
+  private static final int exportInterval = 60;
+  private static final MeasureLong latencyMs = MeasureLong.create(
+      "task_latency",
+      "The task latency in milliseconds",
+      "ms");
+  // Latency in buckets:
+  // [>=0ms, >=100ms, >=200ms, >=400ms, >=1s, >=2s, >=4s]
+  private static final BucketBoundaries boundaries = BucketBoundaries.create(
```
kurtisvg:
```suggestion
  private static final BucketBoundaries LATENCY_BOUNDARIES= BucketBoundaries.create(
```

opencensus/src/main/java/com/example/opencensus/Quickstart.java
```diff
@@ -0,0 +1,87 @@
...
+import io.opencensus.stats.View.Name;
+import io.opencensus.stats.ViewManager;
+
+import java.io.IOException;
+import java.util.Collections;
+import java.util.Random;
+
+public class Quickstart {
+  private static final int exportInterval = 60;
+  private static final MeasureLong latencyMs = MeasureLong.create(
```
kurtisvg:
```suggestion
  private static final MeasureLong LATENCY_MS = MeasureLong.create(
```

opencensus/src/main/java/com/example/opencensus/Quickstart.java
```diff
@@ -0,0 +1,87 @@
...
+import io.opencensus.stats.View;
+import io.opencensus.stats.View.Name;
+import io.opencensus.stats.ViewManager;
+
+import java.io.IOException;
+import java.util.Collections;
+import java.util.Random;
+
+public class Quickstart {
+  private static final int exportInterval = 60;
```
kurtisvg: [Constants](https://google.github.io/styleguide/javaguide.html#s5.2.4-constant-names) should be in capital case:
```suggestion
  private static final int EXPORT_INTERVAL = 60;
```

opencensus/src/main/java/com/example/opencensus/Quickstart.java
```diff
@@ -0,0 +1,87 @@
...
+      "task_latency",
+      "The task latency in milliseconds",
+      "ms");
+  // Latency in buckets:
+  // [>=0ms, >=100ms, >=200ms, >=400ms, >=1s, >=2s, >=4s]
+  private static final BucketBoundaries boundaries = BucketBoundaries.create(
+      Lists.newArrayList(0., 100., 200., 400., 1000., 2000., 4000.));
+  private static final StatsRecorder statsRecorder = Stats.getStatsRecorder();
+
+  public static void main(String[] args) throws IOException, InterruptedException {
```
kurtisvg: There was a conversation about a similiar Stackdriver sample not using main because web frameworks (servlets, SpringBoot, etc) don't typically have an accessible main method. - maybe ask @SurferJeffAtGoogle if that's applicable here.

opencensus/src/main/java/com/example/opencensus/Quickstart.java
```diff
@@ -0,0 +1,87 @@
...
+    // for more details.
+    StackdriverStatsExporter.createAndRegister();
+    // [END setup_exporter]
+
+    // [START example_code]
+    // Record 100 fake latency values between 0 and 5 seconds.
+    long millisPerSecond = 1000;
+    Random rand = new Random();
+    for (int i = 0; i < 100; i++) {
+      long ms = (long) (5 * millisPerSecond * rand.nextDouble());
```
kurtisvg: Use TimeUnit for more readable time conversions
```suggestion
      long ms = (long) (TimeUnit.MILLISECONDS.convert(5, TimeUnit.SECONDS) * rand.nextDouble());
```

opencensus/src/main/java/com/example/opencensus/Quickstart.java
```diff
@@ -0,0 +1,87 @@
...
+      System.out.println(String.format("Latency %d: %d", i, ms));
+      statsRecorder.newMeasureMap().put(latencyMs, ms).record();
+    }
+
+    // The default export interval is 60 seconds. The thread with the StackdriverStatsExporter must
+    // live for at least the interval past any metrics that must be collected, or some risk being
+    // lost if they are recorded after the last export.
+
+    Thread.sleep(millisPerSecond * exportInterval);
```
kurtisvg:
```suggestion
    Thread.sleep(TimeUnit.MILLISECONDS.convert(EXPORT_INTERVAL, TimeUnit.SECONDS));
```

opencensus/src/main/java/com/example/opencensus/Quickstart.java
```diff
@@ -0,0 +1,87 @@
...
+public class Quickstart {
+  private static final int exportInterval = 60;
+  private static final MeasureLong latencyMs = MeasureLong.create(
+      "task_latency",
+      "The task latency in milliseconds",
+      "ms");
+  // Latency in buckets:
+  // [>=0ms, >=100ms, >=200ms, >=400ms, >=1s, >=2s, >=4s]
+  private static final BucketBoundaries boundaries = BucketBoundaries.create(
+      Lists.newArrayList(0., 100., 200., 400., 1000., 2000., 4000.));
```
kurtisvg: Is a decimal type required here? Might be more clear to reflect that with a type suffix. (Eg: `0d, 100d, 200d`)

opencensus/src/main/java/com/example/opencensus/Quickstart.java
dzlier-gcp: Thanks, added.

opencensus/pom.xml
dzlier-gcp: I don't see that in the sample tracker. Only one I see is [this one](https://devrel-prod.googleplex.com/samples/1221?product=monitoring), which tracks `monitoring_opencensus_metrics_quickstart`

opencensus/src/main/java/com/example/opencensus/Quickstart.java
dzlier-gcp: Done

opencensus/src/main/java/com/example/opencensus/Quickstart.java
dzlier-gcp: Done

opencensus/src/main/java/com/example/opencensus/Quickstart.java
dzlier-gcp: Jeffrey, please weigh in. But from my perspective, the point of the sample is the code that's run within the `main` method. People who want to copy paste can just take that part and use it however they see fit. This sample in particular doesn't have any framework integration, so I just have the `exec:java` maven task configured in `pom.xml`.

opencensus/src/main/java/com/example/opencensus/Quickstart.java
dzlier-gcp: Done

opencensus/src/main/java/com/example/opencensus/Quickstart.java
dzlier-gcp: Done

opencensus/src/main/java/com/example/opencensus/Quickstart.java
dzlier-gcp: Done

opencensus/src/main/java/com/example/opencensus/Quickstart.java
dzlier-gcp: Done

opencensus/pom.xml
kurtisvg: According to the samples rubric, each region tag is supposed to be globally unique and begin with the product name (which seems to be `monitoring`). Maybe `monitoring_opencensus_metrics_dependencies` or `monitoring_opencensus_metrics_quickstart_dependencies` to avoid collisions?

dzlier-gcp: Ah I see, so it's not as much for the sample tracker dashboard, it's that the rubric is different from what we've been doing. Will do.

opencensus/src/main/java/com/example/opencensus/Quickstart.java
saturnism: tbh, keep the code simple. If the purpose is to demonstrate OpenCensus, then `main` is fine. If the purpose is to show OpenCensus measuring a Servlet, then use a Servlet.

Keeping things simple & to the point is important. See:

https://github.com/saturnism/grpc-java-by-example
https://github.com/saturnism/opencensus-java-by-example

opencensus/pom.xml
```diff
@@ -0,0 +1,82 @@
...
+        <executions>
+          <execution>
+            <goals>
+              <goal>java</goal>
+            </goals>
+          </execution>
+        </executions>
+        <configuration>
+          <mainClass>com.example.opencensus.Quickstart</mainClass>
+          <cleanupDaemonThreads>false</cleanupDaemonThreads>
```
saturnism: what is the reason for this flag?

opencensus/pom.xml
```diff
@@ -0,0 +1,82 @@
...
+    <maven.compiler.target>1.8</maven.compiler.target>
+    <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+  </properties>
+
+  <dependencies>
+    <!-- [START monitoring_opencensus_metrics_dependencies] -->
+    <dependency>
+      <groupId>io.opencensus</groupId>
+      <artifactId>opencensus-api</artifactId>
+      <version>0.15.0</version>
```
saturnism: use properties for the version.

opencensus/src/main/java/com/example/opencensus/Quickstart.java
```diff
@@ -0,0 +1,88 @@
...
+      long ms = (long) (TimeUnit.MILLISECONDS.convert(5, TimeUnit.SECONDS) * rand.nextDouble());
+      System.out.println(String.format("Latency %d: %d", i, ms));
+      STATS_RECORDER.newMeasureMap().put(LATENCY_MS, ms).record();
+    }
+
+    // The default export interval is 60 seconds. The thread with the StackdriverStatsExporter must
+    // live for at least the interval past any metrics that must be collected, or some risk being
+    // lost if they are recorded after the last export.
+
+    Thread.sleep(TimeUnit.MILLISECONDS.convert(EXPORT_INTERVAL, TimeUnit.SECONDS));
```
saturnism: if the export interval is 60s, shouldn't this wait 60s + n seconds?

opencensus/src/main/java/com/example/opencensus/Quickstart.java
SurferJeffAtGoogle: The purpose of this particular sample is to demonstrate the simplest case of OpenCensus, so let's keep the `main` function.

opencensus/pom.xml
dzlier-gcp: I copied this build from another sample and kept everything the same except the name and mainClass. If it's not needed I can remove it.

opencensus/src/main/java/com/example/opencensus/Quickstart.java
dzlier-gcp: This sleeps for 60 seconds after the last metric was recorded, to make sure at least one export cycle happens after it's done recording.

PR #1242, opened 2018-10-28
Make "Build Project" complete successfully.

prafullkotecha:
Using "IntelliJ Idea CE". Build Project fails with bunch of compilation issues. This commit simply fixes errors, but doesn't address any of the warnings.


appengine-java8/mail/pom.xml
```diff
@@ -46,6 +46,12 @@ Copyright 2016 Google Inc.
       <artifactId>mail</artifactId>
       <version>1.4.7</version>
     </dependency>
+      <dependency>
+          <groupId>com.google.appengine</groupId>
+          <artifactId>appengine-api-1.0-sdk</artifactId>
+          <version>RELEASE</version>
```
kurtisvg: Please change from `RELEASE` to the latest version number instead (for all dependencies). 

We have a tool that incrementally updates to ensure the samples are working correctly (it's not running for this project because it isn't linked in the main pom.)

appengine/datastore/src/main/java/com/example/appengine/AutoValue_Greeting.java
```diff
@@ -0,0 +1,28 @@
+package com.example.appengine;
```
kurtisvg: This class doesn't seem necessary?

kurtisvg:
There are a variety of checkstyle violations as well - please see the Resultstore link in the checks section.

prafullkotecha:
> Thanks for taking the time to submit this. I left a couple of comments that need to be fixed before we can merge.
> 
> If you prefer, please enable us to push commits to this PR and I can make the adjustments.

@kurtisvg - How do I enable that?

kurtisvg:
@prafullkotecha https://help.github.com/articles/allowing-changes-to-a-pull-request-branch-created-from-a-fork/

prafullkotecha:
@kurtisvg - That is already enabled. See attached screenshot.
screen shot 2019-02-08 at 2 14 16 pm

kurtisvg:
@prafullkotecha What command were you using the build this that it failed? I thought maybe it had been skipped by the typical tests, but it looks like it's building correctly on the current master branch

kurtisvg:
So it looks like there are mainly two folders in this PR that were changed `appengine/datastore` and ` appengine-java8/mail`. Both are able to compile fine (both locally and running in our tests) on the current version of master.

I'm going to close this and guess it was resolved in a previous commit - if you are still having issues with this please let me know so we try to troubleshoot. Thanks!

prafullkotecha:
I still some issues in `appengine-java8/mail` project. They look like copy/paste errors between `appengine` and `appengine-java8` modules. I will create a new PR for those fixes.

kurtisvg:
Can you provide any more info on the error? How are you attempting to compile?

kurtisvg:
Ok so it looks like the pom itself didn't have the correct parent declared, so when running from the directory itself was creating an issue. I opened #1333 to address. If you can confirm this is the issue you are running into, that would be great.

prafullkotecha:
Yes. That's one of the issues. I have cloned the whole repo, and imported it as one big project, with multiple modules and module groups into intelliJ. Then, I just run `Build Project` command from the `Build` menu. I get all sorts of error, in all kinds of projects. The links between parent pom, and children poms, and grandchildren poms, all need to be fixed.

PR #1243, opened 2018-10-29
[Firestore] Snippets for document snapshots and paginated queries. 

kurtisvg:
Add Java snippets for the [Paginating Data with Query Cursors](https://cloud.google.com/firestore/docs/query-data/query-cursors) query.

### DIFF
--- a/firestore/src/main/java/com/example/firestore/snippets/QueryDataSnippets.java
+++ b/firestore/src/main/java/com/example/firestore/snippets/QueryDataSnippets.java
@@ -21,16 +21,23 @@ import com.example.firestore.snippets.model.City;
 import com.google.api.core.ApiFuture;
 import com.google.api.core.ApiFutures;
 import com.google.cloud.firestore.CollectionReference;
+import com.google.cloud.firestore.DocumentReference;
 import com.google.cloud.firestore.DocumentSnapshot;
 import com.google.cloud.firestore.Firestore;
 import com.google.cloud.firestore.Query;
 import com.google.cloud.firestore.Query.Direction;
+import com.google.cloud.firestore.QueryDocumentSnapshot;
 import com.google.cloud.firestore.QuerySnapshot;
 import com.google.cloud.firestore.WriteResult;
 
+import com.google.firestore.v1beta1.Document;
+import com.google.protobuf.Api;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
 
 /** Snippets to support firestore querying data documentation. */
 class QueryDataSnippets {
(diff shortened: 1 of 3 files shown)


firestore/src/test/java/com/example/firestore/snippets/QueryDataSnippetsIT.java
```diff
@@ -197,6 +197,20 @@ public void testMultipleCursorConditions() throws Exception {
...
+    Query q = queryDataSnippets.createStartAtSnapshotQueryCursor();
+    List<String> results = getResults(q);
+    List<String> expectedResults = Arrays.asList("SF", "LA", "TOK", "BJ");
+    assertEquals(expectedResults, results);
+  }
+
+  @Test
+  public void testPaginateCursor() throws Exception {
+    // Snippet executes it's own query. Failures result in thrown Exceptions
+    queryDataSnippets.paginateCursor();
```
dzlier-gcp: This test doesn't verify that the API is returning what's expected, just that it doesn't through any exceptions.

This is my concern with methods that don't return anything, except this one doesn't even log anything to stdout either, so there's no way to catch any bugs or changes from one version to another.

Maybe that's not technically a goal or responsibility with our samples, but if a bug gets through the API and doesn't get caught in our samples, that's going to mean people taking our code and using it in their own projects, and it not working for them either.

kurtisvg: That's a fair concern, and one I agree with. The reason I went with this over a more involved test was because: 

1. I followed the [golang](https://github.com/GoogleCloudPlatform/golang-samples/blob/3eebbde2743566722299ac5747aa857008f0f9ff/firestore/firestore_snippets/query.go#L203-L220) example
 2. there are only 4 results in the first page, so the second page is technically empty.

Do you have a better way in mind to test it? I could check for an empty list - would that be preferred? I decided against it at the time, mostly because there are similar issues with some snippets in this folder and figured it would be best just to wait until they are modified to fit the new sample format.

dzlier-gcp: > figured it would be best just to wait until they are modified to fit the new sample format.

That might be the answer for now, depending on how soon all these samples are going to be updated. Are there any queries that are long enough to be paginated? Or can we just reduce the page element count to force the pagination even if it's not practical to get a page of 2 results?

In any case, I'd like it to at least write its findings to stdout, if not return them to be verified by the test.

firestore/src/main/java/com/example/firestore/snippets/QueryDataSnippets.java
```diff
@@ -328,4 +335,51 @@ void createMultipleCursorConditionsQuery() {
...
+    // [START fs_paginate_cursor]
+    // Construct query for first 25 cities, ordered by population.
+    CollectionReference cities = db.collection("cities");
+    ApiFuture<QuerySnapshot> firstPage = cities
+        .orderBy("population")
+        .limit(25)
+        .get();
+
+    // Wait for results.
+    List<QueryDocumentSnapshot> docs = firstPage.get(30, TimeUnit.SECONDS).getDocuments();
```
frankyn: This is performing another `get`? I'm a little confused by snippet given it calls `get`: one here and another on the previous line.

firestore/src/main/java/com/example/firestore/snippets/QueryDataSnippets.java
```diff
@@ -328,4 +335,51 @@ void createMultipleCursorConditionsQuery() {
...
+   * Create a query using a snapshot as a start point.
+   *
+   * @return query
+   */
+  Query createStartAtSnapshotQueryCursor()
+      throws InterruptedException, ExecutionException, TimeoutException {
+    // [START fs_document_snapshot_cursor]
+    // Fetch the Snapshot
+    ApiFuture<DocumentSnapshot> future = db.collection("cities").document("SF").get();
+    DocumentSnapshot snapshot = future.get(30, TimeUnit.SECONDS);
```
frankyn: What does the magic number represent other than 30 seconds?

frankyn: Could you use variable names instead of a magic number to state the same information?

firestore/src/main/java/com/example/firestore/snippets/QueryDataSnippets.java
frankyn: Thanks!

firestore/src/main/java/com/example/firestore/snippets/QueryDataSnippets.java
frankyn: Thanks, that helps!

firestore/src/test/java/com/example/firestore/snippets/QueryDataSnippetsIT.java
kurtisvg: I've updated the snippet slightly to return the queries and rerun for the tests.

Issue #1225, opened 2018-10-05
urlfetch sample for appengine-java8 does not work


ludoch:
Getting this error in the log:
2018-10-05 16:27:57.321 PDT
com.google.apphosting.runtime.jetty9.JettyLogger warn: / (JettyLogger.java:29)
java.io.FileNotFoundException: http://jsonplaceholder.typicode.com/posts/777
	at sun.reflect.NativeConstructorAccessorImpl.newInstance0(Native Method)
	at sun.reflect.NativeConstructorAccessorImpl.newInstance(NativeConstructorAccessorImpl.java:62)
	at sun.reflect.DelegatingConstructorAccessorImpl.newInstance(DelegatingConstructorAccessorImpl.java:45)
	at java.lang.reflect.Constructor.newInstance(Constructor.java:423)

lesv:
@averikitsch This used to work, you might wish to take a look.

PR #1249, opened 2018-11-04
Update UrlFetch sample

averikitsch:
Fixes Issue #1225

* Fixed validation for id
* Updated HTTP content type
* Add url stream handler to pom

### DIFF
--- a/appengine-java8/urlfetch/src/main/webapp/WEB-INF/appengine-web.xml
+++ b/appengine-java8/urlfetch/src/main/webapp/WEB-INF/appengine-web.xml
@@ -2,4 +2,5 @@
 <appengine-web-app xmlns="http://appengine.google.com/ns/1.0">
   <runtime>java8</runtime>
   <threadsafe>true</threadsafe>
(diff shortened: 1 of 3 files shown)


lesv:
I notice that Kokoro isn't happy

appengine-java8/urlfetch/src/main/webapp/WEB-INF/appengine-web.xml
```diff
@@ -2,4 +2,5 @@
 <appengine-web-app xmlns="http://appengine.google.com/ns/1.0">
   <runtime>java8</runtime>
   <threadsafe>true</threadsafe>
+  <url-stream-handler>urlfetch</url-stream-handler>
```
lesv: This is only needed if you wish to use our service to do it, why do you need this?

averikitsch: I read our documentation/error wrong...

lesv:
LGTM

PR #1257, opened 2018-11-09
Vision Product Search - GA

nirupa-kumar:
Vision Product Search beta - > GA

### DIFF
--- a/vision/product-search/cloud-client/src/main/java/com/example/vision/ImportProductSets.java
+++ b/vision/product-search/cloud-client/src/main/java/com/example/vision/ImportProductSets.java
@@ -18,14 +18,13 @@ package com.example.vision;
 
 // [START vision_product_search_tutorial_import]
 import com.google.api.gax.longrunning.OperationFuture;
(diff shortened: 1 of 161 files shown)


nirupa-kumar:
@nnegrey Please review

vision/product-search/cloud-client/src/main/java/com/example/vision/ImportProductSets.java
```diff
@@ -60,7 +57,7 @@ public static void importProductSets(String projectId, String computeRegion, Str
     ProductSearchClient client = ProductSearchClient.create();
```
nnegrey: I'm noticing this is leftover from the beta. 
Should we be using the autoclosing methods for the client.
```
try (ProductSearchClient client = ProductSearchClient.create()) {
   // All the code
   // No need to call client.close() now. 
}

nnegrey: Wasn't sure if ProductSearch is doing stuff a little differently or not. :)

PR #1259, opened 2018-11-15
Update SendGrid Sample due to library changes

averikitsch:


appengine-java8/sendgrid/pom.xml
```diff
@@ -1,5 +1,5 @@
 <!--
-Copyright 2015 Google Inc.
+Copyright 2018 Google Inc.
```
dzlier-gcp: If updating the copyright, please also change `Inc.` to `LLC` (in other files as well)

averikitsch: done.

averikitsch: Kokoro is set to look for Google Inc. not Google LLC.

dzlier-gcp:
Ah, I see that broke a checkstyle check - this should be fixed by updating shared-configuration to 1.0.10.

PR #1263, opened 2018-11-16
Included Java Sample code for OCR 

karan6190:
I have added Sample Spring code using Vision API for performing OCR.
This sample accepts the image and Upload that to google cloud storage and in response send the text written in that Image.


karan6190:
I signed it!

PR #1267, opened 2018-11-20
Update KMS samples to use the GAPIC client

bdhess:
Done in a series of commits to improve reviewability; happy to squash.

We should hold off on merging until we have similar changed ready to go for Golang and Python (and possible NodeJS as well).


bdhess:
FYI, @DanSanche

bdhess:
Closing in favor of #1288

PR #1269, opened 2018-11-21
Add code snippet to delete rows.

RobinRH:


spanner/cloud-client/src/main/java/com/example/spanner/SpannerSample.java
```diff
@@ -251,6 +251,30 @@ static void writeExampleData(DatabaseClient dbClient) {
...
+    for (Singer singer : SINGERS) {
+      mutations.add(
+          Mutation.delete("Singers", 
+              KeySet.singleKey(Key.newBuilder().append(singer.singerId).build())));
+    }
+
+    // Because Albums is created with ON DELETE CASCADE,
+    // you can delete all the rows in both tables
+    // by deleting the entire Singers table.
+    // mutations.add(Mutation.delete("Singers", KeySet.all()));
```
tswast: Intentionally commented out?

It looks like it'll be a no-op since both tables are already removed, but it it's valid to include, uncommenting it will ensure the sample syntax is correct and it compiles.

Alternatively, remove this whole comment block. It's a bit confusing to see two ways to do the same thing in the same sample.

RobinRH: I deleted it. It's another way of doing it and it was intentionally commented out. I can write about it in the topic text instead. It's the easier way to delete both tables, but the more difficult part of deletion is setting up the keys correctly. I wanted to show how to set up the keys.

PR #1270, opened 2018-11-21
[Asset] Add quickstart code for BatchGetAssetsHistory.

peter-zheng-g:


asset/cloud-client/src/test/java/com/example/asset/QuickStartIT.java
```diff
@@ -81,4 +81,14 @@ public void testExportAssetExample() throws Exception {
     String got = bout.toString();
     assertThat(got).contains(String.format("uri: \"%s\"", assetDumpPath));
   }
+
+  @Test
+  public void testBatchGetAssetsHistory() throws Exception {
+    String bucketAssetName = String.format("//storage.googleapis.com/%s", bucketName);
+    BatchGetAssetsHistoryExample.main(bucketAssetName);
+    String got = bout.toString();
+    if (!got.isEmpty()) {
```
dzlier-gcp: Is there any behavior we want to check if "got" _is_ empty?

peter-zheng-g: If "got" is empty, it means the bucket resource change has not been updated into our backend yet. It's rare case but it happens. I don't think there's much we can do for it. Fortunately, it means the API call was successfully, otherwise error response will be returned.

peter-zheng-g:
Thanks for the approval. Do we need to trigger the merge manually or will it merge automatically?

dzliergaard:
We do need to trigger it, I wasn't sure if you would want to do it, I will start it now.

peter-zheng-g:
Thanks for the approval and merge!

PR #1283, opened 2018-12-05
Update README.md

nnegrey:


nnegrey:
Agreed the samples need a bit of an update, but I think that might have to wait.

kurtisvg:
You can set this specific sample to build with a consistent version number by just adding `<version>1.0.0</version>` to the `pom.xml`. Currently it's just using the version from the parent, which happens to be 1.0.9.

nnegrey:
Oh it's in the pom.xml, but the README wasn't updated and it wasn't checked. 
https://github.com/GoogleCloudPlatform/java-docs-samples/blob/master/language/cloud-client/pom.xml

Issue #1291, opened 2018-12-07
Cloud Vision README points to docs  link with no Java snippets


leahecole:
## In which file did you encounter the issue?

https://github.com/GoogleCloudPlatform/java-docs-samples/blob/master/vision/README.md

### Did you change the file? If so, how?

No

## Describe the issue

The Quickstart Walkthrough links to this page https://cloud.google.com/vision/docs/face-tutorial, which does not have any Java code snippets in it

PR #1292, opened 2018-12-07
Add dependencies region tag to vision/face-detection/pom.xml

dzlier-gcp:
#1291 

Needed to submit doc change.


dzlier-gcp:
Yeah I'll do that.

PR #1297, opened 2018-12-12
Infinite Stream recognition

nirupa-kumar:
@nnegrey 
Please review

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -27,7 +27,7 @@
     <parent>
         <groupId>com.google.cloud.samples</groupId>
         <artifactId>shared-configuration</artifactId>
-        <version>1.0.9</version>
+        <version>1.0.10</version>
     </parent>
 
     <properties>
--- a/pom.xml
+++ b/pom.xml
@@ -27,7 +27,7 @@
     <parent>
         <groupId>com.google.cloud.samples</groupId>
         <artifactId>shared-configuration</artifactId>
-        <version>1.0.10</version>
+        <version>1.0.9</version>
     </parent>
 
     <properties>
--- a/speech/cloud-client/src/main/java/com/example/speech/InfiniteStreamRecognize.java
+++ b/speech/cloud-client/src/main/java/com/example/speech/InfiniteStreamRecognize.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2018 Google Inc.
+ * Copyright 2018 Google LLC
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
--- a/speech/cloud-client/src/main/java/com/example/speech/InfiniteStreamRecognize.java
+++ b/speech/cloud-client/src/main/java/com/example/speech/InfiniteStreamRecognize.java
@@ -1,0 +1,181 @@
+/*
+ * Copyright 2018 Google Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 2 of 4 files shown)


speech/cloud-client/src/main/java/com/example/speech/Recognize.java
```diff
@@ -55,6 +57,11 @@
 
 public class Recognize {
 
+  // Creating shared object
+  private static volatile BlockingQueue<byte[]> sharedQueue = new LinkedBlockingQueue();
+  private static TargetDataLine targetDataLine;
+  private static int BYTES_PER_BUFFER = 6400; // buffer size in bytes
```
nnegrey: Is it possible to move these to inside the function?

That way they are included as part of the region tags include statement. If not, we may need to create a separate class file, but no need to do arg parsing. You could follow the proposed new style here: https://github.com/GoogleCloudPlatform/java-docs-samples/pull/1232

nnegrey:
Is there a way to test this? 
Did we test the microphone sample?

speech/cloud-client/src/main/java/com/example/speech/Recognize.java
```diff
@@ -830,4 +839,136 @@ public static void transcribeModelSelectionGcs(String gcsUri) throws Exception {
...
+          }
+
+          clientStream.send(request);
+        }
+      } catch (Exception e) {
+        System.out.println(e);
+      }
+      responseObserver.onComplete();
+
+      clientStream.closeSend();
```
nnegrey: Lines 968-970
If we do a "Ctrl-C" to exit, these lines won't be called, right?
Is there a way to still exit and call them?

speech/cloud-client/src/main/java/com/example/speech/Recognize.java
nirupa-kumar: Switched to the new format in the latest commit. Also needed to move it to it's own class file since that looks like the best approach. Please let me know if you would like me to further look into keeping it in the Recognize file itself.

nirupa-kumar:
> Is there a way to test this?
> Did we test the microphone sample?

No we do not have tests for the microphone sample and the only way to test this currently is to do it manually.

speech/cloud-client/src/main/java/com/example/speech/Recognize.java
nirupa-kumar: Removed the `responseObserver.onComplete()` since I haven't really defined any particular action that needs to be performed. Checked the `clientStream` functionality and found out it gets closed by default when we hit `Ctrl-C` . Tested it to make sure. Yay! Please let me know if there maybe anything else that is necessary.

nirupa-kumar:
@nnegrey 
Thanks for reviewing.Fixed the comments. Please let me know if there is anything else that maybe required.

speech/cloud-client/src/main/java/com/example/speech/Recognize.java
```diff
@@ -46,6 +46,8 @@
 import java.util.ArrayList;
 import java.util.List;
 
+import java.util.concurrent.BlockingQueue;
+import java.util.concurrent.LinkedBlockingQueue;
```
nnegrey: Let's reset this file back to it's original format. (Since we pulled the sample out)

speech/cloud-client/src/main/java/com/example/speech/InfiniteStreamRecognize.java
```diff
@@ -0,0 +1,181 @@
+/*
+ * Copyright 2018 Google Inc.
```
nnegrey: Google LLC

nnegrey: If this throws a style issue. May need to update the pom.xml line 29 to `<version>1.0.11</version>`? (or maybe just 10)

pom.xml
```diff
@@ -27,7 +27,7 @@
     <parent>
         <groupId>com.google.cloud.samples</groupId>
         <artifactId>shared-configuration</artifactId>
-        <version>1.0.9</version>
+        <version>1.0.10</version>
```
nnegrey: Oh not this one. 

The speech/cloud-client/pom.xml
Similar to the beta file: https://github.com/GoogleCloudPlatform/java-docs-samples/blob/master/speech/beta/pom.xml#L26-L30

PR #1326, opened 2019-02-01
vision: address changes in vision annotations.

shollyman:
This updates the vision tests to 
* Use a new local copy of a logo for local testing of logo detection.
* Update test expectations for vision annotations that have changed
* Stops using non-public GCS resources associated to the build environment for gcs testing, 
  and instead references data in the public "cloud-samples-data" bucket.

This PR doesn't address the issue of synchronizing path patterns between the GCS and local resources.  Still considering further changes here to improve test maintenance, like storing test expectations centrally in the bucket in some portable form, and providing a larger set of test resources.


vision/cloud-client/src/test/java/com/example/vision/DetectIT.java
```diff
@@ -84,29 +85,29 @@ public void testFacesGcs() throws Exception {
...
   public void testLabelsGcs() throws Exception {
     // Act
-    String[] args = {"labels", "gs://" + BUCKET + "/vision/wakeupcat.jpg"};
+    String[] args = { "labels", "gs://" + ASSET_BUCKET + "/vision/label/wakeupcat.jpg" };
     Detect.argsHelper(args, out);
 
     // Assert
     String got = bout.toString();
-    assertThat(got).contains("whiskers");
+    assertThat(got).contains("Whiskers");
```
dzlier-gcp: If this (and others below) is going to be case sensitive, might we want to just lowercase the `got` and `expected` values for the comparison?

shollyman: Will relax further.

PR #1329, opened 2019-02-06
Revise artifactId to match dir change

llatif:


kurtisvg:
The parent pom for all of the samples is actually supposed to be this:

```
<!--
  -->
  <parent>
    <version>1.0.9</version>
  </parent>
```

This makes sure the tests get run correctly. 

Can you update to that instead?

kurtisvg:
Closing due to inactivity. Please feel free to repo if still needed.

PR #1330, opened 2019-02-07
Endpoints Auth Between Services

DanSanche:
Adding java samples for [endpoints auth between services
](https://cloud.google.com/endpoints/docs/openapi/service-account-authentication)

related to b/109951097 and cl/228405078


endpoints/getting-started/clients/pom.xml
```diff
@@ -0,0 +1,50 @@
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+  xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  
+  <modelVersion>4.0.0</modelVersion>
+  <groupId>com.example</groupId>
+  <artifactId>example</artifactId>
+  <version>1.0-SNAPSHOT</version>
+  <packaging>jar</packaging>
+
```
kurtisvg: Please add the following as a parent in all poms: 
```
<!--
  -->
  <parent>
    <version>1.0.9</version>
  </parent>
```

endpoints/getting-started/clients/src/main/java/com/example/app/JWTClient.java
```diff
@@ -0,0 +1,123 @@
...
+    final String audience, final int expiryLength)
+    throws FileNotFoundException, IOException {
+
+    Date now = new Date();
+    Date expTime = new Date(System.currentTimeMillis() + TimeUnit.SECONDS.toMillis(expiryLength));
+
+    // build jwt
+    JWTCreator.Builder token = JWT.create()
+        .withIssuedAt(now)
+        // expires after 'expirary_length' seconds.
```
kurtisvg: Please use capitalization and be consistent with punctuation (either use it for all comments or no comments) 

(same for the rest of the comments)

endpoints/getting-started/clients/src/main/java/com/example/app/JWTClient.java
```diff
@@ -0,0 +1,123 @@
...
+   * utilizes com.auth0.jwt.
+   */
+  public static String generateJWT(final String saKeyfile, final String saEmail,
+    final String audience, final int expiryLength)
+    throws FileNotFoundException, IOException {
+
+    Date now = new Date();
+    Date expTime = new Date(System.currentTimeMillis() + TimeUnit.SECONDS.toMillis(expiryLength));
+
+    // build jwt
```
kurtisvg: This comment is unnecessary - either expand or remove.

endpoints/getting-started/clients/src/main/java/com/example/app/JWTClient.java
```diff
@@ -0,0 +1,123 @@
...
+        // swagger spec (e.g. service account email). It can be any string
+        .withIssuer(saEmail)
+        // must be either your Endpoints service name, or match the value
+        // specified as the 'x-google-audience' in the OpenAPI document.
+        .withAudience(audience)
+        // subject and email should match the service account's email
+        .withSubject(saEmail)
+        .withClaim("email", saEmail);
+
+    // sign jwt
```
kurtisvg: This comment is unnecessary - either expand or remove.

endpoints/getting-started/clients/src/main/java/com/example/app/JWTClient.java
```diff
@@ -0,0 +1,123 @@
...
+    while ((line = buffReader.readLine()) != null) {
+       result.append(line);
+    }
+    buffReader.close();
+    return result.toString();
+  }
+// [END endpoints_jwt_request]
+
+  public static void main(final String[] args) throws Exception {
+    String keyPath = args[0];
```
kurtisvg: If you are going to add command line argument, please use a proper parser so that there is some reasonable error handling ([here](https://github.com/GoogleCloudPlatform/java-docs-samples/blob/65970b8cb8aef28e5356bf6b58ac694f16e530ba/dlp/src/main/java/com/example/dlp/Inspect.java#L614) is an example)

Additionally, there is no readme instructing the user with how to execute this docs. In general we have been moving away from CLI executed samples, there was a new format suggested here (but there hasn't been any DPE-wide consensus yet) : https://github.com/GoogleCloudPlatform/java-docs-samples/pull/1232

DanSanche:
@kurtisvg I addressed your comments, let me know if there's anything else

kurtisvg:
@DanSanche Just needs a test for the two snippets, then LGTM

DanSanche:
The issue with testing these is that it requires an endpoints backed to be deployed for it to work as intended. If you think it's necessary, I could test the jwt creation function in isolation, but it feels like that would be more a test of the jwt/cyptography libraries I'm using rather than the sample itself

kurtisvg:
No, that seems like a reasonable explanation.

PR #1332, opened 2019-02-08
Update Audio Profiles to beta

nnegrey:


texttospeech/cloud-client/src/main/java/com/example/texttospeech/SynthesizeText.java
```diff
@@ -134,17 +180,29 @@ public static void main(String... args) throws Exception {
...
             .defaultHelp(true)
-            .description("Synthesize a text or ssml.");
+            .description("Synthesize a text, text with audio effect profiles, or ssml.");
 
     MutuallyExclusiveGroup group = parser.addMutuallyExclusiveGroup().required(true);
-    group.addArgument("--text").help("The text file from which to synthesize speech.");
+    group
+        .addArgument("--text")
+        .help("The text file from which to synthesize speech.")
+        .nargs("+")
```
kurtisvg: This should probably be a different argument? `.addArgument("--effect-profile").required(false)` or something?

nnegrey: I think I'll just drop the CLI entirely like we did for Dialogflow

kurtisvg: 👍

texttospeech/cloud-client/README.md
```diff
@@ -46,30 +46,8 @@ Synthesize text to an output audio file. [Java Code](https://github.com/GoogleCl
 mvn exec:java -DQuickstart
```
nirupa-kumar: Just one nit. Should we be removing this one too ?
```
mvn exec:java -DQuickstart
```

nnegrey: I don't think so, cus the quick start is still in a main() method entirely.

But I did totally delete the pom.xml profile for it, so I should put that back in.
I think it's nice that a quickstart is still runnable, but @kurtisvg thoughts?

kurtisvg: Is there anything mention in the Cloud Docs with regards to it? If so, I would leave it in. If not, it probably doesn't matter.

nnegrey: It does not. 
I'll remove it.

nnegrey:
Friendly Ping @kurtisvg

PR #1342, opened 2019-02-19
BigQuery: add sample for bigquerystorage API.

shollyman:
Adds a sample for demonstrating how to invoke the BQ Storage API.

### DIFF
--- a/bigquery/bigquerystorage/pom.xml
+++ b/bigquery/bigquerystorage/pom.xml
@@ -1,5 +1,5 @@
 <!--
-  Copyright 2018 Google LLC
+  Copyright 2019 Google LLC
 
   Licensed under the Apache License, Version 2.0 (the "License");
   you may not use this file except in compliance with the License.
--- a/bigquery/bigquerystorage/pom.xml
+++ b/bigquery/bigquerystorage/pom.xml
@@ -26,7 +26,7 @@
   <parent>
     <groupId>com.google.cloud.samples</groupId>
     <artifactId>shared-configuration</artifactId>
-    <version>1.0.9</version>
+    <version>1.0.10</version>
   </parent>
 
   <properties>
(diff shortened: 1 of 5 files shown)


bigquery/bigquerystorage/pom.xml
```diff
@@ -0,0 +1,65 @@
+<!--
+  Copyright 2018 Google LLC
```
dzlier-gcp: 2019

shollyman: done

bigquery/bigquerystorage/src/main/java/com/example/bigquerystorage/StorageSample.java
```diff
@@ -0,0 +1,142 @@
...
+          .setFormat(DataFormat.AVRO)
+          .build();
+
+      ReadSession session = client.createReadSession(readSessionRequest);
+
+      SimpleRowReader reader = new SimpleRowReader(
+          new Schema.Parser().parse(session.getAvroSchema().getSchema()));
+
+      // Assert that there are streams available in the session.  If this fails, consider writing
+      // results of a query to a named table rather than consuming cached results directly.
```
tswast: Since this can also fail due to normal empty tables, we should probably mention those first and then talk about the known issue.

bigquery/bigquerystorage/src/test/java/com/example/bigquerystorage/QuickstartSampleIT.java
```diff
@@ -0,0 +1,57 @@
+/*
+ * Copyright 2019 Google Inc.
```
tswast: Nit: Google LLC

bigquery/bigquerystorage/src/main/java/com/example/bigquerystorage/StorageSample.java
shollyman: Updated comment to reflect the empty table case.

PR #1355, opened 2019-03-15
[Cloud Tasks] Add HTTP push queue sample

averikitsch:


averikitsch:
Ping @dzlier-gcp, do you mind reviewing today or tomorrow?

dzlier-gcp:
Taking a look now.

appengine-java8/tasks/src/main/java/com/example/task/CreateHttpTask.java
```diff
@@ -0,0 +1,182 @@
...
+
+import org.apache.commons.cli.CommandLine;
+import org.apache.commons.cli.CommandLineParser;
+import org.apache.commons.cli.DefaultParser;
+import org.apache.commons.cli.HelpFormatter;
+import org.apache.commons.cli.Option;
+import org.apache.commons.cli.Options;
+import org.apache.commons.cli.ParseException;
+
+public class CreateHttpTask {
```
dzlier-gcp: Was there a specific directive to make this into a CLI sample instead of the new sample format? If it's just to keep it consistent with the CreateTask sample that already exists here, I don't think that's enough of a reason to do the whole CLI thing instead of following the new template.

averikitsch: This is a quickstart, so It is nice to have the CLI to run it.

averikitsch: Is there an easy way to compile and run if the correct variables are added?

averikitsch: I removed the CLI.

appengine-java8/tasks/src/main/java/com/example/task/CreateHttpTask.java
```diff
@@ -0,0 +1,77 @@
...
+    String projectId = System.getenv("PROJECT_ID");
+    String queueName = System.getenv("QUEUE_ID");
+    String location = System.getenv("LOCATION_ID");
+    String url = System.getenv("URL");
+
+    // [START cloud_tasks_create_http_task]
+    // Instantiates a client.
+    try (CloudTasksClient client = CloudTasksClient.create()) {
+
+      // Variables provided by the CLI.
```
dzlier-gcp: by system variables

appengine-java8/tasks/src/main/java/com/example/task/CreateHttpTask.java
```diff
@@ -0,0 +1,77 @@
...
+
+public class CreateHttpTask {
+
+  public static void main(String[] args) throws Exception {
+    String projectId = System.getenv("PROJECT_ID");
+    String queueName = System.getenv("QUEUE_ID");
+    String location = System.getenv("LOCATION_ID");
+    String url = System.getenv("URL");
+
+    // [START cloud_tasks_create_http_task]
```
dzlier-gcp: Please put the START tag above import statements so that the dependency paths are included when the code is hosted on c.g.c. (and move the END tag past close brackets as necessary).

appengine-java8/tasks/README.md
```diff
@@ -89,3 +92,18 @@ mvn exec:java -Dexec.mainClass="com.example.task.CreateTask" \
     -Dexec.args="--project-id $GOOGLE_CLOUD_PROJECT \
     --queue $QUEUE_ID --location $LOCATION_ID --payload hello --in-seconds 30"
 ```
+
+### Using HTTP Push Queues
```
dzlier-gcp: So in line with not having the samples be a CLI, I think we're also not really including run instructions in the README for running the samples from the command line, as these code snippets are really for hosting on c.g.c documentation and not meant to be run directly by the users after cloning the whole repo.

appengine-java8/tasks/src/main/java/com/example/task/CreateHttpTask.java
```diff
@@ -0,0 +1,77 @@
...
+import org.apache.commons.cli.CommandLineParser;
+import org.apache.commons.cli.DefaultParser;
+import org.apache.commons.cli.HelpFormatter;
+import org.apache.commons.cli.Option;
+import org.apache.commons.cli.Options;
+import org.apache.commons.cli.ParseException;
+
+public class CreateHttpTask {
+
+  public static void main(String[] args) throws Exception {
```
dzlier-gcp: Following the new sample template, this wouldn't be a main method, but rather a named one like "createHttpTask", and instead of getting the System env var values below, it would include those 4 things as named parameters. Then there would be commented out declarations of them below, e.g.

```
public static void createHttpTask(String projectId, String queueName, string location, string url) {
    // String projectId = "YOUR-PROJECT-ID";
    // etc
```

Then in your test file (unless there's a reason not to include a test with this sample, I don't have full context), you would get those input parameters from required environment variables and supply them in the method call.

averikitsch: This sample will be a part of a quickstart as of right now; therefore I would like to keep the instructions in the README and the main method. I'm adding a bug for myself to update this complete sample after the launch to re-evaluate the CLI in CreateTask.java and add tests.

averikitsch:
In order to be consistent with other languages I moved the HTTP task sample to it's own directory.

appengine-java8/tasks/src/main/java/com/example/task/CreateHttpTask.java
averikitsch: done.

appengine-java8/tasks/src/main/java/com/example/task/CreateHttpTask.java
averikitsch: done.

appengine-java8/tasks/README.md
averikitsch: done.

averikitsch:
ping @dzlier-gcp (Sorry to bug you again, but trying to get this done by the EOW)

tasks/pom.xml
```diff
@@ -0,0 +1,64 @@
...
+  <artifactId>tasks-samples</artifactId>
+
+  <!--
+    The parent pom defines common style checks and testing strategies for our samples.
+    Removing or replacing it should not affect the execution of the samples in anyway.
+  -->
+  <parent>
+    <groupId>com.google.cloud.samples</groupId>
+    <artifactId>shared-configuration</artifactId>
+    <version>1.0.10</version>
```
kurtisvg: Don't forget to update the version of this.

tasks/src/main/java/com/example/task/CreateHttpTask.java
```diff
@@ -0,0 +1,69 @@
...
+
+  public static void main(String[] args) throws Exception {
+    String projectId = System.getenv("PROJECT_ID");
+    String queueName = System.getenv("QUEUE_ID");
+    String location = System.getenv("LOCATION_ID");
+    String url = System.getenv("URL");
+
+    // Instantiates a client.
+    try (CloudTasksClient client = CloudTasksClient.create()) {
+
```
kurtisvg: nix Newline

appengine-java8/tasks/README.md
```diff
@@ -69,12 +69,15 @@ location is "us-central1").
 export LOCATION_ID=<YOUR_ZONE>
 ```
 
+### Using App Engine Queues
 Create a task, targeted at the `/tasks/create` endpoint, with a payload specified:
 
 ```
 mvn exec:java -Dexec.mainClass="com.example.task.CreateTask" \
```
kurtisvg: Are these instructions still correct? It looks like you are using environment variables now

tasks/pom.xml
averikitsch: done.

tasks/src/main/java/com/example/task/CreateHttpTask.java
averikitsch: done.

appengine-java8/tasks/README.md
averikitsch: This is correct. This readme is associated with the App Engine sample.

PR #1363, opened 2019-04-02
Update Kokoro Build Config

averikitsch:
Create separate jobs for Java 8 and 11 samples.


.kokoro/tests/diff_tests.sh
```diff
@@ -42,7 +42,14 @@ echo -e "\n******************** TESTING AFFECTED PROJECTS ********************"
 set +e
 # Diff to find out what has changed from master
 RESULT=0
-cd github/java-docs-samples
+
+# Select directory to run tests
+if [[ $PROJECT == appengine-java11 ]]; then
```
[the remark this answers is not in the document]
kurtisvg: Also one small nit on the folder name: The other samples repo (or at least Python and Golang) are using `appengine/$VERSION` to separate samples out - might be worthwhile to remain consistent.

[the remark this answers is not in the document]
averikitsch: Do you have a link to a file that uses `$VERSION`?

averikitsch:
1. I simplified the Java version check (greater or equal to version not implemented) . Extracting the Java version from the pom.xml tags seems variable and I need to test more. 
2. The run_tests.sh runs the tests individually similarly to diff_tests.sh. Let me know if the print out should be changed.

.kokoro/tests/run_tests.sh
```diff
@@ -38,11 +38,44 @@ gcloud auth activate-service-account\
...
-    -Dbigtable.projectID="${GOOGLE_CLOUD_PROJECT}" \
-    -Dbigtable.instanceID=instance
+# For every pom.xml (may break on whitespace)
+for file in **/pom.xml; do
+    # Navigate to project
+    file=$(dirname "$file")
+    pushd "$file" > /dev/null
+
+    # Check if Java version matches test version
+    VERSION=$(grep "source>$JAVA_VERSION" pom.xml -c)
```
kurtisvg: So if each POM we need the following (we can update the java11 samples to use this format over the configuration in the plugin section):

```xml
    <properties>
        <maven.compiler.target>1.11</maven.compiler.target>
        <maven.compiler.source>1.11</maven.compiler.source>
    </properties>
```

Then we can use the following: `VERSION=$(grep -oP '(?<=<maven.compiler.target>).*?(?=</maven.compiler.target>)' pom.xml)`

And then it looks like we can [check against a comma separated list](https://stackoverflow.com/questions/29300806/a-bash-script-to-check-if-a-string-is-present-in-a-comma-separated-list-of-strin) like so: 
```bash
JAVA_VERSIONS="1.8,1.11"
VERSION="1.11"
if [[ ",$JAVA_VERSIONS," = *",$VERSION,"* ]]; then
  // run tests
fi
```

.kokoro/common.cfg
```diff
@@ -17,21 +17,14 @@
 # Download secrets from Cloud Storage.
 gfile_resources: "/bigstore/cloud-devrel-kokoro-resources/java-docs-samples"
 
-# Download trampoline resources. 
+# Download trampoline resources.
 gfile_resources: "/bigstore/cloud-devrel-kokoro-resources/trampoline"
 
 # Use the trampoline script to run in docker.
 build_file: "java-docs-samples/.kokoro/trampoline.sh"
```
tswast: From the looks of the test failure, this might have to move to the `common.cfg` in the sub-directories? I wonder if the Build configs don't inherit the same way that the job configs do?

tswast: I see a note in the Kokoro docs for Build Config "For build config, common.cfg must be put under the same directory as your build config."

PR #1364, opened 2019-04-03
[CloudAsset] Update client library version to use endpoint v1.

peter-zheng-g:


peter-zheng-g:
@dzlier-gcp  All checks have passed. Please help merge. Thanks!

PR #1371, opened 2019-04-04
OpenCensus: Update EXPORT interval

mayurkale22:
OpenCensus Stackdriver exporter exports every 60s by default, we need make sure main thread is running for longer than that.


mayurkale22:
@kurtisvg Please review.

PR #1372, opened 2019-04-04
Add Healthcare samples for Dataset operations.

dzlier-gcp:


healthcare/v1beta1/src/main/java/snippets/healthcare/fhir/resources/FhirResourceConditionalPatch.java
```diff
@@ -17,58 +17,83 @@
...
-  public static void fhirResourceConditionalPatch(String resourceName) throws IOException {
+  public static void fhirResourceConditionalPatch(
+      String fhirStoreName,
+      String resourceType,
+      String data) throws IOException, URISyntaxException {
     // String resourceName =
     //    String.format(
     //        FHIR_NAME, "project-id", "region-id", "dataset-id", "store-id", "fhir-id");
-    String resourceType = "Patient";
+    // String resourceType = "Patient";
```
kurtisvg: Can we leave resourceType and data inside the function? For the [canonical](https://github.com/GoogleCloudPlatform/java-docs-samples/blob/master/dlp/src/main/java/dlp/snippets/InspectImageFile.java#L52), the image type was deemed unnecessary to pass in through the test, so I moved it inline

Also for resourceType, is there a specific set of allowed strings? Or is it entirely up to the user? Might be nice to have a link explaining what the field could be

healthcare/v1beta1/src/main/java/snippets/healthcare/fhir/resources/FhirResourceConditionalUpdate.java
```diff
@@ -17,57 +17,80 @@
...
+    // String data = "{'family': 'Smith'}";
 
     // Initialize the client, which will be used to interact with the service.
     CloudHealthcare client = createClient();
 
-    // Fetch the initial state of the resource.
-    Fhir.Get getRequest =
-        client.projects().locations().datasets().fhirStores().fhir().get(resourceName);
-    HttpBody resource = getRequest.execute();
+    HttpClient httpClient = HttpClients.createDefault();
```
kurtisvg: I don't know if there is a bug tracking this issue, but maybe it should be linked to in the comment? So future readers understand why these samples aren't using the client?

dzlier-gcp:
Just updated with a commit that fixes all healthcare tests and checkstyle issues.

healthcare/v1beta1/src/main/java/snippets/healthcare/fhir/resources/FhirResourceConditionalUpdate.java
dzlier-gcp: There isn't currently a bug tracking it but I can create one. I doubt they're going to prioritize it though.

kurtisvg: Probably a good idea, just to make sure it's somewhere on their radar.

dzlier-gcp: Created bug 133419634

PR #1373, opened 2019-04-04
Update Quickstart.java

mtabasko:
Increase "EXPORT_INTERVAL" from 60 seconds to 70 seconds, to work around https://github.com/census-instrumentation/opencensus-java/issues/1808. (Otherwise, metrics do not actually get exported to Stackdriver.)


kurtisvg:
Looks like this change was also committed elsewhere - closing.

PR #1383, opened 2019-04-12
Add Oauth2 sample for Java 11

averikitsch:


appengine-java11/oauth2/pom.xml
```diff
@@ -0,0 +1,108 @@
+<?xml version="1.0" encoding="UTF-8"?><project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+
+  <modelVersion>4.0.0</modelVersion>
+  <packaging>war</packaging>
+  <version>1.0-SNAPSHOT</version>
+  <groupId>com.example.appengine</groupId>
+  <artifactId>oauth2-sample</artifactId>
+
+  <!-- <parent>
```
kurtisvg: Reminder to uncomment this before merging.

appengine-java11/oauth2/pom.xml
```diff
@@ -0,0 +1,108 @@
...
+  </parent> -->
+
+  <properties>
+    <maven.compiler.source>11</maven.compiler.source>
+    <maven.compiler.target>11</maven.compiler.target>
+  </properties>
+
+  <dependencies>
+    <!-- Compile/runtime dependencies -->
+    <dependency>
```
kurtisvg: A comment around where this can be found / why it's needed would be helpful.

appengine-java11/oauth2/pom.xml
```diff
@@ -0,0 +1,108 @@
...
+      <version>1</version>
+      <scope>provided</scope>
+    </dependency>
+
+    <!-- OAuth 2.0 packages -->
+    <dependency>
+      <groupId>com.google.oauth-client</groupId>
+      <artifactId>google-oauth-client</artifactId>
+      <version>1.28.0</version>
+      <scope>provided</scope>
```
kurtisvg: Do these need to be scoped to provided?

appengine-java11/oauth2/src/main/java/com/example/appengine/OAuth2CallbackServlet.java
```diff
@@ -0,0 +1,60 @@
...
+import javax.servlet.annotation.WebServlet;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+
+@SuppressWarnings("serial")
+@WebServlet(urlPatterns = "/oauth2callback/*")
+public class Oauth2CallbackServlet extends AbstractAuthorizationCodeCallbackServlet {
+
+  @Override
+  protected void onSuccess(HttpServletRequest req, HttpServletResponse resp, Credential credential)
```
kurtisvg: Some comments on what these functions do and how they get called would be helpful - chances are most users looking at this sample won't be familiar with this interface.

appengine-java11/oauth2/src/main/webapp/WEB-INF/web.xml
```diff
@@ -0,0 +1,27 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+ Copyright 2015 Google Inc.
```
kurtisvg: `2019 Google LLC`

appengine-java11/pom.xml
```diff
@@ -0,0 +1,49 @@
+<!--
```
kurtisvg: This pom can be removed.

appengine-java11/oauth2/pom.xml
averikitsch: I believe so for the runtime.

appengine-java11/pom.xml
averikitsch: Don't we need this for the tests to run properly ie parent v child pom?

appengine-java11/oauth2/src/main/webapp/WEB-INF/web.xml
averikitsch: done.

appengine-java11/oauth2/src/main/java/com/example/appengine/OAuth2CallbackServlet.java
averikitsch: done.

appengine-java11/oauth2/pom.xml
averikitsch: done.

appengine-java11/oauth2/pom.xml
averikitsch: done.

appengine-java11/pom.xml
kurtisvg: No - the update you made to Kokoro finds projects by using the pom's themselves instead of relying on the parent POM. 

(If we were still doing it with the parent pom this tests wouldn't trigger anyway because it doesn't list the sub projects as modules and isn't listed in the pom located at root)

appengine-java11/oauth2/pom.xml
kurtisvg: You may want to verify that these are provided by the runtime - Seems weird to me that they would be provided (I would think that would be a potential problem with versioning)

appengine-java11/oauth2/pom.xml
kurtisvg: Same as comment on the other one - please add just a little more context so there is no mistaking where to find that folder.

kurtisvg:
@averikitsch tests on this sample didn't actually run (see the java11 logs)

appengine-java11/oauth2/pom.xml
averikitsch: The user needs to copy the dependencies into ${project.build.directory}/appengine-staging; that folder provides the dependencies to the runtime. This is what I understand from Ludo, but correct me if this is off.

appengine-java11/oauth2/src/main/appengine/app.yaml
```diff
@@ -0,0 +1,7 @@
+runtime: java11
```
kurtisvg: nit: Should probably have a license?

PR #1386, opened 2019-04-18
Testing Kokoro tests

averikitsch:


kurtisvg:
It might be worth adding some logic that when `$VERSION == ""` the overall test fails.

PR #1387, opened 2019-04-18
[Kokoro] Update version check

averikitsch:
Not all of the pom.xml are structured in the same way; therefore creating a bug to make the version consistent and adding a `version == ""` check.

### DIFF
--- a/.kokoro/tests/run_tests.sh
+++ b/.kokoro/tests/run_tests.sh
@@ -55,7 +55,7 @@ for file in **/pom.xml; do
     VERSION=$(grep -oP '(?<=<maven.compiler.target>).*?(?=</maven.compiler.target>)' pom.xml)
 
     # Check for changes to the current folder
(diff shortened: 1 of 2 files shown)


.kokoro/tests/run_tests.sh
```diff
@@ -55,7 +55,7 @@ for file in **/pom.xml; do
     VERSION=$(grep -oP '(?<=<maven.compiler.target>).*?(?=</maven.compiler.target>)' pom.xml)
 
     # Check for changes to the current folder
-    if [ "$PARENT" -eq 0 ] && [ ",$JAVA_VERSIONS," = *",$VERSION,"* ]; then
+    if [ "$PARENT" -eq 0 ] && [[ $JAVA_VERSION = *"$VERSION"* ]]; then
```
kurtisvg: It's usually a good idea to escape variables w/ quotes in bash. 
```suggestion
    if [ "$PARENT" -eq 0 ] && [[ "$JAVA_VERSION" = *"$VERSION"* ]]; then
```

PR #1388, opened 2019-04-19
Misc testing fixes. 

kurtisvg:
* Improve readability of run_tests.sh
* Removed diff_tests since no longer used
* Removed parent poms since no longer used. 
* Update shared config version of everything to 1.0.11
* Update several plugins causing conflicts


kurtisvg:
Forcing this merge - there are some existing test problems and some new plugin conflicts, but we will deal with them at a future date.

PR #1389, opened 2019-04-19
[Cloud Tasks] Add task sample with auth token

averikitsch:


tasks/src/main/java/com/example/task/CreateHttpTaskWithToken.java
```diff
@@ -0,0 +1,70 @@
+/*
+ * Copyright 2018 Google LLC
```
kurtisvg:
```suggestion
 * Copyright 2019 Google LLC
```

PR #1396, opened 2019-04-24
Container analysis improvements

DanSanche:
In preparation for a launch, I made some fixes to the existing samples. New samples coming soon, but I thought it might be easier to review these ones separately

changes:
- updated region tag to add project namespace, as required by samples tracker
- create client as part of the sample instead of instead of accepting it as a parameter
    - also demonstrate graceful shutdown of client
- removed updateNote and updateOccurrence. (These samples weren't used in the docs, so we're hoping to limit the sample surface area for maintenance reasons)
- use occurrence name/project id instead of expecting users to form an occurrenceName string themselves


container-registry/container-analysis/src/main/java/com/example/containeranalysis/Samples.java
```diff
@@ -46,47 +48,56 @@
...
     // Details about the your vulnerability can be added here
     // Example: vulBuilder.setSeverity(Severity.CRITICAL);
     noteBuilder.setVulnerability(vulBuilder);
     Note newNote = noteBuilder.build();
 
-    final String projectName = ProjectName.format(projectId);
-    return client.createNote(projectName, noteId, newNote);
+    GrafeasV1Beta1Client client = GrafeasV1Beta1Client.create();
+    Note result = client.createNote(projectName, noteId, newNote);
+    client.shutdownNow();
```
lesv: Why are we doing this here?  Is this how all users would use the createNote api?  Or is this just a convenience for us in this moment? (if the just for this sample, then maybe we should put it somewhere else or at least put a comment about it here)

DanSanche:
> Not sure why you've moved the shutdown code into each example. That seems the opposite of what we should be teaching, at least w/o comments.

I thought if I was moving client creation into the sample, it would make sense to show how to shut it down when you're finished as well. But I'm happy to remove them if you think it's unnecessary

lesv:
Thanks -- It's more that we should be thinking about how we want users to use things and if we deviate from that for some reason, we should call that out.  (Do you create a client for the duration of a request? for each API request? Cache them?  Are they thread safe?  etc.

DanSanche:
Sounds good, I removed the shutdown lines. Do you think we should still show the creation of the client in each sample? Even that seems like an anti-pattern if taken literally (creating a new client for each request) but we've received feedback that users were confused where the client was coming from in the old samples

lesv:
@DanSanche I don't, but I'm going to loop in @kurtisvg, @dzlier-gcp as somewhere in the back of my mind I think there was a discussion on this.  If you keep it in, then you should comment about it.

kurtisvg:
Going forward, the new sample format requires that the samples should be copy/pastable. The "cannonical" samples for this format are located [here](https://github.com/GoogleCloudPlatform/java-docs-samples/tree/master/dlp/src/main/java/dlp/snippets). 

For client creation this means we show the user how to create the client in the sample, but also comment on how to properly use the client (see [here](https://github.com/GoogleCloudPlatform/java-docs-samples/blob/master/dlp/src/main/java/dlp/snippets/InspectString.java#L41-L44) for an example.)

DanSanche:
@kurtisvg is including imports in each sample part of the official new format? I wasn't sure if that would be feasible for all languages, so I added a region tag for imports. Is that something I should change now, or is it still up in the air?

kurtisvg:
@DanSanche Yes, those formats have been finalized (including imports). It hasn't been decided when we will start requiring new samples in that format, so it's up to you if you want to follow it for now.

DanSanche:
is there a link I can read to catch up on all the changes to the format?

kurtisvg:
Currently not for Java, but there will be soon-ish (1-2 weeks). There is one for Go, which _mostly_ follows similar guidelines [here](https://github.com/GoogleCloudPlatform/golang-samples/blob/master/CONTRIBUTING.md#style).

DanSanche:
I think I'm going to go ahead and follow the new style guide here to make future maintenance easier. 

Since this is looking like a larger scope than the original PR, I'll close this one and eventually open a new one that includes these changes and the new samples

PR #1402, opened 2019-05-02
Container analysis update

DanSanche:
Updated container analysis samples:
- each sample is in its own file, and includes imports
- added new samples:
    - pollDiscoveryOccurrenceFinished
    - findVulnerabilityOccurrencesForImage
    - findHighSeverityVulnerabilitiesForImage
- updated comment style to be in line with new [canonical format](https://github.com/GoogleCloudPlatform/java-docs-samples/blob/master/dlp/src/main/java/dlp/snippets/InspectImageFile.java)
- samples now client creation as part of sample, rather than accepting it as a parameter
- updated region tags to be compliant with sample tracker


container-registry/container-analysis/src/main/java/com/example/containeranalysis/CreateNote.java
```diff
@@ -0,0 +1,51 @@
+/*
+ * Copyright 2018 Google LLC
```
lesv: 2019 for a new file

container-registry/container-analysis/src/main/java/com/example/containeranalysis/CreateNote.java
```diff
@@ -0,0 +1,51 @@
...
+import java.io.IOException;
+import java.lang.InterruptedException;
+
+
+public class CreateNote {
+
+  // Creates and returns a new Note
+  public static Note createNote(String noteId, String projectId)
+      throws IOException, InterruptedException {
+    // String noteId = "my-note";
```
lesv: Why are these still here?

container-registry/container-analysis/src/main/java/com/example/containeranalysis/CreateNote.java
```diff
@@ -0,0 +1,51 @@
...
+    // String projectId = "my-project-id";
+    final String projectName = ProjectName.format(projectId);
+
+    Note.Builder noteBuilder = Note.newBuilder();
+    Vulnerability.Builder vulBuilder = Vulnerability.newBuilder();
+    // Details about the your vulnerability can be added here
+    // Example: vulBuilder.setSeverity(Severity.CRITICAL);
```
lesv: These don't seem to be right somehow.

container-registry/container-analysis/src/main/java/com/example/containeranalysis/CreateOccurrence.java
```diff
@@ -0,0 +1,58 @@
+/*
+ * Copyright 2018 Google LLC
```
lesv: 2019

container-registry/container-analysis/src/main/java/com/example/containeranalysis/CreateOccurrence.java
```diff
@@ -0,0 +1,58 @@
...
+
+public class CreateOccurrence {
+  // Creates and returns a new Occurrence associated with an existing Note
+  public static Occurrence createOccurrence(String resourceUrl, String noteId, 
+      String occProjectId, String noteProjectId) throws IOException, InterruptedException {
+    // String resourceUrl = "https://gcr.io/project/image@sha256:foo";
+    // String noteId = "my-note";
+    // String occProjectId = "my-project-id";
+    // String noteProjectId = "my-project-id";
+    final NoteName noteName = NoteName.of(noteProjectId, noteId);
```
lesv: If that's the new sample style fine, but it seems confusing to have both of them.

container-registry/container-analysis/src/main/java/com/example/containeranalysis/CreateOccurrence.java
```diff
@@ -0,0 +1,58 @@
...
+    final String occProjectName = ProjectName.format(occProjectId);
+
+    Occurrence.Builder occBuilder = Occurrence.newBuilder();
+    occBuilder.setNoteName(noteName.toString());
+    Details.Builder detailsBuilder = Details.newBuilder();
+    // Details about the vulnerability instance can be added here
```
lesv: Perhaps a link to the docs?

container-registry/container-analysis/src/main/java/com/example/containeranalysis/CreateOccurrence.java
```diff
@@ -0,0 +1,58 @@
...
+    occBuilder.setVulnerability(detailsBuilder);
+    // Attach the occurrence to the associated image uri
+    Resource.Builder resourceBuilder = Resource.newBuilder();
+    resourceBuilder.setUri(resourceUrl);
+    occBuilder.setResource(resourceBuilder);
+    Occurrence newOcc = occBuilder.build();
+
+    // Initialize client that will be used to send requests. After completing all of your requests,
```
lesv: Is this thread safe?  Should I have one per instance? Or 1 / concurrent session?

container-registry/container-analysis/src/main/java/com/example/containeranalysis/DeleteNote.java
```diff
@@ -0,0 +1,39 @@
+/*
+ * Copyright 2018 Google LLC
```
lesv: 2019

container-registry/container-analysis/src/main/java/com/example/containeranalysis/CreateNote.java
DanSanche: what doesn't seem right? The code in the comment, or the existence of the comment itself?

container-registry/container-analysis/src/main/java/com/example/containeranalysis/CreateOccurrence.java
DanSanche: by both of them, do you mean occProjectId and noteProjectId?

container-registry/container-analysis/src/main/java/com/example/containeranalysis/CreateOccurrence.java
DanSanche: do you think that information is necessary here? This comment is replicated in each sample, so I'm trying to limit the amount of copy/pasted documentation. Maybe we could replace this  with a [link to the docs?](https://googleapis.github.io/google-cloud-java/google-cloud-clients/apidocs/index.html?com/google/cloud/devtools/containeranalysis/v1beta1/package-summary.html)

container-registry/container-analysis/src/main/java/com/example/containeranalysis/CreateNote.java
lesv: Look at the line below it, it doesn't seem to match.

container-registry/container-analysis/src/main/java/com/example/containeranalysis/CreateOccurrence.java
lesv: Passing in the values in the function and the commented lines as well.

container-registry/container-analysis/src/main/java/com/example/containeranalysis/CreateOccurrence.java
lesv: SGTM

lesv: The docs probably should be expanded on a bit.

container-registry/container-analysis/src/main/java/com/example/containeranalysis/PollDiscoveryOccurrenceFinished.java
```diff
@@ -0,0 +1,78 @@
...
+    // call the "close" method on the client to safely clean up any remaining background resources.
+    GrafeasV1Beta1Client client = GrafeasV1Beta1Client.create();
+
+    // find the discovery occurrence using a filter string
+    Occurrence discoveryOccurrence = null;
+    String filterStr = "kind=\"DISCOVERY\" AND resourceUrl=\"" + resourceUrl + "\"";
+    while (discoveryOccurrence == null) {
+      for (Occurrence o : client.listOccurrences(projectName, filterStr).iterateAll()) {
+        if (o.getDiscovered() != null) {
+          // there should be only one valid discovery occurrence returned by the given filter
```
lesv: why?

container-registry/container-analysis/src/main/java/com/example/containeranalysis/OccurrencesForNote.java
```diff
@@ -0,0 +1,54 @@
...
+public class OccurrencesForNote {  
+  // Retrieves all the Occurrences associated with a specified Note
+  // Here, all Occurrences are printed and counted
+  public static int getOccurrencesForNote(String noteId, String projectId) 
+      throws IOException, InterruptedException {
+    // String noteId = "my-note";
+    // String projectId = "my-project-id";
+    final NoteName noteName = NoteName.of(projectId, noteId);
+
+    ListNoteOccurrencesRequest request = ListNoteOccurrencesRequest.newBuilder()
```
lesv: should the indentation be different?  ie .newBuilder() indented 4 on it's own line?

container-registry/container-analysis/src/main/java/com/example/containeranalysis/HighVulnerabilitiesForImage.java
```diff
@@ -0,0 +1,50 @@
...
+import java.util.List;
+
+public class HighVulnerabilitiesForImage {
+  // Retrieve a list of vulnerability occurrences with a severity level of 'HIGH' or greater
+  public static List<Occurrence> findHighSeverityVulnerabilitiesForImage(String resourceUrl,
+      String projectId) throws IOException {
+    // String resourceUrl = "https://gcr.io/project/image@sha256:foo";
+    // String projectId = "my-project-id";
+    final String projectName = ProjectName.format(projectId);
+    String filterStr = "kind=\"VULNERABILITY\" AND resourceUrl=\"" + resourceUrl + "\"";
```
lesv: Would String.format be useful here?  (No is an ok answer)

container-registry/container-analysis/src/main/java/com/example/containeranalysis/OccurrencesForImage.java
```diff
@@ -0,0 +1,48 @@
...
+
+public class OccurrencesForImage {
+  // Retrieves all the Occurrences associated with a specified image
+  // Here, all Occurrences are simply printed and counted
+  public static int getOccurrencesForImage(String resourceUrl, String projectId)
+      throws IOException, InterruptedException {
+    // String resourceUrl = "https://gcr.io/project/image@sha256:foo";
+    // String projectId = "my-project-id";
+    final String projectName = ProjectName.format(projectId);
+    final String filterStr = "resourceUrl=\"" + resourceUrl + "\"";
```
lesv: String.format?

container-registry/container-analysis/src/main/java/com/example/containeranalysis/PubSub.java
```diff
@@ -0,0 +1,90 @@
...
+      throws IOException, StatusRuntimeException, InterruptedException {
+    // This topic id will automatically receive messages when Occurrences are added or modified
+    String topicId = "container-analysis-occurrences-v1beta1";
+    SubscriptionAdminClient client = SubscriptionAdminClient.create();
+    PushConfig config = PushConfig.getDefaultInstance();
+    ProjectTopicName topicName = ProjectTopicName.of(projectId, topicId);
+    ProjectSubscriptionName subName = ProjectSubscriptionName.of(projectId, subId);
+    Subscription sub = client.createSubscription(subName, topicName, config, 0);
+    return sub;
+  }
```
lesv: Should there be any vertical whitespace in this method?  (it looks dense)

container-registry/container-analysis/src/main/java/com/example/containeranalysis/VulnerabilityOccurrencesForImage.java
```diff
@@ -0,0 +1,46 @@
...
+import java.util.List;
+
+public class VulnerabilityOccurrencesForImage {
+  // Retrieve a list of vulnerability occurrences assoviated with a resource
+  public static List<Occurrence> findVulnerabilityOccurrencesForImage(String resourceUrl, 
+      String projectId) throws IOException {
+    // String resourceUrl = "https://gcr.io/project/image@sha256:foo";
+    // String projectId = "my-project-id";
+    final String projectName = ProjectName.format(projectId);
+    String filterStr = "kind=\"VULNERABILITY\" AND resourceUrl=\"" + resourceUrl + "\"";
```
lesv: String.format ?

container-registry/container-analysis/src/test/java/com/example/containeranalysis/SamplesTest.java
```diff
@@ -17,22 +17,31 @@
...
 import static junit.framework.TestCase.assertEquals;
 
 import com.google.api.gax.rpc.NotFoundException;
 import com.google.cloud.devtools.containeranalysis.v1beta1.GrafeasV1Beta1Client;
 import com.google.cloud.pubsub.v1.Subscriber;
 import com.google.cloud.pubsub.v1.SubscriptionAdminClient;
+import com.google.containeranalysis.v1beta1.NoteName;
+import com.google.containeranalysis.v1beta1.ProjectName;
 import com.google.pubsub.v1.ProjectSubscriptionName;
 import io.grafeas.v1beta1.Note;
```
lesv: doesn't the Java style guide suggest some white space between things here?

container-registry/container-analysis/src/test/java/com/example/containeranalysis/SamplesTest.java
```diff
@@ -245,10 +226,133 @@ public void testPubSub() throws Exception {
...
+        imageUrl, PROJECT_ID, 5);
+    AnalysisStatus foundStatus = found.getDiscovered().getDiscovered().getAnalysisStatus();
+    assertEquals(foundStatus, AnalysisStatus.FINISHED_SUCCESS);
+
+    // clean up
+    String[] nameArr = found.getName().split("/");
+    String occId = nameArr[nameArr.length - 1];
+    DeleteOccurrence.deleteOccurrence(occId, PROJECT_ID);
+    DeleteNote.deleteNote(discNoteId, PROJECT_ID);
+  }
```
lesv: Wow this is dense, should there be more white space?  comments?

container-registry/container-analysis/src/main/java/com/example/containeranalysis/CreateNote.java
DanSanche: The idea that I was trying to put across is that you can set instance-specific vulnerability details metadata here, but this sample keeps the "details" field empty to be more generic.

I'm going to try to improve this to be more clear

container-registry/container-analysis/src/main/java/com/example/containeranalysis/CreateOccurrence.java
DanSanche: Yeah it occurred to me that it might be better at this point to have noteName and occProjectName as the parameters to the function. But I think I'd prefer to keep it as is to be consistent with other languages, and avoid a massive refactor. 

I'd be open to removing the commented lines, but I want to make sure I'm following the new sample style as well

container-registry/container-analysis/src/main/java/com/example/containeranalysis/CreateNote.java
DanSanche: Ok, I updated the function to add some sensible defaults. I have slightly mixed feeling about this, because I'm making the function less generic by adding these implementation details, but I'm also making the sample more useful by showing real-world usage. Let me know if you agree it's a good tradeoff

container-registry/container-analysis/src/main/java/com/example/containeranalysis/CreateNote.java
DanSanche: I believe this is part of the new recommended java format

container-registry/container-analysis/src/main/java/com/example/containeranalysis/CreateOccurrence.java
DanSanche: hmm the url for the client is a bit long to include in a code comment. Do you have any advice for how best to include it?

https://googleapis.github.io/googlecloudjava/googlecloudclients/apidocs/com/google/cloud/devtools/containeranalysis/v1beta1/package-summary.html

lesv: Firebase dynamic links?  https://firebase.google.com/products/dynamic-links

kurtisvg:
> @kurtisvg Since this is in the new format, I'd like you to weigh in. I'm not happy seeing lines like `\\ String projectId = "my-project-id";` I'd prefer to see `// projectId = "my-project-id"` (No String or ;) as it's a comment, not a line that should be commented out. Ideally we'll have standard language we use to explain we are showing "Example values:" (which I consider a huge win)

The idea is to make it as easy as possible to get users up and running. If they copy/paste the snippet, they can either remove the args and uncomment the snippet, or if they are calling it from another file they can copy the comments and use them before calling the method. 

It's not a perfect solution for a variety of reasons, but since Java doesn't do default values it was the best compromise we could find. The only information we pass in as an argument should be whats _absolutely_ required for testing (so probably values related to project info) to minimize this as much as possible.

container-registry/container-analysis/src/main/java/com/example/containeranalysis/CreateNote.java
lesv: SGTM

PR #1405, opened 2019-05-06
[WIP] Java 11 Region Tags

averikitsch:


appengine-java11/guestbook-cloud-firestore/src/main/java/com/example/guestbook/Persistence.java
```diff
@@ -16,27 +16,32 @@
...
 import com.google.cloud.firestore.Firestore;
 import com.google.cloud.firestore.FirestoreOptions;
+// [END gae_java11_firestore_dependencies]
 import java.util.concurrent.atomic.AtomicReference;
 
 /** Create a persistence connection to your Firestore instance. */
 public class Persistence {
 
-  private static AtomicReference<Firestore> firestore = new AtomicReference<>();
+  private static AtomicReference<Firestore> firestoreRef = new AtomicReference<>();
```
lesv: You could move this inside the getFirestore method.

appengine-java11/guestbook-cloud-firestore/src/main/java/com/example/guestbook/Persistence.java
```diff
@@ -16,27 +16,32 @@
...
+      firestoreRef.set(db);
     }
 
-    return firestore.get();
+    return firestoreRef.get();
   }
 
   public static void setFirestore(Firestore firestore) {
-    Persistence.firestore.set(firestore);
+    Persistence.firestoreRef.set(firestore);
```
lesv: And use the getFirestore() method here instead of using the private member.

appengine-java11/guestbook-cloud-firestore/src/main/java/com/example/guestbook/Persistence.java
averikitsch: Do I need the AtomicReference or can I just rebuild the database object? I'm not sure if this needs to be thread-safe.

lesv: Ask the client library folks.

PR #1406, opened 2019-05-06
Add new Sample Format guidelines. 

kurtisvg:
Add a document summarizing some java-specific requirements for the new sample format.

### DIFF
--- a/SAMPLE_FORMAT.md
+++ b/SAMPLE_FORMAT.md
@@ -78,7 +78,9 @@ Example:
 ```java
 try {
   // Do something
-} catch (Exception e) {
+} catch (IllegalArgumentException e) {
+  // IllegalArgumentException's are thrown when an invalid argument has been passed to a function.
+  // This error should be logged to that the root cause can be debugged and prevented in the future.
   System.out.println("Error during functionName: \n" + e.toString());
 }
 ```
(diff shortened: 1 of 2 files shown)


lesv:
You should update CONTRIBUTING.md as well

kurtisvg:
> You should update CONTRIBUTING.md as well

My plan is to wait until a decision is made on when the sample format becomes required before updating that.

SAMPLE_FORMAT.md
```diff
@@ -0,0 +1,130 @@
...
+ exceptions can be handled by logging the exception to `System.out.println`. If the exception is 
+ something the user may commonly encounter, include a comment explaining how to avoid or handle 
+ correctly.
+
+Example:
+```java
+try {
+  // Do something
+} catch (Exception e) {
+  System.out.println("Error during functionName: \n" + e.toString());
```
lesv: It would be better to show a common exception and good comment here.

SAMPLE_FORMAT.md
```diff
@@ -0,0 +1,130 @@
...
+## Additional Best Practices 
+
+The following are some general Java best practices that should be followed in samples to remain
+idiomatic. 
+
+### Time
+Use the `java.time` package when dealing with units of time in some manner. 
+
+### Logging
+Use `java.util.logging` for consistent logging in web applications.
```
lesv: Should we move to https://github.com/google/flogger at some point?

kurtisvg: Added to todo list for discussion.

PR #1412, opened 2019-05-10
Make the HttpHandler example independent of GAE

MaxRossGoogle:


tasks/src/main/java/com/example/task/CreateHttpTask.java
```diff
@@ -37,9 +37,9 @@ public static void main(String[] args) throws Exception {
     try (CloudTasksClient client = CloudTasksClient.create()) {
       // Variables provided by the system variables.
       // projectId = "my-project-id";
-      // queueName = "my-appengine-queue";
+      // queueName = "my-queue";
       // location = "us-central1";
-      // url = "https://<project-id>.appspot.com/tasks/create";
+      // url = "https://example.com/hellohandler";
```
lesv: You might wish to update the README.md file as well.

MaxRossGoogle:
Hmmm, I thought this was supposed to be a non-GAE sample because it's
included in the documentation for non-GAE HTTP tasks, but now I think we
may just need a totally different sample.

*From: *Les Vogel <notifications@github.com>
*Date: *Fri, May 10, 2019 at 10:13 AM
*To: *GoogleCloudPlatform/java-docs-samples
*Cc: *MaxRossGoogle, Author

*@lesv* approved this pull request.

lesv:
@averikitsch Can you take over Max's PR and include any suggestions he has.

PR #1413, opened 2019-05-10
Update streaming sample to use non-deprecated method

nnegrey:


nnegrey:
Not completely deleted, just removed a bunch of old profiles for running the samples, but since we removed the `main()` awhile back, I deleted them.

PR #1422, opened 2019-05-15
Updated Infinite streaming sample

blechdom:
The new sample uses result_end_time to *resend* unfinalized results in subsequent requests.
Would like advice on making a test.


blechdom:
Noticed the Java 8 test failed.
I looked at the logs and it seems to be unrelated to the file I was working on.
This is for recognizeIT punctuation discrepancies in from an audio file:

expected to contain:
    How old is the Brooklyn Bridge?
but was:
    Transcript : how old is the Brooklyn Bridge

Is this blocking my PR even though it's on a different file?

speech/cloud-client/src/test/java/com/example/speech/RecognizeIT.java
```diff
@@ -113,22 +113,22 @@ public void testStreamRecognize() throws Exception {
   @Test
   public void testAutoPunctuation() throws Exception {
     Recognize.transcribeFileWithAutomaticPunctuation(audioFileName);
-    String got = bout.toString();
-    assertThat(got).contains("How old is the Brooklyn Bridge?");
+    String got = bout.toString().toLowerCase();
+    assertThat(got).contains("how old is the brooklyn bridge");
```
nnegrey: @blechdom Hey, sorry just seeing this, can you revert these test changes?

These are testing the Auto Punctuation API and should be verifying capitalization and symbols. Otherwise, we need to notify the product team of an issue on their backend.

kurtisvg: @nnegrey Sorry, that was my fault. I pushed the changes to unblock the tests. Is it just the Auto Punctuation API you want reverted?

nnegrey: Yea, are they failing? I can go check with the product team if they made a change to something.

kurtisvg: Yes - see https://source.cloud.google.com/results/invocations/67658bb6-0025-46cf-8623-96d63e284f1b/targets/cloud-devrel%2Fjava%2Fjava-docs-samples%2Fjava8%2Fpresubmit/log

nnegrey: Filed a bug, if they respond and fix, I can revert the samples to the old state and leave them as is for now. 

Thanks!

PR #1423, opened 2019-05-16
Move files over

nnegrey:
Moving the samples from https://github.com/GoogleCloudPlatform/firebase-appengine-backend to this repo.


flexible/mobile-backend/src/main/webapp/Dockerfiles
```diff
@@ -0,0 +1,2 @@
+FROM gcr.io/google_appengine/jetty9-compat:githubheadasync
```
lesv: That's not really maintained.

flexible/mobile-backend/src/main/java/com/google/cloud/solutions/flexenv/common/Message.java
```diff
@@ -0,0 +1,69 @@
+/*
+ * Copyright 2018 Google LLC
```
lesv: You mentioned you were moving, so this is probably ok, but it seems wrong to checkin a new file w/ this date.

flexible/mobile-backend/pom.xml
```diff
@@ -0,0 +1,95 @@
...
+  <dependencies>
+    <dependency>
+      <groupId>com.google.appengine</groupId>
+      <artifactId>appengine-api-1.0-sdk</artifactId>
+      <version>${appengine.target.version}</version>
+    </dependency>
+    <dependency>
+      <groupId>javax.servlet</groupId>
+      <artifactId>servlet-api</artifactId>
+      <version>2.5</version>
```
lesv: that's kind of old

lesv: 3.1.0 and a different name is expected here.

flexible/mobile-backend/src/main/webapp/Dockerfiles
nnegrey: I'll check with the author on this.

flexible/mobile-backend/src/main/java/com/google/cloud/solutions/flexenv/common/Message.java
nnegrey: Updated

flexible/mobile-backend/pom.xml
nnegrey: Updated

nnegrey:
These should be good to be moved over now.

flexible/mobile-backend/pom.xml
```diff
@@ -0,0 +1,98 @@
...
+    <groupId>com.google.cloud.samples</groupId>
+    <artifactId>shared-configuration</artifactId>
+    <version>1.0.11</version>
+  </parent>
+
+  <properties>
+    <maven.compiler.target>1.8</maven.compiler.target>
+    <maven.compiler.source>1.8</maven.compiler.source>
+
+    <appengine.target.version>1.9.30</appengine.target.version>
```
lesv: That is really old

lesv: While this works and will get updated by DPEBot, we moved to putting the versions explicitly in the dependencies as it's easier for the user to figure out.

flexible/mobile-backend/pom.xml
```diff
@@ -0,0 +1,98 @@
...
+              <filtering>true</filtering>
+              <targetPath>WEB-INF</targetPath>
+            </resource>
+          </webResources>
+        </configuration>
+      </plugin>
+      <plugin>
+        <groupId>com.google.cloud.tools</groupId>
+        <artifactId>appengine-maven-plugin</artifactId>
+        <version>1.3.2</version>
```
lesv: 2.0.0 or so is current.

flexible/mobile-backend/pom.xml
lesv: It probably shouldn't even be used on Flex as compat is deprecated if I recall.

flexible/mobile-backend/src/main/webapp/Dockerfiles
lesv: Still not maintained.  It probably should be gcr.io/google_appengine/jetty9-compat or gcr.io/jetty9 (nothing in 2 years)or gcr.io/jetty

gcr.io/jetty-compat:release-20190619
or
gcr.io/jetty:2019-06-25-204809 (aka :9)

lesv: githubheadasync is very old

ricalo: I tested with:
* `gcr.io/google-appengine/jetty`
* `gcr.io/jetty`

The sample works with both, but I rather use `gcr.io/google-appengine/jetty` which is specified in the docs of the [Java/Jetty runtime](https://cloud.google.com/appengine/docs/flexible/java/dev-jetty9).

flexible/mobile-backend/pom.xml
ricalo: Yes, `2.0.0` is current. However, you need some extra configuration to `pom.xml` to move to that version. Check the additional `deploy.projectId` and `deploy.version` elements in the plugin configuration:

```xml
<plugin>
  <groupId>com.google.cloud.tools</groupId>
  <artifactId>appengine-maven-plugin</artifactId>
  <version>2.0.0</version>
  <configuration>
    <promote>true</promote>
    <deploy.projectId>playchat-241319</deploy.projectId>
    <deploy.version>sample</deploy.version>  </configuration>
</plugin>
```

flexible/mobile-backend/src/main/webapp/Dockerfiles
nnegrey: @ricalo, should this change from 
`FROM gcr.io/google_appengine/jetty9-compat:githubheadasync`
To
`FROM gcr.io/google_appengine/jetty9`?

ricalo: Almost, it's `gcr.io/google-appengine/jetty` (without the trailing 9).

nnegrey: Updated

flexible/mobile-backend/pom.xml
nnegrey: @ricalo, do you think it should be updated and have the user set the `projectId` field?

ricalo: Mmm... on second thought, the `projectId` could be set using:

```shell
$ gcloud config set project [project-id]
```
which is already specified in the original README. The `deploy.version` element is the piece that was missing from the original instructions.

It looks to me like you can hardcode the version in this sample. Using the same version means that the app is overwritten if users redeploy the sample, which should be fine in this scenario.

lesv:
Status update please?  can we close this?

nnegrey:
I think we should probably just close this as these use a lot of old stuff. @ricalo thoughts?

ricalo:
Closing sounds appropriate.

PR #1424, opened 2019-05-17
Update app engine plugin and readmes for Java 11

averikitsch:


ludoch:
BTW, thanks for upgrading the plugin to 2.0.0!!!

PR #1429, opened 2019-05-21
Moving Bigtable samples into this repo

billyjacobson:
From https://github.com/googleapis/google-cloud-java/tree/master/google-cloud-examples/src/main/java/com/google/cloud/examples/bigtable

### DIFF
--- a/bigtable/pom.xml
+++ b/bigtable/pom.xml
@@ -9,9 +9,17 @@
   <artifactId>bigtable</artifactId>
   <version>1.0-SNAPSHOT</version>
 
+  <!--
+    The parent pom defines common style checks and testing strategies for our samples.
+    Removing or replacing it should not affect the execution of the samples in anyway.
+  -->
+  <parent>
+    <groupId>com.google.cloud.samples</groupId>
+    <artifactId>shared-configuration</artifactId>
+    <version>1.0.11</version>
+  </parent>
+
   <name>bigtable</name>
-  <!-- FIXME change it to the project's website -->
-  <url>http://www.example.com</url>
 
   <properties>
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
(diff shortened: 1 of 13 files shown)


bigtable/pom.xml
```diff
@@ -0,0 +1,90 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
lesv: License & Copyright?

bigtable/pom.xml
```diff
@@ -0,0 +1,90 @@
...
+    <groupId>com.google.cloud.samples</groupId>
+    <artifactId>shared-configuration</artifactId>
+    <version>1.0.11</version>
+  </parent>
+
+  <name>bigtable</name>
+
+  <properties>
+    <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+    <maven.compiler.source>1.7</maven.compiler.source>
```
lesv: 1.8 or higher?

bigtable/pom.xml
```diff
@@ -0,0 +1,90 @@
...
+      <groupId>junit</groupId>
+      <artifactId>junit</artifactId>
+      <version>4.12</version>
+      <scope>test</scope>
+    </dependency>
+    <!-- https://mvnrepository.com/artifact/com.google.cloud/google-cloud-bigtable-admin -->
+    <dependency>
+      <groupId>com.google.cloud</groupId>
+      <artifactId>google-cloud-bigtable-admin</artifactId>
+      <version>0.81.0-alpha</version>
```
lesv: Is this current?  I see com.google.api.grpc:grpc-google-cloud-bigtable-admin-v2:0.57.0 from this month in maven-central.

lesv:
Looks like the 'bigtable.project property' isn't set during kokoro, you probably wish to do so, I'd suggest using PROJECT_ID instead, but either way, you'll need to fix tests.

PR #1430, opened 2019-05-21
Cloud Spanner: Specify the name for packaging the JAR

hegemonic:
Right now, the Cloud Spanner docs go stale whenever we upgrade the parent to a new version. The docs are stale right now, in fact; they say `1.0.10`, but the current version is `1.0.11` ([example](https://cloud.google.com/spanner/docs/getting-started/java/)).

With this change, we can update the docs to tell customers to run something like `mvn package -Drevision=0.1.0-SNAPSHOT`, and then use `0.1.0-SNAPSHOT` throughout the examples.

### DIFF
--- a/spanner/cloud-client/pom.xml
+++ b/spanner/cloud-client/pom.xml
@@ -18,7 +18,6 @@ limitations under the License.
   <groupId>com.example.spanner</groupId>
   <artifactId>spanner-google-cloud-samples</artifactId>
   <packaging>jar</packaging>
-  <version>${revision}</version>
 
   <!--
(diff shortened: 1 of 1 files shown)


hegemonic:
FWIW, Cloud Spanner provides a very similar "getting started" doc for every language we support ([.NET example](https://cloud.google.com/spanner/docs/getting-started/csharp/)), and there's a CLI for every language. It would be nice to maintain parity across languages, so I'm not eager to rip out the CLI code.

I'm more than happy, though, to package the JAR in a more idiomatic way! PTAL.

PR #1431, opened 2019-05-21
Writing quickstart + test for Bigtable

billyjacobson:


kurtisvg:
> Isn't there a pom.xml or something to build & test this

Looks like the pom [here](https://github.com/GoogleCloudPlatform/java-docs-samples/blob/e1f37d1bfe1d9d17c329f1eb443e46ca56478f0b/bigtable/pom.xml) is being used.

billyjacobson:
> Isn't there a pom.xml or something to build & test this? I hate to just check in something that can't be built on an automated basis.
> @kurtisvg FYI

Sorry, I put out the stuff with the pom in a separate PR cus that code was already existing, but this code is all new. You already reviewed that other PR though.

kurtisvg:
Oh the reason this isn't running tests is because it's not being merged to master. Do you want to merge them into the master branch?

lesv:
I'm too til Tuesday

bigtable/src/test/java/com/m/examples/bigtable/QuickstartTest.java
```diff
@@ -0,0 +1,110 @@
...
+  private static String projectId;
+  private static String instanceId;
+  private ByteArrayOutputStream bout;
+
+  @BeforeClass
+  public static void beforeClass() throws IOException {
+    projectId = System.getProperty(PROJECT_PROPERTY_NAME);
+    instanceId = System.getProperty(INSTANCE_PROPERTY_NAME);
+
+    BigtableDataSettings settings =
```
kurtisvg: These clients don't look like they are being used, can we just take them out?

bigtable/src/test/java/com/m/examples/bigtable/QuickstartTest.java
```diff
@@ -0,0 +1,110 @@
...
+  private static final String TABLE_ID = "quickstart-table";
+  private static BigtableDataClient dataClient;
+  private static BigtableTableAdminClient adminClient;
+  private static String projectId;
+  private static String instanceId;
+  private ByteArrayOutputStream bout;
+
+  @BeforeClass
+  public static void beforeClass() throws IOException {
+    projectId = System.getProperty(PROJECT_PROPERTY_NAME);
```
kurtisvg: Preferably these should be environment variables instead of system properties.

bigtable/src/test/java/com/m/examples/bigtable/QuickstartTest.java
```diff
@@ -0,0 +1,110 @@
...
+  @AfterClass
+  public static void afterClass() throws Exception {
+    dataClient.close();
+    adminClient.close();
+  }
+
+  @Before
+  public void setup() throws IOException {
+    if (adminClient == null || dataClient == null) {
+      throw new AssumptionViolatedException(
```
kurtisvg: So while this is a good idea, it's going to break our current tests. Kokoro won't fall if the this error it throw (which it should). This is part of a large problem we need to solve with our testing setup - what to do that the user can build a package without tests but without passing on our tests.

For now, can you throw and IllegalStateException or similar instead?

bigtable/src/main/java/com/m/examples/bigtable/Quickstart.java
```diff
@@ -0,0 +1,53 @@
...
+// START [bigtable_quickstart_veneer]
+import com.google.api.gax.rpc.NotFoundException;
+import com.google.cloud.bigtable.data.v2.BigtableDataClient;
+import com.google.cloud.bigtable.data.v2.BigtableDataSettings;
+import com.google.cloud.bigtable.data.v2.models.Row;
+import com.google.cloud.bigtable.data.v2.models.RowCell;
+
+public class Quickstart {
+  public static void quickstart(String projectId, String instanceId, String tableId)
+      throws Exception {
```
kurtisvg: Don't throw exceptions from this function. Instead if you have to, you can catch the exception and log it. See this [example](https://github.com/GoogleCloudPlatform/java-docs-samples/blob/master/SAMPLE_FORMAT.md#exception-handling).

bigtable/src/main/java/com/m/examples/bigtable/Quickstart.java
```diff
@@ -0,0 +1,53 @@
...
+    // String projectId = "my-project-id";
+    // String instanceId = "my-instance-id";
+    // String tableId = "my-table-id";
+
+    BigtableDataSettings settings =
+        BigtableDataSettings.newBuilder().setProjectId(projectId).setInstanceId(instanceId).build();
+
+    // Creates a bigtable data client.
+    BigtableDataClient dataClient = BigtableDataClient.create(settings);
```
kurtisvg: Does this client need to be closed when complete? If so, please use try/catch and make it clear with a comment. See [client initialization](https://github.com/GoogleCloudPlatform/java-docs-samples/blob/master/SAMPLE_FORMAT.md#client-initialization) in the sample guide.

bigtable/src/test/java/com/m/examples/bigtable/QuickstartTest.java
```diff
@@ -0,0 +1,110 @@
...
+        "System property '%s' is required to perform these tests.".format(varName));
+  }
+
+  @BeforeClass
+  public static void checkRequirements() {
+    requireSysProp(PROJECT_PROPERTY_NAME);
+    requireSysProp(INSTANCE_PROPERTY_NAME);
+  }
+
+  private static final String PROJECT_PROPERTY_NAME = "bigtable.project";
```
kurtisvg: Please move the fields to the top of the class.

bigtable/src/main/java/com/m/examples/bigtable/Quickstart.java
```diff
@@ -0,0 +1,53 @@
+/*
+ * Copyright 2019 Google Inc.
```
kurtisvg:
```suggestion
 * Copyright 2019 Google LLC
```

bigtable/src/test/java/com/m/examples/bigtable/QuickstartTest.java
```diff
@@ -0,0 +1,110 @@
+/*
+ * Copyright 2019 Google Inc.
```
kurtisvg:
```suggestion
 * Copyright 2019 Google LLC
```

bigtable/src/test/java/com/m/examples/bigtable/QuickstartTest.java
billyjacobson: Should I define the env variables here? https://github.com/GoogleCloudPlatform/java-docs-samples/blob/master/.kokoro/tests/run_tests.sh

There don't seem to be any product/test specific env variables, so I wasn't sure

kurtisvg: Sorry for the late response -- right now we have files that are sourced to add secrets too as done [here](https://github.com/GoogleCloudPlatform/java-docs-samples/blob/21c141accd4870b27fd890aa3a9ea1b5cf84d712/.kokoro/tests/run_tests.sh#L58). If you send me an email with the credentials you would like to be added, I can add a file to the correct bucket.

billyjacobson: Fixed for this test and the other

bigtable/src/test/java/com/m/examples/bigtable/InstanceAdminExampleTest.java
```diff
@@ -51,9 +50,16 @@
...
+        System.getenv(varName),
+        "System property '%s' is required to perform these tests.".format(varName));
+    return System.getenv(varName);
+  }
+
   @BeforeClass
   public static void beforeClass() throws IOException {
-    projectId = System.getProperty(PROJECT_PROPERTY_NAME);
+    projectId = requireEnv("GOOGLE_CLOUD_PROJECT");
     if (projectId == null) {
```
kurtisvg: `requireEnv` should throw an error if `projectID` is null, so you can remove this check

bigtable/src/test/java/com/m/examples/bigtable/QuickstartTest.java
```diff
@@ -0,0 +1,82 @@
...
+import org.junit.BeforeClass;
+import org.junit.Test;
+
+/**
+ * Integration tests for {@link Quickstart}
+ */
+public class QuickstartTest {
+
+  private static final String INSTANCE_PROPERTY_NAME = "BIGTABLE_TESTING_INSTANCE";
+  private static final String TABLE_ID = "quickstart-table";
```
igorbernstein2: Please don't hardcode the table id, it will cause these tests to be flaky when multiple PRs are opened

bigtable/src/test/java/com/m/examples/bigtable/QuickstartTest.java
```diff
@@ -0,0 +1,82 @@
...
+  }
+
+  @BeforeClass
+  public static void beforeClass() {
+    projectId = requireEnv("GOOGLE_CLOUD_PROJECT");
+    instanceId = requireEnv(INSTANCE_PROPERTY_NAME);
+  }
+
+  @Before
+  public void setupStream() {
```
igorbernstein2: Please use the junit convention of calling this setUp()

bigtable/src/test/java/com/m/examples/bigtable/TableAdminExampleTest.java
```diff
@@ -45,19 +46,25 @@
...
   private static String instanceId;
   private static String projectId;
   private String tableId;
   private TableAdminExample tableAdmin;
 
+  private static String requireEnv(String varName) {
+    assertNotNull(
+        System.getenv(varName),
+        "System property '%s' is required to perform these tests.".format(varName));
+    return System.getenv(varName);
```
igorbernstein2: Same as above

bigtable/src/main/java/com/m/examples/bigtable/Quickstart.java
```diff
@@ -0,0 +1,59 @@
...
+
+import com.google.api.gax.rpc.NotFoundException;
+import com.google.cloud.bigtable.data.v2.BigtableDataClient;
+import com.google.cloud.bigtable.data.v2.BigtableDataSettings;
+import com.google.cloud.bigtable.data.v2.models.Row;
+import com.google.cloud.bigtable.data.v2.models.RowCell;
+
+public class Quickstart {
+
+  public static void quickstart(String projectId, String instanceId, String tableId) {
```
igorbernstein2: Please include a main()

bigtable/src/main/java/com/m/examples/bigtable/Quickstart.java
```diff
@@ -0,0 +1,59 @@
...
+          System.out.printf(
+              "Family: %s    Qualifier: %s    Value: %s%n",
+              cell.getFamily(), cell.getQualifier().toStringUtf8(), cell.getValue().toStringUtf8());
+        }
+      } catch (NotFoundException e) {
+        System.err.println("Failed to read from a non-existent table: " + e.getMessage());
+      }
+
+    } catch (Exception e) {
+      System.out.println("Error during functionName: \n" + e.toString());
```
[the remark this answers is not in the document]
igorbernstein2: Also, please use System.err

bigtable/src/test/java/com/m/examples/bigtable/HelloWorldTest.java
```diff
@@ -50,10 +49,17 @@
   private static String instanceId;
   private HelloWorld helloWorld;
 
+  private static String requireEnv(String varName) {
+    assertNotNull(
+        System.getenv(varName),
+        "System property '%s' is required to perform these tests.".format(varName));
+    return System.getenv(varName);
```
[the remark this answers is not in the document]
billyjacobson: assertNotNull doesn't return the value of System.getenv(varName)

bigtable/src/main/java/com/m/examples/bigtable/Quickstart.java
billyjacobson: we aren't doing that anymore for our samples https://github.com/GoogleCloudPlatform/java-docs-samples/blob/master/SAMPLE_FORMAT.md#samples-format

bigtable/src/test/java/com/m/examples/bigtable/HelloWorldTest.java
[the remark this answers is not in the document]
igorbernstein2: Oops, I thought you were using Preconditions.checkNotNull, which does. Any reason not to use that?

PR #1432, opened 2019-05-21
Cloud Spanner: Fix samples that transfer money between albums

hegemonic:
The `writeWithTransactionUsingDml` method moved $200,000 from Album1 to Album2. However, Album1 is initialized with a budget of $100,000. As a result, you couldn't work through the [getting started doc](https://cloud.google.com/spanner/docs/getting-started/java/) from start to finish. This snippet also checked whether Album1's budget exceeded a minimum, which isn't consistent with the equivalent snippets for other languages.

In addition, the `writeWithTransaction` method moved $200,000, but in the opposite direction, which was somewhat confusing.

Finally, neither method confirmed that the source album's budget was at least as large as the transfer amount.

I made both methods transfer $100,000 from Album1 to Album2. I also fixed the logic errors and updated the tests as needed.


lesv:
Just saw @kurtisvg  request for @jsimonweb to review - please don't merge until then

jsimonweb:
@hegemonic 
Please simplify this PR to match [this example PR](https://github.com/GoogleCloudPlatform/java-docs-samples/compare/master...jsimonweb:updateWriteWithTransactionUsingDml?expand=1):

This will minimize the changes required to code samples, tests and docs across multiple languages.

Thank you!

hegemonic:
@jsimonweb I can do that (and update other languages to match).

One question, though: Should we add an explicit `if (album2Budget >= transfer)` check as well? I know that's redundant with `if (album2Budget >= 300000)`, but in a real system that moves money around, you'd need that sort of check.

Or should we just get rid of `if (album2Budget >= 300000)` entirely, add `if (album2Budget >= transfer)`, and use that conditional to make the same point?

jsimonweb:
Let's get rid of if (album2Budget >= 300000) entirely and add if (album2Budget >= transfer) 
since that makes the intention of this logic more clear.

hegemonic:
Rather than banging my head against the CLA issue, I'm just going to close this pull request and open a new one that includes the requested changes.

PR #1436, opened 2019-05-23
StringBuffer --> StringBuilder

elharo:
@lesv Local variables don't need to be thread safe.


lesv:
@dpebot merge on green

PR #1437, opened 2019-05-23
fixed region tags and refactored some methods

munkhuushmgl:


iot/api-client/manager/src/main/java/com/example/cloud/iot/examples/DeviceRegistryExample.java
```diff
@@ -884,39 +883,14 @@ public static void unbindDeviceFromGateway(
             .execute();
 
     System.out.println(String.format("Device unbound: %s", response.toPrettyString()));
-    // [END unbind_device_from_gateway]
-  }
-
-  public static void attachDeviceToGateway(MqttClient client, String deviceId)
```
gguuss: Why are you removing this from this file?

iot/api-client/manager/src/main/java/com/example/cloud/iot/examples/DeviceRegistryExample.java
```diff
@@ -971,7 +945,7 @@ public static void createDevice(
             .execute();
 
     System.out.println("Created device: " + createdDevice.toPrettyString());
-    // [END create_device]
+    // [END iot_create_device]
```
gguuss: You missed the:

* create_gateway
* list_gateways
* list_devices_for_gateway

functions below

iot/api-client/manager/src/main/java/com/example/cloud/iot/examples/DeviceRegistryExample.java
munkhuushmgl: I moved that into MqttExample, i thought attachDevice related to mqtt

iot/api-client/manager/src/main/java/com/example/cloud/iot/examples/DeviceRegistryExample.java
munkhuushmgl: Ahh, i will add that real quick

iot/api-client/manager/src/main/java/com/example/cloud/iot/examples/DeviceRegistryExample.java
munkhuushmgl: I have noticed in python sample, attach, detach methods are in mqtt example instead of manager.py

Do you think it is good idea to move those into MqttExample? or keep in DeviceRegistryManager?

gguuss: This makes sense, we'll just need to remember to update the docs too.

PR #1438, opened 2019-05-24
Cloud Spanner: Fix samples that transfer money between albums

hegemonic:
The `writeWithTransactionUsingDml` method moved $200,000 from Album1 to Album2. However, Album1 is initialized with a budget of $100,000. As a result, you couldn't work through the [getting started doc](https://cloud.google.com/spanner/docs/getting-started/java/) from start to finish.

Also, there were some unnecessary inconsistencies between `writeWithTransaction` and `writeWithTransactionUsingDml`, and neither method explicitly compared the transfer amount with the source album's existing balance.

I changed the methods so they both transfer $200,000 from Album2 to Album1. I also added explicit checks against the existing balance, and I updated the tests as needed.

(This is an updated version of #1432. @jsimonweb, PTAL.)

### DIFF
--- a/spanner/cloud-client/src/main/java/com/example/spanner/SpannerSample.java
+++ b/spanner/cloud-client/src/main/java/com/example/spanner/SpannerSample.java
@@ -1055,7 +1055,7 @@ public class SpannerSample {
                       "SELECT MarketingBudget from Albums WHERE SingerId = 1 and AlbumId = 1";
                   ResultSet resultSet2 = transaction.executeQuery(Statement.of(sql2));
                   long album1Budget = 0;
-                  while (resultSet.next()) {
+                  while (resultSet2.next()) {
                     album1Budget = resultSet2.getLong("MarketingBudget");
                   }
                   album1Budget += transfer;
--- a/spanner/cloud-client/src/main/java/com/example/spanner/SpannerSample.java
+++ b/spanner/cloud-client/src/main/java/com/example/spanner/SpannerSample.java
@@ -60,8 +60,8 @@ import java.util.concurrent.TimeUnit;
  *   <li>Writing data using a read-write transaction.
  *   <li>Using an index to read and execute SQL queries over data.
  *   <li>Using commit timestamp for tracking when a record was last updated.
- *   <li>Using Google API Extensions for Java to make thread-safe requests via
- *       long-running operations. http://googleapis.github.io/gax-java/
+ *   <li>Using Google API Extensions for Java to make thread-safe requests via long-running
+ *       operations. http://googleapis.github.io/gax-java/
  * </ul>
  */
 public class SpannerSample {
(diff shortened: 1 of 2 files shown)


spanner/cloud-client/src/main/java/com/example/spanner/SpannerSample.java
```diff
@@ -1047,40 +1040,40 @@ public Void run(TransactionContext transaction) throws Exception {
...
-                if (album1Budget >= 300000) {
+                long transfer = 200000;
+                if (album2Budget >= transfer) {
                   String sql2 =
-                      "SELECT MarketingBudget from Albums WHERE SingerId = 2 and AlbumId = 2";
+                      "SELECT MarketingBudget from Albums WHERE SingerId = 1 and AlbumId = 1";
                   ResultSet resultSet2 = transaction.executeQuery(Statement.of(sql2));
-                  long album2Budget = 0;
+                  long album1Budget = 0;
                   while (resultSet.next()) {
```
jsimonweb: This line should be updated so that the next() test runs against "resultSet2":

while (resultSet2.next()) {

spanner/cloud-client/src/test/java/com/example/spanner/SpannerSampleIT.java
```diff
@@ -185,13 +182,13 @@ public void testSample() throws Exception {
     assertThat(out).contains("12 Melissa Garcia");
 
     runSample("writewithtransactionusingdml");
-    out = runSample("querymarketingbudget");    
-    assertThat(out).contains("1 1 1800000");
-    assertThat(out).contains("2 2 200000");
-    
+    out = runSample("querymarketingbudget");
+    assertThat(out).contains("1 1 200000");
```
jsimonweb: Once the update/fix noted in my comment above is made this value should be "2200000". 
Thus this assert statement should be updated to be:

assertThat(out).contains("1 1 2200000");

spanner/cloud-client/src/test/java/com/example/spanner/SpannerSampleIT.java
```diff
@@ -185,13 +182,13 @@ public void testSample() throws Exception {
...
+    assertThat(out).contains("2 2 550000");
+
     runSample("updateusingpartitioneddml");
     out = runSample("querymarketingbudget");
-    assertThat(out).contains("1 1 1800000");
+    assertThat(out).contains("1 1 200000");
```
jsimonweb: Once the update/fix noted in my comment above is made this value should be "2200000".
Thus this assert statement should be updated to be:

assertThat(out).contains("1 1 2200000");

spanner/cloud-client/src/main/java/com/example/spanner/SpannerSample.java
hegemonic: Great catch. Done.

hegemonic:
@jsimonweb PTAL

PR #1441, opened 2019-05-28
Removes no-longer-used beta features section

gguuss:
Quick PR to remove the beta-features, they're no longer used in the docs.


lesv:
You may find that testing fails if a parent pom was doing it.

gguuss:
Oh, just saw that last comment, looked like tests were green

munkhuushmgl:
LGTM

Issue #1312, opened 2019-01-10
Misspelled System.out.println


austinfathman:
https://github.com/GoogleCloudPlatform/java-docs-samples/blob/646cbc66252198a8e62616ad4ef0ff17a7de48e1/vision/automl/src/main/java/com/google/cloud/vision/samples/automl/ModelApi.java#L200

https://github.com/GoogleCloudPlatform/java-docs-samples/blob/646cbc66252198a8e62616ad4ef0ff17a7de48e1/vision/automl/src/main/java/com/google/cloud/vision/samples/automl/ModelApi.java#L202

"Tranning" -> "Training"

nnegrey:
@sirtorry or @andrewferlitsch can one of you take a look?

nnegrey:
Solved in https://github.com/GoogleCloudPlatform/java-docs-samples/pull/1451. Closing.

PR #1451, opened 2019-06-10
fix misspelling reported by user

andrewferlitsch:


vision/automl/src/main/java/com/google/cloud/vision/samples/automl/ModelApi.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright 2018 Google Inc.
+ * Copyright 2018-19 Google LLC
```
nnegrey: You can leave the license file as is, only need to update the year on new files and the java linter will throw an error as it expects a certain year and style based on the configuration.

nnegrey:
Solves: https://github.com/GoogleCloudPlatform/java-docs-samples/issues/1312

kurtisvg:
Fixes #1312

Issue #1313, opened 2019-01-10
Automl tutorial: flower_trainingdata.csv missing


charbull:
Hello,

Trying to do the following tutorial: https://github.com/GoogleCloudPlatform/java-docs-samples/tree/master/vision/automl

It seems that this is not existant: **gs://java-docs-samples-testing/flower_traindata.csv**

Thank you

nnegrey:
@sirtorry or @andrewferlitsch  can one of you take a look?

nnegrey:
Solved in https://github.com/GoogleCloudPlatform/java-docs-samples/pull/1452

PR #1452, opened 2019-06-10
fix wrong path to training example data

andrewferlitsch:


nnegrey:
Solves: https://github.com/GoogleCloudPlatform/java-docs-samples/issues/1313

nnegrey:
Investigating test failures

PR #1458, opened 2019-06-11
vision: address flakes due to collisions.

shollyman:
Multiple tests in the vision suite write output to GCS.  Sadly, all wrote to the
same output prefix in a nondeterministic order, which can cause issues.  This change does two things:

1) ensures each test run uses a unique prefix
2) makes the prefix clearer to disambiguate output from the cloud_client and the beta samples.

I suspect the beta samples should be cleaned up, but it appears there's still linkages into cloud docs which is a more involved change.  This should get the
tests green once more.

### DIFF
--- a/vision/beta/cloud-client/src/main/java/com/example/vision/AsyncBatchAnnotateImagesGcs.java
+++ b/vision/beta/cloud-client/src/main/java/com/example/vision/AsyncBatchAnnotateImagesGcs.java
@@ -17,7 +17,6 @@
 package com.example.vision;
 
 // [START vision_async_batch_annotate_images_beta]
-import com.google.api.core.ApiFuture;
 import com.google.api.gax.longrunning.OperationFuture;
 import com.google.api.gax.paging.Page;
 import com.google.cloud.storage.Blob;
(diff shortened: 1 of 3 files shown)


vision/beta/cloud-client/src/main/java/com/example/vision/AsyncBatchAnnotateImagesGcs.java
```diff
@@ -18,23 +18,16 @@
...
+import com.google.api.gax.longrunning.OperationFuture;
 import com.google.api.gax.paging.Page;
 import com.google.cloud.storage.Blob;
 import com.google.cloud.storage.Bucket;
 import com.google.cloud.storage.Storage;
 import com.google.cloud.storage.Storage.BlobListOption;
 import com.google.cloud.storage.StorageOptions;
-import com.google.cloud.vision.v1p4beta1.AnnotateImageRequest;
-import com.google.cloud.vision.v1p4beta1.AsyncBatchAnnotateImagesRequest;
+import com.google.cloud.vision.v1p4beta1.*;
```
lesv: We don't like .* .   https://google.github.io/styleguide/javaguide.html#s3.3-import-statements

shollyman: IDE decided to be helpful and collapse a long explicit list into a wildcard.

PR #1459, opened 2019-06-11
Add tests to Cloud Tasks Samples

averikitsch:


tasks/src/main/java/com/example/task/CreateHttpTask.java
```diff
@@ -27,19 +27,18 @@
...
-    String url = System.getenv("URL");
 
     // Instantiates a client.
     try (CloudTasksClient client = CloudTasksClient.create()) {
       // Variables provided by the system variables.
       // projectId = "my-project-id";
       // queueName = "my-queue";
       // location = "us-central1";
-      // url = "https://example.com/taskhandler";
+      String url = "https://example.com/taskhandler";
```
lesv: Do you call this out in the README?  (CHANGEME)

tasks/src/main/java/com/example/task/CreateHttpTaskWithToken.java
```diff
@@ -32,24 +32,24 @@ public static void main(String[] args) throws Exception {
...
     // Instantiates a client.
     try (CloudTasksClient client = CloudTasksClient.create()) {
       // Variables provided by the system variables.
       // projectId = "my-project-id";
       // queueName = "my-queue";
       // location = "us-central1";
-      // url = "https://example.com/taskhandler";
-      String payload = "hello";
+      String email = args[0]; // Cloud IAM service account
+      String url =
```
lesv: Should we suggest changing this?

tasks/src/test/java/com/example/task/CreateHttpTaskIT.java
```diff
@@ -0,0 +1,65 @@
...
+import org.junit.runner.RunWith;
+import org.junit.runners.JUnit4;
+
+/** Tests for creating Tasks with HTTP targets. */
+@RunWith(JUnit4.class)
+public class CreateHttpTaskIT {
+
+  private ByteArrayOutputStream bout;
+  private PrintStream out;
+  // @Rule public Timeout globalTimeout = Timeout.seconds(300); // 5 minute timeout
```
dzliergaard: Kept intentionally?

tasks/src/main/java/com/example/task/CreateHttpTaskWithToken.java
averikitsch: Since this is a snippet. I have updated with the sample without a main and highlighted the parameters.

averikitsch:
@lesv please review changes

PR #1463, opened 2019-06-12
Clean up the code and fix the sample tests

nnegrey:
Fixes: 
- https://github.com/GoogleCloudPlatform/java-docs-samples/pull/1451
- https://github.com/GoogleCloudPlatform/java-docs-samples/pull/1452

### DIFF
--- a/vision/automl/src/main/java/com/google/cloud/vision/samples/automl/PredictionApi.java
+++ b/vision/automl/src/main/java/com/google/cloud/vision/samples/automl/PredictionApi.java
@@ -34,7 +34,6 @@ import com.google.cloud.automl.v1beta1.PredictionServiceClient;
 import com.google.protobuf.ByteString;
 
 import java.io.IOException;
-import java.io.PrintStream;
 import java.nio.file.Files;
 import java.nio.file.Paths;
 import java.util.HashMap;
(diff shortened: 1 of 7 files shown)


vision/automl/src/main/java/com/google/cloud/vision/samples/automl/PredictionApi.java
```diff
@@ -65,76 +61,76 @@
...
-    predictParser.addArgument("modelId");
-    predictParser.addArgument("filePath");
-    predictParser.addArgument("scoreThreshold").nargs("?").type(String.class).setDefault("");
+    parser.addArgument("modelId").required(true);
+    parser.addArgument("filePath").required(true);
+    parser.addArgument("scoreThreshold").nargs("?").type(String.class).setDefault("");
 
-    String projectId = System.getenv("PROJECT_ID");
+    String projectId = System.getenv("GOOGLE_CLOUD_PROJECT");
     String computeRegion = System.getenv("REGION_NAME");
```
lesv: I'm fine w/ this, but don't we have a standard of ZONE & REGION, if not, should we?

nnegrey: Sorry, clarification question: do you mean like the var name should be `REGION` or like we should provide a suggested default region: `us-central1`?

PR #1464, opened 2019-06-12
bigquery: elide unexercised variable from SimpleApp.

shollyman:


lesv:
@dpebot merge when green

PR #1468, opened 2019-06-13
Fix FhirStoreDelete region tag

noerog:

PR #1472, opened 2019-06-18
Ignore FHIR resource conditional patch test while java 8 is broken.

dzlier-gcp:
Updating region tag names and disabling test. Filed internal bug to track broken test.


dzlier-gcp:
Created to replace #1468

dzlier-gcp:
@dpebot merge when green

PR #1303, opened 2018-12-19
Update to use shared data bucket for import test in Product Search

nirupa-kumar:
* Fixed the importProductSets test to use the csv file in the shared bucket. 
* Added note for the pre-created product set requirement for Similar Product Search

PR #1473, opened 2019-06-18
Pdt search updates

nnegrey:


nnegrey:
Tried to push to this PR, but guess that didn't work. 
We can close the old PR: https://github.com/GoogleCloudPlatform/java-docs-samples/pull/1303
And use this one instead. 

Nirupa used to have the CLA signed, but she has since left and maybe that undid her CLA.

PR #1477, opened 2019-06-24
Add samples for Google Cloud Functions. 

kurtisvg:


functions/snippets/pom.xml
```diff
@@ -0,0 +1,79 @@
...
+    <maven.compiler.target>11</maven.compiler.target>
+    <maven.compiler.source>11</maven.compiler.source>
+    <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+  </properties>
+
+  <dependencies>
+    <dependency>
+      <groupId>javax.servlet</groupId>
+      <artifactId>javax.servlet-api</artifactId>
+      <version>4.0.1</version>
```
lesv: This is how they spec the API.  So, I'm fine w/ it, but they are actually running inside jetty which is 3.1

Issue #1481, opened 2019-06-24
Flaky Healthcare tests


averikitsch:
## In which file did you encounter the issue?

healthcare/v1beta1/src/test/java/snippets/healthcare/FhirResourceTests.java

## Describe the issue

[Test Fusion](https://fusion.corp.google.com/runanalysis/test/prod%3Acloud-devrel%2Fjava%2Fjava-docs-samples%2Fjava8%2Fperiodic/prod%3Acloud-devrel%2Fjava%2Fjava-docs-samples%2Fjava8%2Fperiodic/KOKORO/569be8a6-9318-4272-a25b-ad463abef9a4/1561316765710/build%20%2375/Targets?target=github%2Fjava-docs-samples%2Fhealthcare%2Fv1beta1%2Ftarget%2Fsurefire-reports&drilldownTab=tests)

gguuss:
@noerog /FYI

lesv:
```
FhirResourceTests(sponge_log).test_FhirResourceConditionalPatch

Run time 0:00:00.260
Error: Exception patching FHIR resource: HTTP/1.1 404 Not Found
java.lang.RuntimeException:
Exception patching FHIR resource: HTTP/1.1 404 Not Found

	at snippets.healthcare.fhir.resources.FhirResourceConditionalPatch.fhirResourceConditionalPatch(FhirResourceConditionalPatch.java:85)
	at snippets.healthcare.FhirResourceTests.test_FhirResourceConditionalPatch(FhirResourceTests.java:188)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:566)
	at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:59)
	at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
	at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:56)
	at org.junit.internal.runners.statements.InvokeMethod.evaluate(InvokeMethod.java:17)
[...]

PR #1492, opened 2019-06-25
Update DicomStoreExport.java

noerog:


lesv:
Java 8 doesn't appear to pass tests

lesv:
@noerog has just changed tags
@averikitsch @kurtisvg do we want to fix the test repos or admin merge (or something else)

kurtisvg:
#1481 already has the test as flaky, so I'm fine with admin merging.

PR #1493, opened 2019-06-25
Add tests for Cloud Tasks app engine targets

averikitsch:


appengine-java11/tasks/pom.xml
```diff
@@ -45,10 +45,21 @@ Copyright 2019 Google LLC
     <dependency>
       <groupId>com.google.cloud</groupId>
       <artifactId>google-cloud-tasks</artifactId>
-      <version>1.6.0</version>
-      <scope>provided</scope>
+      <version>1.3.0</version>
```
lesv: Why the roll back?

appengine-java11/tasks/pom.xml
```diff
@@ -45,10 +45,21 @@ Copyright 2019 Google LLC
...
     </dependency>
     <!-- [END gae_java11_task_pom] -->
+    <!-- Test dependencies -->
+    <dependency>
+      <groupId>junit</groupId>
+      <artifactId>junit</artifactId>
+      <version>4.13-beta-2</version>
```
lesv: `<scope>test</scope>`

appengine-java11/tasks/pom.xml
averikitsch: The library is broken, hence the added tests to catch this next time.

appengine-java11/tasks/pom.xml
averikitsch: done. Thanks for the review.

PR #1499, opened 2019-06-28
Container analysis v1

DanSanche:
Updated Container Analysis samples to use new v1 libraries


container-registry/container-analysis/src/main/java/com/example/containeranalysis/CreateOccurrence.java
```diff
@@ -39,20 +41,29 @@ public static Occurrence createOccurrence(String resourceUrl, String noteId,
...
+    issueBuilder.setAffectedPackage("your-package-here");
+    Version.Builder affectedVersionBuilder = Version.newBuilder();
+    affectedVersionBuilder.setKind(Version.VersionKind.MINIMUM);
+    issueBuilder.setAffectedVersion(affectedVersionBuilder);
+    Version.Builder fixedVersionBuilder = Version.newBuilder();
+    fixedVersionBuilder.setKind(Version.VersionKind.MAXIMUM);
+    issueBuilder.setFixedVersion(fixedVersionBuilder);
+
+    VulnerabilityOccurrence.Builder vulBuilder = VulnerabilityOccurrence.newBuilder();
+    vulBuilder.addPackageIssue(issueBuilder);
```
DanSanche: If anyone has any advice on how to write these complex builder sections in a clean way, that would be much appreciated

kurtisvg: Chain them together, and use build at the end. 
```
PackageIssue issue = PackageIssue.newBuilder()
                                        .setAffectedCpeUri("your-uri-here")
                                        .setAffectedPackage("your-package-here")
                                        ....
                                        .build()
```

Generally, they replace constructors. See an example from the canonical [here](https://github.com/GoogleCloudPlatform/java-docs-samples/blob/3439c96c7ff8661d28210618615d7d95bc4ad7a8/dlp/src/main/java/dlp/snippets/InspectImageFile.java#L73).

DanSanche: thanks, that's much better!

PR #1501, opened 2019-07-02
Add config for auto-assigning issues and PRs to folks. 

kurtisvg:
Little bot that assigns issues or PRs to folks when opened if they don't already have an assignee. 

May still be a little rough around the edges.


kurtisvg:
> Maybe consider renaming the file to something more concrete.

This is just a stopgap measure until the DRIFT team gets something together. This way they know who to blame :)

PR #1504, opened 2019-07-08
Added command line option class + option to pass different lang code as argument

munkhuushmgl:


speech/cloud-client/README.md
```diff
@@ -155,7 +155,11 @@ mvn exec:java -DRecognize -Dexec.args="word-level-conf gs://cloud-samples-tests/
 ```
 
 ## Infinite Streaming
-Continuously stream audio to the speech API over multiple requests 
+Continuously stream audio to the speech API over multiple requests (by default en-US).
 ```
 mvn exec:java -DInfiniteStreamRecognize
 ```
+If stream audio is in different language, you could also pass language code as a command line argument (for example, korea).
```
beccasaurus: `s/korea/Korean/`

speech/cloud-client/README.md
```diff
@@ -155,7 +155,11 @@ mvn exec:java -DRecognize -Dexec.args="word-level-conf gs://cloud-samples-tests/
...
+```
+mvn exec:java -Dexec.args="-lang_code=ko-KR" -DInfiniteStreamRecognize
+```
```
beccasaurus: (Based on your idea) do you want to also link to the full list of language codes?

https://cloud.google.com/speech-to-text/docs/languages

lesv: Or at least english?

munkhuushmgl:
@lesv I will change it to English

kurtisvg:
> @kurtisvg I'm inclined to approve, even though we are trying to avoid CLI's for java. This seems like a reasonable exception.

lesv@ - This appears to have an existing CLI, so small updates to existing sample fall within our current policy.

That said, this change is contrary to the samples best practices. While using a function variable is more useful when running from the command-line, it's less useful when used as a snippet because now there is a variable that is declared out of scope. 

To follow the sample guidelines it should be left as "en-US" or similar valid value, and should include a link to where a list of valid values can be found.

munkhuushmgl:
@kurtisvg I will revert back the code, and will include a link to list of valid values[language codes] inside README

beccasaurus:
Clarification for everyone –

This isn't actually a snippet. This is a CLI sample application used for tutorials and demos.

In the coming months, we hope to send a PR which will refactor this code so that it has region tags to support a website tutorial which is being authored based on this application.

This isn't a snippet, it's a CLI application :)

beccasaurus:
@munkhuushmgl Based on your testing out this application, I'm +1 having this flag for choosing language. It's one of the very first things you found yourself wanting to do, so it'll be important for a good chunk of our users, we shouldn't hard-code this to English. I recommend that we also update the Python and Node.js applications to support the same flag (in the future)

Note: it's not important to me that this PR goes in now, we could also wait to submit a PR with a fully refactored sample ready to be used in a detailed tutorial.

speech/cloud-client/src/main/java/com/example/speech/InfiniteStreamRecognize.java
```diff
@@ -104,75 +123,72 @@ public void run() {
...
       clientStream = client.streamingRecognizeCallable().splitCall(responseObserver);
 
+
       RecognitionConfig recognitionConfig =
-          RecognitionConfig.newBuilder()
-              .setEncoding(RecognitionConfig.AudioEncoding.LINEAR16)
-              .setLanguageCode("en-US")
-              .setSampleRateHertz(16000)
-              .build();
+              RecognitionConfig.newBuilder()
```
beccasaurus: Hey, why did these lines change? Was an automatic formatter run? Could we remove all whitespace changes from this PR so the diff just shows the behavioral differences? :)

speech/cloud-client/src/main/java/com/example/speech/InfiniteStreamRecognize.java
```diff
@@ -66,15 +69,31 @@
   private static ByteString tempByteString;
 
   public static void main(String... args) {
+    InfiniteStreamRecognizeOptions options = InfiniteStreamRecognizeOptions.fromFlags(args);
+    if (options == null) {
+      // Could not parse.
```
beccasaurus: How about printing something instead of the comment?

Print error "Failed to parse options"

Otherwise someone might run the sample + it will exit without a message + they won't know what's going on

speech/cloud-client/src/main/java/com/example/speech/InfiniteStreamRecognize.java
```diff
@@ -66,15 +69,31 @@
...
+
     try {
-      infiniteStreamingRecognize();
+      infiniteStreamingRecognize(options.langCode);
     } catch (Exception e) {
       System.out.println("Exception caught: " + e);
     }
   }
 
+  public static String convertMillisToDate(double milliSeconds, DecimalFormat format) {
```
beccasaurus: Let's move the `DecimalFormat` setup into this method.

This method is only called once and there is no need to customize the `format`

```java
public static String convertMillisToDate(double milliSeconds) {
    long millis = (long) milliSeconds;
    DecimalFormat format = new DecimalFormat();
    format.setMinimumIntegerDigits(2)
    return String.format("%s:%s /",
            format.format(TimeUnit.MILLISECONDS.toMinutes(millis)),
            format.format(TimeUnit.MILLISECONDS.toSeconds(millis)
                    - TimeUnit.MINUTES.toSeconds(TimeUnit.MILLISECONDS.toMinutes(millis)))
    );
}
```

speech/cloud-client/src/main/java/com/example/speech/InfiniteStreamRecognize.java
```diff
@@ -253,13 +270,12 @@ public void onError(Throwable t) {}
                   bridgingOffset = finalRequestEndTime;
                 }
                 int chunksFromMS = (int) Math.floor((finalRequestEndTime
-                                                - bridgingOffset) / chunkTime);
```
beccasaurus: Can these whitespace changes all be removed from the diff?

A follow-up can be sent with reformatting _if reformatting is required_.

PR #1507, opened 2019-07-10
Add flag for UrlFetch

averikitsch:


appengine-java8/urlfetch/src/main/webapp/WEB-INF/appengine-web.xml
```diff
@@ -2,4 +2,7 @@
 <appengine-web-app xmlns="http://appengine.google.com/ns/1.0">
   <runtime>java8</runtime>
   <threadsafe>true</threadsafe>
+  <!-- Enable URL Fetch service for networking classes -->
```
lesv: Enable old appengine Java 7 behavior for Networking.

PR #1509, opened 2019-07-12
BigQuery Storage: update sample application to show new API options.

shollyman:
Adds breadcrumbs for sharding and data format.  Doesn't need to deal with row count change since that's handled directly by the library in java via the stream resumption interface.


shollyman:
@lesv That's from the coveo code formatter, so I expect it to adhere to the style guide.  It seems correct per 4.5.1 (2), breaking after the assignment symbol.

lesv:
SGTM - it looks odd to me, but in time, I'll probably come to love the idea.

PR #1512, opened 2019-07-15
Add back missing break statement

nnegrey:
Looks like this line accidentally got removed awhile back. Putting it back in so the program properly terminates.


nnegrey:
No, looking through the whole commit history of the PR. It changed over time in the PR.

PR #1522, opened 2019-07-23
forgot to change default val to param

munkhuushmgl:
small fix


lesv:
@dpebot merge when green

PR #1523, opened 2019-07-24
Update Firestore Quickstart to use exec:exec

BenWhitehead:
When exec-maven-plugin runs the `exec:java` goal, it executes the java
program in the maven process (there currently isn't an option to fork
instead of running it in process). As a side effect of this, shutdown
hooks are not run when the program has completed running. In the case
of our Quickstart that means the daemon thread for grpc and gax do not
shutdown when the program has completed running, resulting in maven
listing all the threads that are still running and showing a stacktrace.

This change uses `exec:exec` and manually builds the java command to be
run, thereby forcing the program to be ran in a forked process and able
to use the normal shutdown process.

Attached are a logs showing the following
* Errors printed by maven when using `exec:java`
* A thread dump of the maven process when using `exec:java`
* A log without errors after changing to `exec:exec`

[exec_java.log](https://github.com/GoogleCloudPlatform/java-docs-samples/files/3427694/exec_java.log)
[exec_java_thread_dump.log](https://github.com/GoogleCloudPlatform/java-docs-samples/files/3427693/exec_java_thread_dump.log)
[exec_exec.log](https://github.com/GoogleCloudPlatform/java-docs-samples/files/3427692/exec_exec.log)

### DIFF
--- a/firestore/pom.xml
+++ b/firestore/pom.xml
@@ -37,6 +37,9 @@
   <properties>
     <maven.compiler.target>1.8</maven.compiler.target>
     <maven.compiler.source>1.8</maven.compiler.source>
+    <!-- Define a default empty value property which can be overridden on the command line if
+         necessary by passing `-Dfirestore.project.id="some-id"` when running maven exec -->
+    <firestore.project.id/>
   </properties>
 
   <dependencies>
@@ -57,4 +60,30 @@
       <scope>test</scope>
     </dependency>
   </dependencies>
+
+  <build>
+    <plugins>
+      <plugin>
+        <groupId>org.codehaus.mojo</groupId>
+        <artifactId>exec-maven-plugin</artifactId>
+        <version>1.6.0</version>
+        <executions>
+          <execution>
+            <goals>
+              <goal>exec</goal>
+            </goals>
+          </execution>
+        </executions>
+        <configuration>
+          <executable>java</executable>
+          <arguments>
+            <argument>-classpath</argument>
+            <classpath/>
+            <argument>com.example.firestore.Quickstart</argument>
+            <argument>${firestore.project.id}</argument>
+          </arguments>
+        </configuration>
+      </plugin>
+    </plugins>
+  </build>
 </project>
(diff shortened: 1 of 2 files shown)


firestore/pom.xml
```diff
@@ -37,6 +37,9 @@
   <properties>
     <maven.compiler.target>1.8</maven.compiler.target>
     <maven.compiler.source>1.8</maven.compiler.source>
+    <!-- Define a default empty value property which can be overridden on the command line if
+         necessary by passing `-Dfirestore.project.id="some-id"` when running maven exec -->
+    <firestore.project.id/>
```
kurtisvg: Rather than an empty cell, how about a default value that the user can change in the pom (if they prefer) 
```suggestion
    <firestore.project.id>YOUR-PROJECT-ID</firestore.project.id>
```

BenWhitehead: By having it defined as an empty value we can also support the scenario it is run with `mvn exec:exec` and have the project id inherited from the service credentials. From what I understand we're trying to nudge users in that direction, but we're not there yet across the board with docs.

I think having the empty tag is just as easy for a user to update if they're interested without any impact to them if they don't. If we don't want the short tag, I can expand that.

lesv:
@kurtisvg @nnegrey @averikitsch @dzlier-gcp These comments on exec:exec apply to more than just Firebase.  We should be looking for this during reviews (and fix other samples)

firestore/pom.xml
kurtisvg: That makes sense to me. 

FWIW, I think we are moving both away from command line samples and using system properties to configure.

BenWhitehead: yeah, the system property is a bit annoying when dealing with the maven runners.

BenWhitehead:
@lesv Is there anything else I need to do from your perspective before I can merge this? (I'll rebase it to be up to date with master).

lesv:
@BenWhitehead No - I just wanted to call out exec:java to others.  I see this being used a lot.  I'm happy w/ what you are doing.

kurtisvg:
@lesv - I'll open an issue to reflect it in the sample format guidelines.

PR #1526, opened 2019-07-25
Saner Jetty native WebSocket usage

joakime:
+ Eliminate reliance on SimpleContainerScope (it is deprecated
  and has already been removed in Jetty 10).
+ Use HttpClient properly instead of SimpleContainerScope.
+ SslContextFactory belongs to HttpClient.
+ WebSocketClient uses HttpClient (recommended if you need/want SSL).
+ Use WebSocket Endpoint Annotations, not Adapter (this is also Jetty 10 compatible).
+ Use WebSocketCreator, not raw registered classes (this is also Jetty 10 compatible).

Signed-off-by: Joakim Erdfelt <joakim.erdfelt@gmail.com>

### DIFF
--- a/appengine-java8/websocket-jetty/src/main/java/com/example/flexible/websocket/jettynative/SendServlet.java
+++ b/appengine-java8/websocket-jetty/src/main/java/com/example/flexible/websocket/jettynative/SendServlet.java
@@ -26,6 +26,8 @@ import javax.servlet.annotation.WebServlet;
 import javax.servlet.http.HttpServlet;
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
+
+import org.eclipse.jetty.client.HttpClient;
 import org.eclipse.jetty.http.HttpStatus;
 import org.eclipse.jetty.util.ssl.SslContextFactory;
 import org.eclipse.jetty.websocket.api.Session;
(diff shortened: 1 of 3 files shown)


appengine-java8/websocket-jetty/src/main/java/com/example/flexible/websocket/jettynative/SendServlet.java
```diff
@@ -73,22 +77,31 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
...
+  private WebSocketClient createWebSocketClient() {
+    return new WebSocketClient(this.httpClient);
   }
 
   private void sendMessageOverWebSocket(String message) throws Exception {
+    if (!httpClient.isRunning()) {
+      try {
+        httpClient.start();
+      } catch (URISyntaxException e) {
+        e.printStackTrace();
```
kurtisvg: I know this is consistent with `webSocketClient`, but I don't think just printing the stack trace is the best idea.

joakime: Yup, I agree.
My personal opinion is to treat `.start()` exceptions as fatal and prevent further use.

averikitsch:
@kurtisvg This sample was written by @jabubake. I don't have a lot of background, if she cares to comment.

Issue #1524, opened 2019-07-24
Update Sample Guidelines to recommend using exec:exec


kurtisvg:
See #1523 for example and reasoning

lesv:
@gguuss FYI - since you appear to be solving this.

gguuss:
Putting some notes here for edges so far:

# Multiple command utilities

If you have multiple command utilities to be executed by the Maven Plugin, specify them as a profile:

```XML
  <profiles>
    <profile>
      <id>mqtt</id>
      <activation>
        <property>
          <name>mqtt</name>
        </property>
      </activation>
    </profile>
  </profiles>
```

# Required named arguments

Required named arguments are specified within the <configuration>...<arguments>... section of the profiles section when you're using multiple execution runners.

The following example shows one style of specifying a required parameter

```XML
            <configuration>
              <executable>java</executable>
              <arguments>
                <argument>-classpath</argument>
                <classpath/>
                <argument>com.example.cloud.iot.examples.MqttExample</argument>
                <...>
                <argument>-project_id=${project_id}</argument>
                <...>
              </arguments>
            </configuration>
```

This parameter can be passed to the program as just the variable for exec to handle, for example, `-Dproject_id=NOUN-VERB-NUMBER`.

# Optional named arguments 

Because you don't want to specify optional named arguments unless they are passed, specify an argument where the variable alone is passed to the app.  

The following example shows how to pass an optional parameter:

```
            <configuration>
              <executable>java</executable>
              <arguments>
                <argument>-classpath</argument>
                <classpath/>
                <argument>com.example.cloud.iot.examples.MqttExample</argument>
                <...>
                <argument>${cr}</argument>
                <...>
              </arguments>
            </configuration>
```

This parameter is be passed to the program with its named parameter, for example, `-Dcd=cloud_region=us-central1`.

gguuss:
* [Basic example POM.xml](https://github.com/GoogleCloudPlatform/java-docs-samples/blob/master/firestore/pom.xml)
* [Complex example POM.xml](https://github.com/GoogleCloudPlatform/java-docs-samples/blob/master/iot/api-client/manager/pom.xml)

kurtisvg:
Closing this as I believe we decided we no longer want to use this pattern.

PR #1527, opened 2019-07-25
Fixes classnames in inline documentation

gguuss:


iot/api-client/manager/src/main/java/com/example/cloud/iot/examples/DeviceRegistryExample.java
```diff
@@ -84,11 +84,19 @@
  * <code>
  * $ mvn clean compile assembly:single
  * $ mvn exec:java \
```
lesv: We've found that some libraries don't like exec:java as the shutdown code isn't called properly.  We are moving to exec:exec.

iot/api-client/manager/src/main/java/com/example/cloud/iot/examples/MqttExample.java
```diff
@@ -57,12 +57,14 @@
  *
  * <pre>
  *   $ mvn compile
- *   $ mvn exec:java -Dexec.mainClass="com.google.cloud.iot.examples.MqttExample" \
- *       -Dexec.args="-project_id=my-project-id \
- *       -registry_id=my-registry-id \
- *       -device_id=my-device-id \
- *       -private_key_file=/path/to/private_pkcs8 \
- *       -algorithm=RS256"
+ *   $ mvn exec:java \
```
lesv: dito

iot/api-client/manager/src/main/java/com/example/cloud/iot/examples/DeviceRegistryExample.java
gguuss: SGTM, will resolve once updated.

gguuss:
I'll merge once verifying all the commands work as expected.

gguuss:
Looks like a few updates will be necessary :)

gguuss:
@lesv I'm concerned that moving to `mvn exec:exec` is adding a lot of complexity to the sample. To translate 1 of the 3 command line runners (the simple one) to `exec:exec` has added 100 lines of XML and required careful revision to the run arguments because it does not support passing a complete `ARGV` to the main Java class and instead requires each individual parameter be passed as variables to `mvn exec`.

For this PR, I'll be skipping translating the other samples for IoT to this pattern as it will be too much code for a sensible PR.

lesv:
@gguuss I wasn't expecting you to do all of them.  I understood it was annoying and difficult.  #1524 was opened for this process.

gguuss:
> #1524 was opened for this process.

SGTM, I'll start keeping some notes :) I understand a few advantages of `exec:exec` it's just a little more verbose.

Issue #1531, opened 2019-08-03
Cloud Pub/Sub example could result in message loss


kamalaboulhosn:
## In which file did you encounter the issue? 
java-docs-samples/pubsub/cloud-client/src/main/java/com/example/pubsub/SubscriberExample.java

### Did you change the file? 
No.

## Describe the issue

The example presented does not exhibit best practices for using Cloud Pub/Sub. The message is acknowledged as soon as it is added to the BlockingQueue. However, since the queue is not persistent, a crash of the task would result in message loss for any messages added to the queue that were not yet processed. I understand that it is a simple example, but we should make sure we are demonstrating best practices for using Pub/Sub.

nnegrey:
@anguillanneuf, do you know who would be the right person to take a look at this?

kamalaboulhosn:
I took a stab at it in #1534.

anguillanneuf:
Thank you @kamalaboulhosn so much for reporting this.

anguillanneuf:
Closing now as the PR has been merged. 

Message loss can be dealt by a blocking queue with `nack` logic and a processing timeout. One of Kamal's commits shows exactly how to do that in the PR.

kamalaboulhosn:
@anguillanneuf You don't need a blocking queue to handle message loss. The only thing required to avoid message loss is that the ack needs to happen after all processing of the message is complete. In the example prior to this PR, ack was being called at a point that was potentially before the message had been processed after being added to the blocking queue.

anguillanneuf:
Thanks for calling this out!

PR #1534, opened 2019-08-05
Make Pub/Sub subscriber example simpler and more idiomatic

kamalaboulhosn:
Address #1531 by acking only after processing a message.

### DIFF
--- a/pubsub/cloud-client/src/main/java/com/example/pubsub/SubscriberExample.java
+++ b/pubsub/cloud-client/src/main/java/com/example/pubsub/SubscriberExample.java
@@ -24,48 +24,19 @@ import com.google.cloud.pubsub.v1.MessageReceiver;
 import com.google.cloud.pubsub.v1.Subscriber;
 import com.google.pubsub.v1.ProjectSubscriptionName;
 import com.google.pubsub.v1.PubsubMessage;
-import java.util.concurrent.BlockingQueue;
-import java.util.concurrent.LinkedBlockingDeque;
-import java.util.concurrent.TimeUnit;
 
 public class SubscriberExample {
   // use the default project id
   private static final String PROJECT_ID = ServiceOptions.getDefaultProjectId();
 
-  // The amount of time to wait to offer a message to the queue.
-  private static final long QUEUE_OFFER_TIMEOUT_SECONDS = 5;
-
-  private static final BlockingQueue<MessageAndConsumer> queue = new LinkedBlockingDeque<>();
-
-  // A POD object to hold a message and the consumer associated with it.
-  static class MessageAndConsumer {
-    PubsubMessage message;
-    AckReplyConsumer consumer;
-
-    MessageAndConsumer(PubsubMessage message, AckReplyConsumer consumer) {
-      this.message = message;
-      this.consumer = consumer;
-    }
-  }
-
   static class MessageReceiverExample implements MessageReceiver {
 
     @Override
     public void receiveMessage(PubsubMessage message, AckReplyConsumer consumer) {
-      // Attempt to add the message to the queue for processing. If we can't add it within
-      // QUEUE_OFFER_TIMEOUT_SECONDS, then nack it and allow it to be redelivered (to this client
-      // or to another one that might be running and have more capacity)
-      try {
-        if (!queue.offer(
-            new MessageAndConsumer(message, consumer),
-            QUEUE_OFFER_TIMEOUT_SECONDS,
-            TimeUnit.SECONDS)) {
-          consumer.nack();
-        }
-      } catch (InterruptedException e) {
-        System.out.println("Could not add message to queue: " + e);
-        consumer.nack();
-      }
+      System.out.println(
+          "Message Id: " + message.getMessageId() + " Data: " + message.getData().toStringUtf8());
+      // Ack only after all work for the message is complete.
+      consumer.ack();
     }
   }
 
@@ -81,18 +52,10 @@ public class SubscriberExample {
       subscriber =
           Subscriber.newBuilder(subscriptionName, new MessageReceiverExample()).build();
       subscriber.startAsync().awaitRunning();
-      // Continue to listen to queue
-      while (true) {
-        MessageAndConsumer messageAndConsumer = queue.take();
-        System.out.println("Message Id: " + messageAndConsumer.message.getMessageId());
-        System.out.println("Data: " + messageAndConsumer.message.getData().toStringUtf8());
-        // Ack only after all work for the message is complete.
-        messageAndConsumer.consumer.ack();
-      }
-    } finally {
-      if (subscriber != null) {
-        subscriber.stopAsync();
-      }
+      // Allow the subscriber to run indefinitely unless an unrecoverable error occurs.
+      subscriber.awaitTerminated();
+    } catch (IllegalStateException e) {
+      System.out.println("Subscriber unexpectedly stopped: " + e);
     }
   }
 }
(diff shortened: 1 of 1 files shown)


pubsub/cloud-client/src/main/java/com/example/pubsub/SubscriberExample.java
```diff
@@ -28,18 +28,29 @@
 import java.util.concurrent.LinkedBlockingDeque;
 
 public class SubscriberExample {
+  static class MessageAndConsumer {
```
kurtisvg: Please move this class lower so the static final variables remain at the top.

pubsub/cloud-client/src/main/java/com/example/pubsub/SubscriberExample.java
```diff
@@ -28,18 +28,29 @@
...
+      this.consumer = consumer;
+    }
+  }
 
   // use the default project id
   private static final String PROJECT_ID = ServiceOptions.getDefaultProjectId();
 
-  private static final BlockingQueue<PubsubMessage> messages = new LinkedBlockingDeque<>();
+  private static final BlockingQueue<MessageAndConsumer> messages = new LinkedBlockingDeque<>();
+
```
kurtisvg: nit: Remove extra line

pubsub/cloud-client/src/main/java/com/example/pubsub/SubscriberExample.java
```diff
@@ -28,18 +28,29 @@
...
 
   static class MessageReceiverExample implements MessageReceiver {
 
     @Override
     public void receiveMessage(PubsubMessage message, AckReplyConsumer consumer) {
-      messages.offer(message);
-      consumer.ack();
+      if (!messages.offer(new MessageAndConsumer(message, consumer))) {
```
kurtisvg: Can you add a comment to the user explaining the logic? (e.g., "Attempt to insert the message into the queue. If unable, signal that it has failed to process."  

Also, should we consider a timeout here? The cost to receive/`nack()` isn't free, which means this could cause some thrashing if any customers copy this example verbatim.

pubsub/cloud-client/src/main/java/com/example/pubsub/SubscriberExample.java
```diff
@@ -28,18 +28,29 @@
 import java.util.concurrent.LinkedBlockingDeque;
 
 public class SubscriberExample {
+  static class MessageAndConsumer {
+    PubsubMessage message;
+    AckReplyConsumer consumer;
+
+    MessageAndConsumer(PubsubMessage message, AckReplyConsumer consumer) {
+      this.message = message;
```
kurtisvg: This should probably use getters?

pubsub/cloud-client/src/main/java/com/example/pubsub/SubscriberExample.java
```diff
@@ -57,9 +68,10 @@ public static void main(String... args) throws Exception {
       subscriber.startAsync().awaitRunning();
       // Continue to listen to messages
       while (true) {
-        PubsubMessage message = messages.take();
-        System.out.println("Message Id: " + message.getMessageId());
-        System.out.println("Data: " + message.getData().toStringUtf8());
+        MessageAndConsumer messageAndConsumer = messages.take();
```
kurtisvg: nit: might be worth renaming `messages` to `queue` or something a little more correct

kurtisvg:
@anguillanneuf - FYI

pubsub/cloud-client/src/main/java/com/example/pubsub/SubscriberExample.java
kamalaboulhosn: Getters feels a little like overkill, doesn't it? This is just a POD object with these two fields. I don't think defining getters and using them is really going to increase readability here.

pubsub/cloud-client/src/main/java/com/example/pubsub/SubscriberExample.java
```diff
@@ -66,11 +81,12 @@ public static void main(String... args) throws Exception {
...
           Subscriber.newBuilder(subscriptionName, new MessageReceiverExample()).build();
       subscriber.startAsync().awaitRunning();
-      // Continue to listen to messages
+      // Continue to listen to queue
       while (true) {
-        MessageAndConsumer messageAndConsumer = messages.take();
+        MessageAndConsumer messageAndConsumer = queue.take();
         System.out.println("Message Id: " + messageAndConsumer.message.getMessageId());
         System.out.println("Data: " + messageAndConsumer.message.getData().toStringUtf8());
+        // Ack only once all work with the message is complete.
```
anguillanneuf: nit: Should it be "Ack only once [after] all work.."?

kamalaboulhosn: I think the "once" in there is confusing now that I read it, so I changed it.

anguillanneuf:
@kamalaboulhosn We should definitely promote best practices. Thank you so much for your guidance on how to do that. 

Given the complexity, I would like this example to go in our subscriber guide page here under ["Asynchronous Pull"](https://cloud.google.com/pubsub/docs/pull#asynchronous-pull), instead of the quickstart, because generally the more sophisticated cases and samples go to there while the quickstart is meant to show what the API does. 

I propose we keep the original code and its region tags, either add your code and wrap it in a new set of region tags, `pubsub_subscriber_async_pull`, in this repo, or update the code [here](https://github.com/googleapis/google-cloud-java/blob/4234dbc072b08bad13aca02bf5b0632af0061758/google-cloud-examples/src/main/java/com/google/cloud/examples/pubsub/snippets/SubscriberSnippets.java#L101) with your proposed change. 

Sound fair?

kamalaboulhosn:
@anguillanneuf Sorry, I'm not sure what the "region tags" are. I do think it's worth making sure the code in the quickstart does ack at the right point. This came up because of a [Stack Overflow question](https://stackoverflow.com/questions/57334469/can-async-subscriber-example-lose-messages) where someone noted the possibility of message loss. If we really want to make the quickstart code be more minimal, then I would say we should just remove the use of the blocking queue entirely and just print out and ack the message in the callback itself. This is probably what most users would do anyway. Asynchronously passing the message off to a synchronous queue would be the exception, not the norm.

pubsub/cloud-client/src/main/java/com/example/pubsub/SubscriberExample.java
kurtisvg: It's less about readability, and more about promoting best practices. In Java getters and setters are standard and idiomatic, and we want to make sure to reinforce those practices to our users (as some are likely beginners and may not be able to tell when to cut corners and when not)

kurtisvg:
> @anguillanneuf Sorry, I'm not sure what the "region tags" are. I do think it's worth making sure the code in the quickstart does ack at the right point. This came up because of a [Stack Overflow question](https://stackoverflow.com/questions/57334469/can-async-subscriber-example-lose-messages) where someone noted the possibility of message loss. If we really want to make the quickstart code be more minimal, then I would say we should just remove the use of the blocking queue entirely and just print out and ack the message in the callback itself. This is probably what most users would do anyway. Asynchronously passing the message off to a synchronous queue would be the exception, not the norm.

+1 to removing the queue all together for simplicity - I think that might streamline the sample if each messages is processed as it's received.

anguillanneuf:
Okay, I think we have a consensus. 1. simplify current quickstart example by removing the queue. 2. still show Kamal's example but in a different part of the docs. @kamalaboulhosn region tags refer to the `[START ...]` and `[END ..]` comments in the code.

kamalaboulhosn:
I have updated the example to simplify it now. The more I think about it, the less I think the blocking queue is useful as any example, even in the async pull part of the guide. One really shouldn't have to offload to another data structure in the way this example used to do it except in very specific cases that the vast majority of our users shouldn't run into. It's probably better if we don't put it in our examples at all because people will start to think it is the right way to do things when it is really just unnecessary complication.

anguillanneuf:
LGTM!

PR #1536, opened 2019-08-05
Update Healthcare API DICOMweb samples and tests

noerog:
Add retrieve instance and rendered.

Rename the test file to "DicomWebTests."

For retrieval, download the output to a file rather than printing binary data to the console. Add relevant tests.


noerog:
What's the reason for all of the 403s?

PR #1541, opened 2019-08-09
Pub/Sub: new Pub/Sub DataFlow doc sample

anguillanneuf:
This sample will be used in a new Pub/Sub docs page that shows how to stream data from Pub/Sub to Cloud Storage using DataFlow.


pubsub/streaming-analytics/README.md
```diff
@@ -0,0 +1,150 @@
...
+* [PubSubToGCS.java](src/main/java/com/examples/pubsub/streaming/PubSubToGCS.java)
+
+To run the example and write output files in the desired Cloud Storage location.
+
+```bash
+mvn compile exec:java \
+  -Dexec.mainClass=com.examples.pubsub.streaming.PubSubToGCS \
+  -Dexec.cleanupDaemonThreads=false \
+  -Dexec.args="\
+    --windowSize=2 \
```
davidcavazos: Can this be omitted so we use the default value?

pubsub/streaming-analytics/README.md
```diff
@@ -0,0 +1,150 @@
...
+    --inputTopic=projects/${PROJECT}/topics/cron-topic \
+    --output=gs://${BUCKET}/output- \
```
davidcavazos: I would also go ahead and place the outputs inside a directory so it doesn't accidentally overwrite any files, and just keep it more organized.

```
--output=gs://$BUCKET/samples/dataflow-pubsub/output-
```

... or something like that

pubsub/streaming-analytics/README.md
```diff
@@ -0,0 +1,150 @@
...
+    gcloud scheduler jobs delete my-job
+    ```
+
+1. `Ctrl + C` to stop the program in your terminal.
+
+1. Cancel the DataFlow job in [GCP Console Dataflow page]. Stop the pipeline without draining it. 
+
+1. Delete the topic. 
+   ```bash
+   gcloud pubsub topics delete cron-topic
```
davidcavazos: You also should delete the subscription(s).

```
# Delete all the subscriptions associated to the topic
gcloud pubsub topics list-subscriptions cron-topic | grep $PROJECT | xargs gcloud pubsub subscriptions delete
```

You can probably find a simpler way to do it.

pubsub/streaming-analytics/src/main/java/com/examples/pubsub/streaming/PubSubToGCS.java
```diff
@@ -0,0 +1,83 @@
...
+
+    Pipeline pipeline = Pipeline.create(options);
+
+    pipeline
+      /* 1) Read string messages from Pub/Sub. */
+      .apply("Read PubSub Events", PubsubIO.readStrings().fromTopic(options.getInputTopic()))
+      /* 2) Window the messages into fixed minute intervals. */
+      .apply(Window.into(FixedWindows.of(Duration.standardMinutes(options.getWindowSize()))))
+      /* 3) Output the windowed files to GCS with a default value of 1 for numShards. */
+      .apply("Write Files to GCS", new WriteOneFilePerWindow(options.getOutput(), 1));
```
davidcavazos: How about creating a new variable/constant at the beginning of the function for `numShards`, with a comment explaining what it is, and then use that in here instead of the hardcoded `1`. It would make it easier to read.

pubsub/streaming-analytics/src/main/java/com/examples/pubsub/streaming/PubSubToGCS.java
```diff
@@ -0,0 +1,83 @@
...
+
+    // Execute the pipeline and return the result.
+    PipelineResult result = pipeline.run();
```
davidcavazos: Since the `PipelineResult` is not used, this can be simplified:
```
pipeline.run().waitUntilFinish();
```

pubsub/streaming-analytics/src/main/java/com/examples/pubsub/streaming/PubSubToGCS.java
```diff
@@ -0,0 +1,83 @@
...
+      .withValidation()
+      .as(PubSubToGCSOptions.class);
+
+    options.setStreaming(true);
+
+    Pipeline pipeline = Pipeline.create(options);
+
+    pipeline
+      /* 1) Read string messages from Pub/Sub. */
+      .apply("Read PubSub Events", PubsubIO.readStrings().fromTopic(options.getInputTopic()))
```
davidcavazos: How about?
```
.apply("Read Pub/Sub messages", ...
```

pubsub/streaming-analytics/src/main/java/com/examples/pubsub/streaming/PubSubToGCS.java
```diff
@@ -0,0 +1,83 @@
...
+      .fromArgs(args)
+      .withValidation()
+      .as(PubSubToGCSOptions.class);
+
+    options.setStreaming(true);
+
+    Pipeline pipeline = Pipeline.create(options);
+
+    pipeline
+      /* 1) Read string messages from Pub/Sub. */
```
davidcavazos: Can we do single line comments?
```
// 1) Read string messages from a Pub/Sub topic.
```

pubsub/streaming-analytics/src/main/java/com/examples/pubsub/streaming/PubSubToGCS.java
```diff
@@ -0,0 +1,83 @@
...
+      .apply("Read PubSub Events", PubsubIO.readStrings().fromTopic(options.getInputTopic()))
+      /* 2) Window the messages into fixed minute intervals. */
```
davidcavazos:
```
// 2) Window the messages into fixed-sized intervals.
```

pubsub/streaming-analytics/src/main/java/com/examples/pubsub/streaming/PubSubToGCS.java
```diff
@@ -0,0 +1,83 @@
...
+      .apply(Window.into(FixedWindows.of(Duration.standardMinutes(options.getWindowSize()))))
+      /* 3) Output the windowed files to GCS with a default value of 1 for numShards. */
```
davidcavazos:
```
// 3) Write one file to GCS for every window of messages.
```

pubsub/streaming-analytics/src/main/java/com/examples/pubsub/streaming/PubSubToGCS.java
davidcavazos:
```
.apply("Write windows to GCS", ...
```

pubsub/streaming-analytics/src/main/java/com/examples/pubsub/streaming/PubSubToGCS.java
```diff
@@ -0,0 +1,83 @@
...
+      .apply("Write Files to GCS", new WriteOneFilePerWindow(options.getOutput(), 1));
+
+    // Execute the pipeline and return the result.
```
davidcavazos:
```
// Execute the pipeline and wait until it finishes running.
```

pubsub/streaming-analytics/pom.xml
```diff
@@ -0,0 +1,179 @@
...
+      <groupId>org.slf4j</groupId>
+      <artifactId>slf4j-api</artifactId>
+      <version>${slf4j.version}</version>
+    </dependency>
+    <dependency>
+      <groupId>org.slf4j</groupId>
+      <artifactId>slf4j-jdk14</artifactId>
+      <version>${slf4j.version}</version>
+    </dependency>
+      <dependency>
```
davidcavazos: Lets move the `beam-sdks-java-io-google-cloud-platform ` dependency to after the `beam-runners-google-cloud-dataflow-java` dependency. Also make sure it's correctly indented. And for the version, please use `${beam.version}` instead of hardcoding `2.14.0`.

pubsub/streaming-analytics/pom.xml
```diff
@@ -0,0 +1,179 @@
...
+    <maven.compiler.target>1.8</maven.compiler.target>
+    <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+
+    <beam.version>2.14.0</beam.version>
+
+    <maven-compiler-plugin.version>3.8.1</maven-compiler-plugin.version>
+    <maven-exec-plugin.version>1.6.0</maven-exec-plugin.version>
+    <maven-jar-plugin.version>3.1.2</maven-jar-plugin.version>
+    <maven-shade-plugin.version>3.2.1</maven-shade-plugin.version>
+    <pubsub.version>v1-rev20181105-1.27.0</pubsub.version>
```
davidcavazos: Is this really needed? Are you using this somewhere or is it being used by any dependency?

These versions are kind of like variables, that are used below where the dependencies are actually specified.

pubsub/streaming-analytics/pom.xml
```diff
@@ -0,0 +1,179 @@
...
+    <slf4j.version>1.7.26</slf4j.version>
```
davidcavazos: You might want to bump this to `1.7.28`?

pubsub/streaming-analytics/pom.xml
```diff
@@ -0,0 +1,179 @@
...
+  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  See the License for the specific language governing permissions and
+  limitations under the License.
+-->
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+
+  <groupId>com.example</groupId>
```
davidcavazos: This doesn't *have to* match anything, but it would be nice for it to match the package name `com.examples.pubsub.streaming.PubSubToGCS`, notice it being **`com.examples`** instead of `com.example`

pubsub/streaming-analytics/README.md
```diff
@@ -0,0 +1,150 @@
...
+1. Create a Cloud Storage bucket. Grant the Default Compute Engine Service Account Object Admin Access to your bucket.
+
+   ```bash
+   gsutil mb gs://your-gcs-bucket
+   
+   BUCKET=gs://your-gcs-bucket
+   PROJECT=${gcloud config get-value project}
+   PROJECT_NUMBER=$(gcloud projects describe ${PROJECT} --format="value(projectNumber)")
+   
+   gsutil iam ch \
```
davidcavazos: By the way, on a new project by following these instructions, I was able to run everything without having to do this `gsutil iam ch` both in the DirectRunner as well as in DataflowRunner.

pubsub/streaming-analytics/README.md
```diff
@@ -0,0 +1,150 @@
...
+```bash
+mvn compile exec:java \
+  -Dexec.mainClass=com.examples.pubsub.streaming.PubSubToGCS \
+  -Dexec.cleanupDaemonThreads=false \
+  -Dexec.args="\
+    --windowSize=2 \
+    --inputTopic=projects/${PROJECT}/topics/cron-topic \
+    --output=gs://${BUCKET}/output- \
+    --runner=DataflowRunner"
+```
```
davidcavazos: I would also add a note:

> **NOTE**: You can also run locally with the DirectRunner if you don't specify the `--runner` option.

pubsub/streaming-analytics/README.md
anguillanneuf: Luckily when we cancel the Dataflow job, it deletes this subscription. :)

pubsub/streaming-analytics/README.md
anguillanneuf: What about for an existing project that was created a long time ago? I'd like to keep this as optional in case people run into issues like I did.

pubsub/streaming-analytics/README.md
anguillanneuf: I personally prefer to have it here.

pubsub/streaming-analytics/pom.xml
anguillanneuf: I wasn't using it, removing.

pubsub/streaming-analytics/README.md
anguillanneuf: I removed it. Thanks for verifying!

pubsub/streaming-analytics/README.md
```diff
@@ -0,0 +1,162 @@
...
+   ```
+
+## Streaming Analytics
+
+### Google Cloud Pub/Sub to Google Cloud Storage
+
+* [PubSubToGCS.java](src/main/java/com/examples/pubsub/streaming/PubSubToGCS.java)
+
+To run the example in Cloud Dataflow and write output files in the desired Cloud Storage location.
+
```
davidcavazos: Please add a short explanation of what the sample does. Something like:

The following example will run a streaming pipeline. It will read messages from a Pub/Sub topic, then window them into fixed-sized intervals, and write one file per window into a GCS location.

* `--project`: sets the Google Cloud project ID to run the pipeline on
* `--inputTopic`: sets the input Pub/Sub topic to read messages from
* `--output`: sets the output GCS path prefix to write files to
* `--runner` *[optional]*: specifies the runner to run the pipeline, defaults to "DirectRunner"
* `--windowSize` *[optional]*: specifies the window size in minutes, defaults to 1

anguillanneuf: Good call! I improved the cleanup step to match runner selection.

PR #1546, opened 2019-08-14
Adds closing includecode block on MQTT snippet

gguuss:


gguuss:
Yes, there was an issue reported in the docs that the closing braces don't match :)

PR #1548, opened 2019-08-15
DLP tests failing while job running - expand timeout from 1 minute to 2.

dzlier-gcp:


dzlier-gcp:
Oh I added the messages to see what was failing when the PR ran the tests, because it doesn't behave the same way when I run it locally. But then the tests passed |:l. But I was trying to see what the error was in order to identify a strategy here. I guess we'll have the messages the next time they fail.

lesv:
Not sure why this isn't closed, and why it's failing.

dzlier-gcp:
Updating from master, I'll see what state it's in.

PR #1549, opened 2019-08-15
Feat: Adds functions_helloworld_background

grant:
Adds bg functions snippet
https://devrel.corp.google.com/samples/82


ace-n:
N.B: this may be **blocked** on #1506 due to dependency issues (e.g. `gson`).

PR #1550, opened 2019-08-16
feat: Add GCF Java Tips

grant:
Adds GCF Tips snippets:
https://cloud.google.com/functions/docs/bestpractices/tips


grant:
Looks like I need to fix some linting issues:

```
[ERROR] src/main/java/Tips.java:[42] (indentation) Indentation: 'method def' child has incorrect indentation level 6, expected level should be 4.
[ERROR] src/main/java/Tips.java:[43] (indentation) Indentation: 'method def' child has incorrect indentation level 6, expected level should be 4.
[ERROR] src/main/java/Tips.java:[47] (indentation) Indentation: 'method def' child has incorrect indentation level 6, expected level should be 4.
[ERROR] src/main/java/Tips.java:[48] (indentation) Indentation: 'method def' child has incorrect indentation level 6, expected level should be 4.
[ERROR] src/main/java/Tips.java:[63,15] (naming) MethodName: Method name 'ScopesDemo' must match pattern '^[a-z][a-z0-9][a-zA-Z0-9_]*$'.
[ERROR] src/main/java/Tips.java:[74,15] (naming) MethodName: Method name 'TipsRetry' must match pattern '^[a-z][a-z0-9][a-zA-Z0-9_]*$'.
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-checkstyle-plugin:3.0.0:check (default) on project functions-snippets: You have 6 Checkstyle violations. -> [Help 1]
[ERROR]
[ERROR] To see the full stack trace of the errors, re-run Maven with the -e switch.
[ERROR] Re-run Maven using the -X switch to enable full debug logging.
[ERROR]
[ERROR] For more information about the errors and possible solutions, please read the following articles:
[ERROR] [Help 1] http://cwiki.apache.org/confluence/display/MAVEN/MojoFailureException
```

functions/snippets/src/main/java/Tips.java
```diff
@@ -0,0 +1,91 @@
...
+  private static int heavyComputation() {
+      List<Integer> numbers = Arrays.asList(1, 2, 3, 4, 5, 6, 7, 8, 9);
+      return numbers.stream().reduce((t, x) -> t + x).get();
+  }
+
+  // [START functions_tips_scopes]
+  // [START run_tips_global_scope]
+  // Global (instance-wide) scope
+  // This computation runs at instance cold-start
+  public static final int INSTANCE_VAR = heavyComputation();
```
kurtisvg: Class variables should be declared at the top of the class.

functions/snippets/src/main/java/Tips.java
```diff
@@ -0,0 +1,91 @@
...
+
+  /**
```
kurtisvg: You don't need use java-doc since there won't be any docs generated for this snippet - just use a regular comment for brevity

functions/snippets/src/main/java/Tips.java
```diff
@@ -0,0 +1,91 @@
...
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+
+public class Tips {
+
+  // Create a client with some reasonable defaults. This client can be reused for multiple requests.
+  private static HttpClient client =
+      HttpClient.newBuilder().connectTimeout(Duration.ofSeconds(10)).build();
+
+  // [END functions_tips_scopes]
```
kurtisvg: Excluding these functions breaks one of the goals of the new sample format, in that the sample isn't copy-paste-runnable. Can you leave them in (but maybe move them towards the bottom of the class since they aren't important?)

functions/snippets/src/main/java/Tips.java
```diff
@@ -0,0 +1,91 @@
...
+  // [END run_tips_global_scope]
+  private static int lightComputation() {
+      List<Integer> numbers = Arrays.asList(1, 2, 3, 4, 5, 6, 7, 8, 9);
+      return numbers.stream().reduce((t, x) -> t + x).get();
+  }
+
+  private static int heavyComputation() {
```
kurtisvg: These functions are exactly the same - is this intended?

functions/snippets/src/main/java/Tips.java
kurtisvg: Also, this might not be the way we want to showcase this. Normally, global variables aren't used with servlets due to issues with thread-safety. This example is probably okay because it's a final variable, but we probably either want to add a warning or (better yet) update the example to us a `ServletContextListener` for creating start up resources.

functions/snippets/src/main/java/Tips.java
```diff
@@ -0,0 +1,91 @@
...
+    // Send the TipsRetry using the client
+    HttpResponse<String> getResponse = client.send(getRequest, BodyHandlers.ofString());
+
+    // Write the results to the output:
+    PrintWriter writer = response.getWriter();
+    writer.write(String.format("Received code '%s' from url '%s'.", getResponse.statusCode(), url));
+  }
+}
+
+// [END functions_tips_scopes]
```
kurtisvg: These region tags don't match up

functions/snippets/src/main/java/Tips.java
```diff
@@ -0,0 +1,91 @@
...
+    // Per-function scope
+    // This computation runs every time this function is called
+    final int functionVar = lightComputation();
+    PrintWriter writer = response.getWriter();
+    writer.write(String.format("Per instance: %d, per function: %d", INSTANCE_VAR, functionVar));
+  }
+  // [END run_tips_global_scope]
+  // [END functions_tips_scopes]
+
+  public void TipsRetry(HttpServletRequest request, HttpServletResponse response)
```
kurtisvg: I'm not sure what this function is doing - it should be in a different file if it's for a different snippet

functions/snippets/src/main/java/Tips.java
```diff
@@ -0,0 +1,91 @@
...
+      throws IOException, InterruptedException {
+    // Create a GET TipsRetry to "http://example.com"
+    String url = "http://example.com";
+    HttpRequest getRequest = HttpRequest.newBuilder().uri(URI.create(url)).GET().build();
+
+    // Send the TipsRetry using the client
+    HttpResponse<String> getResponse = client.send(getRequest, BodyHandlers.ofString());
```
kurtisvg: What does TipsRetry mean? This seems like just a simple get request?

functions/snippets/src/main/java/Tips.java
grant: Sure.
I was just following the other samples that exclude them.
Can include them at the bottom.

functions/snippets/src/main/java/Tips.java
grant: Ah, fixed.

functions/snippets/src/main/java/Tips.java
grant: Moved to the very top. Not sure if we need the context listener.

functions/snippets/src/main/java/Tips.java
grant: Removed.

functions/snippets/src/main/java/Tips.java
```diff
@@ -0,0 +1,59 @@
...
+    // Per-function scope
+    // This computation runs every time this function is called
+    final int functionVar = lightComputation();
+    PrintWriter writer = response.getWriter();
+    writer.write(String.format("Per instance: %d, per function: %d", INSTANCE_VAR, functionVar));
+  }
+  // [END functions_tips_scopes]
+  // [END run_tips_global_scope]
+
+  private static int lightComputation() {
```
kurtisvg: These functions are still excluded, which means that this snippet isn't copy-paste-runnable (see the [Sample Format Guidelines](https://github.com/GoogleCloudPlatform/java-docs-samples/blob/master/SAMPLE_FORMAT.md#specific-goals)). Older samples might not be following the format, but we are trying to keep all new samples following the guidelines. 

Also, both functions are exactly the same, which makes it a bit confusing to understand when reading. Perhaps it would be better to replace this with something like this:

```java
   // This function represents a heavy computation, which is only called when the instance starts.
   private static int heavyComputation() {
    TimeUnit.SECONDS.sleep(3);
    return System.currentTimeMillis();
  }

   // This function represents a light computation, which might be called on every incoming request.
   private static int lightComputation() {
    TimeUnit.SECONDS.sleep(0);
    return System.currentTimeMillis();
  }
```

functions/snippets/src/main/java/Tips.java
kurtisvg: It should be updated to either use a ContextListener, or include a warning that global values in Servlets are not threadsafe, and a `ContextListener` is preferred for mutable variables. 

I'm not really sure if/how a ContextListener would work with functions, so this snippet might be a good example to showcase that pattern as well.

functions/snippets/src/main/java/Tips.java
grant: - These functions are excluded like they are in the Node sample.
- These functions aren't exactly the same. They are the same as Node.

https://github.com/GoogleCloudPlatform/nodejs-docs-samples/blob/4b2629e759eba9cf7d89aed3302bcf2795e1eecf/functions/tips/index.js#L18

What would you like us to do?

functions/snippets/src/main/java/Tips.java
grant: I don't really know how to proceed. Do you think that GCF will use multiple threads per Function invocation?

> This example is probably okay because it's a final variable

Since this is a `final` variable, isn't it safe to access across multiple threads?

functions/snippets/src/main/java/Tips.java
kurtisvg: > * These functions are excluded like they are in the Node sample.

I believe that the node repo has a similar policy regarding copy-paste-runnable (since it's borrowed from the overarching [Code Snippets Rubric](https://g3doc.corp.google.com/company/teams/cloud-devrel/dpe/samples/snippets.md?cl=head)). It looks like these samples probably predate this policy, but I don't know how the node repo is enforcing those rules.  

Currently, the java-docs-samples policy is that new samples should follow the guidelines, while minor updates to existing samples can be grandfathered in. Since these samples are new, they need to follow the guidelines and include these functions so that users can copy them, paste them, and run them without having to fill out these functions. 

> * These functions are not exactly the same. 

This is my bad, the version I was looking at shows them [both using addition](https://github.com/GoogleCloudPlatform/java-docs-samples/blob/70e8f4319a05ecb77c900650089ac834f62ccbc4/functions/snippets/src/main/java/Tips.java#L41-L49), but I can see at HEAD they are different. 

At the very least, this version of the sample should contain the comments I posted above. (My two cents is still that the intent of the sample is significantly more clear when using sleep, because the user can see that the difference between running those functions is non-trivial. However if you strongly prefer to remain consistent with the Node version, I'm satisfied with the comments clarifying above)

functions/snippets/src/main/java/Tips.java
kurtisvg: Most servlet environments will use multiple threads to handle requests, so

> Do you think that GCF will use multiple threads per Function invocation?

I don't know how GCF handles things, but it would be idiomatic to Java to remain threadsafe when using Servlets since other runtimes may use multiple threads. 

> Since this is a final variable, isn't it safe to access across multiple threads?

I should have been more clear - since this is a `final` and a primitive it's immutable, and thus thread-safe. However, we should either point out _why_ this is okay, or show the user how to handle these properly. If the user tried to do this with an object that isn't thread-safe (even if it's final), they'll run into issues.

grant:
Fixed with #1586 instead.

PR #1551, opened 2019-08-16
feat: add functions_env_vars

grant:
Adds `devrel/samples/1063`.


functions/snippets/src/main/java/Env.java
```diff
@@ -0,0 +1,35 @@
...
+ */
+
+// [START functions_env_vars]
+import java.io.IOException;
+import java.io.PrintWriter;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+
+public class Env {
+  public void env(HttpServletRequest request, HttpServletResponse response)
```
kurtisvg: Please add a comment to this function describing what it showcases.

functions/snippets/src/main/java/Env.java
```diff
@@ -0,0 +1,35 @@
...
+ * limitations under the License.
+ */
+
+// [START functions_env_vars]
+import java.io.IOException;
+import java.io.PrintWriter;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+
+public class Env {
```
kurtisvg:
```suggestion
public class EnviromentVariableExample {
```

functions/snippets/src/test/java/SnippetsTests.java
```diff
@@ -159,4 +159,10 @@ public void helloBackgroundTest() throws IOException {
     new HelloBackground().helloBackground(request, response);
     assertThat(responseOut.toString(), containsString("Hello John!"));    
   }
+
+  @Test
+  public void envTest() throws IOException {
+    new Env().env(request, response);
+    assertThat(responseOut.toString(), containsString(""));
```
kurtisvg: Checking for any empty string doesn't actually verify that this function works - any non-null string will contain and empty string.

grant: I'm not sure how to set an env var in a test and tried doing so.
Do you know how to set an env var?

kurtisvg: Can you stub it with Mockito? 

```java
when(System.getenv("FOO")).thenReturn("someTestValue");
```

functions/snippets/src/main/java/Env.java
grant: Done.

functions/snippets/src/main/java/Env.java
grant: This suggested name is too verbose and not in line with other class names. It is not common to add `Example` to all the other examples. Looking at the other 3 languages, I've changed the name to `EnvVars`.

functions/snippets/src/test/java/SnippetsTests.java
grant: OK. Thanks.

grant:
Not exactly sure what to do in terms of mocking `System.getenv`.

functions/snippets/src/test/java/SnippetsTests.java
kurtisvg: Ok, it looks like we'll have to add an additional dependency to do this: 
```xml
<dependency>
  <groupId>com.github.stefanbirkner</groupId>
  <artifactId>system-rules</artifactId>
  <version>1.19.0</version>
  <scope>test</scope>
</dependency>
```

```java
import org.junit.contrib.java.lang.system.EnvironmentVariables;

public void EnvironmentVariablesTest {
  @Rule
  public final EnvironmentVariables environmentVariables
    = new EnvironmentVariables();

  @Test
  public void setEnvironmentVariable() {
    environmentVariables.set("name", "value");
    assertEquals("value", System.getenv("name"));
  }
}
``` 

Source: https://stackoverflow.com/a/35393623/9414803

grant:
> Ok, it looks like we'll have to add an additional dependency to do this:
> 
> ```
> <dependency>
>   <groupId>com.github.stefanbirkner</groupId>
>   <artifactId>system-rules</artifactId>
>   <version>1.19.0</version>
>   <scope>test</scope>
> </dependency>
> ```
> 
> ```java
> import org.junit.contrib.java.lang.system.EnvironmentVariables;
> 
> public void EnvironmentVariablesTest {
>   @Rule
>   public final EnvironmentVariables environmentVariables
>     = new EnvironmentVariables();
> 
>   @Test
>   public void setEnvironmentVariable() {
>     environmentVariables.set("name", "value");
>     assertEquals("value", System.getenv("name"));
>   }
> }
> ```
> 
> Source: https://stackoverflow.com/a/35393623/9414803

Thanks for the tip and extensive suggestion to fix this. Tests are now logical and pass. (Getting some merge conflicts, but those should be trivial.)
PTAL.

functions/snippets/src/main/java/EnvVars.java
```diff
@@ -0,0 +1,41 @@
...
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+
+public class EnvVars {
+  /**
+   * Returns the environment variable "foo" set during function deployment.
+   * @param request The Servlet HTTP request.
+   * @param response The Servlet HTTP response.
+   * @throws IOException If there is an I/O writer exception.
+   */
```
kurtisvg: nit: This can be a regular comment (we don't need to use javadoc since no docs are generated from these samples)

grant: Done.

grant:
@kurtisvg Please advise as to how to fix the Java 11 test. I don't have permission to see the errors and debug.

functions/snippets/src/main/java/EnvVars.java
grant: Done.

kurtisvg:
@grant - You should be able to access the logs by clicking Details next to the test name, and then clicking on the failed test log. This repo uses resultstore, so it should be open to the public. If it's not please let me know so we can debug.

Here is what is says atm:

> ------------------------------------------------------------
> - testing functions/snippets
> ------------------------------------------------------------
> [ERROR] src/test/java/SnippetsTests.java:[52] (indentation) Indentation: '=' has incorrect indentation level 4, expected level should be 6.
> [ERROR] Failed to execute goal org.apache.maven.plugins:maven-checkstyle-plugin:3.0.0:check (default) on project functions-snippets: You have 1 Checkstyle violation. -> [Help 1]
> [ERROR]
> [ERROR] To see the full stack trace of the errors, re-run Maven with the -e switch.
> [ERROR] Re-run Maven using the -X switch to enable full debug logging.
> [ERROR]
> [ERROR] For more information about the errors and possible solutions, please read the following articles:
> [ERROR] [Help 1] http://cwiki.apache.org/confluence/display/MAVEN/MojoFailureException
>  Testing failed: Maven returned a non-zero exit code.
>

grant:
> @grant - You should be able to access the logs by clicking Details next to the test name, and then clicking on the failed test log. This repo uses resultstore, so it should be open to the public. If it's not please let me know so we can debug.

Oh, there's a bug where the `Invocation Log` tab shows an inaccurate state.

Screen Shot 2019-09-09 at 16 41 44

You just need to wait a few sections until the 15k lines of logs download and then you can see the error. A bit frustrating.

I sent feedback using the feedback tool.

PR #1553, opened 2019-08-16
Update KnowledgeBaseId and update rouge SA roles

nnegrey:
This is mostly just an FYI to all of you, but I've changed the rouge SA role to Project Viewer, in case someone complains. This should be the final time I have to do this fix (I hope).

Otherwise, if Dialogflow is failing again, it is likely that the Test Knowledge Base was deleted and needs to be re-created and these IDs have to be updated each time.


dzlier-gcp:
Thanks!

PR #1554, opened 2019-08-16
Change retrieved study output file to use .multipart extension and add Accept headers

noerog:


healthcare/v1beta1/src/main/java/snippets/healthcare/dicom/DicomWebRetrieveStudy.java
```diff
@@ -65,7 +65,10 @@ public static void dicomWebRetrieveStudy(String dicomStoreName, String studyId)
     // Execute the request and process the results.
     HttpResponse response = request.executeUnparsed();
 
-    String outputPath = "study.dcm";
+    // When specifying the output file, use an extension like ".multipart."
+    // Then, parse the downloaded multipart file to get each individual
+    // DICOM file.
+    String outputPath = "study.multipart";
```
lesv: I think your comment is probably correct english, but incorrect engineering.  Should it be `".multipart".` instead? L68 or L71 is wrong.

noerog: You're correct. Changed to ".multipart". to be clearer.

healthcare/v1beta1/src/main/java/snippets/healthcare/dicom/DicomWebRetrieveStudy.java
```diff
@@ -90,11 +93,14 @@ private static CloudHealthcare createClient() throws IOException {
         GoogleCredential.getApplicationDefault(HTTP_TRANSPORT, JSON_FACTORY)
             .createScoped(Collections.singleton(CloudHealthcareScopes.CLOUD_PLATFORM));
 
+    HttpHeaders headers = new HttpHeaders();
+    headers.set("X-GFE-SSL", "yes");
```
lesv: Probably should explain this.

lesv:
Note - Java 8 is failing.

healthcare/v1beta1/src/main/java/snippets/healthcare/dicom/DicomWebRetrieveStudy.java
noerog: Added an explanation.

noerog:
> Note - Java 8 is failing.

Whoops, typo.

healthcare/v1beta1/src/main/java/snippets/healthcare/dicom/DicomWebRetrieveStudy.java
lesv: where?

lesv: found it.

lesv: I was speaking about the X-GFE-SSL header as well.

lesv:
J8 failed last time, just updated, it doesn't appear that you fixed it.

healthcare/v1beta1/src/main/java/snippets/healthcare/dicom/DicomWebRetrieveStudy.java
noerog: That's in all of the samples that dzlier wrote. Should I add it to every sample as well?

noerog:
> J8 failed last time, just updated, it doesn't appear that you fixed it.

Yup, it was actually a new issue introduced. Fixed now.

healthcare/v1beta1/src/main/java/snippets/healthcare/dicom/DicomWebRetrieveStudy.java
lesv: Is a customer likely to see this w/o seeing dziller's code?  If so, you should explain it.  Frankly, I hadn't seen it before.

We can't explain things enough.  The standard I use is, if a user familiar with coding, the language, comes to this code, will they understand it, if not, we should comment.

noerog: I spoke to dzlier and he said that it's actually not necessary. I'll need to go through all of the other Healthcare samples and remove it wherever it's OK to, but for now I've removed it from this sample.

lesv: Cool

PR #1558, opened 2019-08-23
Update from kvgbot to Blunderbuss. 

kurtisvg:


lesv:
@dpebot merge when green

PR #1559, opened 2019-08-23
Add BigQueryKmsKey Dataflow sample

davidcavazos:
Creating a Dataflow sample to show how to use Customer-Managed Encryption Keys (CMEK) for the [Using customer-managed encryption keys](https://cloud.google.com/dataflow/docs/guides/customer-managed-encryption-keys) docs page.

Also updating the templates README for consistency.

### DIFF
--- a/dataflow/README.md
+++ b/dataflow/README.md
@@ -1,0 +1,117 @@
+# Getting started with Google Cloud Dataflow
+
+[![Open in Cloud Shell](http://gstatic.com/cloudssh/images/open-btn.svg)](https://console.cloud.google.com/cloudshell/editor)
+
+[Apache Beam](https://beam.apache.org/)
+is an open source, unified model for defining both batch and streaming data-parallel processing pipelines.
+This guides you through all the steps needed to run an Apache Beam pipeline in the
+[Google Cloud Dataflow](https://cloud.google.com/dataflow) runner.
+
+## Setting up your Google Cloud project
+
+The following instructions help you prepare your Google Cloud project.
+
+1. Install the [Cloud SDK](https://cloud.google.com/sdk/docs/).
+   > *Note:* This is not required in
+   > [Cloud Shell](https://console.cloud.google.com/cloudshell/editor)
+   > since it already has the Cloud SDK pre-installed.
+
+1. Create a new Google Cloud project via the
+   [*New Project* page](https://console.cloud.google.com/projectcreate),
+   or via the `gcloud` command line tool.
+
+   ```sh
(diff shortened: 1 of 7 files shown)


davidcavazos:
Hi @lesv, can you take a look at this whenever you have a chance? Thanks!

dataflow/README.md
```diff
@@ -0,0 +1,117 @@
...
+   export GOOGLE_APPLICATION_CREDENTIALS=path/to/your/credentials.json
+   ```
+
+## Setting up a Java development environment
+
+The following instructions help you prepare your development environment.
+
+1. Download and install
+   [OpenJDK](https://openjdk.java.net/install/index.html)
+   (alternatively, you can use the propietary
```
lesv: Let's talk about this tomorrow.

dataflow/encryption-keys/pom.xml
```diff
@@ -0,0 +1,176 @@
...
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+
+  <groupId>com.example</groupId>
+  <artifactId>dataflow-bigquery-kms-key</artifactId>
+  <version>1.0</version>
+
+  <properties>
+    <maven.compiler.source>1.8</maven.compiler.source>
+    <maven.compiler.target>1.8</maven.compiler.target>
```
lesv: Fine, but maybe should be Java 11 or 12

lesv: Beam's docs say J8, but Oracle no longer supports the public version.
https://www.oracle.com/technetwork/java/java-se-support-roadmap.html
Perhaps you want to point folks to https://adoptopenjdk.net/

mik-laj:
@davidcavazos  Apache Beam does not have full support for Java 11 [https://beam.apache.org/roadmap/java-sdk/]. The Polidea team is still working on it. Does Dataflow Runner fully support Jave 11?  I am not sure if we should use this version.

davidcavazos:
Hi @mik-laj, although Apache Beam SDK itself isn't in Java 11 yet, the package can still be used by Java 11 applications. I would go ahead and try to stick with the latest Java version while the Beam SDK catches up. This would mean not having to migrate from one version to another once the old version drops support (like what happened with Python 2), which we've seen can be a huge effort.

I haven't run into any issues by using Java 11, so I would still recommend using the latest version to try to keep up to date.

PR #1561, opened 2019-08-23
feat: Add functions_tips_scopes

grant:
[b/132734505](http://b/132734505)


grant:
Ah, this is actually a duplicate PR.

PR #1572, opened 2019-09-06
feat: Add functions_concepts_stateless

grant:
Adds `functions_concepts_stateless`.

devrel/samples/600

Similar to Node:
https://github.com/GoogleCloudPlatform/nodejs-docs-samples/blob/ded1994db4b952a75c6d3163ab5bb98e86d416d5/functions/concepts/index.js#L41-L59

### DIFF
--- a/functions/snippets/src/main/java/Concepts.java
+++ b/functions/snippets/src/main/java/Concepts.java
@@ -18,16 +18,17 @@
 
 import java.io.IOException;
 import java.io.PrintWriter;
+import java.util.concurrent.atomic.AtomicInteger;
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
 
 public class Concepts {
   
-  private int count = 0;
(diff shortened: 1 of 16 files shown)


functions/snippets/src/main/java/Concepts.java
```diff
@@ -0,0 +1,39 @@
...
+// [START functions_concepts_stateless]
+
+import java.io.IOException;
+import java.io.PrintWriter;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+
+public class Concepts {
+  
+  private int count = 0;
```
kurtisvg: This has the same issue as your other PR - this value isn't thread-safe and could cause issues with regular Servlet handling. 

Here, I think you want to use a `private final AtomicInteger count` to accomplish this safely.

grant: I believe I have used the correct thread-safe counter here now.

grant:
@kurtisvg Review again?

PR #1574, opened 2019-09-08
Functions Concepts Filesystem

grant:
Adds a Functions FS snippet.

I'm not sure exactly why we have this snippet or how it should work in Java. I've created a snippet here that lists all the Function snippet files. It uses `snippets/src/main/java/`, which is a bit ugly and fragile imo, but I don't know what else this snippet should do.

devrel/samples/597
b/132734862

### DIFF
--- a/functions/snippets/src/main/java/FileSystem.java
+++ b/functions/snippets/src/main/java/FileSystem.java
@@ -40,6 +40,8 @@ public class FileSystem {
       String files = String.join(", ", fileList);
       PrintWriter writer = response.getWriter();
       writer.write(String.format("Files: %s", files));
+    } catch (IOException e) {
+      e.printStackTrace();
     };
   }
 }
--- a/functions/snippets/src/main/java/FileSystem.java
+++ b/functions/snippets/src/main/java/FileSystem.java
@@ -16,8 +16,8 @@
 
 // [START functions_concepts_filesystem]
 import java.io.IOException;
-import java.io.PrintWriter;
 import java.io.File;
+import java.io.PrintWriter;
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
 
--- a/functions/snippets/src/main/java/FileSystem.java
+++ b/functions/snippets/src/main/java/FileSystem.java
@@ -15,11 +15,8 @@
  */
 
 // [START functions_concepts_filesystem]
-import com.google.gson.Gson;
-import com.google.gson.JsonObject;
 import java.io.IOException;
 import java.io.PrintWriter;
-import java.io.File;
 import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.util.List;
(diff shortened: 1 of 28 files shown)


functions/snippets/src/main/java/FileSystem.java
```diff
@@ -0,0 +1,47 @@
...
+
+public class FileSystem {
+
+  // Lists the files in the current directory.
+  public void listFiles(HttpServletRequest request, HttpServletResponse response)
+      throws IOException {
+
+    String currentDirectory = System.getProperty("user.dir");
+    Path path = Paths.get(currentDirectory, "snippets/src/main/java/");
+    try (Stream<Path> walk = Files.walk(path)) {
```
kurtisvg: nit: This should probably use a catch and log that an error has occurred in some way

functions/snippets/src/main/java/FileSystem.java
```diff
@@ -0,0 +1,47 @@
...
+      List<String> fileList = walk.filter(Files::isRegularFile)
+        .map(file -> file.toString()).collect(Collectors.toList());
+      String files = String.join(", ", fileList);
```
kurtisvg: nit: L40-42 can be outside of the try-catch

grant: We cannot move L40 outside the try block as it is defined within the try block.

functions/snippets/src/main/java/FileSystem.java
grant: Added catch to print the ST.

grant:
> We cannot move L40 outside the try block as it is defined within the try block.

@kurtisvg Please look at this PR again w.r.t. the comment about the try block.

functions/snippets/src/main/java/FileSystem.java
```diff
@@ -0,0 +1,49 @@
...
+import javax.servlet.http.HttpServletResponse;
+
+public class FileSystem {
+
+  // Lists the files in the current directory.
+  public void listFiles(HttpServletRequest request, HttpServletResponse response)
+      throws IOException {
+
+    String currentDirectory = System.getProperty("user.dir");
+    Path path = Paths.get(currentDirectory, "snippets/src/main/java/");
```
kurtisvg: Taking a second look at this, I think this will actually fail if the user tried to deploy it since the source won't be available. The Go sample uses the current working directory, so can you try this instead?

```java
  // Lists the files in the current directory.
  public void listFiles(HttpServletRequest request, HttpServletResponse response) {
     File cur = new File(".");
     File[] files = cur.listFiles();

     PrintWriter writer = response.getWriter();
     writer.write("Files: \n");
     for (File f : files) {
          writer.write("\t" + f + "\n");
     } 
  }
```
The test can probably just get away with checking for `"Files: "`(since cwd will change depending on environment)

grant: Good point. The previous implementation required that file structure.

I've changed the implementation to adhere to your suggestion.

PR #1577, opened 2019-09-11
feat: Add functions_log_stackdriver

grant:
Adds `functions_log_stackdriver` snippet.
- devrel/samples/1156
- b/132734861
---
Note: I'm not sure how to effectively test this method as there is no response.
Other languages do not seem to test this method.


functions/snippets/src/test/java/SnippetsTests.java
```diff
@@ -179,6 +179,15 @@ public void filesTest() throws IOException {
     assertThat(responseOut.toString(), containsString("Files:"));
   }
 
+  @Test
+  public void logEntry() throws IOException {
+    PubSubMessage message = new PubSubMessage();
+    message.data = "data";
+    message.messageId = "id";
+    new LogEntry().helloPubSub(message);
+    assertThat(responseOut.toString(), containsString(""));
```
kurtisvg: This test isn't checking anything

functions/snippets/src/main/java/LogEntry.java
```diff
@@ -0,0 +1,42 @@
...
+  public void helloPubSub(PubSubMessage message) throws IOException {
+    String data = new String(Base64.getDecoder().decode(message.data.getBytes("UTF-8")));
+    if (data.isEmpty()) {
+      data = "World";
+    }
+    LOGGER.info(String.format("Hello, %s", data));
+  }
+}
+
+class PubSubMessage {
```
kurtisvg: This should be a nested class inside LogEntry.

Additionally, there needs to be a comment explaining how this class is derived since this function doesn't match the servlet pattern. In a different PR, it was mentioned that they were using gson to Marshall the HttpRequest into the input class, but it's not clear to me how restrictive that is. Does this class need to have this exact signature to work?  What happens if one of the attributes is misspelled? Will GCF throw an error or just silently drop that attribute?

functions/snippets/src/main/java/LogEntry.java
```diff
@@ -0,0 +1,42 @@
...
+import java.util.Base64;
+import java.util.Map;
+import java.util.logging.Logger;
+
+public class LogEntry {
+
+  private static Logger LOGGER = Logger.getLogger(LogEntry.class.getName());
+
+  public void helloPubSub(PubSubMessage message) throws IOException {
+    String data = new String(Base64.getDecoder().decode(message.data.getBytes("UTF-8")));
```
kurtisvg: Can `message.data` be null?

grant: I've now added a check to ensure no NPEs.

functions/snippets/src/main/java/LogEntry.java
grant: Sure. I've moved to be an inner class.

I've added a comment to this class. It matches the exact default class provided in the eventprocessing invoker for GCF.
I do not know the behavior if the class is misconfigured.

functions/snippets/src/main/java/LogEntry.java
```diff
@@ -0,0 +1,47 @@
...
+      new String(Base64.getDecoder().decode(message.data.getBytes("UTF-8")));
+    if (data.isEmpty()) {
+      data = "World";
+    }
+    LOGGER.info(String.format("Hello, %s", data));
+  }
+
+  /**
+   * A Pub/Sub message.
+   */
```
kurtisvg: nit: Single line comments should be single lines, not block comments. 

However, consider revising to something like the following to give more helpful information to the user:
```
Make sure to include the PubSubMessage class in your function - the GCF environment will marshal the request data into it. 
```

functions/snippets/src/main/java/LogEntry.java
```diff
@@ -0,0 +1,47 @@
...
+  public static class PubSubMessage {
```
kurtisvg: I don't think you want it static if you are instantiating it. 
```suggestion
  public class PubSubMessage {
```

functions/snippets/src/test/java/SnippetsTests.java
```diff
@@ -179,6 +179,15 @@ public void filesTest() throws IOException {
     assertThat(responseOut.toString(), containsString("Files:"));
   }
 
+  @Test
+  public void logEntry() throws IOException {
+    LogEntry.PubSubMessage message = new LogEntry.PubSubMessage();
+    message.data = "data";
```
kurtisvg: It's probably best to use GSON to test this, since that seems to be what they are using in the functions environment. This will make the test more like the actual environment. 

Something like:
```java
PubSubMessage message = gson.("{\"data\":\"data\",\"messageId\":\"id\"}", PubSubMessage.class)
new LogEntry().helloPubSub(message);
```
If GSON isn't a dependency already, you can add it and scope to `test`. 

Additionally, according to the [gson user guide](https://sites.google.com/site/gson/gson-user-guide), it looks like GSON will just setting missing fields to null.

functions/snippets/src/main/java/LogEntry.java
grant: Done.

functions/snippets/src/test/java/SnippetsTests.java
```diff
@@ -181,11 +185,9 @@ public void filesTest() throws IOException {
 
   @Test
   public void logEntry() throws IOException {
-    LogEntry.PubSubMessage message = new LogEntry.PubSubMessage();
-    message.data = "data";
-    message.messageId = "id";
+    LogEntry.PubSubMessage message = gson.fromJson("{\"data\":\"data\",\"messageId\":\"id\"}", LogEntry.PubSubMessage.class);
     new LogEntry().helloPubSub(message);
-    assertThat(responseOut.toString(), containsString("Hello, data"));
+    // assertThat(responseOut.toString(), containsString("Hello, data"));
```
kurtisvg:
```suggestion
    assertThat(responseOut.toString(), containsString("Hello, data"));
```

grant:
Java 11 is still seeing the error:

```
[ERROR] Failures:
[ERROR]   SnippetsTests.logEntry:190
Expected: a string containing "Hello, data"
     but: was ""
```

kurtisvg:
@grant is this still WIP or can we close?

grant:
- I tested this function on GCF Java 8. Works fine.
- I made the Kokoro test pass.
This PR was approved by kvg on Sep 23 (thanks for the patience).
Merging! 🎈

PR #1578, opened 2019-09-11
Updating IAM service account snippets

mdejong3:
I'm working on the following two docs:
https://cloud.google.com/iam/docs/creating-managing-service-accounts
https://cloud.google.com/iam/docs/creating-managing-service-account-keys

The Java code snippets for these docs don't always capture the imports and variables necessary to run the code. To fix this, I added region tags around the imports and variables necessary for each snippet.


iam/api-client/src/main/java/com/google/iam/snippets/ServiceAccountKeys.java
```diff
@@ -1,4 +1,4 @@
-/* Copyright 2018 Google LLC
+/* Copyright 2019 Google LLC
```
lesv: We don't change the copyrights

iam/api-client/src/main/java/com/google/iam/snippets/ServiceAccountKeys.java
```diff
@@ -15,47 +15,54 @@
...
 import com.google.api.services.iam.v1.IamScopes;
-import com.google.api.services.iam.v1.model.CreateServiceAccountKeyRequest;
 import com.google.api.services.iam.v1.model.ServiceAccountKey;
 import java.io.IOException;
 import java.util.Collections;
+// [END iam_create_key]
+// [END iam_list_keys]
+// [END iam_delete_key]
+
+// [START iam_create_key]
```
lesv: [Java Style Guide](https://google.github.io/styleguide/javaguide.html#s3.3-import-statements) Import statements have a required order.

iam/api-client/src/main/java/com/google/iam/snippets/ServiceAccountKeys.java
```diff
@@ -15,47 +15,54 @@
...
-        new Iam.Builder(
-                GoogleNetHttpTransport.newTrustedTransport(),
-                JacksonFactory.getDefaultInstance(),
-                credential)
-            .setApplicationName("service-account-keys")
-            .build();
+    GoogleCredential credential = GoogleCredential.getApplicationDefault()
+        .createScoped(Collections.singleton(IamScopes.CLOUD_PLATFORM));
+
+    service = new Iam.Builder(GoogleNetHttpTransport.newTrustedTransport(), JacksonFactory.getDefaultInstance(),
```
lesv: I don't think that matches our [line breaking rules](https://google.github.io/styleguide/javaguide.html#s4.5.1-line-wrapping-where-to-break)

iam/api-client/src/main/java/com/google/iam/snippets/ServiceAccountKeys.java
```diff
@@ -15,47 +15,54 @@
...
-        service
-            .projects()
-            .serviceAccounts()
-            .keys()
-            .create(
-                "projects/-/serviceAccounts/" + serviceAccountEmail,
-                new CreateServiceAccountKeyRequest())
-            .execute();
+    ServiceAccountKey key = service.projects().serviceAccounts().keys()
+        .create("projects/-/serviceAccounts/" + serviceAccountEmail, new CreateServiceAccountKeyRequest()).execute();
```
lesv: You might wish to ping the author, they were trying to illustrate something w/ these lines, you've made it a bit harder to see that.

iam/api-client/src/main/java/com/google/iam/snippets/ServiceAccountKeys.java
```diff
@@ -65,14 +72,8 @@ public ServiceAccountKey createKey(String serviceAccountEmail) throws IOExceptio
...
-    List<ServiceAccountKey> keys =
-        service
-            .projects()
-            .serviceAccounts()
-            .keys()
-            .list("projects/-/serviceAccounts/" + serviceAccountEmail)
-            .execute()
-            .getKeys();
+    List<ServiceAccountKey> keys = service.projects().serviceAccounts().keys()
+        .list("projects/-/serviceAccounts/" + serviceAccountEmail).execute().getKeys();
```
lesv: ditto

iam/api-client/src/main/java/com/google/iam/snippets/ServiceAccounts.java
```diff
@@ -1,4 +1,4 @@
-/* Copyright 2018 Google LLC
+/* Copyright 2019 Google LLC
```
lesv: 2018

kurtisvg:
mdejong3@ - We have a [new format for samples](https://github.com/GoogleCloudPlatform/java-docs-samples/blob/master/SAMPLE_FORMAT.md) that includes solutions for issues like missing imports. Are you interested in updating these samples to the newer format? I'm happy to answer questions or provide more info on it.

iam/api-client/src/main/java/com/google/iam/snippets/ServiceAccountCreate.java
```diff
@@ -0,0 +1,51 @@
+/* Copyright 2018 Google LLC
```
kurtisvg: These are new files, so they do need to have the most recent year (It needs to be when the file was created, but not edited).

iam/api-client/src/main/java/com/google/iam/snippets/ServiceAccountKeyCreate.java
```diff
@@ -0,0 +1,46 @@
...
+        .createScoped(Collections.singleton(IamScopes.CLOUD_PLATFORM));
+
+        Iam service = new Iam.Builder(GoogleNetHttpTransport.newTrustedTransport(), JacksonFactory.getDefaultInstance(),
+            credential).setApplicationName("service-account-keys").build();
+        
+        ServiceAccountKey key = service.projects().serviceAccounts().keys()
+            .create("projects/-/serviceAccounts/" + serviceAccountEmail, new CreateServiceAccountKeyRequest()).execute();
+    
+        System.out.println("Created key: " + key.getName());
+        return key;
```
kurtisvg: We want to avoid returning anything, and instead just output content to the console.

iam/api-client/src/main/java/com/google/iam/snippets/ServiceAccountDelete.java
```diff
@@ -0,0 +1,42 @@
...
+import com.google.api.client.googleapis.javanet.GoogleNetHttpTransport;
+import com.google.api.client.json.jackson2.JacksonFactory;
+import com.google.api.services.iam.v1.Iam;
+import com.google.api.services.iam.v1.IamScopes;
+import java.util.Collections;
+
+
+public class ServiceAccountDelete {
+
+    public static void deleteServiceAccount(String email) throws Exception {
```
kurtisvg: Avoid throwing Exceptions at the top level, and instead show the user how to properly handle them (using try/catch). See [Exception Handling](https://github.com/GoogleCloudPlatform/java-docs-samples/blob/master/SAMPLE_FORMAT.md#exception-handling).

iam/api-client/src/main/java/com/google/iam/snippets/ServiceAccountKeyDelete.java
```diff
@@ -0,0 +1,42 @@
...
+import com.google.api.services.iam.v1.Iam;
+import com.google.api.services.iam.v1.IamScopes;
+import java.util.Collections;
+
+
+public class ServiceAccountKeyDelete {
+
+    public static void deleteKey(String fullKeyName) throws Exception {
+
+        GoogleCredential credential = GoogleCredential.getApplicationDefault()
```
kurtisvg: Since you are doing this section a lot, it's totally fine if you want to move it into a helper function after the main part of the snippet. This can make reading the snippet less tedious, since some of the boilerplate is removed:
```java
private void initService() throws IOException{
    // Use the Application Default Credentials strategy for authentication. For more info, please see:
   // https://cloud.google.com/docs/authentication/production#finding_credentials_automatically
    GoogleCredential credential = GoogleCredential.getApplicationDefault()
            .createScoped(Collections.singleton(IamScopes.CLOUD_PLATFORM));
    // Initialize the IAM service, which can be used to send requests to the IAM API.  
    Iam service = new Iam.Builder(
        GoogleNetHttpTransport.newTrustedTransport(), 
        JacksonFactory.getDefaultInstance(),
        credential
    ).setApplicationName("service-account-keys")
    .build();
}
```

iam/api-client/src/test/java/com/google/iam/snippets/ServiceAccountTests.java
```diff
@@ -0,0 +1,109 @@
...
+  }
+
+  @After
+  public void tearDown() {
+    System.setOut(null);
+    bout.reset();
+  }
+
+  @Test
+  public void testServiceAccounts() throws Exception {
```
kurtisvg: While these don't need to be in separate files, the tests should ideally be in separate functions. This way when a snippet fails, we know exactly which test is failing.

iam/api-client/src/test/java/com/google/iam/snippets/ServiceAccountTests.java
```diff
@@ -0,0 +1,109 @@
...
+
+  private ByteArrayOutputStream bout;
+
+  private static final String PROJECT_ID = System.getenv("GOOGLE_CLOUD_PROJECT");
+
+  private static final String NAME = "java-test-" + new Random().nextInt(1000);
+
+  private static final String EMAIL =  NAME + "@" + PROJECT_ID + ".iam.gserviceaccount.com";
+
+  private String got;
```
kurtisvg: I think you can remove this variable from global scope - it's not used `beforeTest` or `tearDown`.

iam/api-client/src/test/java/com/google/iam/snippets/ServiceAccountTests.java
```diff
@@ -0,0 +1,109 @@
...
+import org.junit.runners.JUnit4;
+
+
+@RunWith(JUnit4.class)
+public class ServiceAccountTests {
+
+  private ByteArrayOutputStream bout;
+
+  private static final String PROJECT_ID = System.getenv("GOOGLE_CLOUD_PROJECT");
+
```
kurtisvg: You can remove the empty lines between these variables - you only need to space them out if trying to group a a bunch of similar variables together.

iam/api-client/src/test/java/com/google/iam/snippets/ServiceAccountTests.java
mdejong3: I had tried to do that initially, but the tests kept failing because they need to be run sequentially. Is there a way to run them sequentially if they're in separate functions?

iam/api-client/src/main/java/com/google/iam/snippets/ServiceAccountDelete.java
mdejong3: Is there style guidance for how to deal with exceptions that might be thrown when initializing a variable? For example, in most of these snippets, the possible exception is thrown when building the IAM service, which is also when I initialize the service variable. I need to initialize it outside the block somehow or I can't use it.
Do I just initialize it as null?

iam/api-client/src/main/java/com/google/iam/snippets/ServiceAccountKeyCreate.java
mdejong3: I'm struggling with how to do this, since I need to use the created objects (service account/service account key) to test the other methods. Suggestions?

iam/api-client/src/main/java/com/google/iam/snippets/ServiceAccountDelete.java
kurtisvg: There are usually 2 options which are viewed as acceptable:

1. You can use null or avoid assignment to keep the scope function wide:
```java
Iam service;
try {
   GoogleCredential credential = GoogleCredential.getApplicationDefault()
        .createScoped(Collections.singleton(IamScopes.CLOUD_PLATFORM));

    Iam service = new Iam.Builder(GoogleNetHttpTransport.newTrustedTransport(), JacksonFactory.getDefaultInstance(),
            credential).setApplicationName("service-account-keys").build();
} catch (IOException ex) {
   // Here is where you handle the bad thing
   System.out.println("Unable to initialize the IAM service: " + e);
   return
}
```

2. If it's a somewhat short bit of code (and there aren't any other throws conflicting) sometimes folks prefer putting the entire block around it all:
```java
try{
   GoogleCredential credential = GoogleCredential.getApplicationDefault()
        .createScoped(Collections.singleton(IamScopes.CLOUD_PLATFORM));

    Iam service = new Iam.Builder(GoogleNetHttpTransport.newTrustedTransport(), JacksonFactory.getDefaultInstance(),
            credential).setApplicationName("service-account-keys").build();

    ServiceAccountKey key = service.projects().serviceAccounts().keys()
            .create("projects/-/serviceAccounts/" + serviceAccountEmail, new 
    CreateServiceAccountKeyRequest()).execute();

    System.out.println("Created key: " + key.getName());
} catch (IOException ex) {
    // Here is where you handle the bad thing
   System.out.println("Unable to initialize the IAM service: " + e);
}
```

iam/api-client/src/test/java/com/google/iam/snippets/ServiceAccountTests.java
kurtisvg: Ideally you don't want to run tests in order, because you want them to be independent and hermetic. However sometimes this isn't feasible, in which case you case we've had [a few folks use this trick to order tests with JUnit](https://stackoverflow.com/a/24191972/9414803).

iam/api-client/src/main/java/com/google/iam/snippets/ServiceAccountKeyCreate.java
kurtisvg: What you want to do it create a test fixture - you can use the `@Before` (and `@After`)methods to create (and then delete) a resource for each test. 

Very rarely, it's too expensive to create a resource for each test. In this case you would want to use `@BeforeClass` and `@AfterClass` to create/delete a shared resource for the entire class.

iam/api-client/src/main/java/com/google/iam/snippets/ServiceAccountCreate.java
```diff
@@ -0,0 +1,51 @@
...
+import com.google.api.services.iam.v1.Iam;
+import com.google.api.services.iam.v1.IamScopes;
+import com.google.api.services.iam.v1.model.CreateServiceAccountRequest;
+import com.google.api.services.iam.v1.model.ServiceAccount;
+import java.util.Collections;
+
+
+public class ServiceAccountCreate {
+
+    public static ServiceAccount createServiceAccount(String projectId, String name, String displayName) throws Exception {
```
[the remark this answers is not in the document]
mdejong3: Done!

iam/api-client/src/main/java/com/google/iam/snippets/ServiceAccountCreate.java
```diff
@@ -0,0 +1,51 @@
...
+
+        GoogleCredential credential = GoogleCredential.getApplicationDefault()
```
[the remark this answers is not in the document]
mdejong3: Done

iam/api-client/src/main/java/com/google/iam/snippets/ServiceAccountCreate.java
mdejong3: Done

iam/api-client/src/main/java/com/google/iam/snippets/ServiceAccountKeyCreate.java
mdejong3: Done

iam/api-client/src/main/java/com/google/iam/snippets/ServiceAccountKeyDelete.java
mdejong3: Done, let me know if I didn't do it properly

iam/api-client/src/test/java/com/google/iam/snippets/ServiceAccountTests.java
mdejong3: Done

iam/api-client/src/test/java/com/google/iam/snippets/ServiceAccountTests.java
mdejong3: Done

iam/api-client/src/main/java/com/google/iam/snippets/CreateServiceAccount.java
```diff
@@ -0,0 +1,74 @@
...
+  private static Iam service = null;
+
+  // Creates a service account.
+  public static void createServiceAccount(String projectId) throws Exception {
+    // String projectId = "my-project-id";
+
+    try {
+      initService();
+    } catch (Exception e) {
+      System.out.println("Failed to build service: \n" + e.toString());
```
kurtisvg: This needs to `return` after logging so that when an error occurs, it exits the function call.

iam/api-client/src/main/java/com/google/iam/snippets/CreateServiceAccount.java
```diff
@@ -0,0 +1,74 @@
...
+    }
+    ServiceAccount serviceAccount = new ServiceAccount();
+    serviceAccount.setDisplayName("DISPLAY_NAME");
```
kurtisvg: nit: It's nice to signal to the user these are configurable, and not required to be the literal value. 
```suggestion
    serviceAccount.setDisplayName("your-display-name");
```

iam/api-client/src/main/java/com/google/iam/snippets/CreateServiceAccount.java
```diff
@@ -0,0 +1,74 @@
...
+    CreateServiceAccountRequest request = new CreateServiceAccountRequest();
+    request.setAccountId("service-account-name");
```
kurtisvg:
```suggestion
    request.setAccountId("your-service-account-name");
```

iam/api-client/src/main/java/com/google/iam/snippets/CreateServiceAccount.java
```diff
@@ -0,0 +1,74 @@
...
+    request.setServiceAccount(serviceAccount);
+
+    serviceAccount =
+        service.projects().serviceAccounts().create("projects/" + projectId, request).execute();
+
+    System.out.println("Created service account: " + serviceAccount.getEmail());
+  }
+
+  private static void initService() {
+    try {
```
kurtisvg: You don't need a try here, because you are using a try in the main method. This error will bubble up to that try/catch.

iam/api-client/src/test/java/com/google/iam/snippets/ServiceAccountTests.java
```diff
@@ -0,0 +1,143 @@
...
+      String got = bout.toString();
+      assertTrue(got.contains("Updated display name"));
+    } catch (Exception e) {
+      System.out.println("Unable to rename service account: /n" + e.toString());
+    }
+  }
+
+  @Test
+  public void stage2_testServiceAccountKeyCreate() {
+    try {
```
kurtisvg: You don't need to use try/catch in these tests (once you remove the `throws Exceptions` from the function signature, the compiler will quite bugging you for these).

iam/api-client/src/main/java/com/google/iam/snippets/ServiceAccountCreate.java
```diff
@@ -0,0 +1,51 @@
...
+          .createScoped(Collections.singleton(IamScopes.CLOUD_PLATFORM));
+  
+        Iam service = new Iam.Builder(GoogleNetHttpTransport.newTrustedTransport(), JacksonFactory.getDefaultInstance(),
```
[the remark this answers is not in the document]
mdejong3: Done

iam/api-client/src/main/java/com/google/iam/snippets/ServiceAccountCreate.java
```diff
@@ -0,0 +1,51 @@
...
+import com.google.api.client.googleapis.javanet.GoogleNetHttpTransport;
+import com.google.api.client.json.jackson2.JacksonFactory;
+import com.google.api.services.iam.v1.Iam;
+import com.google.api.services.iam.v1.IamScopes;
+import com.google.api.services.iam.v1.model.CreateServiceAccountRequest;
+import com.google.api.services.iam.v1.model.ServiceAccount;
+import java.util.Collections;
+
+
+public class ServiceAccountCreate {
```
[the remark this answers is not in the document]
mdejong3: Done

iam/api-client/src/main/java/com/google/iam/snippets/ServiceAccountDelete.java
mdejong3: Done

iam/api-client/src/test/java/com/google/iam/snippets/ServiceAccountTests.java
mdejong3: Done

iam/api-client/src/main/java/com/google/iam/snippets/CreateServiceAccount.java
```diff
@@ -0,0 +1,74 @@
...
+import com.google.api.services.iam.v1.model.CreateServiceAccountRequest;
+import com.google.api.services.iam.v1.model.ServiceAccount;
+import java.util.Collections;
+
+public class CreateServiceAccount {
+
+  private static Iam service = null;
+
+  // Creates a service account.
+  public static void createServiceAccount(String projectId) throws Exception {
```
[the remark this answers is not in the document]
mdejong3: Done

iam/api-client/src/main/java/com/google/iam/snippets/CreateServiceAccount.java
```diff
@@ -0,0 +1,74 @@
...
+
+    serviceAccount =
+        service.projects().serviceAccounts().create("projects/" + projectId, request).execute();
+
+    System.out.println("Created service account: " + serviceAccount.getEmail());
+  }
+
+  private static void initService() {
+    try {
+      // Use the Application Default Credentials strategy for authentication. For more info, please
```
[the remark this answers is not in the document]
mdejong3: Done!

iam/api-client/src/main/java/com/google/iam/snippets/CreateServiceAccount.java
mdejong3: Done

iam/api-client/src/main/java/com/google/iam/snippets/CreateServiceAccount.java
mdejong3: Done

iam/api-client/src/main/java/com/google/iam/snippets/CreateServiceAccount.java
mdejong3: Done

iam/api-client/src/main/java/com/google/iam/snippets/CreateServiceAccount.java
mdejong3: Done

iam/api-client/src/test/java/com/google/iam/snippets/ServiceAccountTests.java
mdejong3: Done

iam/api-client/src/main/java/com/google/iam/snippets/CreateServiceAccount.java
```diff
@@ -0,0 +1,74 @@
...
+public class CreateServiceAccount {
+
+  private static Iam service = null;
+
+  // Creates a service account.
+  public static void createServiceAccount(String projectId) throws Exception {
+    // String projectId = "my-project-id";
+
+    try {
+      initService();
```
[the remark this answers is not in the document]
mdejong3: Done

iam/api-client/src/main/java/com/google/iam/snippets/CreateServiceAccount.java
```diff
@@ -0,0 +1,80 @@
...
+public class CreateServiceAccount {
+
+  // Creates a service account.
+  public static void createServiceAccount(String projectId) {
+    // String projectId = "my-project-id";
+    Iam service = null;
+    
+    try {
+      service = initService();
+    } catch (GeneralSecurityException e) {
```
kurtisvg: nit: You can combine multiple Exceptions like this: 
```suggestion
    } catch (IOException | GeneralSecurityException e) {
```

iam/api-client/src/main/java/com/google/iam/snippets/CreateServiceAccount.java
```diff
@@ -0,0 +1,80 @@
...
+import java.util.Collections;
+
+public class CreateServiceAccount {
+
+  // Creates a service account.
+  public static void createServiceAccount(String projectId) {
+    // String projectId = "my-project-id";
+    Iam service = null;
+    
+    try {
```
kurtisvg: super nit: I would consider moving the newline on L36 to before `Iam service`. It makes it more clear that the following `try` block is part of the declaration.
```java
    // String projectId = "my-project-id";
    
    Iam service = null;
    try {
```

iam/api-client/src/main/java/com/google/iam/snippets/CreateServiceAccount.java
mdejong3: Done

iam/api-client/src/main/java/com/google/iam/snippets/CreateServiceAccount.java
mdejong3: Done

mdejong3:
Looks like I can't merge until @lesv looks it over again

PR #1581, opened 2019-09-13
Hybrid Glossaries Tutorial

nnegrey:
Taking over samples from this PR: https://github.com/GoogleCloudPlatform/java-docs-samples/pull/1581


translate/cloud-client/src/main/java/com/example/translate/HybridGlossaries.java
```diff
@@ -0,0 +1,301 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.example.translate;
+
+// [START translate_hybrid_imports]
```
kurtisvg: Can we use the new format and move each snippet into it's own file so we don't need to do this?

translate/cloud-client/src/main/java/com/example/translate/HybridGlossaries.java
```diff
@@ -0,0 +1,301 @@
...
+public class HybridGlossaries {
+
+  // [START translate_hybrid_project_id]
+  private static final String projectId = System.getenv("PROJECT_ID");
+  // [END translate_hybrid_project_id]
+
+  // [START translate_hybrid_vision]
+  /**
+   * @param filePath input image file
+   **/
```
kurtisvg: Prefer not to use Javadoc style - we aren't generating any documentation from this snippets.

translate/cloud-client/src/main/java/com/example/translate/HybridGlossaries.java
```diff
@@ -0,0 +1,301 @@
...
+  static String picToText(String filePath) {
+    List<AnnotateImageRequest> requests = new ArrayList<>();
```
kurtisvg: Include commented out parameters (or the new style of overloaded function is OK).

translate/cloud-client/src/main/java/com/example/translate/HybridGlossaries.java
```diff
@@ -0,0 +1,301 @@
...
+
+    try (ImageAnnotatorClient client = ImageAnnotatorClient.create()) {
```
kurtisvg: Can you add this comment to [client initialization](https://github.com/GoogleCloudPlatform/java-docs-samples/blob/master/SAMPLE_FORMAT.md#client-initialization): 
```
  // once, and can be reused for multiple requests. After completing all of your requests, call
  // the "close" method on the client to safely clean up any remaining background resources.
```

translate/cloud-client/src/main/java/com/example/translate/HybridGlossaries.java
```diff
@@ -0,0 +1,301 @@
...
+      ByteString imageBytes = ByteString.readFrom(new FileInputStream(filePath));
+
```
kurtisvg: nit: There are a few too many empty lines here - can we remove ones that aren't needed?

translate/cloud-client/src/main/java/com/example/translate/HybridGlossaries.java
```diff
@@ -0,0 +1,301 @@
...
+      AnnotateImageRequest request =
+              AnnotateImageRequest.newBuilder()
+                      .addFeatures(feature)
+                      .setImage(image)
+                      .build();
+
+      requests.add(request);
+
+      BatchAnnotateImagesResponse response = client.batchAnnotateImages(requests);
```
kurtisvg: Do we need to use Batch? Is there an API for a single image call?

If not, we can at least inline requests:
```suggestion
      BatchAnnotateImagesResponse response = client.batchAnnotateImages(Arrays.asList(request));
```

translate/cloud-client/src/main/java/com/example/translate/HybridGlossaries.java
```diff
@@ -0,0 +1,301 @@
...
+      List<AnnotateImageResponse> responses = response.getResponsesList();
+
+      StringBuilder result = new StringBuilder();
+
+      for (AnnotateImageResponse res : responses) {
+        result.append(String.format("%s\n", res.getFullTextAnnotation().getText()));
+      }
+
+      return result.toString();
```
kurtisvg: Preferable we don't return, and just process the content somehow (log). See [Function Return Type](https://github.com/GoogleCloudPlatform/java-docs-samples/blob/master/SAMPLE_FORMAT.md#function-return-type).

translate/cloud-client/src/main/java/com/example/translate/HybridGlossaries.java
```diff
@@ -0,0 +1,301 @@
...
+  /** Creates a GCP glossary resource
+   * Assumes you've already manually uploaded a glossary to Cloud Storage
+   *
+   * ARGS
+   * @param languages list of languages in the glossary
+   * @param projectId GCP project id
+   * @param glossaryDisplayName name you want to give this glossary resource
+   * @param glossaryUri the uri of the glossary you uploaded to Cloud Storage
+   * 
+   **/
```
kurtisvg: This isn't a javadoc comment (there's content on the first line), but preferably just use a regular comment (and remove args).

translate/cloud-client/src/main/java/com/example/translate/HybridGlossaries.java
```diff
@@ -0,0 +1,301 @@
...
+  static void createGlossary(List<String> languages, String projectId,
```
kurtisvg: Why do we pass projectId here and use a class variable above? 

Preferable we also remove some of these args and only requires what's absolutely needed for testing. See [Function Paramaters](https://github.com/GoogleCloudPlatform/java-docs-samples/blob/master/SAMPLE_FORMAT.md#function-parameters).

translate/cloud-client/src/main/java/com/example/translate/HybridGlossaries.java
```diff
@@ -0,0 +1,301 @@
...
+                             String glossaryDisplayName, String glossaryUri) {
+    try (TranslationServiceClient translationServiceClient = TranslationServiceClient.create()) {
+
+      LocationName locationName =
+              LocationName.newBuilder().setProject(projectId).setLocation("us-central1").build();
```
kurtisvg: I don't know if there is still the possibility to edit this library's API, but having builders for these type of variables is an anti-pattern. 

Builders are useful when you have a bunch of optional arguments and don't want to have a bunch of overloaded APIs. These arguments are all required, so it should just have a method name. 

The API for this should be a static method: `LocationName.of(projectId, "us-central-1");` (this is way nicer to write and way easier to read).

translate/cloud-client/src/main/java/com/example/translate/HybridGlossaries.java
```diff
@@ -0,0 +1,301 @@
...
+                      .setParent(locationName.toString())
+                      .setGlossary(glossary)
+                      .build();
+
+      // Call the API
+      Glossary response = translationServiceClient.createGlossaryAsync(request)
+              .get(300, TimeUnit.SECONDS);
+
+      System.out.format("Created: %s\n", response.getName());
+    } catch (Exception e) {
```
kurtisvg: Preferable we catch a less general Exception - limit to what's being thrown.

translate/cloud-client/src/main/java/com/example/translate/HybridGlossaries.java
```diff
@@ -0,0 +1,301 @@
...
+                      .build();
+
+      CreateGlossaryRequest request =
+              CreateGlossaryRequest.newBuilder()
+                      .setParent(locationName.toString())
+                      .setGlossary(glossary)
+                      .build();
+
+      // Call the API
+      Glossary response = translationServiceClient.createGlossaryAsync(request)
```
kurtisvg: nit: This looks like it might be a style violation. Can you make sure Checkstyle is running?

Something like this might be better: 

```java
Glossary response = translationServiceClient
              .createGlossaryAsync(request)
              .get(300, TimeUnit.SECONDS);
```

translate/cloud-client/src/main/java/com/example/translate/HybridGlossaries.java
```diff
@@ -0,0 +1,301 @@
...
+   * text: String of text to translate
+   * sourceLanguageCode: language of input text
+   * targetLanguageCode: language of output text
+   * projectId: GCP project id
+   * glossaryName: name you gave your project's glossary
+   *     resource when you created it
+   * RETURNS
+   * String of translated text
+   **/
+  static String translateText(String text, String sourceLanguageCode, String targetLanguageCode,
```
kurtisvg: This should be a [style violation](https://google.github.io/styleguide/javaguide.html#s4.5.1-line-wrapping-where-to-break).

Consider running [google-java-format](https://github.com/google/google-java-format).

translate/cloud-client/src/main/java/com/example/translate/HybridGlossaries.java
```diff
@@ -0,0 +1,301 @@
...
+   * function saves the synthetic audio to the designated output file.   
+   *
+   * @param text plaintext input
+   * @param outFile String name of file under which to save audio output
+   *
+   */
+  static void textToAudio(String text, String outFile) {
+    // Replace special characters with HTML Ampersand Character Codes
+    // These codes prevent the API from confusing text with SSML tags
+    // For example, '<' --> '&lt;' and '&' --> '&amp;'
```
kurtisvg:
```suggestion
    // For example, '<' is replaced with '&lt;' and '&' is replaced with '&amp;'
```

translate/cloud-client/src/main/java/com/example/translate/HybridGlossaries.java
```diff
@@ -0,0 +1,301 @@
...
+    String escapedLines = HtmlEscapers.htmlEscaper().escape(text);
+
+    // Convert plaintext to SSML
+    // Tag SSML so that there is a 2 second pause between each address
```
kurtisvg: What does this tag mean? Can we link to the page where these tags are defined?

translate/cloud-client/src/main/java/com/example/translate/HybridGlossaries.java
```diff
@@ -0,0 +1,301 @@
...
+    String expandedNewline = escapedLines.replaceAll("\\n","\n<break time='2s'/>");
+    String ssmlText = "<speak>" + expandedNewline + "</speak>";
+
+    // Instantiates a client
+    try (TextToSpeechClient textToSpeechClient = TextToSpeechClient.create()) {
+      // Set the ssml text input to synthesize
+      SynthesisInput input = SynthesisInput.newBuilder().setSsml(ssmlText).build();
+
+      // Build the voice request, select the language code ("en-US") and
+      // the ssml voice gender ("male")
```
kurtisvg: Some of these comments are broken in weird places. The comment should go to the line limit.

translate/cloud-client/src/main/java/com/example/translate/HybridGlossaries.java
```diff
@@ -0,0 +1,301 @@
...
+      out.write(audioContents.toByteArray());
+      System.out.println("Audio content written to file " + outFile);
+    } catch (Exception e) {
+      throw new RuntimeException("Error converting the text to audio.", e);
+    }
+  }
+  // [END translate_hybrid_tts]
+
+  // [START translate_hybrid_integration]
+  public static void main(String... args) {
```
kurtisvg: Do we need to have this CLI?

translate/cloud-client/src/test/java/com/example/translate/HybridGlossariesIT.java
```diff
@@ -0,0 +1,154 @@
...
+      "gs://cloud-samples-data/translation/bistro_glossary.csv";
+
+  private ByteArrayOutputStream bout;
+  private PrintStream out;
+  private File outputFile;
+
+  @Before
+  public void setUp() {
+    GLOSSARY_NAME = String.format("bistro-glossary-%s", UUID.randomUUID());
+    LANGUAGES = new ArrayList<>();
```
kurtisvg: Can this just be declared above as static? `Arrays.asList("fr", "en");`

translate/cloud-client/src/test/java/com/example/translate/HybridGlossariesIT.java
```diff
@@ -0,0 +1,154 @@
...
+  private static String GLOSSARY_URI =
+      "gs://cloud-samples-data/translation/bistro_glossary.csv";
+
+  private ByteArrayOutputStream bout;
+  private PrintStream out;
+  private File outputFile;
+
+  @Before
+  public void setUp() {
+    GLOSSARY_NAME = String.format("bistro-glossary-%s", UUID.randomUUID());
```
kurtisvg: This can be static as well these don't need to be initialized in the setup.

translate/cloud-client/src/test/java/com/example/translate/HybridGlossariesIT.java
```diff
@@ -0,0 +1,154 @@
...
+    LANGUAGES = new ArrayList<>();
+    LANGUAGES.add(SRC_LANG);
+    LANGUAGES.add(TGT_LANG);
+    // collect project ID from environment
+    bout = new ByteArrayOutputStream();
+    out = new PrintStream(bout);
+    System.setOut(out);
+  }
+
+  @After
```
kurtisvg: Should this happen after each test? or should it be AfterClass?

translate/cloud-client/src/test/java/com/example/translate/HybridGlossariesIT.java
```diff
@@ -0,0 +1,154 @@
...
+              listGlossariesPagedResponse.getPage().getResponse();
+
+      for (Glossary glossary : listGlossariesResponse.getGlossariesList()) {
+        DeleteGlossaryRequest request =
+                DeleteGlossaryRequest.newBuilder().setName(glossary.getName()).build();
+
+        translationServiceClient.deleteGlossaryAsync(request).get(300, TimeUnit.SECONDS);
+      }
+    } catch (Exception e) {
+      System.out.format("Failed to delete the glossary.\n%s\n", e.getMessage());
```
kurtisvg: Don't need to catch the exception here, Junit will handle it. This makes it register as not a test failure.

translate/cloud-client/src/main/java/com/example/translate/HybridGlossaries.java
nnegrey: This one is a tutorial not a snippet, so I wasn't really sure what rules we have for those.

kurtisvg: Can you link to the tutorial? These feel pretty snippet-y to me - they don't really interact much from what I saw.

nnegrey: I believe it is this one: https://cloud.google.com/translate/docs/hybrid-glossaries-tutorial

translate/cloud-client/src/main/java/com/example/translate/HybridGlossaries.java
nnegrey: Yea, there is only batch or asynchronous calls. But updated to use `Arrays.asList`

translate/cloud-client/src/main/java/com/example/translate/HybridGlossaries.java
nnegrey: Updated

translate/cloud-client/src/main/java/com/example/translate/HybridGlossaries.java
```diff
@@ -0,0 +1,250 @@
...
+    String ssmlText = "<speak>" + expandedNewline + "</speak>";
+
+    // Initialize client that will be used to send requests. This client only needs to be created
+    // once, and can be reused for multiple requests. After completing all of your requests, call
+    // the "close" method on the client to safely clean up any remaining background resources.
+    try (TextToSpeechClient textToSpeechClient = TextToSpeechClient.create()) {
+      // Set the ssml text input to synthesize
+      SynthesisInput input = SynthesisInput.newBuilder().setSsml(ssmlText).build();
+
+      // Build the voice request, select the language code ("en-US") andthe ssml voice gender
```
kurtisvg: nit
```suggestion
      // Build the voice request, select the language code ("en-US") and the ssml voice gender
```

translate/cloud-client/src/main/java/com/example/translate/HybridGlossaries.java
kurtisvg: This is still returning

translate/cloud-client/src/test/java/com/example/translate/HybridGlossariesIT.java
kurtisvg: Is this still a problem?

nnegrey: Since 2 tests create glossaries. 
I can either in `setup` create UUID for the glossary names and run it after the class 
or run it each time.

I'll move the GLOSSARY_DISPLAY_NAME back to the setup

translate/cloud-client/src/main/java/com/example/translate/HybridGlossaries.java
nnegrey: Yea as part of the tutorial this gets sent over to the next method.

translate/cloud-client/src/test/java/com/example/translate/HybridGlossariesIT.java
kurtisvg: I'm okay with either, I just wanted to make sure it was intended.

kurtisvg:
@nnegrey Is this still WIP or can we close out?

translate/cloud-client/src/main/java/com/example/translate/HybridGlossaries.java
nnegrey: Is this a hard required change for the tutorial?
I'll need to work with the TW to update the docs and other languages if so.

kurtisvg: I think this is fine since we are already bending a few other rules here. At this point it's more like  a sample application.

PR #1582, opened 2019-09-16
Update sample format guidelines. 

kurtisvg:


dlp/src/main/java/dlp/snippets/InspectImageFile.java
```diff
@@ -34,11 +34,15 @@
 
 public class InspectImageFile {
 
+  public static void inspectImageFile() {
```
lesv: I like this.

dlp/src/main/java/dlp/snippets/InspectImageFile.java
```diff
@@ -48,13 +52,9 @@ public static void inspectImageFile(String projectId, String filePath) {
...
       // Specify the type and content to be inspected.
       ByteString fileBytes = ByteString.readFrom(new FileInputStream(filePath));
-      ByteContentItem byteItem = ByteContentItem.newBuilder()
-          .setType(BytesType.IMAGE)
-          .setData(fileBytes)
-          .build();
-      ContentItem item = ContentItem.newBuilder()
-          .setByteItem(byteItem)
-          .build();
+      ByteContentItem byteItem =
```
lesv: I prefer the line by line over a single line (easier to understand).

kurtisvg: I agree - these changes were done unintentionally by `google-java-format`. I'll revert when I clean this up.

dlp/src/main/java/dlp/snippets/InspectImageFile.java
nnegrey: +1

kurtisvg:
Added examples for the new overloaded format requirement, and did some general cleanup + examples for some other things in `SAMPLE_FORMAT.md`.

Some of the DLP tests are still failing, but it's unrelated to these updates. 

@lesv - I clarified some info about `Exceptions` to try to clear up some bad practices we are seeing. I'm going to leave the exception changes we discussed during the last weekly sync up to you, as I'm still not sure the best way to clarify when to throw vs catch for these type of snippets.

SAMPLE_FORMAT.md
```diff
@@ -44,44 +47,95 @@ Project should have a `pom.xml` that is readably formatted, declares a parent po
...
 Example:
 ```java
 try {
   // Do something
-} catch (IllegalArgumentException e) {
-  // IllegalArgumentException's are thrown when an invalid argument has been passed to a function.
-  // This error should be logged to that the root cause can be debugged and prevented in the future.
-  System.out.println("Error during functionName: \n" + e.toString());
+} catch (IllegalArgumentException ok) {
+  // IllegalArgumentException's are thrown when an invalid argument has been passed to a function. Ok to ignore.
```
lesv: my changes

PR #1586, opened 2019-09-18
feat: Add Functions tips scopes

grant:
Adds a section on Function scopes

Bug: b/132734505
Accidentally closed here: https://github.com/GoogleCloudPlatform/java-docs-samples/pull/1561


functions/snippets/src/main/java/Scopes.java
```diff
@@ -0,0 +1,53 @@
...
+  
+  private static int heavyComputation() {
+    int[] numbers = new int[] { 1, 2, 3, 4, 5, 6, 7, 8, 9 };
+    return Arrays.stream(numbers).reduce((t, x) -> t * x).getAsInt();
+  }
+
+  // [START functions_tips_scopes]
+  // Global (instance-wide) scope
+  // This computation runs at instance cold-start
+  private static final int InstanceVar = heavyComputation();
```
kurtisvg:
```suggestion
  // Warning: Class variables used in Servlet classes must be thread-safe, or else might introduce race conditions in your code. 
  private static final int InstanceVar = heavyComputation();
```

functions/snippets/src/main/java/Scopes.java
```diff
@@ -0,0 +1,53 @@
...
+
+import java.io.IOException;
+import java.io.PrintWriter;
+import java.util.Arrays;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+
+public class Scopes {
+  // [END functions_tips_scopes]
+  private static int lightComputation() {
```
kurtisvg: Since these functions are excluded from the snippet, this snippet is no longer copy-paste-runnable. If you remove the region tag excludes, you can move these functions below `scopeDemo` so that they are still part of the sample, but the user doesn't have to read them first.

This way the sample is still copy-paste-runnable, but the user doesn't have to read through the functions immediately.

grant: None of the languages for `functions_tips_scopes` are copy-paste-runnable.

Do we want to either:
- Keep the sample consistent
- Include these functions but be inconsistent

kurtisvg: We should include these functions. It's fine for older versions of samples to be grandfathered in, but we should aim to keep newer samples up to date with our guidelines, since they were created with the user in mind.

grant: Included the functions.

kurtisvg: nit: I would still suggest moving them both below `scopeDemo` (since `scopeDemo` is really what you want the user to focus on).

grant: Ah. Moved the functions location to the bottom of the class.

grant:
Anyone can merge when green.

PR #1591, opened 2019-09-25
Fix failing tests

nnegrey:


vision/cloud-client/src/test/java/com/example/vision/DetectIT.java
```diff
@@ -312,8 +312,8 @@ public void testCropHints() throws Exception {
     // Assert
     String got = bout.toString();
     assertThat(got).contains("vertices {");
-    assertThat(got).contains("x: 599");
-    assertThat(got).contains("y: 475");
+    assertThat(got).contains("x: 2");  // Assert in the 200 range
```
kurtisvg: You may want to consider using a regex here instead:
```java
assertThat("FooBarBaz", matchesPattern("^Foo"));
```

nnegrey: Oh yea! Let me do that. I always forget about regex.

PR #1595, opened 2019-09-30
Due to API backend changes, update the samples to match

nnegrey:
https://groups.google.com/forum/#!topic/cloud-speech-discuss/gHj5KJMH4TM

### DIFF
--- a/speech/beta/src/test/java/com/example/speech/RecognizeIT.java
+++ b/speech/beta/src/test/java/com/example/speech/RecognizeIT.java
@@ -30,7 +30,7 @@ import org.junit.runners.JUnit4;
 @RunWith(JUnit4.class)
 @SuppressWarnings("checkstyle:abbreviationaswordinname")
 public class RecognizeIT {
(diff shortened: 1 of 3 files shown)


speech/beta/src/test/java/com/example/speech/RecognizeIT.java
```diff
@@ -71,14 +71,16 @@ public void testMetadata() throws Exception {
   public void testTranscribeDiarization() throws Exception {
     Recognize.transcribeDiarization(recognitionAudioFile);
     String got = bout.toString();
-    assertThat(got).contains("Speaker Tag 2:");
+    assertThat(got).contains("Speaker 1: I'm here");
+    assertThat(got).contains("Speaker 2: hi I'd like to buy a Chrome Cast");
```
lesv: I'd bet at some point Chromecast will be a single word.

lesv: The odds are this is too brittle a test, but I'll approve it.

PR #1597, opened 2019-10-02
IAM: Add access code snippets and tests

melaniedejong:
Added access code snippets and tests for these docs: 
https://cloud.google.com/iam/docs/granting-changing-revoking-access
https://cloud.google.com/iam/docs/testing-permissions
(I don't know how to not include previously merged commits, so please forgive the slight clutter)


iam/api-client/src/main/java/com/google/iam/snippets/AddMember.java
```diff
@@ -0,0 +1,45 @@
...
+        b.getMembers().add(member);
+        System.out.println("Member " + member + " added to role " + role);
+        return;
+      }
+    }
+
+    System.out.println("Role not found in policy; member not added");
+  }
+}
+// [END iam_modify_policy_add_member]
```
lesv: You may want a newline here.

iam/api-client/src/main/java/com/google/iam/snippets/AddMember.java
```diff
@@ -0,0 +1,45 @@
...
+ * limitations under the License.
+ */
+
+package com.google.iam.snippets;
+
+import com.google.api.services.cloudresourcemanager.model.Binding;
+import com.google.api.services.cloudresourcemanager.model.Policy;
+import java.util.List;
+
+// [START iam_modify_policy_add_member]
```
lesv: Most doc teams want you to include the imports now.

lesv: Though I guess you included it in the first one, so it may be ok.

iam/api-client/src/main/java/com/google/iam/snippets/GetPolicy.java
```diff
@@ -0,0 +1,75 @@
...
+    // projectId = "my-project-id"
+
+    Policy policy = null;
+
+    CloudResourceManager service = null;
+    try {
+      service = createCloudResourceManagerService();
+    } catch (IOException | GeneralSecurityException e) {
+      System.out.println("Unable to initialize service: \n" + e.toString());
+      return policy;
```
lesv: - Is there any more that should happen here?  
- What should they do if a policy is null?  
- Why catch at all -- it's a catastrophic error, why not just say (`throws ...`) that it's a possible result?

iam/api-client/src/main/java/com/google/iam/snippets/GetPolicy.java
```diff
@@ -0,0 +1,75 @@
...
+
+    try {
+      GetIamPolicyRequest request = new GetIamPolicyRequest();
+      policy = service.projects().getIamPolicy(projectId, request).execute();
+      System.out.println("Policy retrieved: " + policy.toString());
+      return policy;
+    } catch (IOException e) {
+      System.out.println("Unable to get policy: \n" + e.toString());
+      return policy;
+    }
```
lesv: ditto

iam/api-client/src/main/java/com/google/iam/snippets/GetPolicy.java
```diff
@@ -0,0 +1,75 @@
...
+        new CloudResourceManager.Builder(
+                GoogleNetHttpTransport.newTrustedTransport(),
+                JacksonFactory.getDefaultInstance(),
+                credential)
+            .setApplicationName("service-accounts")
+            .build();
+    return service;
+  }
+}
+// [END iam_get_policy]
```
lesv: newline

iam/api-client/src/main/java/com/google/iam/snippets/RemoveMember.java
```diff
@@ -0,0 +1,51 @@
...
+        }
+        return;
+      }
+    }
+
+    System.out.println("Role not found in policy; member not removed");
+    return;
+  }
+}
+// [END iam_modify_policy_remove_member]
```
lesv: newline

iam/api-client/src/main/java/com/google/iam/snippets/SetPolicy.java
```diff
@@ -0,0 +1,73 @@
...
+    // policy = service.Projects.GetIAmPolicy(new GetIamPolicyRequest(), your-project-id).Execute();
+    // projectId = "my-project-id"
+
+    CloudResourceManager service = null;
+    try {
+      service = createCloudResourceManagerService();
+    } catch (IOException | GeneralSecurityException e) {
+      System.out.println("Unable to initialize service: \n" + e.toString());
+      return;
+    }
```
lesv: ditto -- it's bad form to just catch and do nothing, either do something, or let the exception propagate.

iam/api-client/src/main/java/com/google/iam/snippets/SetPolicy.java
```diff
@@ -0,0 +1,73 @@
...
+
+    try {
+      SetIamPolicyRequest request = new SetIamPolicyRequest();
+      request.setPolicy(policy);
+      Policy response = service.projects().setIamPolicy(projectId, request).execute();
+      System.out.println("Policy set: " + response.toString());
+    } catch (IOException e) {
+      System.out.println("Unable to set policy: \n" + e.toString());
+    }
```
lesv: Why not let the exception occur?

iam/api-client/src/main/java/com/google/iam/snippets/TestPermissions.java
```diff
@@ -0,0 +1,80 @@
...
+public static void testPermissions(String projectId) {
+    // projectId = "my-project-id"
+
+    CloudResourceManager service = null;
+    try {
+      service = createCloudResourceManagerService();
+    } catch (IOException | GeneralSecurityException e) {
+      System.out.println("Unable to initialize service: \n" + e.toString());
+      return;
+    }
```
lesv: not idiomatic

iam/api-client/src/test/java/com/google/iam/snippets/AccessTests.java
```diff
@@ -0,0 +1,120 @@
...
+  }
+
+  @Test
+  public void testTestPermissions() {
+    TestPermissions.testPermissions(PROJECT_ID);
+    String got = bout.toString();
+    assertTrue(
+        got.contains("Of the permissions listed in the request, the caller has the following: "));
+  }
+}
```
lesv: newline

iam/api-client/src/main/java/com/google/iam/snippets/GetPolicy.java
melaniedejong: I chose the throw/catch handling based on the service account snippets, which kurtisvg reviewed (https://github.com/GoogleCloudPlatform/java-docs-samples/pull/1578). He advised avoiding throwing exceptions at the top level (see the resolved conversation here: https://github.com/GoogleCloudPlatform/java-docs-samples/pull/1578/files/ebffd067f33084556e27807dd9cbdd2f8fc386ed#diff-a832b7d097d92f54302d74c8f01b8c43).
Please lmk whether to stick to that pattern from the service account docs or change it to throw exceptions instead.

lesv: @kurtisvg Please comment.

I'm fine if Kurtis is happy, but printing stuff out isn't really idiomatic for Java, I'm not suggesting you throw the exception, just allow it to be thrown, and document that it might be.

What you are doing is converting from Java's error system (throw/catch) to errors (like Golang).  For java users, this tends to hide bugs.  We should only catch an error if it's specific and we know what we are doing with it.  (ie handle it, do something else, etc.)

kurtisvg: melaniedejong@ is correct and is following the rubric guidelines. This decision is derived from the goal was that snippets be self contained to make it easier to execute. It shouldn't return any values or throw any errors. 

In an ideal world, we should be showing the user how to correctly handle or recover from an error. However, there are some situations where there isn't anything to be done (such as this one). 

IIRC the reason we went with logging unrecoverable top level errors was that if we decide to put it in the function signature, it's harder for the user to pinpoint where the error might be causing from. If we have an explicit try/catch it's more obvious what part of the code can cause that error, so if the user copies only one portion of the code they can do that.

melaniedejong:
@lesv Any idea why the checks are still pending?

lesv:
@dpebot merge when green
@melaniedejong I've tried forcing Kokoro to run.

lesv:
@kurtisvg I was trying to remove you from the reviewer list and accidentally requested your review.

melaniedejong:
I don't know how it's failing if the only thing that's been changed was merging in the master branch

kurtisvg:
```
- testing iam/api-client
[ERROR] src/main/java/com/google/iam/snippets/TestPermissions.java:[35] (indentation) Indentation: 'method def modifier' has incorrect indentation level 0, expected level should be 2.
[ERROR] src/main/java/com/google/iam/snippets/TestPermissions.java:[46] (sizes) LineLength: Line is longer than 100 characters (found 116).
[ERROR] src/main/java/com/google/iam/snippets/TestPermissions.java:[55] (sizes) LineLength: Line is longer than 100 characters (found 101).
[ERROR] src/main/java/com/google/iam/snippets/TestPermissions.java:[55,101] (whitespace) OperatorWrap: '+' should be on a new line.
[ERROR] src/main/java/com/google/iam/snippets/RemoveMember.java:[36,9] (whitespace) WhitespaceAround: WhitespaceAround: 'if' is not followed by whitespace. Empty blocks may only be represented as {} when not part of a multi-block statement (4.1.3)
[ERROR] src/main/java/com/google/iam/snippets/RemoveMember.java:[36,44] (whitespace) WhitespaceAround: WhitespaceAround: '{' is not preceded with whitespace.
[ERROR] src/main/java/com/google/iam/snippets/RemoveMember.java:[37] (indentation) Indentation: 'if' child has incorrect indentation level 12, expected level should be 10.
[ERROR] src/main/java/com/google/iam/snippets/RemoveMember.java:[38] (indentation) Indentation: 'if' child has incorrect indentation level 12, expected level should be 10.
[ERROR] src/main/java/com/google/iam/snippets/RemoveMember.java:[40,40] (whitespace) WhitespaceAround: WhitespaceAround: '{' is not preceded with whitespace.
[ERROR] src/main/java/com/google/iam/snippets/RemoveMember.java:[41] (indentation) Indentation: 'if' child has incorrect indentation level 12, expected level should be 10.
[ERROR] src/test/java/com/google/iam/snippets/AccessTests.java:[59] (coding) VariableDeclarationUsageDistance: Distance between variable 'bindings' declaration and its first usage is 4, but allowed 3.  Consider making that variable final if you still need to store its value in advance (before method calls that might have side effects on the original value).
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-checkstyle-plugin:3.0.0:check (default) on project iam-snippets: You have 11 Checkstyle violations. -> [Help 1]
```

melaniedejong:
Ah, thank you!

PR #1599, opened 2019-10-03
Fix failing tests:

nnegrey:
Update samples to use try for clients and update model IDs due to API backend changes


nnegrey:
Looking at the tests. 
Because we have to cut the UUID short due to automl word limit, going to try to fix that a bit more, but also noticing things aren't getting deleted like they should, which is why there is a model id overlap.

PR #1601, opened 2019-10-03
Add new sample for deploying a model with a node count

nnegrey:


vision/automl/src/main/java/com/google/cloud/vision/samples/automl/DeployModelNodeCount.java
```diff
@@ -0,0 +1,61 @@
...
+  static void deployModelNodeCount(String projectId, String modelId) {
+    // String projectId = "YOUR_PROJECT_ID";
+    // String modelId = "YOUR_MODEL_ID";
+    
+    // Initialize client that will be used to send requests. This client only needs to be created
+    // once, and can be reused for multiple requests. After completing all of your requests, call
+    // the "close" method on the client to safely clean up any remaining background resources.
+    try (AutoMlClient client = AutoMlClient.create()) {
+      // Get the full path of the model.
+      ModelName modelFullId = ModelName.of(projectId, "us-central1", modelId);
```
lesv: Was the model created earlier?  Why us-central1?  why not somewhere else?

nnegrey: Yea, there are other methods, to create and deploy the model. 
They just added a new feature for node count. 

us-central1 is the only location supported right now

PR #1605, opened 2019-10-07
refactored and added tags to infinite speech streaming sample

bradmiro:
Staging for viewing in tutorial


speech/cloud-client/src/main/java/com/example/speech/InfiniteStreamRecognize.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright 2018 Google LLC
+ * Copyright 2019 Google LLC
```
lesv: We don't change the copyright of the file on modification.  Otherwise LGTM

PR #1606, opened 2019-10-08
Translate automl test fixes

nnegrey:


dzlier-gcp:
Hold off on submitting - I'm confirming that we're OK with handling IOExceptions with just `e.printStackTrace()`.

translate/automl/src/main/java/com/google/cloud/translate/automl/DatasetApi.java
```diff
@@ -54,53 +54,54 @@
...
+              "\tSource language code: %s",
+              dataset.getTranslationDatasetMetadata().getSourceLanguageCode()));
+      System.out.println(
+          String.format(
+              "\tTarget language code: %s",
+              dataset.getTranslationDatasetMetadata().getTargetLanguageCode()));
+      System.out.println("Dataset create time:");
+      System.out.println(String.format("\tseconds: %s", dataset.getCreateTime().getSeconds()));
+      System.out.println(String.format("\tnanos: %s", dataset.getCreateTime().getNanos()));
+    } catch (IOException e) {
```
dzlier-gcp: In our maintenance meeting the other day, we discussed catching exceptions in this way.

The consensus is: for exceptions, if there is some specific advice we can give the user to handle it, then we should include that at least in a comment.

Otherwise, the exception should pass through the function with `throws`.

Please rework this part (and all other places where you removed a `throws` from the function heading) so that it fits one of those two.

PR #1610, opened 2019-10-08
[Cloud Run] System Packages

averikitsch:


averikitsch:
This is reliant on graphviz being installed. I need to figure out how to add that to the testing env.

kurtisvg:
You'll need to either update the kokoro script to install it, or get it inside the container.

averikitsch:
Adding it to the container. Will re-run the tests once the testing container has been updated.

averikitsch:
@kurtisvg ready for review now

run/system-package/pom.xml
```diff
@@ -0,0 +1,96 @@
...
+          <redirectTestOutputToFile>true</redirectTestOutputToFile>
+        </configuration>
+      </plugin>
+      <plugin>
+        <artifactId>maven-compiler-plugin</artifactId>
+        <version>3.8.0</version>
+      </plugin>
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-shade-plugin</artifactId>
```
kurtisvg: Why do we need to shade this package?

averikitsch: That's how the Spark framework recommends creating an uber-jar in order to run the application.

averikitsch: Would you recommend not using it?

kurtisvg: It seems odd to me - shading doesn't just add the dependencies, it remaps them all. Usually this is to avoid conflicts between dependencies. Normally, I would expect creating an uber-jar to use [something like this](https://stackoverflow.com/a/574650/9414803). 

Look at the [SparkJava](http://sparkjava.com/documentation) docs, i don't see anything about creating an uber-jar. The only reference I can find is [this tutorial](https://pconrad-webapps.github.io/topics/sparkjava_getting_started/), which doesn't seem to be official documentation. 

So it seems like shading might be unnecessary, but I don't know enough about Spark to say definitely.

averikitsch: Thanks for explaining your thoughts. I've looked more into it and have removed the shade.

kurtisvg: 👍

PR #1620, opened 2019-10-17
Use deploy redirector

jamesward:


lesv:
merge when green

lesv:
@dpebot merge when green

kurtisvg:
It looks like this was changed in #1620 as well, so closing this since that is already merged.

jamesward:
Whoops!  You're right.  Sorry for the noise.

PR #1621, opened 2019-10-17
Automl cleanup

nnegrey:
The goal is to centralize where all our AutoML Samples go so that in the future there will be less duplication and make it easier to maintain. 

Once in master, I will:
1. update docs to point here
2. Delete old branches


lesv:
@dpebot merge when green

PR #1628, opened 2019-10-21
IAM: fix region tag syntax

melaniedejong:
Changing region tag so that it uses underscores instead of hyphens


dzlier-gcp:
@dpebot  merge when green

Issue #1538, opened 2019-08-07
Remove obsolete deprecation for Java7, it is now gone.


ludoch:
Seeing this comment in README file samples:

Java 7 runtime support on App Engine standard was deprecated on December 13, 2017 and will be shut down entirely on January 16, 2019. It is replaced by the Java 8 Runtime Environment.

We need to remove it now.

ludoch:
Or maybe delete the entire section 
https://github.com/GoogleCloudPlatform/java-docs-samples/tree/2e5996c68440134a79f1511c57529fa5cf987628/appengine/taskqueue/pull-samples
and others under appengine, as the entire directory seems to be for Java7 only.

Related to this deletion is the removal of references to Java7 in our docs (filed an internal bug for that).

gguuss:
I may be missing some of the affected files but I'm seeing most mentions of Java 7 as deprecation notices. There are a couple projects referencing (incorrectly) Java 7.

```
./datastore/README.md:  * Installed [Maven](https://maven.apache.org/) and Java 7 (or above).
./appengine/pusher-chat/README.md:**Java 7 runtime support on App Engine standard was [deprecated](https://cloud.google.com/appengine/docs/deprecations/java7) on
./appengine/endpoints-frameworks-v2/backend/build.gradle:sourceCompatibility = 1.7  // App Engine Standard uses Java 7
./appengine/endpoints-frameworks-v2/backend/build.gradle:targetCompatibility = 1.7  // App Engine Standard uses Java 7
./appengine/endpoints-frameworks-v2/migration-example/build.gradle:sourceCompatibility = 1.7     // App Engine Standard uses Java 7
./appengine/endpoints-frameworks-v2/migration-example/build.gradle:targetCompatibility = 1.7     // App Engine Standard uses Java 7
./appengine/endpoints-frameworks-v2/README.md:**Java 7 runtime support on App Engine standard was [deprecated](https://cloud.google.com/appengine/docs/deprecations/java7) on
./appengine/endpoints-frameworks-v2/guice-example/build.gradle:sourceCompatibility = 1.7  // App Engine Standard uses Java 7
./appengine/endpoints-frameworks-v2/guice-example/build.gradle:targetCompatibility = 1.7  // App Engine Standard uses Java 7
./appengine/datastore/README.md:**Java 7 runtime support on App Engine standard was [deprecated](https://cloud.google.com/appengine/docs/deprecations/java7) on
[...]

PR #1635, opened 2019-10-22
Remove Java7 samples

averikitsch:
Fixes issue #1538
This is ongoing work with the TW team to reorganize and update our repo. The references in the docs need to be removed and updated first.


averikitsch:
Nothing specific. Plus, there will be an additional check that nothing is broken when the docs are updated.

PR #1640, opened 2019-10-23
run/broken: exclude "improved" route from run_broken_service

grayside:
Need to exclude the "fixed" version of the code from the `run_broken_service` region tag.

### DIFF
--- a/run/hello-broken/src/main/java/com/example/cloudrun/App.java
+++ b/run/hello-broken/src/main/java/com/example/cloudrun/App.java
@@ -49,7 +49,7 @@ public class App {
           res.status(200);
           return String.format("Hello %s!", name);
         });
-        // [END run_broken_service]
+    // [END run_broken_service]
     get(
         "/improved",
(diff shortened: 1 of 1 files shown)


run/hello-broken/src/main/java/com/example/cloudrun/App.java
```diff
@@ -63,6 +64,7 @@ public static void main(String[] args) {
           res.status(200);
           return String.format("Hello %s!", name);
         });
+  // [START run_broken_service]
```
averikitsch: [ERROR] src/main/java/com/example/cloudrun/App.java:[67] (indentation) CommentsIndentation: Comment has incorrect indentation level 2, expected is 4, indentation should be the same level as line 52.
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-checkstyle-plugin:3.0.0:check (default) on project hello-broken: You have 1 Checkstyle violation. -> [Help 1]

PR #1641, opened 2019-10-23
[Cloud Run] Add region tags for tutorial

averikitsch:


run/hello-broken/Dockerfile
```diff
@@ -31,8 +31,8 @@ RUN mvn compile assembly:single
 FROM adoptopenjdk/openjdk11:alpine
 
 # Copy the jar to the production image from the builder stage.
-COPY --from=builder /app/target/hello-broken-*dependencies.jar /hello-broken.jar
+COPY --from=builder /app/target/hello-*dependencies.jar /hello-service.jar
 
 # Run the web service on container startup.
-CMD ["java","-jar","/hello-broken.jar"]
```
grayside: Rather than name this hello-service, I think we might consider going with a generic name, e.g., `service.jar`. This makes copying Dockerfile around easier, and the more specific name doesn't add much.

averikitsch: done

PR #1646, opened 2019-10-31
Update library version

elharo:
b/143705168

@lesv


kurtisvg:
Not sure why the tests failed here, but they passed in #1647 so I merged that one instead.

PR #1719, opened 2019-11-11
Update Cloud Tasks library version

averikitsch:


averikitsch:
Changes to the new endpoint need to be pushed, hence the breaking tests.

PR #1734, opened 2019-11-14
feat: adding CreateCluster sample for Dataproc

bradmiro:
This PR adds a sample that creates a Dataproc cluster using the Java Client Library.


dataproc/src/main/java/CreateCluster.java
```diff
@@ -0,0 +1,63 @@
...
+
+    String myEndpoint = String.format("%s-dataproc.googleapis.com:443", region);
+
+    // Configure the settings for the cluster controller client
+    ClusterControllerSettings clusterControllerSettings =
+        ClusterControllerSettings.newBuilder().setEndpoint(myEndpoint).build();
+
+    // Create a cluster controller client with the configured settings
+    try (ClusterControllerClient clusterControllerClient = ClusterControllerClient.create(clusterControllerSettings)) {
+
```
kurtisvg: nit: remove new line

dataproc/src/main/java/CreateCluster.java
```diff
@@ -0,0 +1,63 @@
...
+      // Create the cluster object with the desired cluster config
+      Cluster cluster = Cluster.newBuilder().setClusterName(clusterName).setConfig(clusterConfig).build();
+
+      // Create the cluster
+      Cluster response = clusterControllerClient.createClusterAsync(projectId, region, cluster).get();
+
+      // Print out the response
+      System.out.println(response);
+
```
kurtisvg: nit: Remove extra new line

In general, a new line between each command feels excessive (IMO makes it harder to read). For this sample I would suggest grouping into 3 'sections' to keep similar ideas together:
* Arrange - construct the configs
* Act - send the API call
* Assert - print the response

See [Arrange, Act, Assert](https://github.com/GoogleCloudPlatform/java-docs-samples/blob/master/SAMPLE_FORMAT.md#arrange-act-assert)

dataproc/src/main/java/CreateCluster.java
```diff
@@ -0,0 +1,63 @@
...
+    } catch (Exception e) {
```
kurtisvg: Please catch specific exceptions and how the user to to resolve, or else throw them to the next level - See [Exception Handling](https://github.com/GoogleCloudPlatform/java-docs-samples/blob/master/SAMPLE_FORMAT.md#exception-handling)

dataproc/src/test/java/CreateClusterTest.java
```diff
@@ -0,0 +1,94 @@
...
+import org.junit.runner.RunWith;
+import org.junit.runners.JUnit4;
+
+@RunWith(JUnit4.class)
+public class CreateClusterTest {
+
+  private static final String BASE_CLUSTER_NAME = "test-cluster";
+  private static final String REGION = "us-central1";
+
+  private static String projectId;
```
kurtisvg: This makes it more clear where this is set, and if unset it will be null and still caught in `checkRequirements`.
```suggestion
  private static String projectId = System.getenv("GOOGLE_CLOUD_PROJECT");;
```

dataproc/src/test/java/CreateClusterTest.java
```diff
@@ -0,0 +1,94 @@
...
+        System.getenv(varName),
+        String.format("Environment variable '%s' is required to perform these tests.", varName)
+    );
+  }
+
+  @BeforeClass
+  public static void checkRequirements() {
+    requireEnv("GOOGLE_APPLICATION_CREDENTIALS");
+    requireEnv("GOOGLE_CLOUD_PROJECT");
+    projectId = System.getenv("GOOGLE_CLOUD_PROJECT");
```
kurtisvg: Remove lines L55-L56

dataproc/src/test/java/CreateClusterTest.java
```diff
@@ -0,0 +1,94 @@
...
+
+    bout = new ByteArrayOutputStream();
+    System.setOut(new PrintStream(bout));
+  }
+
+  @Test
+  public void createClusterTest() throws Exception {
+    CreateCluster createCluster = new CreateCluster();
+
+    createCluster.createCluster(projectId, REGION, clusterName);
```
kurtisvg: If you make this method static, you can just call the method (you shouldn't need to initialize a class for this).
```suggestion
    CreateCluster.createCluster(projectId, REGION, clusterName);
```

dataproc/src/main/java/CreateCluster.java
```diff
@@ -0,0 +1,63 @@
...
+          .setNumInstances(2).build();
+
+      ClusterConfig clusterConfig = ClusterConfig.newBuilder().setMasterConfig(masterConfig)
+          .setWorkerConfig(workerConfig).build();
+
+      // Create the cluster object with the desired cluster config
+      Cluster cluster = Cluster.newBuilder().setClusterName(clusterName).setConfig(clusterConfig).build();
+
+      // Create the cluster
+      Cluster response = clusterControllerClient.createClusterAsync(projectId, region, cluster).get();
```
kurtisvg: Please separate out the request creation from the `.get` - We want it to be obvious to users that a future is being returned here.

dataproc/src/main/java/CreateCluster.java
```diff
@@ -0,0 +1,63 @@
...
+
+      // Print out the response
+      System.out.println(response);
```
kurtisvg: Any chance you can print a more specific message then this to show how to interact with the response? Maybe something like `"Cluster created successfully: %s", response.getName()`. This will allow the test to be more explicit, and give users a better idea on how to interact with the object.

dataproc/src/test/java/CreateClusterTest.java
```diff
@@ -0,0 +1,94 @@
...
+    String output = bout.toString();
+
+    assertThat(output, CoreMatchers.containsString("cluster-uuid"));
+  }
+
+  @After
+  public void tearDown() throws IOException {
+    String myEndpoint = String.format("%s-dataproc.googleapis.com:443", REGION);
```
kurtisvg: Does this still need to be here? You aren't using a custom endpoint to build it

dataproc/src/test/java/CreateClusterTest.java
```diff
@@ -0,0 +1,94 @@
...
+
+    ClusterControllerSettings clusterControllerSettings =
+        ClusterControllerSettings.newBuilder().setEndpoint(myEndpoint).build();
+
+    try (ClusterControllerClient clusterControllerClient =
+        ClusterControllerClient.create(clusterControllerSettings)) {
+
+        clusterControllerClient.deleteClusterAsync(projectId, REGION, clusterName).get();
+
```
kurtisvg: nit: Remove whitespace

dataproc/src/test/java/CreateClusterTest.java
```diff
@@ -0,0 +1,94 @@
...
+    } catch (Exception e) {
+      System.out.println("Error during cluster deletion: \n" + e.toString());
+    }
+  }
+
```
kurtisvg: nit: Remove whitespace

dataproc/src/main/java/CreateCluster.java
bradmiro: Do you feel strongly about this? I followed the convention in the doc- [(https://googleapis.dev/java/google-cloud-clients/latest/com/google/cloud/dataproc/v1/ClusterControllerClient.html#createClusterAsync-com.google.cloud.dataproc.v1.CreateClusterRequest-)]

Either way, I feel these should match

bradmiro:
Most of @kurtisvg's suggestions fixed with these commits. I left a comment on one suggestion.

dataproc/src/main/java/CreateCluster.java
kurtisvg: Yes - creation of a future followed immediately by a `.get` is an anti-pattern. It should be clear that the `OperationFuture` is a long running operation and may block for an extended period of time.

bradmiro: Ok, fixed + commited this in the sample and the test, and made a note to fix the javadoc to match this.

dataproc/src/main/java/CreateCluster.java
```diff
@@ -0,0 +1,83 @@
...
+      OperationFuture<Cluster, ClusterOperationMetadata> createClusterAsyncRequest =
+          clusterControllerClient.createClusterAsync(projectId, region, cluster);
+      Cluster response = createClusterAsyncRequest.get();
+
+      // Print out the response
+      System.out.println(
+          String.format("Cluster created successfully: %s", response.getClusterName())
+      );
+
+    } catch (IOException e) {
```
kurtisvg: IOException + InterruptedException are system/ hardware levels exceptions that occur during i/o and threading that should probably just be a `throws` at the Function level. I would not expect that missing or incorrect auth to show up as an IOException. 

(ExecutionException does look good to leave as-is though)

PR #1738, opened 2019-11-15
fix: small updates to dataproc sample, test and pom

bradmiro:
Slight change to pom, changed region tags to include 'dataproc', added endpoint back to test (would not successfully delete the cluster without this)


kurtisvg:
@bradmiro Can you rebase this to the current version of master?

dataproc/src/test/java/CreateClusterTest.java
```diff
@@ -75,8 +76,13 @@ public void createClusterTest() throws IOException, InterruptedException {
 
   @After
   public void tearDown() throws IOException, InterruptedException {
+    String myEndpoint = String.format("%s-dataproc.googleapis.com:443", REGION);
```
kurtisvg: Is there a way to file an issue on the client library for this? 

The region is already passed in the `deleteClusterAsyncRequest`, and it seems pretty dumb to force the user to create a new regional client just to delete a cluster.

bradmiro: I agree and this is consistent across all of the languages. I have this documented on my end.

PR #1742, opened 2019-11-17
Fixed missing cloud_region in MQTT example

jush:
The readme mentions that the project ID `blue-jet-123` is in `asia-east1` but then it forgets to pass the `cloud_region` parameter which results in a `Not authorized to connect (5)` exception since the default region is `us-central1`.


gguuss:
@kurtisvg - Any ideas on why kokoro seems to be getting stuck?

kurtisvg:
@gguuss Kokoro won't trigger tests for PRs from outside GCP. You can manualy run with `kokoro: run` label

PR #1745, opened 2019-11-18
[MySQL] Add Dockerfile and instructions for Cloud Run

averikitsch:
Added new dependencies because of annoying logging warning.


cloud-sql/mysql/servlet/Dockerfile
```diff
@@ -0,0 +1,35 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Use the official maven/Java 11 image to create a build artifact.
+# https://hub.docker.com/_/maven
+FROM maven:3.6.2-jdk-8-slim as builder
+
+# Copy local code to the container image.
+WORKDIR /app
+COPY pom.xml .
```
kurtisvg: Any particular reason for copying the POM separate?

cloud-sql/mysql/servlet/pom.xml
```diff
@@ -64,14 +64,24 @@
       <artifactId>HikariCP</artifactId>
       <version>3.4.1</version>
     </dependency>
+    <dependency>
+       <groupId>org.slf4j</groupId>
```
kurtisvg: Why are these added?

cloud-sql/mysql/servlet/Dockerfile
averikitsch: We've been doing it across all samples in order to reduce the change of copying over files that aren't needed.

cloud-sql/mysql/servlet/pom.xml
averikitsch: There was an annoying logging warning about slf4j not being found. Though it doesn't break anything it was distracting. I can remove if needed.

kurtisvg: LGTM, was just curious.

PR #1748, opened 2019-11-21
Add vision ocr set endpoint snippets

nnegrey:
These samples are going on a special docs page that shows only how to change your endpoint.
Similar to these docs pages: https://cloud.google.com/natural-language/docs/locations#automl_set_endpoint-java


nnegrey:
Ping

vision/cloud-client/src/main/java/com/example/vision/SetEndpoint.java
```diff
@@ -0,0 +1,69 @@
...
+import java.util.ArrayList;
+import java.util.List;
+
+class SetEndpoint {
+
+  // Change your endpoint
+  static void setEndpoint() throws IOException {
+    // [START vision_set_endpoint]
+    ImageAnnotatorSettings settings =
+        ImageAnnotatorSettings.newBuilder().setEndpoint("eu-vision.googleapis.com:443").build();
```
lesv: Should there be some comment on this endpoint?

vision/cloud-client/src/main/java/com/example/vision/SetEndpoint.java
```diff
@@ -0,0 +1,69 @@
...
+
+    // Initialize client that will be used to send requests. This client only needs to be created
+    // once, and can be reused for multiple requests. After completing all of your requests, call
+    // the "close" method on the client to safely clean up any remaining background resources.
+    ImageAnnotatorClient client = ImageAnnotatorClient.create(settings);
```
lesv: Nice - Is the client thread safe?  (ie can it be used concurrently?)

vision/cloud-client/src/main/java/com/example/vision/SetEndpoint.java
```diff
@@ -0,0 +1,69 @@
...
+    Image image = Image.newBuilder().setSource(imgSource).build();
+    Feature feature = Feature.newBuilder().setType(Feature.Type.TEXT_DETECTION).build();
+    AnnotateImageRequest request =
+        AnnotateImageRequest.newBuilder().addFeatures(feature).setImage(image).build();
+    List<AnnotateImageRequest> requests = new ArrayList<>();
+    requests.add(request);
+
+    BatchAnnotateImagesResponse batchResponse = client.batchAnnotateImages(requests);
+
+    for (AnnotateImageResponse response : batchResponse.getResponsesList()) {
```
lesv: If there are multiple response, should each one be called out?

vision/cloud-client/src/main/java/com/example/vision/SetEndpoint.java
nnegrey: It will live on a specialized page like this one: https://cloud.google.com/natural-language/docs/locations

The product teams asked for them to be very minimal samples with specific wording.

vision/cloud-client/src/main/java/com/example/vision/SetEndpoint.java
nnegrey: Normally I'd put it in the `try` resource block, but because the product team didn't want anything else on the docs page for this snippet, the region tag ends right below this.
But I still wanted to make sure this was testable, so the rest of the snippet runs the code then closes the client, but the actual snippet is only the above 2 lines on the docs.

PR #1757, opened 2019-11-26
bigquery_create_dataset sample

stephaniewang526:


bigquery/cloud-client/src/main/java/com/example/bigquery/RunCreateDataset.java
```diff
@@ -0,0 +1,43 @@
...
+  public static void runCreateDataset() {
+    // TODO(developer): Replace these variables before running the sample.
+    String datasetName = "my-dataset-name";
+    createDataset(datasetName);
+  }
+
+  public static Dataset createDataset(String datasetName) {
+    // Initialize client that will be used to send requests. This client only needs to be created
+    // once, and can be reused for multiple requests. After completing all of your requests, call
+    // the "close" method on the client to safely clean up any remaining background resources.
```
tswast: Thoughts on creating the client in `runCreateDataset` and actually showing the `close()` call?

If possible, it'd be cool to show constructing the client in a "try with resources" block in `runCreateDataset`.

bigquery/cloud-client/src/main/java/com/example/bigquery/RunCreateDataset.java
```diff
@@ -0,0 +1,43 @@
...
+
+package com.example.bigquery;
+
+// [START bigquery_create_dataset]
+import com.google.cloud.bigquery.BigQuery;
+import com.google.cloud.bigquery.BigQueryOptions;
+import com.google.cloud.bigquery.Dataset;
+import com.google.cloud.bigquery.DatasetInfo;
+
+public class RunCreateDataset {
```
kurtisvg: nit: This can just be `CreateDataset`

bigquery/cloud-client/src/main/java/com/example/bigquery/RunCreateDataset.java
kurtisvg: You won't be able to create the client in `runCreateDataset` because createDataset is called in the test.

+1 to using try-with-resources.

bigquery/cloud-client/src/main/java/com/example/bigquery/RunCreateDataset.java
```diff
@@ -0,0 +1,43 @@
...
+  }
+
+  public static Dataset createDataset(String datasetName) {
+    // Initialize client that will be used to send requests. This client only needs to be created
+    // once, and can be reused for multiple requests. After completing all of your requests, call
+    // the "close" method on the client to safely clean up any remaining background resources.
+    BigQuery bigquery = BigQueryOptions.getDefaultInstance().getService();
+
+    DatasetInfo datasetInfo = DatasetInfo.newBuilder(datasetName).build();
+    return bigquery.create(datasetInfo);
```
kurtisvg: Same as the other PR - don't return values, instead interact with it (possibly via printing a message) and use the test to verify the message was correct.

stephaniewang526: I am creating a separate feature request to make BigQuery client AutoCloseable -- it is currently not so I'm unable to use try-with-resources. Tracking it here: https://b.corp.google.com/issues/145229155

PR #1758, opened 2019-11-26
bigquery_delete_dataset

stephaniewang526:


bigquery/cloud-client/src/test/java/com/example/bigquery/RunDeleteDatasetIT.java
```diff
@@ -0,0 +1,37 @@
...
+import com.google.cloud.bigquery.BigQueryOptions;
+import com.google.cloud.bigquery.DatasetId;
+import com.google.cloud.bigquery.testing.RemoteBigQueryHelper;
+import org.junit.Test;
+
+public class RunDeleteDatasetIT {
+
+  @Test
+  public void deleteDataset() {
+    BigQuery bigquery = BigQueryOptions.getDefaultInstance().getService();
```
kurtisvg: In order to test without returning a value, just capture the output of the snippet and verify it's printing the correct data. For a good example of how to do this, [see this test](https://github.com/GoogleCloudPlatform/java-docs-samples/blob/master/dlp/src/test/java/dlp/snippets/InspectTests.java#L65).

bigquery/cloud-client/src/main/java/com/example/bigquery/DeleteDataset.java
```diff
@@ -31,14 +31,13 @@ public static void runDeleteDataset() {
...
+  public static void deleteDataset(String projectId, String datasetName) {
     // Initialize client that will be used to send requests. This client only needs to be created
-    // once, and can be reused for multiple requests. After completing all of your requests, call
-    // the "close" method on the client to safely clean up any remaining background resources.
+    // once, and can be reused for multiple requests.
     BigQuery bigquery = BigQueryOptions.getDefaultInstance().getService();
 
     DatasetId datasetId = DatasetId.of(projectId, datasetName);
-    return bigquery.delete(datasetId, DatasetDeleteOption.deleteContents());
+    System.out.println(bigquery.delete(datasetId, DatasetDeleteOption.deleteContents()));
```
kurtisvg: In this case the response doesn't really return anything useful to the user since it's just a boolean - can you do something like this instead? 
```java
boolean success = bigquery.delete(datasetId, DatasetDeleteOption.deleteContents();
if (sucess){
    System.out.println("Dataset deleted successfully");
} else {
    System.out.println("Dataset was not deleted.");
}
```
Otherwise it may be non-obvious that the user has to check the result of the function to verify it was successful (since most other libraries just raise Exceptions when operations fail). 

This might apply to your other PR as well, and I just didn't catch it. Feel free to update it in this PR if you want.

PR #1760, opened 2019-11-27
Revert "Remove Java7 samples"

averikitsch:
Reverts GoogleCloudPlatform/java-docs-samples#1635

TW has not accurately removed all java 7 samples yet ...


kurtisvg:
Rather than revert this commit, can we just pin all the docs references to a previous commit?

averikitsch:
Good idea, let me check

PR #1768, opened 2019-11-27
bq_update_dataset_access

stephaniewang526:


bigquery/cloud-client/src/main/java/com/example/bigquery/UpdateDatasetAccess.java
```diff
@@ -0,0 +1,61 @@
...
+    BigQuery bigquery = BigQueryOptions.getDefaultInstance().getService();
+
+    Dataset dataset = bigquery.getDataset(datasetName);
+    List<Acl> beforeAcls = dataset.getAcl();
+
+    // Make a copy of the ACLs so that they can be modified.
+    ArrayList<Acl> acls = new ArrayList<>(beforeAcls);
+    Acl newEntry = Acl.of(new User("sample.bigquery.dev@gmail.com"), Role.READER);
+    try {
+      acls.add(newEntry);
```
kurtisvg: This can be moved out of the try/catch (see above comment). Generally, you should try to limit try/catches around where the exceptions actually occur.

bigquery/cloud-client/src/main/java/com/example/bigquery/UpdateDatasetAccess.java
```diff
@@ -0,0 +1,61 @@
...
+    updateDatasetAccess(datasetName);
+  }
+
+  public static void updateDatasetAccess(String datasetName) {
+    // Initialize client that will be used to send requests. This client only needs to be created
+    // once, and can be reused for multiple requests.
+    BigQuery bigquery = BigQueryOptions.getDefaultInstance().getService();
+
+    Dataset dataset = bigquery.getDataset(datasetName);
+    List<Acl> beforeAcls = dataset.getAcl();
```
kurtisvg: Does this throw an exception as well? Presumable it's making an API call underneath?

bigquery/cloud-client/src/main/java/com/example/bigquery/UpdateDatasetAccess.java
```diff
@@ -0,0 +1,61 @@
...
+
+    // Make a copy of the ACLs so that they can be modified.
```
kurtisvg: This is somewhat misleading - you aren't making a copy of the ACLs, you are making a copy of the list (which still has references to the original ACL objects). The references in the list still point to the same objects. 

Also, there's no reason to retain a link to the original list, because it doesn't get used anywhere. 

Something like this might be better instead:
```suggestion
  // For more information on the types of ACLs available see: 
  // https://link.to.acl.docs
  // Create a new ACL granting the READER role to "sample.bigquery.dev@gmail.com"
  Acl newEntry = Acl.of(new Acl.User("sample.bigquery.dev@gmail.com"), Role.READER);
   
   // Get a copy of the ACLs list from the dataset and append the new entry
   ArrayList<Acl> acls = new ArrayList<>(dataset.getAcl());
   acls.add(newEntry);
   
   // Then down below add a comment like "Update the dataset to use the new ACLs".
```

Issue #1764, opened 2019-11-27
Broken KMS tests


kurtisvg:
Tests in `kms/` are failing due to lint checks - see #1763 for example.

lesv:
This appears to have been fixed.  Closing.

Issue #1762, opened 2019-11-27
google-cloud-kms dependency way out of date


chingor13:
https://github.com/GoogleCloudPlatform/java-docs-samples/blob/master/kms/pom.xml#L28

The current version of google-cloud-kms is 1.35.0 (it's GA now)

gguuss:
Reassigning to Dan as I believe he authored this sample.

gguuss:
Dan, let me know if you can't get to this within SLO, happy to take a pass on it.

kurtisvg:
It would be good to figure out why Renovate hasn't updated this.

DanSanche:
I updated the version with no issues here https://github.com/GoogleCloudPlatform/java-docs-samples/pull/1769. I'm not familiar enough with the repo to say why it's not being auto-updated though

kurtisvg:
Fixed by #1769

PR #1769, opened 2019-11-27
KMS: update package version

DanSanche:
addresses https://github.com/GoogleCloudPlatform/java-docs-samples/issues/1762


kurtisvg:
Can you fix the style checks? I'd prefer not to merge a major version dump without verifying it works correctly.

DanSanche:
Yup, fixed (or at least ignored. I think we'd want to keep RSA as an abbreviation here)

kurtisvg:
Fixes #1764

PR #1775, opened 2019-12-02
ignore checkstyle generated file

elharo:


lesv:
@elharo Do you need us to merge?

PR #1777, opened 2019-12-02
Bump Spring Cloud GCP version to 1.2

meltsufin:


kurtisvg:
@dzlier-gcp - Any insight into why this version bumps aren't being performed by renovate?

elharo:
At least one of the dependencies doesn't seem to follow anything approximating semver; e.g. HOXTON.release

vision/spring-framework/pom.xml
```diff
@@ -42,7 +42,7 @@ limitations under the License.
 			<dependency>
```
averikitsch: I know the original formatting is off, but would you be able to realign this pom file?

meltsufin: How is it off? Is this about tabs vs. spaces?

meltsufin:
> At least one of the dependencies doesn't seem to follow anything approximating semver; e.g. HOXTON.release

@elharo Hoxton is a release train name in Spring Cloud and they use it as version for the BOM. We don't really have control over it.

vision/spring-framework/pom.xml
averikitsch: It's using tabs not spaces, but just a nit since it renders differently on github.

meltsufin: Done.

Issue #1761, opened 2019-11-27
Use com.google.cloud:libraries-bom for dependency management


chingor13:
Moving forward, we're encouraging users to import the `com.google.cloud:libraries-bom` in the `dependencyManagement` section of the user's `pom.xml` to resolve dependency compatibility. Users will no longer need to specify individual version numbers for google-cloud-x libraries.

averikitsch:
@chingor13 can you add this to our sample format info? Also, is this an ask for updating existing samples?

PR #1778, opened 2019-12-02
Add instructions for using the libraries-bom in dependencies.

kolea2:
Towards #1761


dzlier-gcp:
So are we going to replace all the `[dependencies]` tags through the samples with this bom dependency instead?

kolea2:
@dzlier-gcp yes, that's the end goal. This is step 1 :)

PR #1781, opened 2019-12-03
Use libraries-bom for snippets - part 2

kolea2:
Towards #1761

### DIFF
--- a/container-registry/container-analysis/pom.xml
+++ b/container-registry/container-analysis/pom.xml
@@ -27,34 +27,41 @@
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
   </properties>
 
+  <!-- [START dependencies] -->
+  <dependencyManagement>
+    <dependencies>
+      <dependency>
+        <groupId>com.google.cloud</groupId>
+        <artifactId>libraries-bom</artifactId>
+        <version>3.0.0</version>
+        <type>pom</type>
+        <scope>import</scope>
+      </dependency>
+    </dependencies>
+  </dependencyManagement>
+
   <dependencies>
-    <!-- [START dependencies] -->
     <dependency>
       <groupId>com.google.cloud</groupId>
       <artifactId>google-cloud-containeranalysis</artifactId>
-      <version>0.117.0-beta</version>
     </dependency>
     <dependency>
       <groupId>io.grafeas</groupId>
       <artifactId>grafeas</artifactId>
-      <version>0.21.0</version>
     </dependency>
     <dependency>
       <groupId>com.google.cloud</groupId>
       <artifactId>google-cloud-core</artifactId>
-      <version>1.91.3</version>
     </dependency>
     <dependency>
       <groupId>com.google.cloud</groupId>
       <artifactId>google-cloud-core-grpc</artifactId>
-      <version>1.91.3</version>
     </dependency>
-
     <dependency>
       <groupId>com.google.cloud</groupId>
       <artifactId>google-cloud-pubsub</artifactId>
-      <version>1.100.0</version>
     </dependency>
+
     <dependency>
       <groupId>commons-cli</groupId>
       <artifactId>commons-cli</artifactId>
@@ -70,7 +77,6 @@
       <artifactId>netty-tcnative-boringssl-static</artifactId>
       <version>2.0.27.Final</version>
     </dependency>
-
     <!-- [END dependencies] -->
     <dependency>
       <groupId>junit</groupId>
@@ -78,5 +84,7 @@
       <version>4.12</version>
       <scope>test</scope>
     </dependency>
+    <!-- [START dependencies] -->
   </dependencies>
+  <!-- [END dependencies] -->
 </project>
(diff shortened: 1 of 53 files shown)


dlp/src/main/java/dlp/snippets/InspectImageFile.java
```diff
@@ -33,14 +33,6 @@
 import java.util.List;
 
 public class InspectImageFile {
-
-  public static void inspectString() {
-    // TODO(developer): Replace these variables before running the sample.
-    String projectId = "my-project-id";
-    String textToInspect = "My name is Gary and my email is gary@example.com";
-    inspectString(projectId, textToInspect);
```
kolea2: @kurtisvg this was breaking the build as it could not compile. This inspectString method does not exist in this class.

kurtisvg: Can we update to the following instead?

```java
public static void inspectImageFile() {	
    // TODO(developer): Replace these variables before running the sample.	
    String projectId = "my-project-id";	
    String filePath = "path/to/image/file";	
    inspectString(projectId, filePath);
 }
```

container-registry/container-analysis/pom.xml
```diff
@@ -27,34 +27,41 @@
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
   </properties>
 
+  <!-- [START dependencies] -->
+  <dependencyManagement>
```
kurtisvg: Should we consider adding a comment explaining what it is or why we commend it? May confuse users not familiar with the concept. 

Do we have an existing doc or something that we can link to?

dzlier-gcp: It's talked about in the CONTRIBUTING guide as a change from yesterday, but the same note here wouldn't go amiss.

container-registry/container-analysis/pom.xml
```diff
@@ -70,13 +77,14 @@
...
     </dependency>
-
     <!-- [END dependencies] -->
     <dependency>
       <groupId>junit</groupId>
       <artifactId>junit</artifactId>
       <version>4.12</version>
       <scope>test</scope>
     </dependency>
+    <!-- [START dependencies] -->
```
dzlier-gcp: Remove

container-registry/container-analysis/pom.xml
```diff
@@ -70,13 +77,14 @@
...
   </dependencies>
+  <!-- [END dependencies] -->
```
dzlier-gcp: Remove

firestore/pom.xml
```diff
@@ -59,8 +69,9 @@
       <version>4.13-beta-3</version>
       <scope>test</scope>
     </dependency>
+    <!-- [START fs-maven] -->
```
dzlier-gcp: Remove these and check the rest of your changes for these weird tags, looks like a merge error.

container-registry/container-analysis/pom.xml
kurtisvg: These are intentional - the snippet is showing the following:

```
<dependencyManagement>
 // bom stuff here
</dependencyManagement>

<dependencies>
  <dependency>
    // included dependency
  </dependency>

  // excluded dependencies 

</dependencies>
```

dzlier-gcp: But it's using the same tag name as above in the same file, does that work?

dlp/src/main/java/dlp/snippets/InspectImageFile.java
kolea2: Sorry, took another look and now I think I know what this sample is doing. I believe the correct fix is:
```
public static void inspectImageFile() {	
    // TODO(developer): Replace these variables before running the sample.	
    String projectId = "my-project-id";	
    String filePath = "path/to/image/file";	
    inspectImageFile(projectId, filePath);
 }
```

where `inspectImageFile(String, String)` is found directly below this in this class. Let me know if my understanding is incorrect.

container-registry/container-analysis/pom.xml
kolea2: SG, I can add a link to the libraries-bom page in the comments

container-registry/container-analysis/pom.xml
kolea2: @kurtisvg can you confirm this is ok to do?

kurtisvg: Yes - there are two ways to have "multi-part" snippets. You can use `[START_EXCLUDE]`/`[END_EXCLUDE]` or you can just end the tag and start it in the correct place (see http://go/code-snippets-201#including-code-from-multiple-sections-of-a-file)

dlp/src/main/java/dlp/snippets/InspectImageFile.java
kurtisvg: That looks correct to me 👍

container-registry/container-analysis/pom.xml
dzlier-gcp: Neat, did not know that!

kolea2:
@kurtisvg @dzlier-gcp I think all comments are addressed, mind taking another look?

PR #1785, opened 2019-12-04
bigquery_list_datasets sample

stephaniewang526:


bigquery/cloud-client/src/main/java/com/example/bigquery/ListDatasets.java
```diff
@@ -0,0 +1,51 @@
...
+    BigQuery bigquery = BigQueryOptions.getDefaultInstance().getService();
+
+    // List datasets in a specified project
+    try {
+      Page<Dataset> datasets = bigquery.listDatasets(projectId, DatasetListOption.pageSize(100));
+      for (Dataset dataset : datasets.iterateAll()) {
+        System.out.println(dataset.getDatasetId() + " dataset in project listed successfully");
+      }
+    } catch (BigQueryException e) {
+      System.out.println("Project does not contain any datasets \n" + e.toString());
```
kurtisvg: nit: Is this the only reason a BigQueryException would return? Might want to generalize the error message.

PR #1790, opened 2019-12-06
fix: use IdTokenCredentials for IAP request

### LINKED ISSUE #1643: BuildIapRequest fails when using workload Identity
## In which file did you encounter the issue?
https://github.com/GoogleCloudPlatform/java-docs-samples/blob/master/iap/src/main/java/com/example/iap/BuildIapRequest.java

We're trying to change our applications to use workload identity instead of explicitly setting the GOOGLE_APPLICATION_CREDENTIALS path like we were doing before.
But we've hit an issue when one of our apps running on GKE tries to invoke an airflow DAG running in cloud composer.

Previously, we followed the documentation below to invoke the airflow DAG running in Cloud Composer which is Cloud IAP-secured.
https://cloud.google.com/iap/docs/authentication-howto#authenticating_from_a_service_account 

The documentation refers to this BuildIapRequest.java class in the google java-docs-samples and if you look specifically at line 61, it just throws an exception if GoogleCredentials.getApplicationDefault() returns anything other than ServiceAccountCredentials.
https://github.com/GoogleCloudPlatform/java-docs-samples/blob/master/iap/src/main/java/com/example/iap/BuildIapRequest.java#L61 

```java
if (credentials == null || !(credentials instanceof ServiceAccountCredentials)) {
  throw new Exception("Google credentials : service accounts credentials expected");
}
```

This is the root of our problem. When we explicitly set the GOOGLE_APPLICATION_CREDENTIALS, it was getting a ServiceAccountCredentials and was working as expected.
When we switched to use workload identity, it started getting a ComputeEngineCredentials instead and failing because of the check above and we couldn't find a new version of this class with support for ComputeEngineCredentials.

After digging into the source code for how the SignedJWT object is being created and signed in the BuildIapRequest.java class, we were able to recreate it using the following approach...
(The main change was to use credentials.sign() instead of trying to access the private key which was not available in ComputeEngineCredentials)

Old code:
```java
JWSSigner signer = new RSASSASigner(credentials.getPrivateKey());
SignedJWT signedJwt = new SignedJWT(jwsHeader, claims);
signedJwt.sign(signer);
return signedJwt.serialize();
```

New Code:
```java
SignedJWT signedJwt = new SignedJWT(jwsHeader, claims);
String signedJwtStr = Base64.getEncoder().encodeToString(credentials.sign(signedJwt.getSigningInput()));
return new String(signedJwt.getSigningInput())+"."+signedJwtStr;
```

Now, the problem is the new code works fine when we're dealing with ServiceAccountCredentials. But when the credentials object returned from GoogleCredentials.getApplicationDefault() is a ComputeEngineCredentials (as is the case when workload identity is enabled) it doesn't.
We've attached 3 log excerpts for different scenarios tested:
Successful DAG invocation (when dealing with ServiceAccountCredentials) -> app-success.log
Failed DAG invocation (when dealing with ComputeEngineCredentials)   -> app-failure.log
In this scenario, the ID token was getting created but the call to composer was returning a "401 Unauthorized" error which suggests that the signed data returned is different in this scenario.
Failed DAG invocation with updated  google-auth-library-oauth2-http library -> app-failure-with-lib-update.log
After 2, we tried updating the "google-auth-library-oauth2-http" library version to 0.16.2 as we did in the other ticket mentioned and even more interestingly, it gives a totally different error.
It claims some permission is missing but doesn't give any information on the permission that is needed. We already added the iam.serviceAccounts.signBlob role while testing scenario 2.
IOException: Error code 403 trying to sign provided bytes: The caller does not have permission.
The conclusion we've come with at this point is that "credentials.sign()" returns a different/unexpected output if the credentials is a ServiceAccountCredentials vs ComputeEngineCredentials.

### Did you change the file? If so, how?
No

## Describe the issue

Able to invoke composer rest endpoint from kubernetes which has workload identity enabled

chingor13:
Fixes #1643

In google-auth-library 0.17.0 (the latest 0.18.0 is preferred though), we introduced a new interface `IdTokenProvider` which a subset of `GoogleCredentials` types can provide (`ServiceAccountCredentials` and `ComputeEngineCredentials` both implement `IdTokenProvider`). The interface abstract the mechanism for fetching an Open ID token using the means available for that type of credential.

This should also drastically simplify the sample code and not require you to handle the JWT encoding/Open ID implementation.


lesv:
Do you need us to merge?

PR #1803, opened 2019-12-11
Use test endpoint for datalabeling integration tests.

dzlier-gcp:


.kokoro/tests/run_tests.sh
```diff
@@ -58,6 +58,7 @@ if [[ "$SCRIPT_DEBUG" != "true" ]]; then
     export GOOGLE_APPLICATION_CREDENTIALS=${KOKORO_GFILE_DIR}/service-acct.json
     export QUEUE_ID=my-appengine-queue
     export LOCATION_ID=us-east1
+	export DATALABELING_ENDPOINT=test-datalabeling.sandbox.googleapis.com:443
```
averikitsch: nit: remove extra tab

datalabeling/beta/cloud-client/src/main/java/com/example/datalabeling/CreateAnnotationSpecSet.java
```diff
@@ -39,7 +40,15 @@ static void createAnnotationSpecSet(String projectId) {
     annotationLabels.put("label_1", "label_1_description");
     annotationLabels.put("label_2", "label_2_description");
 
-    try (DataLabelingServiceClient dataLabelingServiceClient = DataLabelingServiceClient.create()) {
+    String endpoint = System.getenv("DATALEBELING_ENDPOINT");
```
averikitsch:
```suggestion
    String endpoint = System.getenv("DATALABELING_ENDPOINT");
```

datalabeling/beta/cloud-client/src/main/java/com/example/datalabeling/CreateAnnotationSpecSet.java
```diff
@@ -39,29 +40,38 @@ static void createAnnotationSpecSet(String projectId) {
     annotationLabels.put("label_1", "label_1_description");
     annotationLabels.put("label_2", "label_2_description");
 
-    try (DataLabelingServiceClient dataLabelingServiceClient = DataLabelingServiceClient.create()) {
+    String endpoint = System.getenv("DATALABELING_ENDPOINT");
+    if (endpoint == null) {
+      endpoint = DataLabelingServiceSettings.getDefaultEndpoint();
+    }
 
+    try (DataLabelingServiceClient dataLabelingServiceClient =
```
averikitsch: Please add comments to each file, that this is unnecessary for users to do or something to that effect.

averikitsch:
Tests are not passing now that the variable typo was fixed.

datalabeling/beta/cloud-client/src/main/java/com/example/datalabeling/CreateAnnotationSpecSet.java
kurtisvg: +1. We should also add a comment why this is necessary for these snippets, and try to hide this portion from being staged in the documentation.

averikitsch: nit: It would be nice to still add a comment here (doesn't need to be in the staged version) to explain the extra code.

dzlier-gcp:
Justin suggested talking to Jeff Ching about moving this sample into the Datalabeling repo so that the tests are only triggered when an actual change to datalabeling goes through, instead of changing the endpoint we're targeting, so I'm going to check with him before submitting.

kurtisvg:
@dzlier-gcp Perhaps we should submit this now so that the current integration tests stop, and then move it in a different PR

PR #1804, opened 2019-12-11
Clean up of App engine java 11 pom files

averikitsch:


averikitsch:
Updated the spanner sample to use the BOM. All other samples are correctly using the BOM.

PR #1807, opened 2019-12-12
feat: add functions tips retries samples

grant:
Adds 2 samples related to GCF function retries:
- b/132733794
- b/132733814

### DIFF
--- a/functions/snippets/src/main/java/PubSub.java
+++ b/functions/snippets/src/main/java/PubSub.java
@@ -1,0 +1,83 @@
+/*
+ * Copyright 2019 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+// [START functions_tips_infinite_retries]
+import java.sql.Timestamp;
+// [START functions_tips_retry]
+import java.text.ParseException;
+import java.util.Map;
+
+
+import com.google.gson.Gson;
+import com.google.gson.JsonObject;
+
+public class PubSub {
+  // Use GSON (https://github.com/google/gson) to parse JSON content.
+  private Gson gsonParser = new Gson();
+
+  // [END functions_tips_infinite_retries]
+  // [START functions_tips_retry]
+  public String retryPubSub(PubSubMessage message) throws Exception, ParseException {
+    JsonObject body = gsonParser.fromJson(message.data, JsonObject.class);
+    
+    boolean retry = false;
+    if (body.has("retry")) {
+      retry = body.get("retry").getAsBoolean();
+    }
+    
+    if (retry) {
+      throw new Exception("Retrying...");
+    } else {
+      throw new Exception("Not retrying...");
+    }
+  }
+  // [END functions_tips_retry]
+
+  // [START functions_tips_infinite_retries]
+  /**
+   * Background Cloud Function that only executes within
+   * a certain time period after the triggering event
+   */
+  public void infiniteRetryPubSub(PubSubMessage message) throws Exception, ParseException {
+    JsonObject body = gsonParser.fromJson(message.data, JsonObject.class);
+    final long MAX_AGE = 10_000;
+
+    long eventAge = 0;
+    if (body.has("timestamp")) {
+      long ts = Timestamp.valueOf(body.get("timestamp").getAsString()).getTime();
+      eventAge = System.currentTimeMillis() - ts;
+    }
+
+    // Ignore events that are too old
+    if (eventAge > MAX_AGE) {
+      System.out.printf("Dropping event %s with age %ld ms.", body, eventAge);
+      return;
+    }
+
+    // Do what the function is supposed to do
+    System.out.printf("Processing event %s with age %ld ms.", body, eventAge);
+  }
+
+  // [START functions_tips_retry]
+  public class PubSubMessage {
+    String data;
+    Map<String, String> attributes;
+    String messageId;
+    String publishTime;
+  }
+}
+// [END functions_tips_retry]
+// [END functions_tips_infinite_retries]
--- a/functions/snippets/src/main/java/PubSub.java
+++ b/functions/snippets/src/main/java/PubSub.java
@@ -15,17 +15,16 @@
  */
 
 // [START functions_tips_infinite_retries]
-import java.sql.Timestamp;
 // [START functions_tips_retry]
+import com.google.gson.Gson;
+import com.google.gson.JsonObject;
+import java.time.Duration;
+import java.time.LocalDateTime;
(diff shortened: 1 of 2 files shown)


functions/snippets/src/main/java/PubSub.java
```diff
@@ -0,0 +1,83 @@
...
+  // [START functions_tips_infinite_retries]
+  /**
+   * Background Cloud Function that only executes within
+   * a certain time period after the triggering event
+   */
+  public void infiniteRetryPubSub(PubSubMessage message) throws Exception, ParseException {
+    JsonObject body = gsonParser.fromJson(message.data, JsonObject.class);
+    final long MAX_AGE = 10_000;
+
+    long eventAge = 0;
```
eamonnmcmanus: Could we do something like this?

```java
LocalDateTime timestamp = LocalDateTime.now();
if (body.has("timestamp")) {
  timestamp = LocalDateTime.parse(body.get("timestamp").getAsString());
}
long eventAge = Duration.between(timestamp, LocalDateTime.now()).toMillis();
```

functions/snippets/src/main/java/PubSub.java
```diff
@@ -0,0 +1,83 @@
...
+  public String retryPubSub(PubSubMessage message) throws Exception, ParseException {
+    JsonObject body = gsonParser.fromJson(message.data, JsonObject.class);
+    
+    boolean retry = false;
+    if (body.has("retry")) {
+      retry = body.get("retry").getAsBoolean();
+    }
+    
+    if (retry) {
+      throw new Exception("Retrying...");
```
eamonnmcmanus: This doesn't seem right. I think what we should have is that if you want to trigger a retry then you throw an exception, which will result in an HTTP 500 response, and that should trigger the retry logic. If you don't want to retry then you just return. But here we're throwing an exception in both cases.

kurtisvg: We should probably throw something more specific than `Exception` as well.

functions/snippets/src/main/java/PubSub.java
kurtisvg: +1 - also please use `java.time` for `eventAge` as well.

functions/snippets/src/main/java/PubSub.java
```diff
@@ -0,0 +1,83 @@
...
+      System.out.printf("Dropping event %s with age %ld ms.", body, eventAge);
+      return;
+    }
+
+    // Do what the function is supposed to do
+    System.out.printf("Processing event %s with age %ld ms.", body, eventAge);
+  }
+
+  // [START functions_tips_retry]
+  public class PubSubMessage {
```
kurtisvg: This class needs a comment explaining Functions will marshall it using `gson`.

functions/snippets/src/main/java/PubSub.java
grant: - I'm following the canonical sample for this snippet:
  - https://github.com/GoogleCloudPlatform/nodejs-docs-samples/blob/master/functions/tips/index.js#L166
- I've changed the `Exception` to a `RuntimeException`.

functions/snippets/src/main/java/PubSub.java
grant: Done.

functions/snippets/src/main/java/PubSub.java
```diff
@@ -0,0 +1,83 @@
...
+public class PubSub {
+  // Use GSON (https://github.com/google/gson) to parse JSON content.
+  private Gson gsonParser = new Gson();
+
+  // [END functions_tips_infinite_retries]
+  // [START functions_tips_retry]
+  public String retryPubSub(PubSubMessage message) throws Exception, ParseException {
+    JsonObject body = gsonParser.fromJson(message.data, JsonObject.class);
+    
+    boolean retry = false;
```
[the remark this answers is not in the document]
grant: Done.

functions/snippets/src/main/java/PubSub.java
```diff
@@ -0,0 +1,83 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+// [START functions_tips_infinite_retries]
+import java.sql.Timestamp;
```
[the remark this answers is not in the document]
grant: Thanks, I'll use `LocalDateTime`.

functions/snippets/src/main/java/PubSub.java
grant: I've changed to the suggested snippet. Not sure what Kurtis' comment means.

functions/snippets/src/main/java/PubSub.java
```diff
@@ -0,0 +1,83 @@
...
+// [START functions_tips_retry]
+import java.text.ParseException;
+import java.util.Map;
+
```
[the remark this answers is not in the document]
grant: Done.

functions/snippets/src/main/java/PubSub.java
```diff
@@ -15,33 +15,32 @@
...
   private Gson gsonParser = new Gson();
 
   // [END functions_tips_infinite_retries]
   // [START functions_tips_retry]
   public String retryPubSub(PubSubMessage message) throws Exception, ParseException {
     JsonObject body = gsonParser.fromJson(message.data, JsonObject.class);
     
-    boolean retry = false;
+    boolean retry = body.has("retry") && body.get("retry").getAsBoolean();
     if (body.has("retry")) {
```
eamonnmcmanus: Did you mean to delete these three lines?

functions/snippets/src/main/java/PubSub.java
```diff
@@ -0,0 +1,83 @@
...
+    
+    boolean retry = body.has("retry") && body.get("retry").getAsBoolean();
+    if (body.has("retry")) {
+      retry = body.get("retry").getAsBoolean();
+    }
+    
+    if (retry) {
+      throw new RuntimeException("Retrying...");
+    } else {
+      throw new RuntimeException("Not retrying...");
```
eamonnmcmanus: I still don't think this is right. You want to throw an exception in the retry case, which will result in a 500 status to the client (whatever part of the system is delivering the event). That tells it that it may retry. But in the non-retry case, you want to return normally, which will result in a 200 status indicating that the event has been successfully handled. In the [node](https://github.com/GoogleCloudPlatform/nodejs-docs-samples/blob/master/functions/tips/index.js#L166) equivalent, we do `call(err)` in the retry case and plain `call()` in the non-retry case, and I'm assuming that results in equivalent 500/200 behaviour.

lesv:
Seems like this PR has been abandoned?  If there isn't a response, I'll close the PR in 48h.

ace-n:
@lesv caught you just in time 😄

@grant can you update this PR, or do you want me to take these on?

grant:
There were issues testing this code locally. It's almost there.
Ace, you can follow-up on this as I haven't been prioritizing the Java Functions work. I'll close the PR.

ace-n:
Got it - I'll take it from here. 🙂

Mind if I reopen this PR, in case it's easier to pick this up where you left off (vs. start a new PR)?

grant:
> Got it - I'll take it from here. 🙂
> 
> Mind if I reopen this PR, in case it's easier to pick this up where you left off (vs. start a new PR)?

Go ahead and revive it!

PR #1810, opened 2019-12-12
Move springboot app from getting-started-java

averikitsch:


dzlier-gcp:
Where is this going to be featured in our docs? Wondering since we're moving away from App Engine Flex to Standard or Run.

averikitsch:
That would be a question for the Java working group. This is the start of the work to remove the getting started java repo and this is linked twice in the flex docs.

PR #1811, opened 2019-12-12
Update pom files, update samples

nnegrey:


lesv:
Appears like Java11 failed.

nnegrey:
Bom is using an old library version, waiting on the next release of the bom.

PR #1813, opened 2019-12-13
Add information about necessary permissions

elharo:
@DanSanche I finally got the KMS tests to pass. This information was not easy to deduce. I may have overgranted the permissions, but this seems to work.


kms/README.md
```diff
@@ -19,11 +19,19 @@ You must [enable the Google Cloud KMS API](https://console.cloud.google.com/flow
 
 You must set your project ID in order to run the tests
 
-`GOOGLE_CLOUD_PROJECT=<your-project-id-here>`
+`$export GOOGLE_CLOUD_PROJECT=<your-project-id-here>`
 
 ### Grant Permissions
 
-You must ensure that the [user account or service account](https://cloud.google.com/iam/docs/service-accounts#differences_between_a_service_account_and_a_user_account) you used to authorize your gcloud session has the proper permissions to edit KMS resources for your project. More information can be found in the [Google KMS Docs](https://cloud.google.com/kms/docs/reference/permissions-and-roles)
+You must ensure that the [user account or service account](https://cloud.google.com/iam/docs/service-accounts#differences_between_a_service_account_and_a_user_account) you used to authorize your gcloud session has the proper permissions to edit KMS resources for your project. In the Cloud Console uner IAM, add the following permissions to the project whose service account you're using to test:
```
DanSanche: This isn't quite right - the list contains a list of roles, not permissions (each role grants a set of permissions). Cloud KMS Admin is the most permissive role, so it should be the only role that's necessary to do everything with KMS

elharo: That didn't work for me. I had thought that, but the ITs didn't work. It wasn't until I added the additional roles on this list that they started passing.

DanSanche: hmm, looks like you're right. The docs say

> The Admin role is meant for users to manage keys but not use them, and the Encrypter or Decrypter roles are meant for services who can use a key to encrypt and decrypt data, but not manage them.

so it looks like admin alone won't unlock everything. I'd like to find the minimal set of roles required if we're going to add it to the README though. I believe we can remove Cloud KMS CryptoKey Public Key Viewer and possibly Cloud KMS Importer


>

DanSanche: Also, this should say "add following roles to the project", since the list consists of roles, not permissions

kms/README.md
```diff
@@ -19,11 +19,19 @@ You must [enable the Google Cloud KMS API](https://console.cloud.google.com/flow
 
 You must set your project ID in order to run the tests
 
-`GOOGLE_CLOUD_PROJECT=<your-project-id-here>`
+`$export GOOGLE_CLOUD_PROJECT=<your-project-id-here>`
```
lesv: a space between '$' and 'export' makes it easier for users to copy

kms/README.md
```diff
@@ -35,15 +43,15 @@ Build your project with:
 
 You can run the quickstart with:
 
-    java -cp target/kms-samples-1.0.0-jar-with-dependencies.jar \
+    java -cp target/kms-samples-1.0.11-jar-with-dependencies.jar \
```
lesv: that seems rather specific and not really sustainable.

PR #1829, opened 2019-12-17
Fix flaky speech test for speaker diarization

nnegrey:
Speaker tags can be swapped between 2 speakers for different calls to the api.

Call 1:
Person A gets speaker tag 1
Person B gets speaker tag 2
Call 2:
Person A gets speaker tag 2
Person B gets speaker tag 1

Fix: https://github.com/GoogleCloudPlatform/java-docs-samples/issues/1826

Filed a FR to have first person talking always get 1 and increment additional speakers from there.


nnegrey:
It appears the endpoint will sometimes just return 1 speaker tag. Updating the tests for this.

PR #1903, opened 2019-12-20
fix: Add ORDER BY to guarantee output order

olavloite:
Add `ORDER BY SingerId DESC` to the sample query to guarantee the order of the returned values. This will fix the failing `SpannerSample.queryWithArrayOfStruct` sample.


averikitsch:
Does this affect the meaning of the docs that this sample is used in?

olavloite:
No, the meaning of the docs stays the same and is not affected by the additional `ORDER BY` statement. See https://cloud.google.com/spanner/docs/structs#querying_data_with_an_array_of_struct_objects

(The code was also already changed for Node.js as the same problem occurred there)

PR #1908, opened 2019-12-26
feat: dataproc quickstart and createcluster

bradmiro:
* adding quickstart sample + test
* updating createCluster + test
* updating pom


dataproc/src/main/java/Quickstart.java
```diff
@@ -0,0 +1,166 @@
...
+      // Submit an asynchronous request to execute the job
+      Job request = jobControllerClient.submitJob(projectId, region, job);
+      String jobId = request.getReference().getJobId();
+      System.out.println(String.format("Submitted job \"%s\"", jobId));
+
+      // Wait for the job to finish
+      CompletableFuture<Job> finishedJobFuture =
+          CompletableFuture.supplyAsync(() ->
+           waitForJobCompletion(jobControllerClient, projectId, region, jobId));
+      int timeout = 10;
```
kurtisvg: nit: Can we inline timeout? It looks like it's only ever used once

dataproc/src/test/java/CreateClusterTest.java
```diff
@@ -60,35 +60,32 @@ public static void checkRequirements() {
 
   @Before
   public void setUp() {
-    clusterName = String.format("%s-%s", BASE_CLUSTER_NAME, UUID.randomUUID().toString());
-
     bout = new ByteArrayOutputStream();
     System.setOut(new PrintStream(bout));
+    standardOutOrig = System.out;
```
kurtisvg: You don't need this, but if you do keep it in it needs to set `standardOutOrig` before you call `System.setOut` - otherwise you are just getting a reference to the value you just set.

dataproc/src/test/java/CreateClusterTest.java
```diff
@@ -38,18 +38,18 @@
 @RunWith(JUnit4.class)
 public class CreateClusterTest {
 
-  private static final String BASE_CLUSTER_NAME = "test-cluster";
+  private static final String CLUSTER_NAME =
+      String.format("java-dataproc-create-cluster-test-%s", UUID.randomUUID().toString());
```
kurtisvg: 👍

dataproc/src/test/java/CreateClusterTest.java
```diff
@@ -38,18 +38,18 @@
...
   private static final String REGION = "us-central1";
+  private static final String PROJECT_ID = System.getenv("GOOGLE_CLOUD_PROJECT");
 
-  private static String projectId = System.getenv("GOOGLE_CLOUD_PROJECT");
-  private String clusterName;
   private ByteArrayOutputStream bout;
+  private PrintStream standardOutOrig;
 
   private static void requireEnv(String varName) {
     assertNotNull(
```
kurtisvg: Discovered in another PR we've been doing this backwards - it needs to be `assertNotNull(msg, value)` to work correctly.

dataproc/src/main/java/Quickstart.java
bradmiro: It's used twice, hence my reasoning for not wanting to inline it: once when calling `.get()`, the other in the `catch` block (line 152).

dataproc/pom.xml
```diff
@@ -49,6 +61,10 @@
       <artifactId>google-cloud-dataproc</artifactId>
       <version>0.118.0</version>
     </dependency>
+    <dependency>
+      <groupId>com.google.cloud</groupId>
+      <artifactId>google-cloud-storage</artifactId>
```
kurtisvg: nit: Preferably, this should be above the test dependencies comment.

dataproc/src/main/java/Quickstart.java
```diff
@@ -0,0 +1,166 @@
...
+            String.format("Job timed out after %d minutes: %s", timeout, e.getMessage()));
+      }
+
+      // Delete the cluster.
+      OperationFuture<Empty, ClusterOperationMetadata> deleteClusterAsyncRequest =
+          clusterControllerClient.deleteClusterAsync(projectId, region, clusterName);
+      deleteClusterAsyncRequest.get();
+      System.out.println(String.format("Cluster \"%s\" successfully deleted.", clusterName));
+
+    } catch (ExecutionException e) {
```
kurtisvg: If `ExecutionException` occurs from interacting with the API (like it's thrown if you don't have permission or if your request is misconfigured), keep it and clarify that it's an error from interacting with the service with the `println`. 

If it's not from the API, please let it bubble up (and be consistent with the other samples - it's bubbled up in a few other places).

kurtisvg:
You may also need to truncate your UUID:
> INVALID_ARGUMENT: Cluster name 'java-dataproc-create-cluster-test-facb5e75-2e17-486c-9e22-e108bf8e8e63' must match pattern (?:[a-z](?:[-a-z0-9]{0,49}[a-z0-9]

dataproc/src/main/java/Quickstart.java
bradmiro: Ahh yes, I see what you're saying. Errors caused by our API bubble up to be an Execution Exception, so I think it should definitely be caught. I'll add it the other sample.

PR #1910, opened 2019-12-27
Update renovate config. 

kurtisvg:
1. Remove label `kokoro:run` (it is being applied by the trusted contributors bot)
2. Removes `prCreation: notPending` (doesn't seem to be doing anything)
3. Groups a few similar packages together (less PRs for us to review)
4. Enables the "masterIssue" feature - I turned this on for Python, and [it looks useful for tracking what Renovate plans to update](https://github.com/GoogleCloudPlatform/python-docs-samples/issues/2653)


renovate.json
```diff
@@ -2,16 +2,25 @@
...
+      "packagePatterns": [ "^io.grpc:grpc-" ],
       "groupName": "gRPC packages"
+    },
+    {
+      "packagePatterns": [ "^io.vertx:vertex-" ],
+      "groupName": "Vertx packages"
+    },
+    {
+      "packagePatterns": [ "^io.ktor:ktor-" ],
+      "groupName": "Ktor packages"
```
chingor13: An option for `ktor` is to import the [`io.ktor:ktor-bom`](https://search.maven.org/artifact/io.ktor/ktor-bom) in a single place which should include all of the ktor artifacts.

kurtisvg: @averikitsch FYI (I believe these dependencies are in `appengine-java11/`)

averikitsch: Can you add micronaut as well.

kurtisvg: The [only instance of micronaut I can find is already using the bom](https://github.com/GoogleCloudPlatform/java-docs-samples/blob/e95a9ede3bf45bc70f17fd4279bb431c64af745d/appengine-java11/micronaut-helloworld/pom.xml#L41).

Should I remove ktor from here and just change to the bom instead?

averikitsch: I've updated Ktor to use the bom PR #1915

dzlier-gcp:
With not-pending we're going to have to monitor these more actively instead of relying on email notifications, since kokoro doesn't seem to want to run the tests by default without the kokoro:run label.

kurtisvg:
> With not-pending we're going to have to monitor these more actively instead of relying on email notifications, since kokoro doesn't seem to want to run the tests by default without the kokoro:run label.

I don't think this will have any effect with our current workflow. The docs describes the following for `prCreation=not-pending`:
```
not-pending: Renovate will wait until status checks have completed (passed or failed) before raising the PR
```

We aren't actually running tests on any branch unless it's in master or when a PR is opened, so I don't think this is doing anything for us.

renovate.json
kurtisvg: Updated this PR.

kurtisvg:
> I'm good w/ this, can you ask it to not update the servlet spec as well?

We closed that PR, so the default behavior is that it won't open again (unless it we ask it to).

Issue #1479, opened 2019-06-24
Flaky DLP tests


averikitsch:
## In which file did you encounter the issue?

dlp/src/test/java/com/example/dlp/InspectIT.java

## Describe the issue

[Test Fusion](https://fusion.corp.google.com/runanalysis/test/prod%3Acloud-devrel%2Fjava%2Fjava-docs-samples%2Fjava8%2Fperiodic/prod%3Acloud-devrel%2Fjava%2Fjava-docs-samples%2Fjava8%2Fperiodic/KOKORO/569be8a6-9318-4272-a25b-ad463abef9a4/1561316765710/build%20%2375/Targets?target=github%2Fjava-docs-samples%2Fdlp%2Ftarget%2Ffailsafe-reports&drilldownTab=tests)

PR #1911, opened 2019-12-27
Cleanup DLP Inspect snippets.

kurtisvg:
Clean up of DLP Inspect snippets.

Addresses #1479.


dlp/src/main/java/com/example/dlp/QuickStart.java
```diff
@@ -78,7 +77,7 @@ public static void main(String[] args) throws Exception {
               .build();
       ContentItem contentItem = ContentItem.newBuilder().setByteItem(byteContentItem).build();
 
-      String projectId = ServiceOptions.getDefaultProjectId();
+      String projectId = "";
```
averikitsch: nit: should a comment be added here to instruct the user?

kurtisvg: This file needs to be cleaned up yet - `ServiceOptions` seems to have been removed and it won't compile currently - I'll move it into an overloaded function in a separate PR.

kurtisvg:
There are also some tests outside of the InspectTests and RedactTests still failing. Will continue to fix in separate PRs.

PR #1912, opened 2019-12-27
Cleanup DLP Quickstart.

kurtisvg:
Working towards  #1479


dlp/src/test/java/dlp/snippets/QuickstartTests.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright 2017 Google Inc.
+ * Copyright 2019 Google LLC
```
lesv: We don't update copyright dates

dlp/src/main/java/dlp/snippets/QuickStart.java
```diff
@@ -14,7 +14,9 @@
  * limitations under the License.
  */
 
-package com.example.dlp;
+package dlp.snippets;
```
lesv: Ok, but not really the standard for either us or Java.

dlp/src/test/java/dlp/snippets/QuickstartTests.java
kurtisvg: This is actually a new file - Git just thinks it was moved because it was similar to a deleted file.

dlp/src/main/java/dlp/snippets/QuickStart.java
kurtisvg: We don't have guidance on what the snippet package should be (but this is the package the canonical snippet has been in for a while.)

What would you like to be?

PR #1928, opened 2020-01-03
automl: break tests into individual test files and remove bom from pom for v1 library support

nnegrey:


automl/cloud-client/src/test/java/com/example/automl/DeleteModelTest.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright 2019 Google LLC
+ * Copyright 2020 Google LLC
```
nnegrey: Will revert back to 2019 as GitHub decided to just call this one a file rename.

automl/cloud-client/src/test/java/com/example/automl/GetModelEvaluationTest.java
```diff
@@ -0,0 +1,81 @@
...
+import org.junit.After;
+import org.junit.Before;
+import org.junit.BeforeClass;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.junit.runners.JUnit4;
+
+@RunWith(JUnit4.class)
+public class GetModelEvaluationTest {
+  private static final String PROJECT_ID = System.getenv("GOOGLE_CLOUD_PROJECT");
```
averikitsch: Does this need to be the AUTOML_PROJECT?

nnegrey: Yea, I'm working on the commits now to change those.

PR #1935, opened 2020-01-08
asset: Update real time feed sample code to user the latest package

cwxie-google:
I have to remove the bom management because our library hasn't been released to bom and we're projected to launch next week, I will update it once we have the library in bom


lesv:
Do you need us to merge this?

cwxie-google:
Thanks @lesv, yes please, we'd like to merge it.

PR #1938, opened 2020-01-09
translate: add basic translation samples

nnegrey:


translate/cloud-client/src/test/java/com/example/translate/TranslateTextTests.java
```diff
@@ -0,0 +1,71 @@
...
+  private static void requireEnvVar(String varName) {
+    assertNotNull(
+        "Environment variable '%s' is required to perform these tests.".format(varName),
+        System.getenv(varName));
+  }
+
+  @BeforeClass
+  public static void checkRequirements() {
+    requireEnvVar("GOOGLE_APPLICATION_CREDENTIALS");
+    requireEnvVar("GOOGLE_CLOUD_PROJECT");
```
lesv: I imagine the project is needed, but G_A_C shouldn't be?

nnegrey: V3 of translate became a resourceful API and requires credentials now.

PR #1939, opened 2020-01-09
translate: add v3 samples for translate text with model

nnegrey:


translate/cloud-client/src/main/java/com/example/translate/BatchTranslateTextWithModel.java
```diff
@@ -0,0 +1,104 @@
...
+    // once, and can be reused for multiple requests. After completing all of your requests, call
+    // the "close" method on the client to safely clean up any remaining background resources.
+    try (TranslationServiceClient client = TranslationServiceClient.create()) {
+      // Supported Locations: `global`, [glossary location], or [model location]
+      // Glossaries must be hosted in `us-central1`
+      // Custom Models must use the same location as your model. (us-central1)
+      String location = "us-central1";
+      LocationName parent = LocationName.of(projectId, location);
+
+      GcsSource gcsSource = GcsSource.newBuilder().setInputUri(inputUri).build();
```
kurtisvg: nit: I think some comments for on each section might add value. Something like `// Configure the source of the file from a GCS bucket`, `// Configure the output file in a GCS bucket`, etc.

translate/cloud-client/src/main/java/com/example/translate/BatchTranslateTextWithModel.java
```diff
@@ -0,0 +1,104 @@
...
+          BatchTranslateTextRequest.newBuilder()
+              .setParent(parent.toString())
+              .setSourceLanguageCode(sourceLanguage)
+              .addTargetLanguageCodes(targetLanguage)
+              .addInputConfigs(inputConfig)
+              .setOutputConfig(outputConfig)
+              .putModels(targetLanguage, modelPath)
+              .build();
+
+      OperationFuture<BatchTranslateResponse, BatchTranslateMetadata> future =
```
kurtisvg: nit: A comment here clarifying that this will send the request to the server to start the job asynchronously would also be useful.

PR #1947, opened 2020-01-14
bigtable hbase: read and filter snippets

billyjacobson:


bigtable/hbase/snippets/src/test/java/com/example/bigtable/ReadsTest.java
```diff
@@ -0,0 +1,346 @@
...
+    projectId = requireEnv("GOOGLE_CLOUD_PROJECT");
+    instanceId = requireEnv(INSTANCE_ENV);
+
+    try (BigtableTableAdminClient adminClient =
+        BigtableTableAdminClient.create(projectId, instanceId)) {
+      CreateTableRequest createTableRequest =
+          CreateTableRequest.of(TABLE_ID).addFamily(COLUMN_FAMILY_NAME);
+      adminClient.createTable(createTableRequest);
+
+      try (BigtableDataClient dataClient = BigtableDataClient.create(projectId, instanceId)) {
```
lesv: Should there be a discussion of key design?  Having a common prefix is ok for tests, but it's exactly the pattern we are trying to discourage users from doing.

billyjacobson: The schema is deviceType#deviceId#YYYYMMDD this was discussed and reviewed with members of the Bigtable team. This is only a small subset of data so we can't really show much variety

PR #1948, opened 2020-01-14
Add secretmanager samples

sethvargo:


sethvargo:
Could someone with permissions please enable the Secret Manager API on the builders:

```text
$ gcloud services enable secretmanager.googleapis.com
```

secretmanager/src/main/java/com/example/ListSecretVersions.java
```diff
@@ -0,0 +1,50 @@
...
+    try (SecretManagerServiceClient client = SecretManagerServiceClient.create()) {
+      // Build the parent name.
+      SecretName parent = SecretName.of(projectId, secretId);
+
+      // Create the request.
+      ListSecretVersionsRequest request =
+          ListSecretVersionsRequest.newBuilder().setParent(parent.toString()).build();
+
+      // Get all versions, paginated.
+      ListSecretVersionsPagedResponse iterator = client.listSecretVersions(request);
```
kurtisvg: nit:
```suggestion
      ListSecretVersionsPagedResponse pagedResonse = client.listSecretVersions(request);
```

secretmanager/src/main/java/com/example/ListSecretVersions.java
```diff
@@ -0,0 +1,50 @@
...
+
+      // List all versions and their state.
```
kurtisvg: Does this iterate all versions, or just the versions in this page?

secretmanager/src/main/java/com/example/ListSecretVersions.java
```diff
@@ -0,0 +1,50 @@
...
+      for (SecretVersion version : iterator.iterateAll()) {
+        System.out.printf("Secret version %s, %s\n", version.getName(), version.getState());
```
kurtisvg: Prefer using `forEach` (this is backwards in the style guide, need to update)

something like:
```java
pagedResponse.iterateAll().forEach(version -> System.out.printf("Secret version %s, %s\n", version.getName(), version.getState());
```

kurtisvg: Same elsewhere

secretmanager/src/main/java/com/example/AccessSecretVersion.java
```diff
@@ -0,0 +1,52 @@
...
+import com.google.cloud.secretmanager.v1beta1.AccessSecretVersionResponse;
+import com.google.cloud.secretmanager.v1beta1.SecretManagerServiceClient;
+import com.google.cloud.secretmanager.v1beta1.SecretVersionName;
+import java.io.IOException;
+
+public class AccessSecretVersion {
+
+  // Access the payload for the given secret version if one exists. The version
+  // can be a version number as a string (e.g. "5") or an alias (e.g. "latest").
+  public AccessSecretVersionResponse accessSecretVersion(
```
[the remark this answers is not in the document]
kurtisvg: same elsewhere

secretmanager/src/main/java/com/example/AccessSecretVersion.java
```diff
@@ -0,0 +1,52 @@
...
+      String projectId, String secretId, String versionId) throws IOException {
+    // Create a Secret Manager client with cleanup.
+    try (SecretManagerServiceClient client = SecretManagerServiceClient.create()) {
```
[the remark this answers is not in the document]
kurtisvg: same elsewhere

secretmanager/src/main/java/com/example/Quickstart.java
```diff
@@ -0,0 +1,96 @@
...
+import com.google.cloud.secretmanager.v1beta1.Replication;
+import com.google.cloud.secretmanager.v1beta1.Secret;
+import com.google.cloud.secretmanager.v1beta1.SecretManagerServiceClient;
+import com.google.cloud.secretmanager.v1beta1.SecretPayload;
+import com.google.cloud.secretmanager.v1beta1.SecretVersion;
+import com.google.protobuf.ByteString;
+
+public class Quickstart {
+
+  public static void main(String... args) throws Exception {
```
kurtisvg: We have a strong preference to not having CLIs in snippets (most java developers use an IDE, and prefer copy/pasting/running from that) . If you feel strongly about keeping it, use main instead of an overloaded function (else add an overloaded function, as mentioned above).

secretmanager/src/main/java/com/example/Snippets.java
```diff
@@ -0,0 +1,164 @@
...
+import org.kohsuke.args4j.Argument;
+import org.kohsuke.args4j.CmdLineException;
+import org.kohsuke.args4j.CmdLineParser;
+import org.kohsuke.args4j.Option;
+import org.kohsuke.args4j.spi.SubCommand;
+import org.kohsuke.args4j.spi.SubCommandHandler;
+import org.kohsuke.args4j.spi.SubCommands;
+
+public class Snippets {
+  public static void main(String... args) throws IOException {
```
kurtisvg: As I mentioned above, strong preference to not including CLI for maintainability reasons.

secretmanager/src/test/java/com/example/QuickstartIT.java
```diff
@@ -0,0 +1,81 @@
...
+  public static void beforeAll() throws Exception {
+    Assert.assertFalse("missing GOOGLE_CLOUD_PROJECT", Strings.isNullOrEmpty(PROJECT_ID));
+  }
+
+  @AfterClass
+  public static void afterAll() throws Exception {
+    // Avoid running cleanup if project id was not given. This prevents an
+    // obtuse error from appearing in test output when the user forgets to set
+    // GOOGLE_CLOUD_PROJECT.
+    if (Strings.isNullOrEmpty(PROJECT_ID)) {
```
kurtisvg: Looks like we are asserting this to be false above - should prevent this scenario from happening?

secretmanager/src/main/java/com/example/GetSecret.java
```diff
@@ -0,0 +1,55 @@
...
+
+      // Create the request.
+      GetSecretRequest request = GetSecretRequest.newBuilder().setName(name.toString()).build();
+
+      // Create the secret.
+      Secret secret = client.getSecret(request);
+
+      // Get the replication policy.
+      String replication = "";
+      if (secret.getReplication().getAutomatic() != null) {
```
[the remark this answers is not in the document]
sethvargo: Yea, it's a thwart in the API. It's not an enum because it's a `oneOf`. For example, userManaged is actually a struct with more fields.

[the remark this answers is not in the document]
sethvargo: It's my goal to add tiny hand-written helpers for `isAutomatic` and `isUserManaged` in all the libraries.

secretmanager/src/main/java/com/example/ListSecretVersions.java
sethvargo: Umm - I was under the impression that it would iterate over all objects in all pages. Am I misunderstanding? I changed it to use forEach, does that change the behavior?

secretmanager/src/main/java/com/example/Quickstart.java
sethvargo: Oh okay - I was following KMS that has executable snippets. Removed.

secretmanager/src/main/java/com/example/Snippets.java
sethvargo: Done

secretmanager/src/main/java/com/example/CreateSecret.java
```diff
@@ -0,0 +1,58 @@
...
+      // Build the parent name from the project.
+      ProjectName parent = ProjectName.of(projectId);
+
+      // Create the request.
+      CreateSecretRequest request =
+          CreateSecretRequest.newBuilder()
+              .setParent(parent.toString())
+              .setSecretId(secretId)
+              .setSecret(
+                  Secret.newBuilder()
```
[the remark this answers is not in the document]
sethvargo: Done

secretmanager/src/test/java/com/example/QuickstartIT.java
sethvargo: afterAll seems to always run, even if beforeAll fails.

secretmanager/src/main/java/com/example/AccessSecretVersion.java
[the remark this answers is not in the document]
sethvargo: Done

sethvargo:
@kurtisvg thank you for the review. I've updated the PR. Please take another look.

secretmanager/src/main/java/com/example/ListSecretVersions.java
sethvargo: Verified it's all pages. Updated.

Issue #2011, opened 2020-01-21
Move AutoML Video samples


nnegrey:
Move AutoML Video samples out of branch into master into automl directory and update the samples.

nnegrey:
Fix: https://github.com/GoogleCloudPlatform/java-docs-samples/pull/1370

PR #2014, opened 2020-01-21
automl-beta: update samples and move them to master for video

nnegrey:
Some beta samples haven't been merged to master and are on branches.
This PR takes the published V1 samples and copies them over to the beta folder for the products still in GA.

Partial fix for: #2011


nnegrey:
@stephaniewang526 ping

PR #2015, opened 2020-01-22
Adds functions_helloworld_http sample

grant:
Adds functions_helloworld_http in addition to functions_helloworld_get. (Not sure why we need both)
b/132732935

### DIFF
--- a/functions/snippets/src/main/java/HelloWorld.java
+++ b/functions/snippets/src/main/java/HelloWorld.java
@@ -22,12 +22,28 @@ import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
 
 public class HelloWorld {
+  // [END functions_helloworld_get]
+  // [START functions_helloworld_http]
+  // HTTP Cloud Function.
+  public void helloGet(HttpServletRequest request, HttpServletResponse response)
+      throws IOException {
+    String name = request.getParameter("name");
+    if (name == null) {
+      name = "World";
+    }
+    PrintWriter writer = response.getWriter();
+    writer.write(String.format("Hello %s!", name));
+  }
+  // [END functions_helloworld_http]
+  
+  // [START functions_helloworld_get]
   // Simple function to return "Hello World"
   public void helloGet(HttpServletRequest request, HttpServletResponse response)
       throws IOException {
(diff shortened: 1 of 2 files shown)


functions/snippets/src/main/java/HelloWorld.java
```diff
@@ -14,19 +14,37 @@
...
+// [START functions_helloworld_http]
 // [START functions_helloworld_get]
 import java.io.IOException;
 import java.io.PrintWriter;
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
 
 public class HelloWorld {
+  // [END functions_helloworld_get]
+  // [START functions_helloworld_http]
```
lesv: Your missing an END of this from L17

functions/snippets/src/test/java/SnippetsTests.java
```diff
@@ -81,11 +81,18 @@ public void afterTest() {
 
   @Test
   public void helloWorldTest() throws IOException {
-    new HelloWorld().helloGet(request, response);
+    new HelloWorld().helloWorld(request, response);
```
lesv: Looks like this change was wrong.

functions/snippets/src/main/java/HelloWorld.java
```diff
@@ -14,19 +14,37 @@
...
  */
 
+// [START functions_helloworld_http]
 // [START functions_helloworld_get]
 import java.io.IOException;
 import java.io.PrintWriter;
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
 
 public class HelloWorld {
```
eamonnmcmanus: This should use the new API (i.e., implement HttpFunction).

lesv:
Should we close this PR?

grant:
There have been lots of updates to the Java Functions Framework since the creation of this PR on Jan 22. I haven't focused on this language as the FF has been changing.
https://github.com/GoogleCloudPlatform/functions-framework-java

Thanks for all the comments and linking the `SAMPLE_FORMAT.md` doc. A new PR would be better to continue adding these snippets. Thus closing in favor of a new PR.

PR #2029, opened 2020-01-24
Update java8 tests to run against java11

averikitsch:
Sample edits to force sample run.


kurtisvg:
I see the following:
```
Skipping tests: Java version (${java.version}) not required (1.8, 11)
```

kurtisvg:
Looks like you've found a flaw in our test script - it's not running in either project because a property is being used to pass in the java version. 

We should probably add a specific check for if the $JAVA_POM isn't something we expect and make sure that we log that it's invalid and either fail or run the test anyway.

averikitsch:
Can I hard code the values we expect?

kurtisvg:
> Can I hard code the values we expect?

Yeah that seems reasonable.

averikitsch:
I've removed the sample. The one I chose is broken so I added an issue #2030

PR #2034, opened 2020-01-25
Update springboot app for Flex

averikitsch:


appengine-java8/helloworld-springboot/pom.xml
```diff
@@ -1,5 +1,6 @@
+<?xml version="1.0" encoding="UTF-8"?>
 <!--
-~ Copyright (c) 2013 Google Inc.
+~ Copyright (c) 2020 Google LLC
```
lesv: Doesn't need to be updated, but I guess the license checker asked you to do this.
@kurtisvg this should pass.

appengine-java8/helloworld-springboot/README.md
```diff
@@ -84,7 +84,7 @@ These settings should be revisited for production use.
 
 ## Deploy to App Engine flexible environment
 
-1. `mvn appengine:deploy`
+1. `mvn clean package appengine:deploy -Dapp.deploy.projectId=projectId`
```
lesv: Is app.deploy.projectId really required?  I thought you would just use GCLOUD_CONFIG

appengine-java8/helloworld-springboot/pom.xml
```diff
@@ -33,51 +30,52 @@
...
-  </build>
-</project>
+        <version>2.2.2.RELEASE</version>
+      </plugin>
+      <plugin>
+        <groupId>com.google.cloud.tools</groupId>
+        <artifactId>appengine-maven-plugin</artifactId>
+        <version>2.2.0</version>
+        <configuration>
+          <version>GCLOUD_CONFIG</version>
```
lesv: I think this is supposed to be for Project_ID?
https://github.com/GoogleCloudPlatform/app-maven-plugin/blob/master/USER_GUIDE.md#deploy

appengine-java8/helloworld-springboot/pom.xml
averikitsch: Since I re-wrote most of this sample, would it be ok to update this?

appengine-java8/helloworld-springboot/README.md
averikitsch: Project Id is required. I'm having the use pass it in as an arg instead of changing the pom file.

appengine-java8/helloworld-springboot/pom.xml
averikitsch: Added a comment here.

lesv:
If this is for Flex, why is it located in appengine-java8?

PR #2044, opened 2020-01-29
Heathcare API: fix resource path and console output for HL7v2MessageList

noerog:
The tests didn't catch this because the tests use their own resource path, not the path used in HL7v2MessageList.java.


noerog:
Can someone please merge this?

PR #2047, opened 2020-01-30
feat: add cli functionality to dataproc quickstart

bradmiro:
I made the Dataproc quickstart runnable and provided instructions on how to run it with maven.


bradmiro:
Given that this is an E2E tutorial, I discussed with the TW and we agreed that all of the "quickstart" samples in all languages should be runnable as-is without modifying any code (including uncommenting and setting variables). This is the only sample we intend to do this for and the rest will follow the traditional framework, [such as the createCluster sample](https://github.com/GoogleCloudPlatform/java-docs-samples/blob/master/dataproc/src/main/java/CreateCluster.java). I definitely acknowledge we're making atypical trade-offs to create this functionality.

PR #2050, opened 2020-02-03
feat: add functions_firebase_auth

grant:
Adds default Firebase Auth GCF handler.

### DIFF
--- a/functions/snippets/src/main/java/FirebaseAuth.java
+++ b/functions/snippets/src/main/java/FirebaseAuth.java
@@ -1,0 +1,32 @@
+/*
+ * Copyright 2019 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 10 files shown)


functions/snippets/src/main/java/FirebaseAuth.java
```diff
@@ -0,0 +1,32 @@
...
+ */
+
+// [START functions_firebase_auth]
+import com.google.cloud.functions.Context;
+import com.google.cloud.functions.RawBackgroundFunction;
+import java.util.logging.Logger;
+
+public class FirebaseAuth implements RawBackgroundFunction {
+
+  private static final Logger logger = Logger.getLogger(FirebaseAuth.class.getName());
```
kurtisvg:
```suggestion
  private static final Logger LOGGER = Logger.getLogger(FirebaseAuth.class);
```

functions/snippets/src/main/java/FirebaseAuth.java
```diff
@@ -0,0 +1,32 @@
...
+
+  @Override
+  public void accept(String json, Context context) {
```
kurtisvg: nit: Perhaps you should show the user how to parse the JSON?

functions/snippets/src/main/java/FirebaseAuth.java
grant: Done.

functions/snippets/src/main/java/FirebaseAuth.java
grant: That would change the functionality of the snippet and would make Java different than the other languages.
I've added a stub for parsing JSON though.

Issue #2051, opened 2020-02-03
Translate Tests failing


averikitsch:
https://fusion.corp.google.com/runanalysis/buildlogs/prod%3Acloud-devrel%2Fjava%2Fjava-docs-samples%2Fjava11%2Fcontinuous/prod%3Acloud-devrel%2Fjava%2Fjava-docs-samples%2Fjava11%2Fcontinuous/KOKORO/ffc5bf22-2fb4-476e-8c99-b443a1012cc8/0/prod%3Acloud-devrel%2Fjava%2Fjava-docs-samples%2Fjava11%2Fcontinuous/Targets

```
- testing translate/cloud-client
------------------------------------------------------------
[ERROR] Tests run: 1, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 4.822 s <<< FAILURE! - in com.example.translate.BatchTranslateTextTests
[ERROR] testBatchTranslateText(com.example.translate.BatchTranslateTextTests)  Time elapsed: 4.822 s  <<< ERROR!
java.util.concurrent.ExecutionException: com.google.api.gax.rpc.InvalidArgumentException: Operation with name "projects/779844219229/locations/us-central1/operations/20200203-14521580770358-5e36e468-0000-2932-95fe-240588736054" failed with status = GrpcStatusCode{transportCode=INVALID_ARGUMENT} and message = Output uri prefix must be an empty bucket
	at com.google.common.util.concurrent.AbstractFuture.getDoneValue(AbstractFuture.java:552)
	at com.google.common.util.concurrent.AbstractFuture.get(AbstractFuture.java:431)
	at com.google.common.util.concurrent.FluentFuture$TrustedFuture.get(FluentFuture.java:91)
	at com.google.common.util.concurrent.ForwardingFuture.get(ForwardingFuture.java:68)
	at com.google.api.gax.longrunning.OperationFutureImpl.get(OperationFutureImpl.java:133)
	at com.example.translate.BatchTranslateText.batchTranslateText(BatchTranslateText.java:89)
	at com.example.translate.BatchTranslateTextTests.testBatchTranslateText(BatchTranslateTextTests.java:104)
```

PR #2053, opened 2020-02-04
translate: use unique gcs folders for batch translate tests

nnegrey:


nnegrey:
Fix: https://github.com/GoogleCloudPlatform/java-docs-samples/issues/2051

translate/cloud-client/src/test/java/com/example/translate/BatchTranslateTextWithGlossaryAndModelTests.java
```diff
@@ -137,7 +139,7 @@ public void testBatchTranslateTextWithGlossaryAndModel()
         "en",
         "ja",
         INPUT_URI,
-        "gs://" + PROJECT_ID + "/BATCH_TRANSLATION_OUTPUT/",
+        "gs://" + PROJECT_ID + "/" + PREFIX_PATH,
```
lesv: Ok - but as I look at this, I wonder if String.format should be used as being a bit clearer, I've decided that it's just longer, but I wish we'd get rid of all the string contraction.

nnegrey: Updated to use a variable.
```
private static final String OUTPUT_URI =
          String.format("gs://%s/%s/%s/", PROJECT_ID, PREFIX, UUID.randomUUID());
```

PR #2054, opened 2020-02-06
Cleanup DLP Metadata and Jobs snippets

shubha-rajan:
Working towards #1479
Also updated integration tests.


dlp/src/main/java/dlp/snippets/InspectGcsFile.java
```diff
@@ -51,7 +51,7 @@
   public static void inspectGcsFile() throws InterruptedException, ExecutionException, IOException {
     // TODO(developer): Replace these variables before running the sample.
     String projectId = "your-project-id";
-    String gcsUri = "gs://" + "your-bucket-name" + "/path/to/your/image.png";
+    String gcsUri = "gs://" + "your-bucket-name" + "/path/to/your/test.txt";
```
kurtisvg: nit: 
```suggestion
    String gcsUri = "gs://" + "your-bucket-name" + "/path/to/your/file.txt";
```

dlp/src/main/java/dlp/snippets/InfoTypesList.java
```diff
@@ -0,0 +1,61 @@
...
+
+import java.io.IOException;
+import java.util.List;
+
+public class InfoTypesList  {
+
+    public static void listInfoTypes() throws IOException {
+        // TODO(developer): Replace these variables before running the sample.
+        String filter = "supported_by=INSPECT";
+        String languageCode= "en-US";
```
kurtisvg: These two fields (`filter` and `languageCode`) aren't required that the user set to run, so we should inline them in the sample below instead.

dlp/src/main/java/dlp/snippets/InfoTypesList.java
```diff
@@ -0,0 +1,61 @@
...
+        try (DlpServiceClient dlpClient = DlpServiceClient.create()) {
+
+            // Construct the request to be sent by the client
+            ListInfoTypesRequest listInfoTypesRequest =
+                    ListInfoTypesRequest.newBuilder().setFilter(filter).setLanguageCode(languageCode).build();
+
+            // Use the client to send the API request.
+            ListInfoTypesResponse infoTypesResponse = dlpClient.listInfoTypes(listInfoTypesRequest);
+
+            // Parse the response and process the results */
```
kurtisvg:
```suggestion
            // Parse the response and process the results
```

dlp/src/main/java/dlp/snippets/InfoTypesList.java
```diff
@@ -0,0 +1,61 @@
...
+            List<InfoTypeDescription> infoTypeDescriptions = infoTypesResponse.getInfoTypesList();
+            for (InfoTypeDescription infoTypeDescription : infoTypeDescriptions) {
```
kurtisvg: Consider inlining the `getInfoTypesList` and adding a top level print statement here:
System.out.println("Infotypes found:");
```suggestion
            for (InfoTypeDescription infoTypeDescription : response.getInfoTypesList()) {
```

dlp/src/main/java/dlp/snippets/InfoTypesList.java
kurtisvg: We should also add links to the docs for these resources, so users know other options that are available for them.

dlp/src/main/java/dlp/snippets/InspectTextFile.java
```diff
@@ -37,7 +37,7 @@
   public static void inspectTextFile() {
     // TODO(developer): Replace these variables before running the sample.
     String projectId = "my-project-id";
-    String filePath = "path/to/image.png";
+    String filePath = "path/to/textfile.txt";
```
kurtisvg: super nit:
```suggestion
    String filePath = "path/to/text/file.txt";
```

dlp/src/main/java/dlp/snippets/JobsDelete.java
```diff
@@ -0,0 +1,55 @@
...
+import com.google.cloud.dlp.v2.DlpServiceClient;
+import com.google.privacy.dlp.v2.DeleteDlpJobRequest;
+import com.google.privacy.dlp.v2.DlpJobName;
+
+import java.io.IOException;
+
+public class JobsDelete  {
+    public static void deleteJobs() throws IOException {
+        // TODO(developer): Replace these variables before running the sample.
+        String projectID = "your-project-id";
```
kurtisvg: nit: Do we use caps in other places? We should be consistent with values we expect users to replace.

dlp/src/main/java/dlp/snippets/JobsList.java
```diff
@@ -0,0 +1,65 @@
...
+
+public class JobsList {
+
+    public static void listJobs() throws IOException {
+        // TODO(developer): Replace these variables before running the sample.
+        // For more info on filters and job types,
+        // see https://cloud.google.com/dlp/docs/reference/rest/v2/projects.dlpJobs/list
+        String projectId = "your-project-id";
+        String filter = "state=DONE";
+        String jobType = "INSPECT_JOB";
```
kurtisvg: Same here - please move `filter` and `jobType` to the sample

dlp/src/test/java/dlp/snippets/InfoTypesTests.java
```diff
@@ -0,0 +1,66 @@
...
+    public void tearDown() {
+        System.setOut(null);
+        bout.reset();
+    }
+
+    @Test
+    public void testListInfoTypes() throws Exception {
+        InfoTypesList.listInfoTypes("supported_by=INSPECT", "en-US");
+        String output = bout.toString();
+        assertTrue(output.contains("Name") && output.contains("Display name"));
```
kurtisvg: protip: These tests are easier to debug if you use this format:
```java
Assert.assertThat(x, CoreMatchers.containsString("foo"));
```

The error message will be much more descriptive and easier to debug (Expected x to contain foo, but was "" instead).

dlp/src/test/java/dlp/snippets/RedactTests.java
```diff
@@ -62,10 +67,14 @@ public void tearDown() {
...
   @Test
-  public void testRedactImage() {
+  public void testRedactImage() throws IOException {
     RedactImageFile.redactImageFile(PROJECT_ID, "src/test/resources/test.png");
 
     String output = bout.toString();
     assertThat(output, containsString("Redacted image written"));
+
+    // Clean up test output
+    Path outputFile = Paths.get("redacted.png");
```
kurtisvg: This should be in a separate cleanup function with the `After` tag, so it still runs of the test fails.

dlp/src/test/java/dlp/snippets/JobsTests.java
```diff
@@ -0,0 +1,113 @@
...
+    @After
+    public void tearDown() {
+        System.setOut(null);
+        bout.reset();
+    }
+
+    @Test
+    public void testListJobs() throws Exception {
+        // Ensure that there is at least one job to list
+        InspectGcsFile.inspectGcsFile(PROJECT_ID,
```
kurtisvg: If there isn't a job, we should make sure there is with a `setUp` function that creates one that runs first.

Or, alternatively make the test pass if no jobs are found.

dlp/src/test/java/dlp/snippets/RedactTests.java
shubha-rajan: moved this to a cleanup function in 65be2be

dlp/src/test/java/dlp/snippets/InfoTypesTests.java
shubha-rajan: reformatted to use `Assert.assertThat()` in 66b34ca

dlp/src/main/java/dlp/snippets/JobsList.java
shubha-rajan: inlined these in the sample in 66b34ca

dlp/src/main/java/dlp/snippets/JobsDelete.java
shubha-rajan: The existing reformatted snippets weren't consistent with this. I updated all of them to use lowercase letters in the last commit, because project ids tend to be lowercase, but I can also make them all uppercase if needed

dlp/src/main/java/dlp/snippets/InfoTypesList.java
shubha-rajan: I inlined them and added comments to provide context on other options, but couldn't really find a good link for more info on filters. I got the supported filter types from here: https://cloud.google.com/dlp/docs/reference/rpc/google.privacy.dlp.v2#google.privacy.dlp.v2.InfoTypeSupportedBy

dlp/src/test/java/dlp/snippets/JobsTests.java
shubha-rajan: I think the `setUp` approach is better, since the delete method is tested in the same class, and that has to have an existing job so that we can test it.

kurtisvg:
Hmm looks like `testDeleteJobs` and `testListJobs` are both failing with null pointers. Any idea why this might be?

shubha-rajan:
> Hmm looks like `testDeleteJobs` and `testListJobs` are both failing with null pointers. Any idea why this might be?

Took me a sec to figure this one out since they were passing locally for me. You need to set a GCS_PATH env variable that links to where you upload the test.txt file. `checkRequirements()` should have caught that and printed out an error message, but it didn't run before 'setUp()' did.  I updated CheckRequirements to use the @BeforeClass tag instead

shubha-rajan:
Sort of on a related note: is there a Java way of making something like a config.py or a .env so we don't have to manually set so many env variables?

dlp/src/test/java/dlp/snippets/JobsTests.java
```diff
@@ -0,0 +1,105 @@
...
+
+import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.assertTrue;
+
+@RunWith(JUnit4.class)
+public class JobsTests {
+
+    private static final Pattern JOB_ID_PATTERN = Pattern.compile("projects/.*/dlpJobs/i-\\d+");
+    private static final String PROJECT_ID = System.getenv("GOOGLE_CLOUD_PROJECT");
+    private static final String GCS_PATH = "gs://" + PROJECT_ID + "/dlp";
```
kurtisvg: I think you were probably right the fist time and this should be a provided env var. More generally, any time we ask the user to provide a specific resources we should probably be asking for an environment variable in a test. 

This probably applies to `PUB_SUB_TOPIC_ID` and `PUB_SUB_SUBSCRIPTION_ID` as well. 

If you want to address this in this PR, I can walk you through how to set these during our next 1:1. If not, please open an issue and assign it to yourself and we'll fix at a later time.

dlp/src/test/java/dlp/snippets/JobsTests.java
```diff
@@ -0,0 +1,105 @@
...
+
+        // Check that the output contains jobIds
+        Matcher matcher = JOB_ID_PATTERN.matcher(bout.toString());
+        assertTrue("List must contain results.", matcher.find());
+    }
+
+    @Test
+    public void testDeleteJobs() throws Exception {
+        // Get a list of JobIds, and extract one to delete
+        JobsList.listJobs(PROJECT_ID);
```
kurtisvg: I think this pattern is going to introduce a race condition when this test is run in parallel. It's possible while each test is going to create a new job, they could both make a request and get the same job, which one might fail. 

Additionally, JUNIT doesn't actually guarantee any specific order for tests. deletejobs could run after listjobs.

I think this means we have to do 2 things:
1. Update ListJobs to pass as long as the request was successful (even if it's empty)
2. Create a job with a specific UUID that the deleteJob can delete.

shubha-rajan: Made these changes! One interesting thing to note, after some debugging and inspecting in pantheon, i found out I had to append "i-" to the UUID string I generated in order to get `deleteJob` to work. Is this possibly a bug in the `DlpJobName.of()` method?

dlp/src/test/java/dlp/snippets/JobsTests.java
shubha-rajan: Decided to keep the env vars. Let's go over how to set these for kokoro on Monday.

kurtisvg:
Looks like a few tests are failing yet - can you take a look?

dlp/src/test/java/dlp/snippets/JobsTests.java
kurtisvg: It could be - can you create a short script to show the issue and open an issue against the repo that contains it?

kurtisvg:
Forcing merge - failing tests are still being tracked in #1479

Issue #2132, opened 2020-02-18
Datacatalog Build Failing


nnegrey:
GoogleCloudPlatform/java-docs-samples#2102

https://source.cloud.google.com/results/invocations/9479d637-f0f9-4e16-bc22-d0d28f83d079/targets

https://source.cloud.google.com/results/invocations/0eb0ea20-3c6d-4d29-b6f0-92bc7920ae29/targets

Reaching out to sample writer

nnegrey:
@mesmacosta

lesv:
@nnegrey what is the status of this issue?

PR #2137, opened 2020-02-18
Fix breaking changes on Data Catalog library update

mesmacosta:
On the new Data Catalog library version there was a breaking change on the package naming, to make it clearer what version is being used, hence the tests started failing. Updating the sample codes as requested on the PR: #2102


nnegrey:
Fix: https://github.com/GoogleCloudPlatform/java-docs-samples/issues/2132

nnegrey:
@mesmacosta, it seems these tests don't seem to quite finish? Are they just really long running?

mesmacosta:
@nnegrey They are really quick, you have to tag the PR with a label kokoro:run to start the tests on the PR.

PR #2166, opened 2020-02-20
fix:  bump arrow-vector to address integration failures

shollyman:
DO NOT SUBMIT

related: https://issues.apache.org/jira/browse/ARROW-5412


shollyman:
Bungled branches between efforts, going to abandon this one and send a followup PR.

PR #2188, opened 2020-02-21
second part 2 of Cloud Talent solution v4

munkhuushmgl:
There is no unit test for CreateCLientEvent because it exposes full path of jobs.


jobs/v4/src/main/java/com/example/jobs/JobSearchCommuteSearch.java
```diff
@@ -0,0 +1,94 @@
...
+import com.google.cloud.talent.v4beta1.SearchJobsRequest;
+import com.google.cloud.talent.v4beta1.SearchJobsResponse;
+import com.google.cloud.talent.v4beta1.TenantName;
+import com.google.cloud.talent.v4beta1.TenantOrProjectName;
+import com.google.protobuf.Duration;
+import com.google.type.LatLng;
+
+import java.io.IOException;
+
+public class JobSearchCommuteSearch {
```
lesv: nit - class name seems a bit wordy.  JobByCommuteSearch perhaps.  (not required)

munkhuushmgl: Only class or all other search classes in general? I can remove the prefix [JobSearch] from Search classes

PR #2243, opened 2020-02-26
Renovate/com.google.api client google api client 1.x

lesv:
1. Update to Java 11
2. Get rid of Findbugs (no longer supported)
3. WAR's should use the latest war plugin.
4. Remvoe a bad / wrong import in MqttExampleOptions

PR replaces #2070 which was broken.

### DIFF
--- a/compute/cmdline/pom.xml
+++ b/compute/cmdline/pom.xml
@@ -30,8 +30,8 @@ limitations under the License.
   </parent>
 
   <properties>
-    <maven.compiler.target>1.8</maven.compiler.target>
-    <maven.compiler.source>1.8</maven.compiler.source>
+    <maven.compiler.target>11</maven.compiler.target>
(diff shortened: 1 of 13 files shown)


compute/cmdline/pom.xml
```diff
@@ -30,8 +30,8 @@ limitations under the License.
   </parent>
 
   <properties>
-    <maven.compiler.target>1.8</maven.compiler.target>
-    <maven.compiler.source>1.8</maven.compiler.source>
+    <maven.compiler.target>11</maven.compiler.target>
```
kurtisvg: Why are we bumping these to 11?

lesv: It's supported.  I had to fix stuff anyway, and it seemed like a reasonable thing to do.

lesv: Reverted - PTAL

PR #2282, opened 2020-02-29
Corrected local execution cmd-line

kintel:
Looks like bad copypasta. The updated cmd-line works correctly for me.


lesv:
Kokoro pretends it likes this, but I never see results.

PR #2300, opened 2020-03-03
video: move samples out of branch

nnegrey:


video/beta/src/main/java/com/example/video/DetectLogo.java
```diff
@@ -0,0 +1,147 @@
+/*
+ * Copyright 2019 Google LLC
```
lesv: Should this be 2020?

nnegrey: ooops yes.

PR #2301, opened 2020-03-03
automl: move video object tracking samples out of branch

nnegrey:


lesv:
Oops - missed that this was a draft

PR #2347, opened 2020-03-06
feat: add QueryOptions samples for JDBC

olavloite:
Adds samples for `QueryOptions` for Cloud Spanner JDBC. This PR also adds basic infrastructure for samples for Cloud Spanner JDBC.

The PR currently fails to build as the changes that are needed have not yet been added to the JDBC driver.


spanner/jdbc/README.md
```diff
@@ -0,0 +1,36 @@
+# Cloud Spanner JDBC Example
```
lesv: Nice, simple and clear!

spanner/jdbc/pom.xml
```diff
@@ -0,0 +1,94 @@
...
+  </properties>
+
+  <!--
+    The parent pom defines common style checks and testing strategies for our samples.
+    Removing or replacing it should not affect the execution of the samples in anyway.
+  -->
+  <parent>
+    <groupId>com.google.cloud.samples</groupId>
+    <artifactId>shared-configuration</artifactId>
+    <version>1.0.11</version>
```
lesv: Try 1.0.12 - we are trying to roll it out.

spanner/jdbc/pom.xml
```diff
@@ -0,0 +1,94 @@
...
+    <dependency>
+      <groupId>io.grpc</groupId>
+      <artifactId>grpc-api</artifactId>
+      <version>1.27.2</version>
+    </dependency>
+    <!-- The Spanner JDBC driver dependency -->
+    <dependency>
+      <groupId>com.google.cloud</groupId>
+      <artifactId>google-cloud-spanner-jdbc</artifactId>
+      <version>1.13.1-SNAPSHOT</version>
```
lesv: Don't use -SNAPSHOT's

spanner/jdbc/pom.xml
```diff
@@ -0,0 +1,94 @@
...
+  <!-- [START spanner_jdbc_snapshot_repo] -->
+  <repositories>
+    <repository>
+      <id>snapshots-repo</id>
+      <url>https://oss.sonatype.org/content/repositories/snapshots</url>
+      <releases><enabled>false</enabled></releases>
+      <snapshots><enabled>true</enabled></snapshots>
+    </repository>
+  </repositories>
+  <!-- [END spanner_jdbc_snapshot_repo] -->
```
lesv: Not a good idea.

spanner/jdbc/src/main/java/com/example/spanner/JdbcSample.java
```diff
@@ -0,0 +1,214 @@
...
+
+    String command = args[0];
+    String projectId = SpannerOptions.getDefaultProjectId();
+    String instanceId = args[1];
+    String databaseId = args[2];
+
+    String connectionUrl =
+        String.format(
+            "jdbc:cloudspanner:/projects/%s/instances/%s/databases/%s",
+            projectId, instanceId, databaseId);
```
lesv: Nice!

spanner/jdbc/src/main/java/com/example/spanner/JdbcSample.java
```diff
@@ -0,0 +1,214 @@
...
+    System.err.println("    JdbcSample <command> <instance_id> <database_id>");
+    System.err.println("");
+    System.err.println("Examples:");
+    System.err.println("    JdbcSample createtable my-instance example-db");
+    System.err.println("    JdbcSample insertdata my-instance example-db");
+    System.err.println("    JdbcSample connectionwithqueryoptions my-instance example-db");
+    System.err.println("    JdbcSample setqueryoptions my-instance example-db");
+
+    System.exit(1);
+  }
```
lesv: I think we try to avoid using CLI tools for java samples.  I'll approve this since I couldn't find a prohibition in https://github.com/GoogleCloudPlatform/java-docs-samples/blob/master/SAMPLE_FORMAT.md

spanner/jdbc/src/test/java/com/example/spanner/JdbcSampleIT.java
```diff
@@ -0,0 +1,91 @@
...
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.junit.runners.JUnit4;
+
+/** Unit tests for {@code JdbcSample} */
+@RunWith(JUnit4.class)
+public class JdbcSampleIT {
+  // The instance needs to exist for tests to pass.
+  private final String instanceId = System.getProperty("spanner.test.instance");
+  private final String databaseId = formatForTest(System.getProperty("spanner.sample.database"));
```
lesv: Do we need to add something in Java-Docs-Samples-Testing for this?

olavloite:
> There is only one or two things that need any action on.
> 
> Overall I like this. Let me know what we might need to do the get testing to pass. It will probably be easier if we use EnvVar's instead of java properties. And we really don't want customers to consider using -SNAPSHOT's.

@lesv 
This PR will only build once https://github.com/googleapis/java-spanner-jdbc/pull/76 has been merged. There's nothing actionable for you or your team to make this pass at this moment.

spanner/jdbc/README.md
olavloite: Thanks :-)

spanner/jdbc/pom.xml
olavloite: Done. Works on my local development environment, so I assume it will also work once the PR this depends on has been merged into the JDBC driver.

spanner/jdbc/pom.xml
olavloite: I'll change that as soon as a version of the JDBC driver with support for `QueryOptions` has been released. Until then this PR won't build anyways, so it can't be merged by accident either.

spanner/jdbc/pom.xml
olavloite: Removed.

spanner/jdbc/src/test/java/com/example/spanner/JdbcSampleIT.java
olavloite: No, AFAICT this should work out of the box. This is also modeled to work exactly as the existing `SpannerSampleIT`: https://github.com/GoogleCloudPlatform/java-docs-samples/blob/c3c4600b0848a48dece316cf809f95721c427f44/spanner/cloud-client/src/test/java/com/example/spanner/SpannerSampleIT.java#L39-L40

spanner/jdbc/src/main/java/com/example/spanner/JdbcSample.java
olavloite: This is modeled to work in the same way as the existing `SpannerSample` samples: https://github.com/GoogleCloudPlatform/java-docs-samples/blob/c3c4600b0848a48dece316cf809f95721c427f44/spanner/cloud-client/src/main/java/com/example/spanner/SpannerSample.java#L1597-L1610

spanner/jdbc/src/main/java/com/example/spanner/JdbcSample.java
```diff
@@ -0,0 +1,214 @@
...
+      System.out.printf("Insert counts: %s\n", Arrays.toString(updateCounts));
+    } finally {
+      connection.setAutoCommit(originalAutoCommit);
+    }
+  }
+
+  // [START spanner_jdbc_connection_with_query_options]
+  static void connectionWithQueryOptions(String projectId, String instanceId, String databaseId)
+      throws SQLException {
+    String optimizerVersion = "LATEST";
```
skuruppu:
```suggestion
    String optimizerVersion = "1";
```

skuruppu:
@olavloite if you could look at the tests that would be great. Looks like there are some failures.

spanner/jdbc/src/main/java/com/example/spanner/JdbcSample.java
olavloite: Done

olavloite:
@lesv 
> Overall I like this. Let me know what we might need to do the get testing to pass. It will probably be easier if we use EnvVar's instead of java properties. And we really don't want customers to consider using -SNAPSHOT's.

All the required dependencies have now been merged, so it should now be possible to get the test cases up and running. For that the integration test needs access to a Spanner instance and database. I assumed that the Java system properties `spanner.test.instance` and `spanner.test.database` would automatically have been set by the build environment, as those are also used by the `SpannerSampleIT`, but apparently it's not. I have no specific preference for either a Java system property or an Environment Variable. Could you arrange for one or the other being set for the integration tests of this build? If it's an EnvVar, I'll modify the IT accordingly to use that.

spanner/jdbc/pom.xml
```diff
@@ -0,0 +1,72 @@
...
+  </properties>
+
+  <!--
+    The parent pom defines common style checks and testing strategies for our samples.
+    Removing or replacing it should not affect the execution of the samples in anyway.
+  -->
+  <parent>
+    <groupId>com.google.cloud.samples</groupId>
+    <artifactId>shared-configuration</artifactId>
+    <version>1.0.12</version>
```
lesv: 1.0.13 now

spanner/jdbc/pom.xml
```diff
@@ -0,0 +1,72 @@
...
+    <dependency>
+      <groupId>junit</groupId>
+      <artifactId>junit</artifactId>
+      <version>4.13</version>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>com.google.truth</groupId>
+      <artifactId>truth</artifactId>
+      <version>1.0</version>
```
lesv: 1.0.1

lesv:
Great - other than it doesn't compile.

olavloite:
> Great - other than it doesn't compile.

It depends on `System.getProperty("spanner.test.instance")` and `System.getProperty("spanner.sample.database")` which are available for the normal Spanner samples build, but apparently not for this build. I'm open for any other way way of getting a reference to a valid test instance, such as the previously mentioned EnvVar.

lesv:
You can hard code these in your test.  If you need me to do something in java-docs-samples-testing, let me know.  We have default-instance in us-central1 and the following DB's:
```

  | Name | CPU utilization | Size |  
-- | -- | -- | -- | --
  | mysample-4ca24702-bd50-41a4-a | 0.62% | 474 B |  
  | mysample-73af71b0-b212-4794-b | 0.38% | 603 B |  
  | quickstart-db | 0.32% | 89 B |  
  | restored-mysample-16ae8df9-7ec | 0.23% | 1.81 KB |  
  | sejzn-3d2384a5-1344-4bff-8 | 0.29% | 13.46 KB |  
  | vzdqp-da80840c-38c5-430b-a
```

spanner/jdbc/pom.xml
olavloite: Done.

spanner/jdbc/pom.xml
olavloite: Done

PR #2360, opened 2020-03-09
feat: dataproc instantiate inline workflow sample

bradmiro:
Adding a sample showing how to instantiate an inline workflow for dataproc.

### DIFF
--- a/dataproc/src/main/java/InstantiateInlineWorkflow.java
+++ b/dataproc/src/main/java/InstantiateInlineWorkflow.java
@@ -1,0 +1,118 @@
+/*
+ * Copyright 2020 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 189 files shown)


dataproc/src/main/java/InstantiateInlineWorkflow.java
```diff
@@ -0,0 +1,118 @@
...
+    // Create a workflow template service client with the configured settings. The client only
+    // needs to be created once and can be reused for multiple requests. Using a try-with-resources
+    // closes the client, but this can also be done manually with the .close() method.
+    try (WorkflowTemplateServiceClient workflowTemplateServiceClient =
+        WorkflowTemplateServiceClient.create(workflowTemplateServiceSettings)) {
+
+      // Configure the jobs within the workflow.
+      HadoopJob teragenHadoopJob =
+          HadoopJob.newBuilder()
+              .setMainJarFileUri("file:///usr/lib/hadoop-mapreduce/hadoop-mapreduce-examples.jar")
```
lesv: Is this pre-installed?
Why is this a useful example if it's already there?

bradmiro: The point we're trying to get across is how you can structure a workflow and run it against the cluster. We sometimes use these toy examples instead of say a PySpark example on GCS. 

I'm also referencing the canonical YAML example from the [docs](https://cloud.google.com/dataproc/docs/concepts/workflows/using-yamls) (which is not to say this sample couldn't be improved either).

lesv: SG

lesv:
Do you need me to merge?

bradmiro:
I can merge! I'm finishing these up in a couple other languages and will merge when all are done, probably tomorrow :) thanks!

PR #2373, opened 2020-03-11
dataflow/flex-templates: Add Streaming BeamSQL and Kafka-to-BigQuery samples

davidcavazos:
Adding samples for Flex Templates


davidcavazos:
There are 2 samples:

1. Pub/Sub to BigQuery
2. Kafka to BigQuery

@anguillanneuf can you take a look at the Pub/Sub sections?
@stephaniewang526 can you take a look at the BigQuery sections?

stephaniewang526:
The BQ part looks like it makes sense - however you're using Apache Beam/ BQ APIs instead of BQ Java client (which I write samples for) so I will ask @lesv or @kurtisvg to take a closer look to make sure it's all good.

dataflow/flex-templates/kafka_to_bigquery/.gcloudignore
```diff
@@ -0,0 +1,7 @@
+# Ignore everything from the Cloud Build command.
```
lesv: A bit unexpected and incompatible with CloudBuild - you might wish to mention that in the README, but otherwise OK.

dataflow/flex-templates/kafka_to_bigquery/README.md
```diff
@@ -0,0 +1,337 @@
...
+2. An Apache Beam streaming pipeline running in Dataflow Flex Templates.
+   This subscribes to a Kafka topic, consumes the messages that are published
+   to that topic, processes them, and writes them into a BigQuery table.
+
+    * [Dockerfile](Dockerfile)
+    * [KafkaToBigQuery.java](src/main/java/org/apache/beam/samples/KafkaToBigQuery.java)
+    * [pom.xml](pom.xml)
+    * [metadata.json](metadata.json)
+
+### Starting the Kafka server
```
lesv: Wow that's a lot of steps, I wonder if it wouldn't be easier on users to just use a script that asks them what it needs to know?

dataflow/flex-templates/kafka_to_bigquery/Dockerfile
```diff
@@ -0,0 +1,8 @@
+FROM gcr.io/dataflow-templates-base/java11-template-launcher-base:latest
```
lesv: Copyright & License?

dataflow/flex-templates/kafka_to_bigquery/.gcloudignore
lesv: Copyright & Licenses - general rule - if a human creates it, then it needs a copyright & license.

dataflow/flex-templates/kafka_to_bigquery/kafka/Dockerfile
```diff
@@ -0,0 +1,37 @@
+FROM openjdk:8-jre-alpine
```
lesv: Copyright & License?

dataflow/flex-templates/kafka_to_bigquery/kafka/create-topic.sh
```diff
@@ -0,0 +1,20 @@
+#!/bin/sh
```
lesv: Copyright & License?

dataflow/flex-templates/kafka_to_bigquery/kafka/start-kafka.sh
```diff
@@ -0,0 +1,48 @@
+#!/bin/sh
+
```
lesv: Copyright & License?

dataflow/flex-templates/kafka_to_bigquery/metadata.json
```diff
@@ -0,0 +1,32 @@
+{
```
lesv: Ideally should have Copyright & License unless whatever reading it has a problem w/ that.
 "_Copyright" : "Copyright 2020 Google, LLC",
"_License":"....",
If that doesn't work, then it's ok to skip.

dataflow/flex-templates/streaming_beam_sql/.gcloudignore
```diff
@@ -0,0 +1,7 @@
+# Ignore everything from the Cloud Build command.
```
lesv: You might wish to mention this in the README; Copyright & License

dataflow/flex-templates/streaming_beam_sql/Dockerfile
```diff
@@ -0,0 +1,8 @@
+FROM gcr.io/dataflow-templates-base/java11-template-launcher-base:latest
```
lesv: Copyright & License

dataflow/flex-templates/streaming_beam_sql/README.md
```diff
@@ -0,0 +1,315 @@
...
+
+```sh
+export TEMPLATE_IMAGE="$PROJECT/samples/dataflow/streaming-beam-sql:latest"
+
+# Build the image into Container Registry, this is roughly equivalent to:
+#   gcloud auth configure-docker
+#   docker image build -t $TEMPLATE_IMAGE .
+#   docker push $TEMPLATE_IMAGE
+gcloud builds submit --tag "gcr.io/$TEMPLATE_IMAGE" .
+```
```
lesv: So this is interesting w/ that .cloudignore above -- I presume you've verified that things work.

dataflow/flex-templates/streaming_beam_sql/README.md
```diff
@@ -0,0 +1,315 @@
...
+{"url": "https://cloud.google.com/bigquery/", "review": "positive"}
+```
+
+### Cleaning up
+
+After you've finished this tutorial, you can clean up the resources you created
+on Google Cloud so you won't be billed for them in the future.
+The following sections describe how to delete or turn off these resources.
+
+#### Clean up the Flex template resources
```
lesv: I wonder if you shouldn't just have a cleanup.sh script somewhere?

dataflow/flex-templates/streaming_beam_sql/README.md
```diff
@@ -0,0 +1,315 @@
...
+
+[![Open in Cloud Shell](http://gstatic.com/cloudssh/images/open-btn.svg)](https://console.cloud.google.com/cloudshell/editor)
+
+📝 Docs: [Using Flex Templates](https://cloud.google.com/dataflow/docs/guides/templates/using-flex-templates)
+
+Samples showing how to create and run an
+[Apache Beam](https://beam.apache.org/) template with a custom Docker image on
+[Google Cloud Dataflow](https://cloud.google.com/dataflow/docs/).
+
+## Before you begin
```
lesv: Wow - this is long, I understand your trying to teach here, but I wonder if providing a script that does it and explains as it runs might be easier on users.  `./setup.sh`

dataflow/flex-templates/streaming_beam_sql/metadata.json
```diff
@@ -0,0 +1,23 @@
+{
```
lesv: Same comment about Copyright & Licenses at earlier.

dataflow/flex-templates/streaming_beam_sql/pom.xml
```diff
@@ -0,0 +1,175 @@
...
+      <name>Apache Development Snapshot Repository</name>
+      <url>https://repository.apache.org/content/repositories/snapshots/</url>
+      <releases>
+        <enabled>false</enabled>
+      </releases>
+      <snapshots>
+        <enabled>true</enabled>
+      </snapshots>
+    </repository>
+  </repositories>
```
lesv: Why are we sending users to SNAPSHOT repositories?  That's not really good practice.

dataflow/flex-templates/streaming_beam_sql/README.md
davidcavazos: I agree, it's a lot. I played a bit with creating a script, but now I would have to handle user confirmations when creating/deleting resources and much more validation. That would also make doing the setup/cleanup harder to port to windows. I will leave the instructions in the README for simplicity for now.

These instructions are also present in the docs, so we would prefer to keep them as compatible as possible.

dataflow/flex-templates/kafka_to_bigquery/README.md
davidcavazos: Decided not to script this in the end. Find explanation below.

dataflow/flex-templates/streaming_beam_sql/README.md
lesv: Why not link to the docs then?  Though I prefer README's where it's included, but a single source of truth is also a good idea.

dataflow/flex-templates/streaming_beam_sql/pom.xml
lesv: Typically yes-  *-SNAPSHOT's are often not really versioned or rebuildable.  They are really only useful (on an OSS project) if you are following top of trunk, but it leads to non reproducible builds.

dataflow/flex-templates/streaming_beam_sql/README.md
davidcavazos: They want to merge fast for the launch and I don't want to stall the launch since a script like that would take several weeks to develop. I agree that a script would be very handy, but at this point we all just want this to merge and get it over with.

dataflow/flex-templates/streaming_beam_sql/pom.xml
davidcavazos: Got it, removed. Thank you!

dataflow/flex-templates/streaming_beam_sql/README.md
davidcavazos: Hi @lesv, sorry for my last response, I was a little frustrated with how all the deadlines the product team wanted to meet. I did play around for just under an hour trying to implement a setup script, here's what I learned from it.

* We have to check if a resource already exists before creating it
* We need to ask for user confirmation before creating any resources
* Right now Windows users only have to make small changes to the commands, mainly the `\` for line continuations. But having a bash script would make it much more harder for them to run the sample.
* All the complexity adds up and the script was turning out to be as large as the sample itself
* Also having the explanations in Markdown makes the text clearer to read with all the nice formatting
* Most of the instructions (except for Cloud Scheduler), are actually instructions that users need to follow to use Flex Templates, so it makes sense to have them in the README
* The docs are somewhat simplified to make them simpler and easier to follow, while the README has more detail on some of the commands and why/how to run them, so it makes sense to keep both.

I had a chat with Meredith and we agreed to keep it like this for now. If we see there is a need from users in the future, we can allocate extra resources into creating these scripts.

lesv: Sg - I already approved the PR

lesv:
Let me know if you need me to merge.

PR #2378, opened 2020-03-11
vision: update product search tests

nnegrey:
Fix: https://github.com/GoogleCloudPlatform/java-docs-samples/issues/2355


nnegrey:
Gotta update another test to use a UUID

PR #2394, opened 2020-03-13
GCF: Add Slack sample + clean up imports

ace-n:
~**Do not merge** until this has actually been tested with Slack.~

(@lesv I added you since Kurtis is OOO - LMK if I should ping someone else!)


functions/snippets/pom.xml
```diff
@@ -26,20 +26,32 @@
...
-    <!-- Required to for com.example.functions.ParseContentType -->
+    <!-- Required for com.example.functions.ParseContentType -->
     <dependency>
       <groupId>com.google.code.gson</groupId>
       <artifactId>gson</artifactId>
       <version>2.8.6</version>
     </dependency>
 
-    <!-- Required to for com.example.functions.RetrieveLogs -->
+    <!-- Required for com.example.functions.RetrieveLogs -->
```
lesv: I would hope this isn't true.  Can't you also write to stderr / stdout?

functions/snippets/src/main/java/com/example/functions/SlackSlashCommand.java
```diff
@@ -0,0 +1,172 @@
...
+  private static final Logger LOGGER = Logger.getLogger(HelloHttp.class.getName());
+  private SlackSignature.Verifier verifier;
+  private Gson gson = new Gson();
+
+  public SlackSlashCommand() throws IOException, GeneralSecurityException {
+    kgClient = new Kgsearch.Builder(
+        GoogleNetHttpTransport.newTrustedTransport(), new JacksonFactory(), null).build();
+
+    // Read + parse config file
+    Path configPath = Path.of(System.getProperty("user.dir"), "config.json");
```
lesv: OK, but isn't this against 12 factor?  (Shouldn't config come from EnvVar's)?

functions/snippets/src/main/java/com/example/functions/SlackSlashCommand.java
```diff
@@ -0,0 +1,172 @@
...
+    if (kgResults.size() == 0) {
+      attachmentJson.addProperty("text","No results match your query...");
+
+      attachments.add(attachmentJson);
+      responseJson.add("attachments", attachmentJson);
+
+      return gson.toJson(responseJson);
+    }
+
+    ArrayMap entity = (ArrayMap) ((ArrayMap) kgResults.get(0)).get("result");
```
lesv: Not sure what to recommend here - this has lots of stuff wrong.

functions/snippets/src/main/java/com/example/functions/SlackSlashCommand.java
```diff
@@ -0,0 +1,172 @@
...
+  }
+
+  private void addPropertyIfPresent(
+      JsonObject target, String targetName, ArrayMap source, String sourceName) {
+    if (source.containsKey(sourceName)) {
+      target.addProperty(targetName, source.get(sourceName).toString());
+    }
+  }
+
+  private String formatSlackMessage(List<Object> kgResults, String query) {
```
lesv: List<Object> is rarely a good idea.

functions/snippets/src/main/java/com/example/functions/SlackSlashCommand.java
```diff
@@ -0,0 +1,172 @@
...
+      return gson.toJson(responseJson);
+    }
+
+    ArrayMap entity = (ArrayMap) ((ArrayMap) kgResults.get(0)).get("result");
+
+    // Construct Knowledge Graph response attachment
+    String title = entity.get("name").toString();
+    if (entity.containsKey("description")) {
+      title = String.format("%s: %s", title, entity.get("description").toString());
+    }
```
lesv: Lots of .toString()'s are often a flag of bad design.  .toString() on an object isn't guaranteed to provide something presentable.

functions/snippets/src/main/java/com/example/functions/SlackSlashCommand.java
```diff
@@ -0,0 +1,172 @@
...
+
+    // Construct top level response
+    attachments.add(attachmentJson);
+    responseJson.add("attachments", attachmentJson);
+
+    return gson.toJson(responseJson);
+  }
+
+  private List<Object> searchKnowledgeGraph(String query) throws IOException {
```
lesv: You rarely need `private`, and `List<Object>` is rarely correct.

functions/snippets/src/main/java/com/example/functions/SlackSlashCommand.java
```diff
@@ -0,0 +1,172 @@
...
+    Path configPath = Path.of(System.getProperty("user.dir"), "config.json");
+    JsonObject configJson = (new Gson()).fromJson(Files.readString(configPath), JsonObject.class);
+
+    SLACK_SECRET = configJson.get("SLACK_SECRET").getAsString();
+    API_KEY = configJson.get("KG_API_KEY").getAsString();
+
+    verifier = new SlackSignature.Verifier(new SlackSignature.Generator(SLACK_SECRET));
+  }
+
+  private boolean isValidSlackWebhook(HttpRequest request, String requestBody) throws IOException {
```
lesv: `private` probably isn't needed.

functions/snippets/src/main/java/com/example/functions/SlackSlashCommand.java
```diff
@@ -0,0 +1,172 @@
...
+
+    if (!isValidSlackWebhook(request, bodyString)) {
+      response.setStatusCode(HttpURLConnection.HTTP_UNAUTHORIZED);
+      return;
+    }
+
+    String query = body.get("text").getAsString();
+
+    // Call knowledge graph API
+    List<Object> kgResults = searchKnowledgeGraph(query);
```
lesv: `List<Object>` rarely a good idea.

functions/snippets/src/main/java/com/example/functions/SlackSlashCommand.java
ace-n: I'm not sure the Java runtime (easily) supports file uploads anyway - so that or properties in `pom.xml` are probably the way to go here.

(The other samples use `config.json` files, and I was copying what they did.)

functions/snippets/src/main/java/com/example/functions/SlackSlashCommand.java
ace-n: Would it be better if this [API response] was converted to JSON, and processed that way?

(Not sure how possible/easy that conversion is off the top of my head, but it may be easier to conceptualize/work with.)

That would also fix the `toString()` and `List<Object>` usage.

ace-n:
(Related question: should we document how to include files alongside your function like we do for other languages?)

functions/snippets/pom.xml
ace-n: That particular sample shows how to _retrieve_ logs from to Stackdriver.

The existing samples use `java.util.logging.Logger` though - would you want me to replace that with `System.out.println`?

lesv: No - My comment wasn't to discourage use, only that most users can just use stdout/stderr and get it logged.  It's only if you want something specialized do you need Stackdriver.  

Perhaps a different comment would be in order, but that's up to you.

functions/snippets/src/main/java/com/example/functions/SlackSlashCommand.java
lesv: Yea - either to Json, or pass a map with a well defined set of values.  The key is if you are using Object and casting, then you really should check before you actually do the cast.  Java is a strongly typed language, if you skip that, you are missing the point. (note - I'm not saying this is a good thing, only what is Idiomatic)

functions/snippets/src/main/java/com/example/functions/SlackSlashCommand.java
lesv: Sounds good.

lesv:
re-disabling surefire, we should enable but set appropriate defaults so users can run.

functions/snippets/pom.xml
ace-n: Ah - so you're thinking of the `functions_log_helloworld` sample in `LogHelloWorld.java`.

Fair point - I'll add some non-Stackdriver examples there (in a separate PR).

.kokoro/tests/run_tests.sh
```diff
@@ -65,12 +65,14 @@ if [[ "$SCRIPT_DEBUG" != "true" ]]; then
...
+    # shellcheck source=src/dlp_secrets.txt
     source "${KOKORO_GFILE_DIR}/dlp_secrets.txt"
-    # shellcheck source=src/bigtable_secrets.sh
+    # shellcheck source=src/bigtable_secrets.txt
     source "${KOKORO_GFILE_DIR}/bigtable_secrets.txt"
-    # shellcheck source=src/automl_secrets.sh
+    # shellcheck source=src/automl_secrets.txt
     source "${KOKORO_GFILE_DIR}/automl_secrets.txt"
+    # shellcheck source=src/functions_secrets.txt
+    source "${KOKORO_GFILE_DIR}/functions_secrets.txt"
```
lesv: Thanks - probably should have been in it's own PR

functions/snippets/config.slack.json
```diff
@@ -0,0 +1,4 @@
+{
```
lesv: Should this now be removed?

ace-n:
I've tried to trigger profiles based on three things:
1) _Maven archetypes_ - the [Java functions framework](https://github.com/GoogleCloudPlatform/functions-framework-java) deploys the function using `mvn functions:deploy` as shown [here](https://github.com/GoogleCloudPlatform/functions-framework-java/blob/9644a41184daddc2009316567a73e63f842307c1/invoker/function-maven-plugin/src/main/java/com/google/cloud/functions/plugin/DeployFunction.java#L33). However, [this StackOverflow post](https://stackoverflow.com/questions/15964170/how-to-activate-profile-based-on-goal-being-executed) says _goal_-dependent triggering is not possible.
2) _Environment variables_ - these are set _after_ the function build (and test run) is complete, so you can't use them as a proxy for "am I on `dev` or `prod`?"
3) **File presence** - if `.gcloudignore` is present, skip the tests. This **does** work, but it seems kinda hacky - especially vs. option #1.

@lesv thoughts?

functions/snippets/pom.xml
```diff
@@ -102,10 +114,36 @@
...
-  <!-- Required for Java 8 (Alpha) functions in the inline editor -->
+  <!-- Disable tests when deploying to GCF -->
+  <!-- (where .gcloudignore isn't present) -->
+  <profiles>
+    <profile>
+      <activation>
+        <file>
+          <missing>.gcloudignore</missing>
+        </file>
+      </activation>
```
lesv: probably not the best idea.

lesv: .gcloudignore isn't going to github, so you are turning off our tests.

ace-n: Found an alternative - see b/151989174 for a proposed longer-term solution.

PR #2401, opened 2020-03-13
(containeranalysis) Expanded tutorial sample

Attil:
Added missing info on how to get SHA digest of a container.

Documentation could use some info on how to get the requested digest. I have added a `curl` method, but if someone is using Java, they probably want to use Java for that too...


lesv:
Also, please take a look at our [SAMPLE_FORMAT](https://github.com/GoogleCloudPlatform/java-docs-samples/blob/master/SAMPLE_FORMAT.md) doc.

Attil:
Thanks for the reply, I'll update the PR soon.

About gcr.io/v2/project - this is Docker registry API (https://docs.docker.com/registry/spec/api/) with which GCR is compatible.

lesv:
Are you still planning on finishing this PR, or should we close it?

lesv:
ping @Attil

PR #2411, opened 2020-03-13
fix: SampleTests: customAttributeSampleTest

munkhuushmgl:
Fixing this issue: https://github.com/GoogleCloudPlatform/java-docs-samples/issues/2261

Since there is Thread.sleep(10000) in the main, the problem is that JUnit needs to wait until the whole thing is finished.

### DIFF
--- a/jobs/v3/src/test/java/SampleTests.java
+++ b/jobs/v3/src/test/java/SampleTests.java
@@ -39,6 +39,7 @@ import org.junit.runners.JUnit4;
 public class SampleTests {
 
   private static ByteArrayOutputStream bout;
+  private long timeInMillis;
 
   @BeforeClass
   public static void setUp() {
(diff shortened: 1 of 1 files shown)


lesv:
Mike - can you also bring the client library up to date (for v3)?

munkhuushmgl:
> Mike - can you also bring the client library up to date (for v3)?

Okay, I will do that here

munkhuushmgl:
> Mike - can you also bring the client library up to date (for v3)?

I may have misunderstood what you meant by updating the client library. 
Could you be bit more descriptive on updating the client library?

munkhuushmgl:
> > Mike - can you also bring the client library up to date (for v3)?
> 
> I may have misunderstood what you meant by updating the client library.
> Could you be bit more descriptive on updating the client library?

library version is already up-to-date.

PR #2443, opened 2020-03-19
update shared config

averikitsch:


lesv:
Care to update to 1.0.13?

averikitsch:
@lesv @kvg This PR's tests take a long time which means by the time they pass the branch needs to be updated. Can we force merge it knowing the tests are passing?

PR #2447, opened 2020-03-19
Fix Flaky DLP tests

shubha-rajan:
Fixes: #2302, #2303, #2317 

The tests keep timing out, so I followed the pattern implemented in the corresponding python samples to create a job and immediately cancel it.

### DIFF
--- a/dlp/src/test/java/dlp/snippets/InspectTests.java
+++ b/dlp/src/test/java/dlp/snippets/InspectTests.java
@@ -26,6 +26,7 @@ import java.io.ByteArrayOutputStream;
 import java.io.IOException;
 import java.io.PrintStream;
 import java.util.concurrent.ExecutionException;
+import java.util.concurrent.TimeUnit;
 import org.junit.After;
 import org.junit.Before;
 import org.junit.Test;
@@ -102,7 +103,9 @@ public class InspectTests {
 
   @Test
   public void testInspectGcsFile() throws InterruptedException, ExecutionException, IOException {
+
     InspectGcsFile.inspectGcsFile(PROJECT_ID, GCS_PATH, TOPIC_ID, SUBSCRIPTION_ID);
+    TimeUnit.SECONDS.sleep(3);
 
     String output = bout.toString();
     assertThat(output, containsString("Job created: "));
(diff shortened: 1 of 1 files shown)


dlp/src/test/java/dlp/snippets/InspectTests.java
```diff
@@ -103,8 +105,13 @@ public void testInspectGcsFile() throws InterruptedException, ExecutionException
...
     String output = bout.toString();
-    assertThat(output, containsString("Info type: PHONE_NUMBER"));
-    assertThat(output, containsString("Info type: EMAIL_ADDRESS"));
+    assertThat(output, containsString("Job created: "));
+
+    String jobId = output.split("Job created: ")[1].split("\n")[0];
+    CancelDlpJobRequest request = CancelDlpJobRequest.newBuilder().setName(jobId).build();
+    try (DlpServiceClient client = DlpServiceClient.create()) {
+      client.cancelDlpJob(request);
+    }
```
lesv: Fine if it works, but I'd bet these things can't happen right after the other.  (some hysteresis). 
(and it would be a flaky test again)

dlp/src/test/java/dlp/snippets/InspectTests.java
```diff
@@ -114,8 +121,13 @@ public void testInspectDatastoreEntity()
...
```
lesv: Fine if it works, but I'd bet these things can't happen right after the other.  (some hysteresis). 
(and it would be a flaky test again)

dlp/src/test/java/dlp/snippets/InspectTests.java
```diff
@@ -125,6 +137,12 @@ public void testInspectBigQueryTable()
...
 
     String output = bout.toString();
-    assertThat(output, containsString("Info type: PHONE_NUMBER"));
+    assertThat(output, containsString("Job created: "));
+
+    String jobId = output.split("Job created: ")[1].split("\n")[0];
+    CancelDlpJobRequest request = CancelDlpJobRequest.newBuilder().setName(jobId).build();
+    try (DlpServiceClient client = DlpServiceClient.create()) {
+      client.cancelDlpJob(request);
+    }
```
lesv: Fine if it works, but I'd bet these things can't happen right after the other.  (some hysteresis). 
(and it would be a flaky test again)

shubha-rajan:
@lesv I think you're right about the delay. It might create a race condition, so that it fails at the assertion before the job is created.  I added a 3s delay before the assertion and cancelling, but I'm not sure if that's enough. What do you think? I feel like timing always introduces a bit of flakiness to tests.

lesv:
You aren't testing canceling, so delay / cancel @ 1s, 4s, & 10s-  if it still fails, ignore it.  Comment that we don't really care about canceling, and we are only doing it to conserve quota.

PR #2450, opened 2020-03-20
[WIP] Add Firebase samples

ace-n:
**Do not merge** until:
- [x] ~**Firebase Remote Config**: b/151857384 is fixed~ (found a workaround! 🙂)
- 🚫**Firebase Analytics**: is tested on GCF itself **(will move into a separate PR)**
- [x] **Firebase Reactive sample**: is added


ace-n:
@lesv mind reviewing? I wasn't able to test the analytics sample (the trigger seems broken), so I'll follow up with @samtstern and see if I can fix that.

functions/snippets/pom.xml
```diff
@@ -52,6 +52,13 @@
       <version>3.4.2</version>
     </dependency>
 
+    <!-- Required for com.example.functions.MakeUpperCaseRTDB -->
```
samtstern: `MakeUpperCaseRTDB` implies Realtime Database (RTDB) but I see Firestore here.  Not sure if just the comment is wrong or something else needs to change.

functions/snippets/src/main/java/com/example/functions/FirebaseAnalytics.java
```diff
@@ -0,0 +1,59 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.example.functions;
+
+// [START functions_firebase_auth]
```
samtstern: I think this `START` tag has the wrong name.  Maybe s/auth/analytics?

functions/snippets/src/main/java/com/example/functions/FirebaseRemoteConfig.java
```diff
@@ -0,0 +1,54 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.example.functions;
+
+// [START functions_firebase_rtdb]
```
samtstern: Snippet name says `rtdb` but example shows remote config.

functions/snippets/pom.xml
ace-n: Ah - this is out-of-date. Fixed.

functions/snippets/src/main/java/com/example/functions/FirebaseFirestoreReactive.java
```diff
@@ -0,0 +1,75 @@
...
+    newFields.put("original", newValue);
+
+    String affectedDoc = context.resource().split("/documents/")[1].replace("\"", "");
+
+    LOGGER.info(String.format("Replacing value: %s --> %s", currentValue, newValue));
+    try {
+      firestore.document(affectedDoc).set(newFields, SetOptions.merge()).get();
+    } catch (Exception e) {
+      throw new RuntimeException(e);
+    }
```
lesv: Why do you bother with this?  Don't bother catching or even the try if you aren't going to do something useful with it.  It will automatically propagate to the parent.

ace-n: This throws `InterruptedException` and `ConcurrentException`, but this `accept` method can only throw `RuntimeException`.

I'll change this handler to be more specific, though (and only catch the two exception types I mentioned above).

functions/snippets/src/main/java/com/example/functions/FirebaseFirestoreReactive.java
```diff
@@ -0,0 +1,69 @@
...
+    newFields.put("original", newValue);
+
+    String affectedDoc = context.resource().split("/documents/")[1].replace("\"", "");
+
+    LOGGER.info(String.format("Replacing value: %s --> %s", currentValue, newValue));
+    try {
+      firestore.document(affectedDoc).set(newFields, SetOptions.merge()).get();
+    } catch (ExecutionException | InterruptedException e) {
+      throw new RuntimeException(e);
+    }
```
lesv: Sortof ok, but better to just comment that these might occur and let them propoage.

functions/snippets/src/main/java/com/example/functions/FirebaseFirestoreReactive.java
```diff
@@ -0,0 +1,69 @@
...
+  private Gson gsonParser = new Gson();
+
+  private static final Logger LOGGER = Logger.getLogger(FirebaseFirestoreReactive.class.getName());
+  private static final Firestore firestore = FirestoreOptions.getDefaultInstance().getService();
+
+  @Override
+  public void accept(String json, Context context) throws RuntimeException {
+    // Get the recently-written value
+    JsonObject body = gsonParser.fromJson(json, JsonObject.class);
+    JsonObject fields = body.getAsJsonObject("value").getAsJsonObject("fields");
```
lesv: Normally, you'd check between these, or use an operator that checks / skips if null.

functions/snippets/src/main/java/com/example/functions/FirebaseFirestoreReactive.java
```diff
@@ -0,0 +1,69 @@
...
+    if (!fields.has("original")) {
+      return;
+    }
+
+    // Convert recently-written value to ALL CAPS
+    String currentValue = fields.getAsJsonObject("original").get("stringValue").getAsString();
```
lesv: Same issue, if 'original' isn't present, then you'll get an error getting 'stringValue'

functions/snippets/src/main/java/com/example/functions/FirebaseFirestoreReactive.java
ace-n: That causes a compiler error - so we **have** to cast the exception type (or modify the Functions Framework itself).

functions/snippets/src/main/java/com/example/functions/FirebaseFirestoreReactive.java
ace-n: Fixed, PTAL.

functions/snippets/src/main/java/com/example/functions/FirebaseFirestoreReactive.java
ace-n: Fixed, PTAL.

PR #2487, opened 2020-03-25
feat: add backups samples

olavloite:
Add samples for Spanner backups.

### DIFF
--- a/spanner/cloud-client/src/main/java/com/example/spanner/SpannerSample.java
+++ b/spanner/cloud-client/src/main/java/com/example/spanner/SpannerSample.java
@@ -1719,6 +1719,7 @@ public class SpannerSample {
     }
 
     // List backups using pagination.
+    System.out.println("All backups, listed using pagination:");
     Page<Backup> page = instance.listBackups(Options.pageSize(10));
     while (true) {
       for (Backup backup : page.getValues()) {
--- a/spanner/cloud-client/src/main/java/com/example/spanner/SpannerSample.java
+++ b/spanner/cloud-client/src/main/java/com/example/spanner/SpannerSample.java
@@ -2091,7 +2091,6 @@ public class SpannerSample {
                 + clientProject);
         printUsageAndExit();
       }
-      // Get the most recent sample backup of the sample database.
       // Generate a backup id for the sample database.
       String backupName =
           String.format(
--- a/spanner/cloud-client/src/main/java/com/example/spanner/SpannerSample.java
+++ b/spanner/cloud-client/src/main/java/com/example/spanner/SpannerSample.java
@@ -1660,8 +1660,9 @@ public class SpannerSample {
   static void listDatabaseOperations(InstanceAdminClient instanceAdminClient, DatabaseAdminClient dbAdminClient, DatabaseId databaseId) {
     Instance instance = instanceAdminClient.getInstance(databaseId.getInstanceId().getInstance());
     // Get optimize restored database operations for the sample database.
-    String filter = "(metadata.@type:type.googleapis.com/"
-                  + "google.spanner.admin.database.v1.OptimizeRestoredDatabaseMetadata)";
+    String filter = String.format("(name:%s) AND (metadata.@type:type.googleapis.com/"
+                    + "google.spanner.admin.database.v1.OptimizeRestoredDatabaseMetadata)",
+                    databaseId.getDatabase());
     for (Operation op : instance.listDatabaseOperations(Options.filter(filter)).iterateAll()) {
       try {
         OptimizeRestoredDatabaseMetadata metadata =
@@ -1692,6 +1693,7 @@ public class SpannerSample {
         System.err.println(e.getMessage());
       }
     }
+    System.out.println(String.format("Listed all optimize operations for database %s", databaseId.getDatabase()));
   }
   // [END spanner_list_database_operations]
 
--- a/spanner/cloud-client/src/main/java/com/example/spanner/SpannerSample.java
+++ b/spanner/cloud-client/src/main/java/com/example/spanner/SpannerSample.java
@@ -33,10 +33,9 @@ import com.google.cloud.spanner.Database;
 import com.google.cloud.spanner.DatabaseAdminClient;
 import com.google.cloud.spanner.DatabaseClient;
 import com.google.cloud.spanner.DatabaseId;
-import com.google.cloud.spanner.DatabaseInfo.State;
-import com.google.cloud.spanner.ErrorCode;
 import com.google.cloud.spanner.Instance;
 import com.google.cloud.spanner.InstanceAdminClient;
+import com.google.cloud.spanner.InstanceId;
 import com.google.cloud.spanner.Key;
 import com.google.cloud.spanner.KeySet;
 import com.google.cloud.spanner.Mutation;
@@ -56,7 +55,6 @@ import com.google.cloud.spanner.TransactionContext;
 import com.google.cloud.spanner.Type;
 import com.google.cloud.spanner.Value;
 import com.google.common.io.BaseEncoding;
-import com.google.common.util.concurrent.Uninterruptibles;
 import com.google.longrunning.Operation;
 import com.google.protobuf.InvalidProtocolBufferException;
 import com.google.spanner.admin.database.v1.CreateBackupMetadata;
@@ -172,7 +170,9 @@ public class SpannerSample {
 
   /** Get a database id to restore a backup to from the sample database id. */
   static String createRestoredSampleDbId(DatabaseId database) {
-    String restoredDbId = database.getDatabase().replace("mysample", "restored");
+    int index = database.getDatabase().indexOf('-');
+    String prefix = database.getDatabase().substring(0, index);
+    String restoredDbId = database.getDatabase().replace(prefix, "restored");
     if (restoredDbId.length() > 30) {
       restoredDbId = restoredDbId.substring(0, 30);
     }
@@ -1524,18 +1524,10 @@ public class SpannerSample {
 
   // [START spanner_create_backup]
   static void createBackup(
-      DatabaseAdminClient dbAdminClient, DatabaseId databaseId) {
-    String backupName =
-        String.format(
-            "%s_%02d",
-            databaseId.getDatabase(), LocalDate.now().get(ChronoField.ALIGNED_WEEK_OF_YEAR));
-    BackupId backupId = BackupId.of(databaseId.getInstanceId(), backupName);
+      DatabaseAdminClient dbAdminClient, DatabaseId databaseId, BackupId backupId) {
     // Set expire time to 14 days from now.
-    long currentTimeInMicroSeconds =
-        TimeUnit.MICROSECONDS.convert(System.currentTimeMillis(), TimeUnit.MILLISECONDS);
-    long deltaTimeInMicroseconds = TimeUnit.MICROSECONDS.convert(14L, TimeUnit.DAYS);
-    Timestamp expireTime =
-        Timestamp.ofTimeMicroseconds(currentTimeInMicroSeconds + deltaTimeInMicroseconds);
+    Timestamp expireTime = Timestamp.ofTimeMicroseconds(TimeUnit.MICROSECONDS.convert(
+        System.currentTimeMillis() + TimeUnit.DAYS.toMillis(14), TimeUnit.MILLISECONDS));
     Backup backup =
         dbAdminClient
             .newBackupBuilder(backupId)
@@ -1550,18 +1542,11 @@ public class SpannerSample {
       backup = op.get();
       System.out.println("Created backup [" + backup.getId() + "]");
     } catch (ExecutionException e) {
-      // If the operation failed during execution, expose the cause.
       throw (SpannerException) e.getCause();
     } catch (InterruptedException e) {
-      // Throw when a thread is waiting, sleeping, or otherwise occupied,
-      // and the thread is interrupted, either before or during the activity.
       throw SpannerExceptionFactory.propagateInterrupt(e);
     }
-    // Check that the backup has been created and is ready for use.
-    if (!backup.isReady()) {
-      System.out.println("Backup is not ready for use.");
-      throw new RuntimeException("Backup is not ready for use.");
-    }
+
     // Reload the metadata of the backup from the server.
     backup = backup.reload();
     System.out.println(
@@ -1578,18 +1563,10 @@ public class SpannerSample {
 
   // [START spanner_cancel_create_backup]
   static void cancelCreateBackup(
-      DatabaseAdminClient dbAdminClient, DatabaseId databaseId) {
-    String backupName =
-        String.format(
-            "%s_%02d_cancel",
-            databaseId.getDatabase(), LocalDate.now().get(ChronoField.ALIGNED_WEEK_OF_YEAR));
-    BackupId backupId = BackupId.of(databaseId.getInstanceId(), backupName);
+      DatabaseAdminClient dbAdminClient, DatabaseId databaseId, BackupId backupId) {
     // Set expire time to 14 days from now.
-    long currentTimeInMicroSeconds =
-        TimeUnit.MICROSECONDS.convert(System.currentTimeMillis(), TimeUnit.MILLISECONDS);
-    long deltaTimeInMicroseconds = TimeUnit.MICROSECONDS.convert(14L, TimeUnit.DAYS);
-    Timestamp expireTime =
-        Timestamp.ofTimeMicroseconds(currentTimeInMicroSeconds + deltaTimeInMicroseconds);
+    Timestamp expireTime = Timestamp.ofTimeMicroseconds(TimeUnit.MICROSECONDS.convert(
+        System.currentTimeMillis() + TimeUnit.DAYS.toMillis(14), TimeUnit.MILLISECONDS));
 
     // Create a backup instance.
     Backup backup =
@@ -1609,21 +1586,20 @@ public class SpannerSample {
       // for the current status of the backup operation.
       RetryingFuture<OperationSnapshot> pollingFuture = op.getPollingFuture();
       // Wait for the operation to finish.
-      // isDone will return true if the operation has finished successfully or if it was cancelled
-      // or any other error occurred..
+      // isDone will return true when the operation is complete, regardless of whether it was
+      // successful or not.
       while (!pollingFuture.get().isDone()) {
         System.out.println("Waiting for the cancelled backup operation to finish...");
         Thread.sleep(TimeUnit.MILLISECONDS.convert(5, TimeUnit.SECONDS));
       }
       if (pollingFuture.get().getErrorCode() == null) {
         // Backup was created before it could be cancelled. Delete the backup.
-        System.out.println("Deleting backup [" + backup.getId() + "]...");
         backup.delete();
       } else if (pollingFuture.get().getErrorCode().getCode() == StatusCode.Code.CANCELLED) {
         System.out.println("Backup operation for [" + backup.getId() + "] successfully cancelled");
       }
     } catch (ExecutionException e) {
-      throw (SpannerException) e.getCause();
+      throw SpannerExceptionFactory.newSpannerException(e.getCause());
     } catch (InterruptedException e) {
       throw SpannerExceptionFactory.propagateInterrupt(e);
     }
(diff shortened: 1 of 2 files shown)


olavloite:
@skuruppu The samples are the same as in the private repo, but the `restoreDatabase` sample has been slightly updated to also take into account that it can fail if there are already pending restore operations on the instance.

spanner/cloud-client/src/main/java/com/example/spanner/SpannerSample.java
```diff
@@ -1473,9 +1503,371 @@ static void queryWithQueryOptions(DatabaseClient dbClient) {
...
+      // Wait for the operation to finish.
+      // isDone will return true if the operation has finished successfully or if it was cancelled
+      // or any other error occurred..
+      while (!pollingFuture.get().isDone()) {
+        System.out.println("Waiting for the cancelled backup operation to finish...");
+        Thread.sleep(TimeUnit.MILLISECONDS.convert(5, TimeUnit.SECONDS));
+      }
+      if (pollingFuture.get().getErrorCode() == null) {
+        // Backup was created before it could be cancelled. Delete the backup.
+        System.out.println("Deleting backup [" + backup.getId() + "]...");
```
skuruppu: I think the comment is sufficient and we can get rid of the println.

spanner/cloud-client/src/main/java/com/example/spanner/SpannerSample.java
```diff
@@ -1473,9 +1503,371 @@ static void queryWithQueryOptions(DatabaseClient dbClient) {
...
+    try {
+      // Wait for the backup operation to complete.
+      backup = op.get();
+      System.out.println("Created backup [" + backup.getId() + "]");
+    } catch (ExecutionException e) {
+      // If the operation failed during execution, expose the cause.
+      throw (SpannerException) e.getCause();
+    } catch (InterruptedException e) {
+      // Throw when a thread is waiting, sleeping, or otherwise occupied,
+      // and the thread is interrupted, either before or during the activity.
```
skuruppu: I realize we have this type of comment in the `createDatabase` sample but I'm wondering if we could get rid of it here to keep things concise.

spanner/cloud-client/src/main/java/com/example/spanner/SpannerSample.java
```diff
@@ -1473,9 +1503,371 @@ static void queryWithQueryOptions(DatabaseClient dbClient) {
...
+  // [END spanner_list_backup_operations]
+
+  // [START spanner_list_database_operations]
+  static void listDatabaseOperations(
+      InstanceAdminClient instanceAdminClient,
+      DatabaseAdminClient dbAdminClient,
+      DatabaseId databaseId) {
+    Instance instance = instanceAdminClient.getInstance(databaseId.getInstanceId().getInstance());
+    // Get optimize restored database operations for the sample database.
+    String filter = String.format("(name:%s) AND (metadata.@type:type.googleapis.com/"
```
skuruppu: In the other samples, we ended up listing all optimize operations so if you could remove the name filter that would be great. Hope it doesn't mess up the tests too much.

spanner/cloud-client/src/main/java/com/example/spanner/SpannerSample.java
```diff
@@ -1473,9 +1503,371 @@ static void queryWithQueryOptions(DatabaseClient dbClient) {
...
+        DatabaseId restoredDbId = DatabaseId.of(
+            databaseId.getInstanceId(),
+            metadata.getName().substring(metadata.getName().lastIndexOf('/') + 1));
+        Database restoredDb = new Database(restoredDbId, State.UNSPECIFIED, dbAdminClient);
+        if (restoredDb.exists()) {
+          RestoreInfo restore = restoredDb.getRestoreInfo();
+          if (restore != null) {
+            System.out.println(
+                String.format(
+                    "Database %s restored from backup %s is optimized.",
```
skuruppu: I think to keep things simpler we can just do:

`"Database %s restored from backup is %s%% optimized"`

This is what we ended up doing in other languages. This way, you don't need to get the `restoreDb` or the `restore` info, etc.

spanner/cloud-client/src/main/java/com/example/spanner/SpannerSample.java
```diff
@@ -1473,9 +1503,371 @@ static void queryWithQueryOptions(DatabaseClient dbClient) {
...
+                  + "but the database no longer exists.",
+                  metadata.getName()));
+        }
+      } catch (InvalidProtocolBufferException e) {
+        // The returned operation does not contain OptimizeRestoredDatabaseMetadata.
+        System.err.println(e.getMessage());
+      }
+    }
+    System.out.println(
+        String.format("Listed all optimize operations for database %s", databaseId.getDatabase()));
```
skuruppu: Can probably remove this too.

spanner/cloud-client/src/main/java/com/example/spanner/SpannerSample.java
```diff
@@ -1473,9 +1503,371 @@ static void queryWithQueryOptions(DatabaseClient dbClient) {
...
+    }
+
+    // List all backups with a create time after a certain timestamp and that are also ready.
+    System.out.println("All databases created after 2019-10-18T02:56:53Z and that are ready:");
+    for (Backup backup :
+        instance
+            .listBackups(Options.filter("create_time >= \"2019-10-18T02:56:53Z\" AND state:READY"))
+            .iterateAll()) {
+      System.out.println(backup);
+    }
```
skuruppu: I know this is how we defined it in the spec but we realized after that the samples tests won't be meaningful since these filter won't show anything. So we ended up doing the following:

1. Get the backup ID as input to listBackups and filter on name:backupId.
2. Database filter was changed to filter on the databaseId given as input.
3. Expire time filter was changed to expire_time < (time now + 30 days).
4. size_bytes > X where X depends on the size of the database you're creating in other parts of the test. PHP is using X = 500 and Node.js is using X = 100.
5. create_time >= (time now - 1 day)

Hope it's ok to make these changes.

spanner/cloud-client/src/main/java/com/example/spanner/SpannerSample.java
```diff
@@ -1473,9 +1503,371 @@ static void queryWithQueryOptions(DatabaseClient dbClient) {
...
+  static void restoreBackup(
+      DatabaseAdminClient dbAdminClient,
+      DatabaseId sourceDatabaseId,
+      DatabaseId restoreToDatabase) {
+    // Get the most recent sample backup of the sample database.
+    String backupName =
+        String.format(
+            "%s_%02d",
+            sourceDatabaseId.getDatabase(), LocalDate.now().get(ChronoField.ALIGNED_WEEK_OF_YEAR));
+    BackupId backupId = BackupId.of(sourceDatabaseId.getInstanceId(), backupName);
```
skuruppu: We ended up passing in the backupId as an argument in the other languages. Could we also do that here?

spanner/cloud-client/src/main/java/com/example/spanner/SpannerSample.java
```diff
@@ -1473,9 +1503,371 @@ static void queryWithQueryOptions(DatabaseClient dbClient) {
...
+    Backup backup = dbAdminClient.newBackupBuilder(backupId).build();
+    // Initiate the request which returns an OperationFuture.
+    System.out.println(String.format(
+        "Restoring backup [%s] to database [%s]...",
+        backup.getId().toString(),
+        restoreToDatabase.toString()));
+    // Try the restore operation in a retry loop, as there is a limit on the number of restore
+    // operations that is allowed to execute simultaneously, and we should retry if we hit this,
+    // limit.
```
skuruppu: Are you hitting this in the samples tests because a single instance is shared? We didn't do this for other languages but I think that might be because we're creating and destroying an instance for each test run.

I'm a bit worried about the complexity of this in terms of a code sample since we just want to show users the simplest possible example.

spanner/cloud-client/src/main/java/com/example/spanner/SpannerSample.java
```diff
@@ -1473,9 +1503,371 @@ static void queryWithQueryOptions(DatabaseClient dbClient) {
...
+  // [END spanner_restore_backup]
+
+  // [START spanner_update_backup]
+  static void updateBackup(DatabaseAdminClient dbAdminClient, DatabaseId databaseId) {
+    // Get the most recent sample backup of the sample database.
+    String backupName =
+        String.format(
+            "%s_%02d",
+            databaseId.getDatabase(), LocalDate.now().get(ChronoField.ALIGNED_WEEK_OF_YEAR));
+    BackupId backupId = BackupId.of(databaseId.getInstanceId(), backupName);
```
skuruppu: Same here, it would be good to get the backupId as an input.

spanner/cloud-client/src/main/java/com/example/spanner/SpannerSample.java
```diff
@@ -1473,9 +1503,371 @@ static void queryWithQueryOptions(DatabaseClient dbClient) {
...
+  // [END spanner_update_backup]
+
+  // [START spanner_delete_backup]
+  static void deleteBackup(DatabaseAdminClient dbAdminClient, DatabaseId databaseId) {
+    // Get the most recent sample backup of the sample database.
+    String backupName =
+        String.format(
+            "%s_%02d",
+            databaseId.getDatabase(), LocalDate.now().get(ChronoField.ALIGNED_WEEK_OF_YEAR));
+    BackupId backupId = BackupId.of(databaseId.getInstanceId(), backupName);
```
skuruppu: Same here regarding getting the backupId as in input.

spanner/cloud-client/src/test/java/com/example/spanner/SpannerSampleIT.java
```diff
@@ -241,6 +241,36 @@ public void testSample() throws Exception {
...
+    assertThat(out).contains("optimize");
+
+    out = runSample("listbackups");
+    assertThat(out).contains(databaseId);
+
+    out = runSample("restorebackup");
+    assertThat(out).contains("Restored database [");
+
+    out = runSample("updatebackup");
+    assertThat(out).contains("Updated backup [");
```
skuruppu: We ended up checking the outputs (like the backupId in listbackups) in the tests for other languages. I don't know if that's conventional in the Java tests but would it be possible to do that here?

At least in Node.js, I found it helpful to know when we hadn't defined the types correctly and it printed out "undefined".

spanner/cloud-client/src/main/java/com/example/spanner/SpannerSample.java
```diff
@@ -1473,9 +1503,371 @@ static void queryWithQueryOptions(DatabaseClient dbClient) {
...
   // [END spanner_query_with_query_options]
 
+  // [START spanner_create_backup]
+  static void createBackup(
+      DatabaseAdminClient dbAdminClient, DatabaseId databaseId) {
+    String backupName =
+        String.format(
+            "%s_%02d",
+            databaseId.getDatabase(), LocalDate.now().get(ChronoField.ALIGNED_WEEK_OF_YEAR));
+    BackupId backupId = BackupId.of(databaseId.getInstanceId(), backupName);
```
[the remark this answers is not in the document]
olavloite: done

spanner/cloud-client/src/main/java/com/example/spanner/SpannerSample.java
```diff
@@ -1473,9 +1503,371 @@ static void queryWithQueryOptions(DatabaseClient dbClient) {
...
+    // Set expire time to 14 days from now.
+    long currentTimeInMicroSeconds =
+        TimeUnit.MICROSECONDS.convert(System.currentTimeMillis(), TimeUnit.MILLISECONDS);
+    long deltaTimeInMicroseconds = TimeUnit.MICROSECONDS.convert(14L, TimeUnit.DAYS);
+    Timestamp expireTime =
+        Timestamp.ofTimeMicroseconds(currentTimeInMicroSeconds + deltaTimeInMicroseconds);
```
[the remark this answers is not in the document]
olavloite: That works well.

spanner/cloud-client/src/main/java/com/example/spanner/SpannerSample.java
```diff
@@ -1473,9 +1503,371 @@ static void queryWithQueryOptions(DatabaseClient dbClient) {
...
+    } catch (InterruptedException e) {
+      // Throw when a thread is waiting, sleeping, or otherwise occupied,
+      // and the thread is interrupted, either before or during the activity.
+      throw SpannerExceptionFactory.propagateInterrupt(e);
+    }
+    // Check that the backup has been created and is ready for use.
+    if (!backup.isReady()) {
+      System.out.println("Backup is not ready for use.");
+      throw new RuntimeException("Backup is not ready for use.");
+    }
```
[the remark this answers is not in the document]
olavloite: done

spanner/cloud-client/src/main/java/com/example/spanner/SpannerSample.java
```diff
@@ -1473,9 +1503,371 @@ static void queryWithQueryOptions(DatabaseClient dbClient) {
...
+      backup = op.get();
+      System.out.println("Created backup [" + backup.getId() + "]");
+    } catch (ExecutionException e) {
+      // If the operation failed during execution, expose the cause.
+      throw (SpannerException) e.getCause();
+    } catch (InterruptedException e) {
+      // Throw when a thread is waiting, sleeping, or otherwise occupied,
+      // and the thread is interrupted, either before or during the activity.
+      throw SpannerExceptionFactory.propagateInterrupt(e);
+    }
```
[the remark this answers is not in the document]
olavloite: done

spanner/cloud-client/src/main/java/com/example/spanner/SpannerSample.java
```diff
@@ -1473,9 +1503,371 @@ static void queryWithQueryOptions(DatabaseClient dbClient) {
...
+        String.format(
+            "%s_%02d_cancel",
+            databaseId.getDatabase(), LocalDate.now().get(ChronoField.ALIGNED_WEEK_OF_YEAR));
+    BackupId backupId = BackupId.of(databaseId.getInstanceId(), backupName);
+    // Set expire time to 14 days from now.
+    long currentTimeInMicroSeconds =
+        TimeUnit.MICROSECONDS.convert(System.currentTimeMillis(), TimeUnit.MILLISECONDS);
+    long deltaTimeInMicroseconds = TimeUnit.MICROSECONDS.convert(14L, TimeUnit.DAYS);
+    Timestamp expireTime =
+        Timestamp.ofTimeMicroseconds(currentTimeInMicroSeconds + deltaTimeInMicroseconds);
```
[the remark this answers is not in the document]
olavloite: done

spanner/cloud-client/src/main/java/com/example/spanner/SpannerSample.java
```diff
@@ -1473,9 +1503,371 @@ static void queryWithQueryOptions(DatabaseClient dbClient) {
...
+    try {
+      // Try to cancel the backup operation.
+      System.out.println("Cancelling create backup operation for [" + backup.getId() + "]...");
+      dbAdminClient.cancelOperation(op.getName());
+      // Get a polling future for the running operation. This future will regularly poll the server
+      // for the current status of the backup operation.
+      RetryingFuture<OperationSnapshot> pollingFuture = op.getPollingFuture();
+      // Wait for the operation to finish.
+      // isDone will return true if the operation has finished successfully or if it was cancelled
+      // or any other error occurred..
```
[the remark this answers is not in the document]
olavloite: done

spanner/cloud-client/src/main/java/com/example/spanner/SpannerSample.java
```diff
@@ -1473,9 +1503,371 @@ static void queryWithQueryOptions(DatabaseClient dbClient) {
...
+      // If the operation failed during execution, expose the cause.
+      throw (SpannerException) e.getCause();
+    } catch (InterruptedException e) {
+      // Throw when a thread is waiting, sleeping, or otherwise occupied,
+      // and the thread is interrupted, either before or during the activity.
+      throw SpannerExceptionFactory.propagateInterrupt(e);
+    }
+    // Check that the backup has been created and is ready for use.
+    if (!backup.isReady()) {
+      System.out.println("Backup is not ready for use.");
```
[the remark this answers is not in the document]
olavloite: done

spanner/cloud-client/src/main/java/com/example/spanner/SpannerSample.java
olavloite: done

spanner/cloud-client/src/main/java/com/example/spanner/SpannerSample.java
olavloite: agree, I removed it.

spanner/cloud-client/src/main/java/com/example/spanner/SpannerSample.java
olavloite: done

spanner/cloud-client/src/main/java/com/example/spanner/SpannerSample.java
olavloite: done

spanner/cloud-client/src/main/java/com/example/spanner/SpannerSample.java
olavloite: done

spanner/cloud-client/src/main/java/com/example/spanner/SpannerSample.java
olavloite: Done

spanner/cloud-client/src/main/java/com/example/spanner/SpannerSample.java
olavloite: done

spanner/cloud-client/src/main/java/com/example/spanner/SpannerSample.java
olavloite: Yep, the Java CI env runs two parallel tests (Java 8 and Java 11) on the same instance, so the restore operation would fail for one of them each time. What I could do is remove this error handling here and add it to the test case. That keeps the sample simple. The test will then run the sample repeatedly until it succeeds, fails with a different error than the expected error, or more than X attempts have failed with the 'more than X pending restores' error.

spanner/cloud-client/src/main/java/com/example/spanner/SpannerSample.java
olavloite: done.

spanner/cloud-client/src/main/java/com/example/spanner/SpannerSample.java
olavloite: done.

spanner/cloud-client/src/main/java/com/example/spanner/SpannerSample.java
olavloite: I've moved the retry logic away from the sample and into the test case.

spanner/cloud-client/src/test/java/com/example/spanner/SpannerSampleIT.java
olavloite: I've added checks for the backup ids in the logs for Java as well.

olavloite:
> Sorry for all the feedback. We ended up making a few more changes to make things consistent between the languages.

No problem. I've processed the comments and made the corresponding changes. PTAL.

spanner/cloud-client/src/test/java/com/example/spanner/SpannerSampleIT.java
```diff
@@ -241,6 +252,83 @@ public void testSample() throws Exception {
...
+        String.format(
+            "Database %s restored from backup",
+            DatabaseId.of(
+                dbId.getInstanceId(),
+                SpannerSample.createRestoredSampleDbId(dbId))
+            .getName()));
+
+    out = runSample("updatebackup");
+    assertThat(out).contains(
+        String.format("Updating expire time of backup [%s]", backupId.toString()));
```
skuruppu: I think it's better to check for "Updated backup [%s]".

spanner/cloud-client/src/test/java/com/example/spanner/SpannerSampleIT.java
```diff
@@ -241,6 +252,83 @@ public void testSample() throws Exception {
...
+
+    out = runSample("listbackupoperations");
+    assertThat(out).contains(
+        String.format(
+            "Backup %s on database %s pending:",
+            backupId.getName(),
+            dbId.getName()));
+
+    out = runSample("listbackups");
+    assertThat(out).contains(databaseId);
```
skuruppu: I think it would be good to test at least some of outputs with various filters. Let me know if you think it would be too flaky.

spanner/cloud-client/src/test/java/com/example/spanner/SpannerSampleIT.java
olavloite: Good spot, fixed.

spanner/cloud-client/src/test/java/com/example/spanner/SpannerSampleIT.java
olavloite: With the new filters, the output should now be predictable, so I've added checks to verify that the backup that was created is returned by all list operations.

PR #2490, opened 2020-03-25
Update Bigtable+Beam hello world examples and move them into java-docs-samples

billyjacobson:
Existing samples are here: https://github.com/GoogleCloudPlatform/cloud-bigtable-examples/tree/master/java/dataflow-connector-examples/src/main/java/com/google/cloud/bigtable/dataflow/example


bigtable/beam/helloworld/src/main/java/HelloWorldWrite.java
```diff
@@ -0,0 +1,94 @@
...
+                    row.addColumn(
+                        Bytes.toBytes("stats_summary"),
+                        Bytes.toBytes("os_build"),
+                        timestamp,
+                        Bytes.toBytes("android"));
+                    out.output(row);
+                  }
+                }))
+        .apply(CloudBigtableIO.writeToTable(bigtableTableConfig));
+    // [START bigtable_beam_helloworld_write_transforms]
```
lesv: Should be end, sorry to hit merge

PR #2531, opened 2020-03-30
Add missing helloworld samples + delete excess logging code

ace-n:
✅Add `functions_helloworld_error`
✅Add `functions_helloworld_gcs_generic`
✅Delete excess logging code in `SnippetsTests.java`

### DIFF
--- a/functions/snippets/src/main/java/com/example/functions/HelloGcsGeneric.java
+++ b/functions/snippets/src/main/java/com/example/functions/HelloGcsGeneric.java
@@ -1,0 +1,38 @@
+/*
+ * Copyright 2020 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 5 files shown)


functions/snippets/src/main/java/com/example/functions/HelloGcsGeneric.java
```diff
@@ -0,0 +1,38 @@
...
+
+  @Override
+  public void accept(GcsEvent event, Context context) {
+    LOGGER.info("Event: " + context.eventId());
+    LOGGER.info("Event Type: " + context.eventType());
+    LOGGER.info("Bucket: " + event.getBucket());
+    LOGGER.info("File: " + event.getName());
+    LOGGER.info("Metageneration: " + event.getMetageneration());
+    LOGGER.info("Created: " + event.getTimeCreated());
+    LOGGER.info("Updated: " + event.getUpdated());
```
lesv: Ok, but one would imagine such a thing as doing a single string.Format() and log just once.

ace-n: Even if it's across multiple lines?

lesv: This is grouped data, you want it to appear in the log together.  It will on Functions, so it's probably Ok in this use case, but other use cases where things are multi-threaded and concurrent, not so much.  A simple string makes it be just a single log write, how it actually appears in the logs might not have the \n's, but at least they would all be together.  There are other techniques to do this, but a single string is the easiest.

ace-n:
@lesv FYI: added a `functions_helloworld_method` sample that I realized I forgot.

PR #2552, opened 2020-04-01
Add functions/ocr sample

ace-n:
✅This has already been tested on Cloud Functions itself

Also, I wasn't sure if the `OcrTranslateApiMessage` POJO was the best way to go - so I'd appreciate thoughts on that.

### DIFF
--- a/functions/snippets/pom.xml
+++ b/functions/snippets/pom.xml
@@ -33,7 +33,7 @@
     <dependency>
       <groupId>com.google.cloud</groupId>
       <artifactId>libraries-bom</artifactId>
-      <version>4.0.0</version>
+      <version>4.4.0</version>
       <type>pom</type>
       <scope>import</scope>
      </dependency>
--- a/functions/snippets/pom.xml
+++ b/functions/snippets/pom.xml
@@ -8,14 +8,14 @@
   <artifactId>functions-snippets</artifactId>
   <version>1.0-SNAPSHOT</version>
 
-  <!--
-   The parent pom defines common style checks and testing strategies for our samples.
-   Removing or replacing it should not affect the execution of the samples in anyway.
- -->
-  <parent>
-    <groupId>com.google.cloud.samples</groupId>
-    <artifactId>shared-configuration</artifactId>
-    <version>1.0.13</version>
+  <!--  
+   The parent pom defines common style checks and testing strategies for our samples. 
+   Removing or replacing it should not affect the execution of the samples in anyway. 
+ -->  
+  <parent>  
+    <groupId>com.google.cloud.samples</groupId> 
+    <artifactId>shared-configuration</artifactId> 
+    <version>1.0.13</version> 
   </parent>
 
   <properties>
@@ -27,7 +27,36 @@
     <groovy.version>3.0.2</groovy.version>
   </properties>
 
+  <dependencyManagement>
+    <dependencies>
+      <dependency>
+        <groupId>com.google.cloud</groupId>
+        <artifactId>libraries-bom</artifactId>
+        <version>4.0.0</version>
+        <type>pom</type>
+        <scope>import</scope>
+      </dependency>
+    </dependencies>
+  </dependencyManagement>
+
   <dependencies>
+    <dependency>
+      <groupId>com.google.cloud</groupId>
+      <artifactId>google-cloud-storage</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>com.google.cloud</groupId>
+      <artifactId>google-cloud-vision</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>com.google.cloud</groupId>
+      <artifactId>google-cloud-translate</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>com.google.cloud</groupId>
+      <artifactId>google-cloud-pubsub</artifactId>
+    </dependency>
+
     <!-- Required for com.example.functions.ParseContentType -->
     <dependency>
       <groupId>com.google.code.gson</groupId>
--- a/functions/snippets/pom.xml
+++ b/functions/snippets/pom.xml
@@ -13,8 +13,8 @@
    Removing or replacing it should not affect the execution of the samples in anyway.
  -->
   <parent>
-    <groupId>com.google.cloud.samples</groupId> 
-    <artifactId>shared-configuration</artifactId> 
+    <groupId>com.google.cloud.samples</groupId>
+    <artifactId>shared-configuration</artifactId>
     <version>1.0.15</version>
   </parent>
 
@@ -200,7 +200,7 @@
       <plugin>
         <groupId>org.codehaus.gmavenplus</groupId>
         <artifactId>gmavenplus-plugin</artifactId>
-        <version>1.8.1</version>
+        <version>1.9.0</version>
         <executions>
           <execution>
             <id>groovy-compile</id>
--- a/functions/snippets/pom.xml
+++ b/functions/snippets/pom.xml
@@ -15,7 +15,7 @@
   <parent>
     <groupId>com.google.cloud.samples</groupId> 
     <artifactId>shared-configuration</artifactId> 
-    <version>1.0.14</version>
+    <version>1.0.15</version>
   </parent>
 
   <properties>
--- a/functions/snippets/pom.xml
+++ b/functions/snippets/pom.xml
@@ -130,7 +130,7 @@
     <dependency>
       <groupId>com.google.truth</groupId>
       <artifactId>truth</artifactId>
-      <version>0.42</version>
+      <version>1.0.1</version>
       <scope>test</scope>
     </dependency>
     <dependency>
--- a/functions/snippets/pom.xml
+++ b/functions/snippets/pom.xml
@@ -27,19 +27,21 @@
     <groovy.version>3.0.2</groovy.version>
   </properties>
 
+  <!-- Required for com.example.functions.ImageMagick -->
   <dependencyManagement>
-    <dependencies>
-      <dependency>
-        <groupId>com.google.cloud</groupId>
-        <artifactId>libraries-bom</artifactId>
-        <version>4.0.0</version>
-        <type>pom</type>
-        <scope>import</scope>
-      </dependency>
-    </dependencies>
+   <dependencies>
+    <dependency>
+      <groupId>com.google.cloud</groupId>
+      <artifactId>libraries-bom</artifactId>
+      <version>4.0.0</version>
+      <type>pom</type>
+      <scope>import</scope>
+     </dependency>
+   </dependencies>
   </dependencyManagement>
 
   <dependencies>
+    <!-- Required for com.example.functions.ImageMagick -->
     <dependency>
       <groupId>com.google.cloud</groupId>
       <artifactId>google-cloud-storage</artifactId>
--- a/functions/snippets/pom.xml
+++ b/functions/snippets/pom.xml
@@ -2,7 +2,7 @@
 <project xmlns="http://maven.apache.org/POM/4.0.0"
     xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
     xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
-  <modelVersion>4.4.0</modelVersion>
+  <modelVersion>4.0.0</modelVersion>
 
   <groupId>com.example.cloud.functions</groupId>
   <artifactId>functions-snippets</artifactId>
@@ -27,21 +27,19 @@
     <groovy.version>3.0.2</groovy.version>
   </properties>
 
-  <!-- Required for com.example.functions.ImageMagick -->
   <dependencyManagement>
-   <dependencies>
-    <dependency>
-      <groupId>com.google.cloud</groupId>
-      <artifactId>libraries-bom</artifactId>
-      <version>4.4.0</version>
-      <type>pom</type>
-      <scope>import</scope>
-     </dependency>
-   </dependencies>
+    <dependencies>
+      <dependency>
+        <groupId>com.google.cloud</groupId>
+        <artifactId>libraries-bom</artifactId>
+        <version>4.4.0</version>
+        <type>pom</type>
+        <scope>import</scope>
+      </dependency>
+    </dependencies>
   </dependencyManagement>
 
   <dependencies>
-    <!-- Required for com.example.functions.ImageMagick -->
     <dependency>
       <groupId>com.google.cloud</groupId>
       <artifactId>google-cloud-storage</artifactId>
@@ -132,7 +130,7 @@
     <dependency>
       <groupId>com.google.truth</groupId>
       <artifactId>truth</artifactId>
-      <version>1.0.1</version>
+      <version>0.42</version>
       <scope>test</scope>
     </dependency>
     <dependency>
--- a/functions/snippets/pom.xml
+++ b/functions/snippets/pom.xml
@@ -2,20 +2,20 @@
 <project xmlns="http://maven.apache.org/POM/4.0.0"
     xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
     xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
-  <modelVersion>4.0.0</modelVersion>
+  <modelVersion>4.4.0</modelVersion>
 
   <groupId>com.example.cloud.functions</groupId>
   <artifactId>functions-snippets</artifactId>
   <version>1.0-SNAPSHOT</version>
 
-  <!--  
-   The parent pom defines common style checks and testing strategies for our samples. 
-   Removing or replacing it should not affect the execution of the samples in anyway. 
- -->  
-  <parent>  
+  <!--
+   The parent pom defines common style checks and testing strategies for our samples.
+   Removing or replacing it should not affect the execution of the samples in anyway.
+ -->
+  <parent>
     <groupId>com.google.cloud.samples</groupId> 
     <artifactId>shared-configuration</artifactId> 
-    <version>1.0.13</version> 
+    <version>1.0.14</version>
   </parent>
 
   <properties>
--- a/functions/snippets/src/main/java/com/example/functions/OcrProcessImage.java
+++ b/functions/snippets/src/main/java/com/example/functions/OcrProcessImage.java
@@ -103,11 +103,13 @@ public class OcrProcessImage implements BackgroundFunction<GcsEvent> {
       }
 
       if (visionResponse.hasError()) {
-        // Throw IOException; to be handled by catch block below
-        throw new IOException(visionResponse.getError().getMessage());
+        // Log error
+        LOGGER.log(
+            Level.SEVERE, "Error in vision API call: " +visionResponse.getError().getMessage());
+        return;
       }
     } catch (IOException e) {
-      // Log error (since IOException cannot be thrown by a function)
+      // Log error (since IOException cannot be thrown by a Cloud Function)
       LOGGER.log(Level.SEVERE, "Error detecting text: " + e.getMessage(), e);
       return;
     }
--- a/functions/snippets/src/main/java/com/example/functions/OcrProcessImage.java
+++ b/functions/snippets/src/main/java/com/example/functions/OcrProcessImage.java
@@ -61,7 +61,7 @@ public class OcrProcessImage implements BackgroundFunction<GcsEvent> {
 
   // [START functions_ocr_process]
   @Override
-  public void accept(GcsEvent gcsEvent, Context context) throws RuntimeException {
+  public void accept(GcsEvent gcsEvent, Context context) {
 
     // Validate parameters
     String bucket = gcsEvent.getBucket();
@@ -78,7 +78,7 @@ public class OcrProcessImage implements BackgroundFunction<GcsEvent> {
   // [END functions_ocr_process]
 
   // [START functions_ocr_detect]
-  private void detectText(String bucket, String filename) throws RuntimeException {
+  private void detectText(String bucket, String filename) {
     LOGGER.info("Looking for text in image " + filename);
 
     List<AnnotateImageRequest> visionRequests = new ArrayList<>();
@@ -123,12 +123,14 @@ public class OcrProcessImage implements BackgroundFunction<GcsEvent> {
     DetectLanguageResponse languageResponse;
     try (TranslationServiceClient client = TranslationServiceClient.create()) {
       languageResponse = client.detectLanguage(languageRequest);
-    } catch (IOException e1) {
-      // Cast to RuntimeException
-      throw new RuntimeException(e1);
+    } catch (IOException e) {
+      // Log error
+      LOGGER.severe("Error detecting language: " + e.getMessage());
+      e.printStackTrace();
+      return;
     }
 
-    if (languageResponse == null || languageResponse.getLanguagesCount() == 0) {
+    if (languageResponse.getLanguagesCount() == 0) {
       LOGGER.info("No languages were detected for text: " + text);
       return;
     }
@@ -144,9 +146,11 @@ public class OcrProcessImage implements BackgroundFunction<GcsEvent> {
       PubsubMessage pubsubApiMessage = PubsubMessage.newBuilder().setData(byteStr).build();
       try {
         publisher.publish(pubsubApiMessage).get();
-      } catch (InterruptedException | ExecutionException e2) {
-        // Cast to RuntimeException
-        throw new RuntimeException(e2);
+      } catch (InterruptedException | ExecutionException e) {
+        // Log error
+        LOGGER.severe("Error publishing translation request: " + e.getMessage());
+        e.printStackTrace();
+        return;
       }
     }
   }
--- a/functions/snippets/src/main/java/com/example/functions/OcrProcessImage.java
+++ b/functions/snippets/src/main/java/com/example/functions/OcrProcessImage.java
@@ -41,6 +41,8 @@ import java.util.List;
 import java.util.concurrent.ExecutionException;
 import java.util.logging.Logger;
 
+// [END functions_ocr_process]
+
 public class OcrProcessImage implements BackgroundFunction<GcsEvent> {
   // TODO<developer> set these environment variables
   private static final String PROJECT_ID = System.getenv("GCP_PROJECT");
@@ -57,6 +59,7 @@ public class OcrProcessImage implements BackgroundFunction<GcsEvent> {
         ProjectTopicName.of(PROJECT_ID, TRANSLATE_TOPIC_NAME)).build();
   }
 
+  // [START functions_ocr_process]
   @Override
   public void accept(GcsEvent gcsEvent, Context context) throws RuntimeException {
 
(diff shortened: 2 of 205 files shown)


functions/snippets/pom.xml
```diff
@@ -8,14 +8,14 @@
...
-    <artifactId>shared-configuration</artifactId>
-    <version>1.0.13</version>
+  <!--  
+   The parent pom defines common style checks and testing strategies for our samples. 
+   Removing or replacing it should not affect the execution of the samples in anyway. 
+ -->  
+  <parent>  
+    <groupId>com.google.cloud.samples</groupId> 
+    <artifactId>shared-configuration</artifactId> 
+    <version>1.0.13</version>
```
lesv, line 18: 1.0.14

functions/snippets/pom.xml
```diff
@@ -27,7 +27,36 @@
     <groovy.version>3.0.2</groovy.version>
   </properties>
 
+  <dependencyManagement>
+    <dependencies>
+      <dependency>
+        <groupId>com.google.cloud</groupId>
+        <artifactId>libraries-bom</artifactId>
+        <version>4.0.0</version>
```
lesv, line 35: 4.4.0

functions/snippets/src/main/java/com/example/functions/OcrProcessImage.java
```diff
@@ -0,0 +1,152 @@
...
+import com.google.gson.Gson;
+import com.google.protobuf.ByteString;
+import com.google.pubsub.v1.ProjectTopicName;
+import com.google.pubsub.v1.PubsubMessage;
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.concurrent.ExecutionException;
+import java.util.logging.Logger;
+
```
lesv, line 43: I would expect you'd want to `[END functions_ocr_process]` here and start it just before `accept()`.

functions/snippets/src/main/java/com/example/functions/OcrProcessImage.java
```diff
@@ -0,0 +1,152 @@
...
+  private static Publisher publisher;
+  private static final Gson gson = new Gson();
+
+  public OcrProcessImage() throws IOException {
+    publisher = Publisher.newBuilder(
+        ProjectTopicName.of(PROJECT_ID, TRANSLATE_TOPIC_NAME)).build();
+  }
+
+  @Override
+  public void accept(GcsEvent gcsEvent, Context context) throws RuntimeException {
```
lesv, line 61: This method seems a bit tiny and pointless as a teaching example - it really only has a single significant line.  I wonder if you might want to do this a bit differently?

functions/snippets/src/main/java/com/example/functions/OcrProcessImage.java
```diff
@@ -0,0 +1,152 @@
...
+    try (ImageAnnotatorClient client = ImageAnnotatorClient.create()) {
+      visionResponse = client.batchAnnotateImages(visionRequests).getResponses(0);
+      if (visionResponse.hasError()) {
+        LOGGER.severe("Error detecting text: " + visionResponse.getError().getMessage());
+        return;
+      }
+    } catch (IOException e) {
+      // Cast to RuntimeException
+      throw new RuntimeException(e);
+    }
```
lesv, line 103: Why do we do this?

lesv: Just because you use a try w/ resources doesn't mean you need to ever catch anything.

functions/snippets/src/main/java/com/example/functions/OcrProcessImage.java
```diff
@@ -0,0 +1,152 @@
...
+        AnnotateImageRequest.newBuilder().addFeatures(textFeature).setImage(img).build();
+    visionRequests.add(visionRequest);
+
+    // Detect text in an image using the Cloud Vision API
+    AnnotateImageResponse visionResponse;
+    try (ImageAnnotatorClient client = ImageAnnotatorClient.create()) {
+      visionResponse = client.batchAnnotateImages(visionRequests).getResponses(0);
+      if (visionResponse.hasError()) {
+        LOGGER.severe("Error detecting text: " + visionResponse.getError().getMessage());
+        return;
```
lesv, line 98: Why don't you throw something here?

functions/snippets/src/main/java/com/example/functions/OcrProcessImage.java
```diff
@@ -0,0 +1,152 @@
...
+            .setMimeType("text/plain")
+            .setContent(text)
+            .build();
+    DetectLanguageResponse languageResponse;
+    try (TranslationServiceClient client = TranslationServiceClient.create()) {
+      languageResponse = client.detectLanguage(languageRequest);
+    } catch (IOException e) {
+      // Cast to RuntimeException
+      throw new RuntimeException(e);
+    }
```
lesv, line 125: Why?

functions/snippets/src/main/java/com/example/functions/OcrProcessImage.java
```diff
@@ -0,0 +1,152 @@
...
+        LOGGER.severe("Error detecting text: " + visionResponse.getError().getMessage());
+        return;
+      }
+    } catch (IOException e) {
+      // Cast to RuntimeException
+      throw new RuntimeException(e);
+    }
+
+    if (visionResponse == null || !visionResponse.hasFullTextAnnotation()) {
+      return;
```
lesv, line 106: Is this a success?

functions/snippets/src/main/java/com/example/functions/OcrProcessImage.java
```diff
@@ -0,0 +1,152 @@
...
+    DetectLanguageResponse languageResponse;
+    try (TranslationServiceClient client = TranslationServiceClient.create()) {
+      languageResponse = client.detectLanguage(languageRequest);
+    } catch (IOException e) {
+      // Cast to RuntimeException
+      throw new RuntimeException(e);
+    }
+
+    if (languageResponse == null || languageResponse.getLanguagesCount() == 0) {
+      return;
```
lesv, line 128: Is this a success?

functions/snippets/src/main/java/com/example/functions/OcrProcessImage.java
```diff
@@ -0,0 +1,152 @@
...
+      LOGGER.info("Sending translation request for language " + targetLanguage);
+      OcrTranslateApiMessage message = new OcrTranslateApiMessage(text, filename, targetLanguage);
+      ByteString byteStr = ByteString.copyFrom(message.toPubsubData());
+      PubsubMessage pubsubApiMessage = PubsubMessage.newBuilder().setData(byteStr).build();
+      try {
+        publisher.publish(pubsubApiMessage).get();
+      } catch (InterruptedException | ExecutionException e) {
+        // Cast to RuntimeException
+        throw new RuntimeException(e);
+      }
```
lesv, line 145: Why?

functions/snippets/src/main/java/com/example/functions/OcrTranslateApiMessage.java
```diff
@@ -0,0 +1,77 @@
...
+    return text;
+  }
+
+  public String getFilename() {
+    return filename;
+  }
+
+  public String getLang() {
+    return lang;
+  }
```
lesv, line 49: Why?

functions/snippets/src/main/java/com/example/functions/OcrTranslateApiMessage.java
```diff
@@ -0,0 +1,77 @@
...
+
+  public static OcrTranslateApiMessage fromPubsubData(byte[] data)
+      throws IllegalArgumentException {
+    String jsonStr = new String(Base64.getDecoder().decode(data));
+    OcrTranslateApiMessage ocrMessage = gson.fromJson(jsonStr, OcrTranslateApiMessage.class);
+
+    // Get + verify parameters
+    String text = ocrMessage.getText();
+    String filename = ocrMessage.getFilename();
+    String targetLang = ocrMessage.getLang();
```
lesv, line 59: Since it's your class, you can access directly.  No reason to use the accessors.

functions/snippets/src/main/java/com/example/functions/OcrTranslateText.java
```diff
@@ -0,0 +1,99 @@
...
+    try {
+      ByteString byteStr = ByteString.copyFrom(translateMessage.toPubsubData());
+      PubsubMessage pubsubApiMessage = PubsubMessage.newBuilder().setData(byteStr).build();
+
+      publisher.publish(pubsubApiMessage).get();
+      LOGGER.info("Text translated to " + targetLang);
+    } catch (InterruptedException | ExecutionException e) {
+      // Cast to RuntimeException
+      throw new RuntimeException(e);
+    }
```
lesv, line 96: Why?

functions/snippets/src/main/java/com/example/functions/OcrTranslateText.java
```diff
@@ -0,0 +1,99 @@
...
+            .addContents(text)
+            .build();
+
+    TranslateTextResponse response;
+    try (TranslationServiceClient client = TranslationServiceClient.create()) {
+      response = client.translateText(request);
+    } catch (IOException e) {
+      // Cast to RuntimeException
+      throw new RuntimeException(e);
+    }
```
lesv, line 75: Why?

functions/snippets/src/main/java/com/example/functions/OcrProcessImage.java
ace-n: The Base `BackgroundFunction` type in the Functions Framework can **only** throw subclasses of `RuntimeException`, which these are not.

(IntelliJ requires the exception to be addressed in a `catch` block - even if that block merely ignores the error.)

functions/snippets/src/main/java/com/example/functions/OcrProcessImage.java
ace-n: See comment above.

functions/snippets/src/main/java/com/example/functions/OcrTranslateText.java
ace-n: Ditto.

functions/snippets/src/main/java/com/example/functions/OcrTranslateText.java
ace-n: Ditto.

functions/snippets/src/main/java/com/example/functions/OcrTranslateApiMessage.java
ace-n: Done.

functions/snippets/src/main/java/com/example/functions/OcrTranslateApiMessage.java
ace-n: @kurtisvg has a preference for getters/setters - but I can drop these if you'd prefer.

functions/snippets/src/main/java/com/example/functions/OcrProcessImage.java
ace-n: I was [copying the Node example](https://github.com/GoogleCloudPlatform/nodejs-docs-samples/blob/master/functions/ocr/app/index.js#L113-L136).

(Though I agree this might be worth refactoring - `detectText` is a borderline "God method", so maybe we should move the Pub/Sub publishing code to this method instead?)

functions/snippets/src/main/java/com/example/functions/OcrProcessImage.java
ace-n: Done.

functions/snippets/pom.xml
ace-n: Done.

functions/snippets/pom.xml
ace-n: Done.

functions/snippets/src/main/java/com/example/functions/OcrProcessImage.java
ace-n: Switched to `RuntimeException`.

functions/snippets/src/main/java/com/example/functions/OcrProcessImage.java
ace-n: Ditto.

functions/snippets/src/main/java/com/example/functions/OcrProcessImage.java
ace-n: Added a log statement. (See above comment for more info.)

functions/snippets/src/main/java/com/example/functions/OcrProcessImage.java
ace-n: Added a log statement.

I'm reluctant to throw an exception here, as this could be caused by bad _user_ input (as opposed to _programmer_ error.)

functions/snippets/src/main/java/com/example/functions/OcrProcessImage.java
lesv: Let's keep things simple and only ask the user to follow one method.

functions/snippets/src/main/java/com/example/functions/OcrProcessImage.java
lesv: https://google.github.io/styleguide/javaguide.html#s6.2-caught-exceptions
Also see Effective Java v3 Exceptions
* Item 69: Use exceptions only for exceptional conditions 
* Item 70: Use checked exceptions for recoverable conditions and runtime exceptions for programming errors
* Item 71: Avoid unnecessary use of checked exceptions 
* Item 72: Favor the use of standard exceptions 
* Item 73: Throw exceptions appropriate to the abstraction
* Item 74: Document all exceptions thrown by each method 
* Item 77: Don’t ignore exceptions

functions/snippets/src/main/java/com/example/functions/OcrTranslateApiMessage.java
lesv: Your not using them outside your class, so I'd prefer dropping them, but feel free to ask kurtis about my comment.

functions/snippets/src/main/java/com/example/functions/OcrProcessImage.java
lesv: What if you just say `throws IOException` ?

ace-n: Ah - so (some of) these seem more like "checked" exceptions.

Would it be better to log them and exit the function?

functions/snippets/src/main/java/com/example/functions/OcrTranslateText.java
lesv, line 75: This is still very wrong.

functions/snippets/src/main/java/com/example/functions/OcrTranslateApiMessage.java
ace-n: Will do. Also, these _are_ used outside the class (e.g. [here](https://github.com/GoogleCloudPlatform/java-docs-samples/pull/2552/files#diff-7400aeba097e7758edf3c3f4cc8a9b7aR41-R43))

functions/snippets/src/main/java/com/example/functions/OcrTranslateText.java
ace-n: Compiler error: `Overridden method accept(...) does not throw IOException`

functions/snippets/src/main/java/com/example/functions/OcrTranslateApiMessage.java
kurtisvg: Spoke with Les - we should use getters/setter from outside of the class (like where you linked) but make sure to use the internal reference (like in `fromPubSubData`).

ace-n:
**N.B:** I fixed some of the relevant `spotbugs` errors, but the _try-with-resources_ pattern is causing issues as documented [here](https://github.com/spotbugs/spotbugs/issues/756).

functions/snippets/src/main/java/com/example/functions/OcrTranslateApiMessage.java
```diff
@@ -0,0 +1,77 @@
...
+      throw new IllegalArgumentException("Missing filename parameter");
+    }
+    if (targetLang == null) {
+      throw new IllegalArgumentException("Missing lang parameter");
+    }
+
+    return ocrMessage;
+  }
+
+  public byte[] toPubsubData() {
```
kurtisvg, line 73: Any reason not to use `com.google.pubsub.v1.PubsubMessage` over byte? seems like a weird place to split up the helper functions.

functions/snippets/src/main/java/com/example/functions/OcrSaveResult.java
```diff
@@ -0,0 +1,58 @@
...
+  private static final Logger LOGGER = Logger.getLogger(OcrSaveResult.class.getName());
+
+  @Override
+  public void accept(PubSubMessage pubSubMessage, Context context) {
+    OcrTranslateApiMessage ocrMessage = OcrTranslateApiMessage.fromPubsubData(
+        pubSubMessage.data.getBytes());
+
+    String text = ocrMessage.getText();
+    String filename = ocrMessage.getFilename();
+    String lang = ocrMessage.getLang();
```
kurtisvg, line 43: Any reason not to inline these three variables?

functions/snippets/src/main/java/com/example/functions/OcrTranslateText.java
```diff
@@ -0,0 +1,99 @@
...
+    if (response == null || response.getTranslationsCount() == 0) {
```
kurtisvg, line 76: Why are we checking if `response == null`? Does the client actually return that as a result? 

Can we add a comment inside function explaining why it's returning?

functions/snippets/src/test/java/com/example/functions/OcrTests.java
```diff
@@ -0,0 +1,150 @@
...
+
+  @Test(expected = IllegalArgumentException.class)
+  public void functionsOcrProcess_shouldValidateParams() throws IOException {
+    new OcrProcessImage().accept(new GcsEvent(), null);
+  }
+
+  @Test
+  public void functionsOcrProcess_shouldDetectText() throws IOException {
+    GcsEvent gcsEvent = new GcsEvent();
+    gcsEvent.bucket = FUNCTIONS_BUCKET;
```
kurtisvg, line 72: Should use either a setter (or constructor) to set these fields.

functions/snippets/src/test/java/com/example/functions/OcrTests.java
```diff
@@ -0,0 +1,150 @@
...
+    // Check log messages
+    List<LogRecord> logs = LOG_HANDLER.getStoredLogRecords();
+    String expectedMessage = String.format(
+        "Saving result to %s in bucket %s", resultFilename, RESULT_BUCKET);
+    assertThat(LOG_HANDLER.getStoredLogRecords().get(1).getMessage()).isEqualTo(expectedMessage);
+
+    // Check for (recently) written file
+    // (Using timestamps instead of exists() saves us a deletion step)
+    BlobInfo resultBlob = storage.get(RESULT_BUCKET, resultFilename);
+    Long fiveMinutesAgo = System.currentTimeMillis() - 60000 * 5;
```
kurtisvg, line 146: Please use `java.time` for this instead.  You can use something like `Instance.now().minus(5, MINUTES)`

Additionally, this check doesn't actually verify it worked when run in parallel (as we do with our unit tests).

kurtisvg:
> **N.B:** I fixed some of the relevant `spotbugs` errors, but the _try-with-resources_ pattern is causing issues as documented [here](https://github.com/spotbugs/spotbugs/issues/756).

I opened https://github.com/GoogleCloudPlatform/java-repo-tools/issues/123 to surpress this everywhere, but for now you should be able to use something like the following on the classes affected:
```java
@SuppressFBWarnings(value = "RCN_REDUNDANT_NULLCHECK_WOULD_HAVE_BEEN_A_NPE",
        justification = "Scanning generated code of try-with-resources")
```

functions/snippets/src/main/java/com/example/functions/OcrProcessImage.java
```diff
@@ -0,0 +1,156 @@
...
+  private static final Gson gson = new Gson();
+
+  public OcrProcessImage() throws IOException {
+    publisher = Publisher.newBuilder(
+        ProjectTopicName.of(PROJECT_ID, TRANSLATE_TOPIC_NAME)).build();
+  }
+
+  // [START functions_ocr_process]
+  @Override
+  public void accept(GcsEvent gcsEvent, Context context) throws RuntimeException {
```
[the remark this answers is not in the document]
ace-n: Done.

functions/snippets/src/main/java/com/example/functions/OcrTranslateApiMessage.java
```diff
@@ -0,0 +1,77 @@
...
+
+  public String getFilename() {
+    return filename;
+  }
+
+  public String getLang() {
+    return lang;
+  }
+
+  public static OcrTranslateApiMessage fromPubsubData(byte[] data)
```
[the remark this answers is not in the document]
ace-n: There's also a `PubSubMessage` (capitalized `s`) POJO here, that's necessary for deserializing incoming Pub/Sub events.

(I tried deserializing into the `com.google.pubsub.v1.PubsubMessage` type, but the naive `Gson` deserializer couldn't handle it. Longer-term, we should probably write a custom deserializer for this class.)

functions/snippets/src/main/java/com/example/functions/OcrTranslateApiMessage.java
```diff
@@ -0,0 +1,77 @@
...
+      throws IllegalArgumentException {
+    String jsonStr = new String(Base64.getDecoder().decode(data));
```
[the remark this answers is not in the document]
ace-n: See conversation above - it doesn't (yet) work with Cloud Functions' event de-serialization logic.

functions/snippets/src/main/java/com/example/functions/OcrTranslateApiMessage.java
```diff
@@ -0,0 +1,77 @@
...
+  public String getFilename() {
+    return filename;
+  }
+
+  public String getLang() {
+    return lang;
+  }
+
+  public static OcrTranslateApiMessage fromPubsubData(byte[] data)
+      throws IllegalArgumentException {
```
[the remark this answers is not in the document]
ace-n: Per your comment below, I moved this into the constructor.

functions/snippets/src/main/java/com/example/functions/OcrTranslateApiMessage.java
```diff
@@ -0,0 +1,77 @@
...
+    String jsonStr = new String(Base64.getDecoder().decode(data));
+    OcrTranslateApiMessage ocrMessage = gson.fromJson(jsonStr, OcrTranslateApiMessage.class);
```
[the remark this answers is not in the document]
ace-n: Moved validation into the constructor - PTAL?

functions/snippets/src/main/java/com/example/functions/OcrTranslateText.java
ace-n: This was a defensive programming reflex - removed. 🙂

functions/snippets/src/test/java/com/example/functions/OcrTests.java
ace-n: Done.

functions/snippets/src/test/java/com/example/functions/OcrTests.java
ace-n: Replaced this with a UUID'd filename + existence check. 🙂

(The downside is that this also required implementing teardown code.)

ace-n:
@kurtisvg thanks for the tip RE `spotbugs`.

It's triggering on other `functions` samples, so I'll probably fix all of them in a separate PR (to avoid cluttering this one).

I've addressed the **non-NPE** issues in with _these_ samples, however.

functions/snippets/src/main/java/com/example/functions/OcrProcessImage.java
```diff
@@ -0,0 +1,156 @@
...
+import com.google.protobuf.ByteString;
+import com.google.pubsub.v1.ProjectTopicName;
+import com.google.pubsub.v1.PubsubMessage;
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.concurrent.ExecutionException;
+import java.util.logging.Logger;
+
+// [END functions_ocr_process]
```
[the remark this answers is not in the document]
ace-n: The existing samples do that. (@lesv had similar feedback)

If you'd like, I can move the Pub/Sub publishing code to `functions_ocr_process` since `detectText` is quite long.

functions/snippets/src/main/java/com/example/functions/OcrProcessImage.java
```diff
@@ -0,0 +1,161 @@
...
+      LOGGER.info("Sending translation request for language " + targetLanguage);
+      OcrTranslateApiMessage message = new OcrTranslateApiMessage(text, filename, targetLanguage);
+      ByteString byteStr = ByteString.copyFrom(message.toPubsubData());
+      PubsubMessage pubsubApiMessage = PubsubMessage.newBuilder().setData(byteStr).build();
+      try {
+        publisher.publish(pubsubApiMessage).get();
+      } catch (InterruptedException | ExecutionException e) {
+        // Log error
+        LOGGER.severe("Error publishing translation request: " + e.getMessage());
+        e.printStackTrace();
```
kurtisvg, line 152: printStackTrace only goes to stdout
```suggestion
        LOGGER.log(Level.SEVERE, "Error publishing translation request: " + e.getMessage(), e);
```

functions/snippets/src/main/java/com/example/functions/OcrProcessImage.java
kurtisvg: Please log over `RuntimeException` everywhere.

functions/snippets/src/main/java/com/example/functions/OcrProcessImage.java
kurtisvg: (please fix everywhere else)

ace-n: Done.

functions/snippets/pom.xml
```diff
@@ -124,7 +130,7 @@
     <dependency>
       <groupId>com.google.truth</groupId>
       <artifactId>truth</artifactId>
-      <version>1.0.1</version>
+      <version>0.42</version>
```
kurtisvg, line 133: nit: Did you mean to downgrade this?

functions/snippets/src/main/java/com/example/functions/OcrProcessImage.java
```diff
@@ -0,0 +1,161 @@
...
+    try (ImageAnnotatorClient client = ImageAnnotatorClient.create()) {
+      visionResponse = client.batchAnnotateImages(visionRequests).getResponses(0);
+      if (visionResponse == null || !visionResponse.hasFullTextAnnotation()) {
+        LOGGER.info(String.format("Image %s contains no text", filename));
+        return;
+      }
+
+      if (visionResponse.hasError()) {
+        // Throw IOException; to be handled by catch block below
+        throw new IOException(visionResponse.getError().getMessage());
```
kurtisvg, line 107: We shouldn't rewrap errors like this - the `visionResponse` isn't an `IOException`.

functions/snippets/pom.xml
ace-n: Oops - fixed!

functions/snippets/src/main/java/com/example/functions/OcrProcessImage.java
ace-n: Added a separate logging statement instead.

ace-n:
@lesv please LGTM **and** merge if this looks good to you - thanks!

PR #2557, opened 2020-04-02
[Cloud Run] Auth Snippet

averikitsch:
Fixes #<issue_number_goes_here>

- [ ] Tests pass

### DIFF
--- a/run/authentication/src/test/java/com/example/cloudrun/AuthenticationTest.java
+++ b/run/authentication/src/test/java/com/example/cloudrun/AuthenticationTest.java
@@ -1,0 +1,59 @@
+/*
+ * Copyright 2020 Google LLC
+ *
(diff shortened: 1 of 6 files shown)


run/authentication/src/test/java/com/example/cloudrun/AuthenticationTest.java
```diff
@@ -0,0 +1,59 @@
...
+  @After
+  public void tearDown() {
+    System.setOut(null);
+  }
+
+  @Test
+  public void canMakeGetRequest() throws IOException {
+    String url = "http://example.com/";
+    String expectedResp;
+    if (System.getenv("GOOGLE_CLOUD_PROJECT") != null) {
```
grayside, line 49: If run locally, `GOOGLE_CLOUD_PROJECT` is completely optional. If run without that variable set it may be worth logging some corrective instruction.

PR #2559, opened 2020-04-03
docs: add additional samples for Spanner JDBC

olavloite:
Adds additional samples for the Cloud Spanner open source JDBC driver.


lesv:
ping?

olavloite:
> ping?

Sorry for the slow response, and thanks for the review. It might take a little time to get all the region tags. In the meantime I'll make sure the other issues you pointed out are fixed.

olavloite:
I'm looking into this now, but I'm getting what I would say is a false positive warning from spotbugs. Spotbugs complains that the following code is not cleaning up all resources:

```java
  static void createConnection(String projectId, String instanceId, String databaseId)
      throws SQLException {
    try (Connection connection =
        DriverManager.getConnection(
            String.format(
                "jdbc:cloudspanner:/projects/%s/instances/%s/databases/%s",
                projectId, instanceId, databaseId))) {
      try (Statement statement = connection.createStatement()) {
        try (ResultSet rs = statement.executeQuery("SELECT CURRENT_TIMESTAMP()")) {
          while (rs.next()) {
            System.out.printf(
                "Connected to Cloud Spanner at [%s]", rs.getTimestamp(1).toString());
          }
        }
      }
    }
  }
```

This seems to be a known issue with spotbugs: https://github.com/spotbugs/spotbugs/issues/293

olavloite:
@lesv 

* I've refactored the samples to use the new style and fixed all issues reported by `mvn clean package verify`. One problem with that is that spotbugs seems to [report false positives](https://github.com/GoogleCloudPlatform/java-docs-samples/pull/2559#issuecomment-611978649) when using try-with-resources. This forces me to put `@SuppressFBWarnings` on every sample, as all JDBC samples use this idiom. Do you know of any other solution to this?
* Removed all Javadoc.
* The integration tests pass.

Region tags are still missing for most samples.

lesv:
If you give me a list, I'll make sure to add it to the [exclude filter](https://github.com/GoogleCloudPlatform/java-repo-tools/blob/master/checkstyle-config/src/main/resources/c.g.c.s/spotBugsExcludeFilter.xml).

spanner/jdbc/src/main/java/com/example/spanner/jdbc/CreateConnectionWithCustomHostExample.java
```diff
@@ -0,0 +1,53 @@
...
+      value = "OBL_UNSATISFIED_OBLIGATION",
+      justification = "https://github.com/spotbugs/spotbugs/issues/293")
+  // Creates a JDBC connection to a Cloud Spanner database on a custom host.
+  static void createConnectionWithCustomHost(
+      String projectId, String instanceId, String databaseId, int port) throws SQLException {
+    try (Connection connection =
+        DriverManager.getConnection(
+            String.format(
+                "jdbc:cloudspanner://localhost:%d/projects/%s/instances/%s/databases/%s"
+                    + "?usePlainText=true",
```
lesv, line 46: I'm sure there will be text explaining this in the docs, but a comment or two might be helpful here.

spanner/jdbc/src/main/java/com/example/spanner/jdbc/TransactionWithRetryLoopExample.java
```diff
@@ -0,0 +1,85 @@
...
+                  .build();
+          spannerConnection.bufferedWrite(Arrays.asList(mutationSingers, mutationAlbums));
+          spannerConnection.commit();
+          System.out.printf(
+              "Transaction committed at [%s]%n", spannerConnection.getCommitTimestamp().toString());
+          break;
+        } catch (JdbcAbortedException e) {
+          // Transaction aborted, retry.
+          System.out.println("Transaction aborted, starting retry");
+        }
```
lesv, line 81: It appears the assumption is that it will eventually succeed, that might not be true in all cases.  Or is it?

spanner/jdbc/src/main/java/com/example/spanner/jdbc/TransactionWithRetryLoopUsingOnlyJdbcExample.java
```diff
@@ -0,0 +1,81 @@
...
+          }
+          break;
+        } catch (SQLException e) {
+          if (e.getErrorCode() == Code.ABORTED.value()) {
+            // Transaction aborted, retry.
+            System.out.println("Transaction aborted, starting retry");
+          } else {
+            throw e;
+          }
+        }
```
lesv, line 77: I really like this, but I wonder again if an infinite loop is possible here?

spanner/jdbc/src/main/java/com/example/spanner/jdbc/TransactionWithRetryLoopExample.java
olavloite: It will always eventually succeed, or fail with a different error than `Aborted`. The JDBC transactions use a Spanner `TransactionManager` under the hood, which should be used with a loop like this:

https://github.com/googleapis/java-spanner/blob/c8be7b5d43c01f6170eb02c6e8cc146cb2597ecc/google-cloud-spanner/src/main/java/com/google/cloud/spanner/DatabaseClient.java#L252-L279

spanner/jdbc/src/main/java/com/example/spanner/jdbc/TransactionWithRetryLoopUsingOnlyJdbcExample.java
olavloite: This will also always eventually succeed or fail with a different error code than `Aborted`.

spanner/jdbc/src/main/java/com/example/spanner/jdbc/CreateConnectionWithCustomHostExample.java
olavloite: Good point, added a comment explaining the `usePlainText` property.

olavloite:
> If you give me a list, I'll make sure to add it to the [exclude filter](https://github.com/GoogleCloudPlatform/java-repo-tools/blob/master/checkstyle-config/src/main/resources/c.g.c.s/spotBugsExcludeFilter.xml).

The following samples currently require this annotation:

```
@SuppressFBWarnings(
      value = "OBL_UNSATISFIED_OBLIGATION",
      justification = "https://github.com/spotbugs/spotbugs/issues/293")
```

---

```java
AbortBatchExample.abortBatch
AutocommitUpdateDataExample.update
BatchDdlExample.batchDdl
BatchDdlUsingSqlStatementsExample.batchDdlUsingSqlStatements
BatchDmlExample.batchDml
BatchDmlUsingSqlStatementsExample.batchDmlUsingSqlStatements
ConnectionWithQueryOptionsExample.connectionWithQueryOptions
CreateConnectionExample.createConnection
CreateConnectionWithCredentialsExample.createConnectionWithCredentials
CreateConnectionWithCustomHostExample.createConnectionWithCustomHost
CreateConnectionWithDataSourceExample.createConnectionWithDataSource
CreateConnectionWithDefaultProjectIdExample.createConnectionWithDefaultProjectId
CreateConnectionWithPropertiesExample.createConnectionWithProperties
CreateConnectionWithUrlPropertiesExample.createConnectionWithUrlProperties
CreateTableExample.createTable
GetCommitTimestampExample.getCommitTimestamp
GetReadTimestampExample.getReadTimestamp
InsertDataExample.insertData
PartitionedDmlExample.partitionedDml
ReadOnlyTransactionExample.readOnlyTransaction
ReadWriteTransactionExample.readWriteTransaction
SetQueryOptionsExample.setQueryOptions
SingleUseReadOnlyExample.singleUseReadOnly
SingleUseReadOnlyTimestampBoundExample.singleUseReadOnlyTimestampBound
```

olavloite:
@lesv Thanks for the review. I will submit a new PR to remove the `@SuppressFBWarnings` when the samles have been added to the exlude filter.

PR #2564, opened 2020-04-04
Fix [some] GCF spotbugs errors

ace-n:
This fixes **some** of the GCF `spotbugs` errors. The others likely need to be whitelisted, either globally or on a per-class basis.

The following rules were creating lots of false positives, and should (probably) be whitelisted globally:
- `IMC_IMMATURE_CLASS_NO_TOSTRING`
- `CRLF_INJECTION_LOGS`
- _possibly_ `CLI_CONSTANT_LIST_INDEX`

Once these rules are whitelisted (and `shared-configuration` is updated accordingly), I'll re-run `spotbugs` and fix any issues I missed. 🙂

### DIFF
--- a/functions/snippets/src/main/java/com/example/functions/CorsEnabled.java
+++ b/functions/snippets/src/main/java/com/example/functions/CorsEnabled.java
@@ -33,7 +33,7 @@ public class CorsEnabled implements HttpFunction {
   public void service(HttpRequest request, HttpResponse response)
       throws IOException {
     // Set CORS headers for preflight requests
-    if (request.getMethod().equals("OPTIONS")) {
+    if ("OPTIONS".equals(request.getMethod())) {
       response.appendHeader("Access-Control-Allow-Origin", "*");
       response.appendHeader("Access-Control-Allow-Methods", "POST");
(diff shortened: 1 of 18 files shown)


functions/snippets/src/main/java/com/example/functions/CorsEnabled.java
```diff
@@ -33,7 +33,7 @@
   public void service(HttpRequest request, HttpResponse response)
       throws IOException {
     // Set CORS headers for preflight requests
-    if (request.getMethod().equals("OPTIONS")) {
+    if ("OPTIONS".equals(request.getMethod())) {
```
lesv, line 36: I'd be interested in hearing a bit more about this, it's a tiny bit surprising, but probably correct.

ace-n:
I think the philosophy here is to avoid an NPE if the non-literal is
`null`. (Even if the programmer knows this can't happen, the compiler may
not.)

lesv:
Ace - do you plan to close this?

ace-n:
Yes - I'll close this in favor of a new PR. #tooManyMergeConflicts

PR #2601, opened 2020-04-08
Split files into separate packages

ace-n:
(Apologies for the long PR - I thought this would be 10x {easier, quicker, more concise} than it was. That said, **let me know if I should shard it into separate PRs.** I'd recommend **ignoring** the `GcsEvent.java`, `MockContext.java`, and `PubSubMessage.java` files - as they're all identical.)

**Motivation**
Builds are failing on GCF, in part due to memory constraints. Hopefully splitting each function (set) into its own file helps with this.

**Highlights** _(skim the diff and read this carefully instead)_
- I added the `eventpojos` subpackage(s) for GCF's event-storing POJOs. Ideally, these would be in a separate Maven-repo'd package - but given that a) we're not there yet and b) I wanted to avoid "sibling dependencies", this was the only option. (They're easy enough to remove anyway, when the time comes.)
- Some minor refactoring was done (namely `fields` --> `getters/setters`, `checkstyle` line wrapping), but only where necessary to get the build passing. (Unfortunately, the changes were not as simple as `git mv`.) **Everything else was copy-pasted from existing samples.**


functions/snippets/concepts/pom.xml
```diff
@@ -0,0 +1,149 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
lesv, line 1: Shouldn't there be a copyright here?

functions/snippets/concepts/pom.xml
```diff
@@ -0,0 +1,149 @@
...
+      <properties>
+        <skipTests>true</skipTests>
+      </properties>
+    </profile>
+  </profiles>
+
+  <build>
+    <plugins>
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
```
lesv, line 113: This is already in share-configuration w/o the skipTests.  Ok to keep.

functions/snippets/firebase/pom.xml
```diff
@@ -0,0 +1,130 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
lesv, line 1: Copyright & License?

functions/snippets/firebase/pom.xml
```diff
@@ -0,0 +1,130 @@
...
+      <groupId>com.google.cloud</groupId>
+      <artifactId>google-cloud-firestore</artifactId>
+      <version>1.32.5</version>
+    </dependency>
+
+    <!-- Required for Function primitives -->
+    <dependency>
+      <groupId>com.google.cloud.functions</groupId>
+      <artifactId>functions-framework-api</artifactId>
+      <version>1.0.0-alpha-2-rc3</version>
```
lesv, line 34: That version string isn't really appropriate.

functions/snippets/helloworld/pom.xml
```diff
@@ -0,0 +1,231 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
lesv, line 1: Copyright & license
```
<!--
blah blah blah
-->
```

functions/snippets/http/pom.xml
```diff
@@ -0,0 +1,150 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
lesv, line 1: Copyright & License?

functions/snippets/imagemagick/pom.xml
```diff
@@ -0,0 +1,181 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
lesv, line 1: Copyright & License?

functions/snippets/logging/pom.xml
```diff
@@ -0,0 +1,141 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
lesv, line 1: Copyright & Licenses?

functions/snippets/ocr/pom.xml
```diff
@@ -0,0 +1,127 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
lesv, line 1: Copyright & License?

functions/snippets/pom.xml
```diff
@@ -8,6 +8,18 @@
   <artifactId>functions-snippets</artifactId>
   <version>1.0-SNAPSHOT</version>
 
+  <packaging>pom</packaging>
```
lesv, line 11: This may execute multiple times in J-D-S as it does a search for pom.xml files, and not this style.

(ie. this might be redundant)

functions/snippets/slack/pom.xml
```diff
@@ -0,0 +1,297 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
lesv, line 1: Copyright & License?

functions/snippets/firebase/pom.xml
eamonnmcmanus: It's what we have, but yes, it needs to be changed and will be.

functions/snippets/helloworld/src/main/scala/com/example/functions/helloworld/ScalaHelloBackground.scala
```diff
@@ -1,9 +1,10 @@
-package com.example.functions
+package com.example.functions.helloworld
 
 // [START functions_helloworld_background]
-import com.google.cloud.functions.{BackgroundFunction, Context, HttpRequest}
 import java.util.logging.Logger
 
+import com.google.cloud.functions.{BackgroundFunction, Context, HttpRequest}
+
 class ScalaHelloBackground extends BackgroundFunction[HttpRequest] {
```
eamonnmcmanus, line 8: We can address it later, but this seems wrong to me. `HttpRequest` isn't the sort of class you would expect to see as the payload of an event.

kurtisvg:
> PTAL - the only thing that's important is the copyright & license on the pom.xml.
> The use of a parent pom is probably not required, but is probably ok.

Please do not use a parent pom. Each pom should be separate and use the shared-config as the parent.

functions/snippets/pom.xml
ace-n: This file was deleted, per @kurtisvg's suggestion below. 🙂

functions/snippets/concepts/src/main/java/com/example/functions/concepts/eventpojos/GcsEvent.java
```diff
@@ -0,0 +1,72 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.example.functions.concepts.eventpojos;
+
+import java.util.Date;
+
+// [START functions_helloworld_gcs_event]
```
grant, line 21: We're going to have multiple region tag source in DRIFT if we have multiple files that have the same tag. Do we need it in multiple places?

functions/snippets/concepts/src/main/java/com/example/functions/concepts/eventpojos/GcsEvent.java
```diff
@@ -0,0 +1,72 @@
...
+public class GcsEvent {
```
grant, line 22: Do we expect users to copy and paste this class? Why are there multiple copies of this class in the samples rather than just one copy?

ace-n: Ideally we'd move them into a separate package in the central Maven repo - but copying is the only way to avoid "sibling dependencies" for now.

functions/snippets/concepts/src/main/java/com/example/functions/concepts/eventpojos/GcsEvent.java
ace-n: Commented below.

Issue #2252, opened 2020-02-27
Update Dialogflow tests


nnegrey:
Break up dialogflow samples into individual files
Update tests to be standalone.

PR #2483, opened 2020-03-24
dialogflow: remove unused samples and tests

Strykrol:
Tests are currently unaltered, as the tests have variable dependencies. An example being the `listContexts` function stores a count into a `contexts` variable, which is then used to verify `deleteContext` functions as expected. 

If I can get help refactoring test files that would be great. Otherwise, I will take a stab at them later this week.

PR #2602, opened 2020-04-08
dialogflow: delete samples that are not in docs

nnegrey:
fix the tests that are broken after the deletions
Update this PR: https://github.com/GoogleCloudPlatform/java-docs-samples/pull/2483

Fixes https://github.com/GoogleCloudPlatform/java-docs-samples/issues/2252

- [x] Tests pass


dialogflow/cloud-client/src/test/java/com/example/dialogflow/DetectIntentKnowledgeTest.java
```diff
@@ -37,20 +35,16 @@
...
- */
+
 @RunWith(JUnit4.class)
 @SuppressWarnings("checkstyle:abbreviationaswordinname")
-public class KnowledgeBaseManagementIT {
+public class DetectIntentKnowledgeTest {
 
   private static String PROJECT_ID = System.getenv().get("GOOGLE_CLOUD_PROJECT");
   private static String TEST_KNOWLEDGE_BASE_ID = "MTA4MzE0ODY5NTczMTQzNzU2ODA";
   private static String TEST_DOCUMENT_ID = "MTUwNjk0ODg1NTU4NzkzMDExMg";
```
lesv, line 45: Should we be concerned about these constants and running testing jobs in parallel. (ie. j8 & j11)

nnegrey: Not for this one. I used to have these tests dynamically create these resources, but the API has to do some ML training for an indeterminate time on the created resources which would create a flaky test occasionally. There are some more involved work-arounds to get around this, but would significantly complicate the tests.

So instead having this resource be static and used for "prediction" should handle any number of parallel requests and all of them should get an identical response and only a change to the Document we created should break this.

PR #2615, opened 2020-04-10
GCF CI/CD: add cloudbuild.yaml

ace-n:


functions/snippets/ci_cd/cloudbuild.yaml
```diff
@@ -0,0 +1,9 @@
+# [START functions_ci_cd_cloud_build]
```
lesv, line 1: Shouldn't there be a license and copyright?

ace-n: Done.

PR #2630, opened 2020-04-10
GCF: misc. idiomatic fixes

ace-n:
**Note:** I split up each fix into its own commit (to make this easier to review)

h/t @saturnism for the suggestions.


functions/concepts/src/main/java/com/example/functions/concepts/LazyFields.java
```diff
@@ -30,20 +30,26 @@
   // Always initialized (at cold-start)
   // Warning: Class variables used in Servlet classes must be thread-safe,
   // or else might introduce race conditions in your code.
-  private static final int nonLazyGlobal = fileWideComputation();
+  private static final int NON_LAZY_GLOBAL = fileWideComputation();
```
kurtisvg, line 33: nit: This says "fileWideComputation", but it's really "perStartupComputation". It's not done once per file, it's done once per class initialization.

functions/firebase/src/main/java/com/example/functions/firebase/FirebaseFirestoreReactive.java
```diff
@@ -70,7 +70,7 @@ public void accept(String json, Context context) throws RuntimeException {
 
     LOGGER.info(String.format("Replacing value: %s --> %s", currentValue, newValue));
     try {
-      firestore.document(affectedDoc).set(newFields, SetOptions.merge()).get();
+      FIRESTORE.document(affectedDoc).set(newFields, SetOptions.merge()).get();
     } catch (ExecutionException | InterruptedException e) {
       throw new RuntimeException(e);
```
kurtisvg, line 75: nit: I thought we had settled on logging this over wrapping in RuntimeException (@lesv FYI)

functions/firebase/src/main/java/com/example/functions/firebase/FirebaseFirestoreReactive.java
```diff
@@ -33,15 +33,15 @@
 public class FirebaseFirestoreReactive implements RawBackgroundFunction {
 
   // Use GSON (https://github.com/google/gson) to parse JSON content.
-  private Gson gsonParser = new Gson();
+  private Gson GSON_PARSER = new Gson();
```
kurtisvg, line 36: This isn't static, or final. (nit: Also I almost always see it called just `gson`)

functions/firebase/src/main/java/com/example/functions/firebase/FirebaseFirestoreReactive.java
ace-n: Will fix this more broadly.

functions/firebase/src/main/java/com/example/functions/firebase/FirebaseFirestoreReactive.java
ace-n: Will fix this more broadly.

ace-n:
`spotbugs` should be passing now, except for the whitelisted bugs described below. **IMO we should get this whitelist integrated into the repo ASAP**, to avoid modifying each module's `pom.xml` individually.

```
<?xml version="1.0" encoding="UTF-8"?>
 <FindBugsFilter
              xmlns="https://github.com/spotbugs/filter/3.0.0"
              xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
              xsi:schemaLocation="https://github.com/spotbugs/filter/3.0.0 https://raw.githubusercontent.com/spotbugs/spotbugs/3.1.0/spotbugs/etc/findbugsfilter.xsd">

    <Match> <!-- Global excludes -->
        <Or>
            <Bug pattern="IMC_IMMATURE_CLASS_NO_TOSTRING" />
            <Bug pattern="CRLF_INJECTION_LOGS" />
            <Bug pattern="CLI_CONSTANT_LIST_INDEX" />
            <Bug pattern="LI_LAZY_INIT_STATIC" />
            <Bug pattern="PCOA_PARTIALLY_CONSTRUCTED_OBJECT_ACCESS" />
            <Bug pattern="RCN_REDUNDANT_NULLCHECK_WOULD_HAVE_BEEN_A_NPE" />
        </Or>
    </Match>

    <!-- Functions samples -->
    <Match>
        <Or>
            <Class name="com.example.functions.helloworld.GroovyHelloWorld" />
            <Class name="com.example.functions.helloworld.GroovyHelloBackground" />
        </Or>
        <Bug pattern="MS_SHOULD_BE_FINAL" />
    </Match>

    <Match>
        <Or>
            <Class name="com.example.functions.helloworld.HelloError" />
            <Class name="com.example.functions.concepts.RetryPubSub" />
            <Class name="com.example.functions.ocr.OcrProcessImage" />
            <Class name="com.example.functions.ocr.OcrTranslateApiMessage" />
        </Or>
        <Bug pattern="WEM_WEAK_EXCEPTION_MESSAGING" />
    </Match>

    <Match>
        <Class name="com.example.functions.helloworld.KotlinHelloBackground" />
        <Bug pattern="DLS_DEAD_LOCAL_STORE" />
    </Match>

    <Match>
        <Class name="com.example.functions.imagemagick.ImageMagick" />
        <Or>
            <Bug pattern="PATH_TRAVERSAL_IN" />
            <Bug pattern="DMI_HARDCODED_ABSOLUTE_FILENAME" />
            <Bug pattern="COMMAND_INJECTION" />
        </Or>
    </Match>


    <Match>
        <Class name="com.example.functions.concepts.LazyFields" />
        <Bug pattern="UPM_UNCALLED_PRIVATE_METHOD" />
    </Match>

    <Match>
        <Class name="com.example.functions.concepts.FileSystem" />
        <Bug pattern="NP_NULL_ON_SOME_PATH_FROM_RETURN_VALUE" />
    </Match>

    <!-- TODO(ace-n): switch Date objects in eventpojos to java.time ones -->
    <Match>
        <Or>
            <Bug pattern="EI_EXPOSE_REP" />
            <Bug pattern="EI_EXPOSE_REP2" />
        </Or>
    </Match>

</FindBugsFilter>
```

functions/concepts/src/main/java/com/example/functions/concepts/FileSystem.java
```diff
@@ -34,9 +34,9 @@ public void service(HttpRequest request, HttpResponse response)
     File currentDirectory = new File(".");
     File[] files = currentDirectory.listFiles();
     BufferedWriter writer = response.getWriter();
```
saturnism, line 36: Consider using a `PrintWriter`, e.g.,

```java
PrintWriter writer = new PrintWriter(response.getWriter());
writer.println(...)
```

functions/concepts/src/main/java/com/example/functions/concepts/LazyFields.java
```diff
@@ -30,20 +30,28 @@
...
   // Declared at cold-start, but only initialized if/when the function executes
-  private static Integer lazyGlobal = null;
+  // Uses the "initialization-on-demand holder" idiom
+  // More information: https://en.wikipedia.org/wiki/Initialization-on-demand_holder_idiom
+  private static class LazyGlobalHolder {
+    // This value is initialized only if (and when) the getLazyGlobal() function below is called
+    private static final Integer LAZY_GLOBAL = functionSpecificComputation();
+  }
+
+  private static Integer getLazyGlobal() {
```
saturnism, line 43: The common idiom is..

```java
private static class LazyGlobalHolder {
  private LazyGlobalHolder() {} // remove the possibility to instantiate.
  private static final Integer INSTANCE = ...;
  private static Integer getInstance();
}
```

saturnism, line 43: this should go into the `LazyGlobalHolder` class

functions/concepts/src/main/java/com/example/functions/concepts/LazyFields.java
```diff
@@ -30,20 +30,28 @@
...
+  }
 
   @Override
   public void service(HttpRequest request, HttpResponse response)
       throws IOException {
-    // This value is initialized only if (and when) the function is called
-    if (lazyGlobal == null) {
-      lazyGlobal = functionSpecificComputation();
-    }
+    Integer lazyGlobal = getLazyGlobal();
```
saturnism, line 50: this should be `LazyGlobalHolder.getInstance()`

functions/http/pom.xml
```diff
@@ -61,11 +61,15 @@
       <artifactId>gson</artifactId>
       <version>2.8.6</version>
     </dependency>
+    <dependency>
+      <groupId>javax.servlet</groupId>
+      <artifactId>javax.servlet-api</artifactId>
+      <version>4.0.1</version>
```
saturnism, line 67: what is this used for? in generally we should have no dependency on the servlet api

functions/slack/src/main/java/com/example/functions/slack/SlackSlashCommand.java
```diff
@@ -167,14 +162,14 @@ JsonObject searchKnowledgeGraph(String query) throws IOException {
   public void service(HttpRequest request, HttpResponse response) throws IOException {
 
     // Validate request
-    if (request.getMethod() != "POST") {
+    if (!"POST".equals(request.getMethod())) {
```
saturnism, line 165: good catch!

functions/http/pom.xml
lesv: And it shouldn't be 4.* anyway, unless you are using a 4.* container.

saturnism: preferably we don't have this at all.

functions/imagemagick/src/main/java/com/example/functions/imagemagick/ImageMagick.java
```diff
@@ -86,7 +92,7 @@ public void accept(GcsEvent gcsEvent, Context context) {
           LOGGER.info(String.format("Detected %s as OK.", gcsEvent.getName()));
         }
       }
-    } catch (Exception e) {
+    } catch (IOException e) {
       LOGGER.info(String.format("Error with Vision API: %s", e.getMessage()));
     }
```
lesv, line 97: Can you be more specific?
Is there a form of mitigation?
Should you even be catching this?  (rather than providing a warning it might happen and let it propagate)

functions/pom.xml
```diff
@@ -0,0 +1,49 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
lesv, line 1: This isn't needed and actually makes things difficult for us.

functions/slack/pom.xml
```diff
@@ -155,6 +154,20 @@
...
     <plugins>
+      <plugin>
+        <!--
+          Google Cloud Functions Framework Maven plugin
+
+          This plugin allows you to run Cloud Functions Java code
+          locally. Use the following terminal command to run a
+          given function locally:
+
+          mvn function:run -Drun.functionTarget=your.package.yourFunction
```
lesv, line 165: Should also be in README's.

functions/pom.xml
saturnism: i would keep it. it's really standard.

lesv: As @kurtisvg mentioned in the doc, we used to do this everywhere, users got confused and it created problems.  Getting rid of it increased CSAT.

lesv: Note - I agree that many users use this and it is standard, but for many in our target audience this is very difficult for them to use.

functions/http/pom.xml
ace-n: We don't. It's been deleted.

functions/slack/src/main/java/com/example/functions/slack/SlackSlashCommand.java
ace-n: `spotbugs` 🙂

functions/imagemagick/src/main/java/com/example/functions/imagemagick/ImageMagick.java
ace-n: Switched to a `LOGGER.log()` format that allows printing the entire exception.

("catching this and letting it propagate" is probably the _right_ answer, but **will not compile** without a cast to `RuntimeException` due to [this issue](https://github.com/GoogleCloudPlatform/functions-framework-java/issues/21) until the release is updated)

kurtisvg:
@ace-n - Please break PRs like this into smaller pieces - The per commit breakdown you did was great, but it would have been easier to review if each one was it's own PR instead. 

There are a number of changes in this PR that would be ready to merge, but as changes and feedback continue to pile on it's a pain separate those changes out. 

(FYI - @lesv)

functions/pom.xml
ace-n: Removed from this + child POMs.

functions/slack/pom.xml
ace-n: Added to parent-level README.

(There are no per-sample "child" READMEs - happy to discuss if that's something you want.)

functions/concepts/src/main/java/com/example/functions/concepts/LazyFields.java
ace-n: Thanks - done.

functions/concepts/src/main/java/com/example/functions/concepts/LazyFields.java
ace-n: Thanks - done.

functions/concepts/src/main/java/com/example/functions/concepts/LazyFields.java
ace-n: Thanks - done.

functions/concepts/src/main/java/com/example/functions/concepts/FileSystem.java
ace-n: `response.getWriter()` expects a `BufferedWriter`, not a `PrintWriter`.

(I tried using `new BufferedWriter(printWriter)`, but those stubs didn't produce any output in the tests.)

ace-n:
@kurtisvg normally I'd agree, but my concern was that these changes would depend on each other + PR separation risked merge conflicts.

Perhaps there is a better way to go about this though?

ace-n:
Per our meeting just now - this PR **does not** make the recommended `package` changes.

@lesv @kurtisvg LMK if you'd like me to do that here, or in a separate PR.

functions/concepts/src/main/java/com/example/functions/concepts/LazyFields.java
```diff
@@ -36,18 +36,21 @@
   // Uses the "initialization-on-demand holder" idiom
   // More information: https://en.wikipedia.org/wiki/Initialization-on-demand_holder_idiom
   private static class LazyGlobalHolder {
+    // Making the default constructor private prohibits instantiation of this class
```
ace-n, line 39: **N.B:** wrong commit name, should be `[comments] use proper IDOH pattern`

functions/imagemagick/src/main/java/com/example/functions/imagemagick/ImageMagick.java
```diff
@@ -74,7 +81,7 @@ public void accept(GcsEvent gcsEvent, Context context) {
       List<AnnotateImageResponse> responses = response.getResponsesList();
       for (AnnotateImageResponse res : responses) {
         if (res.hasError()) {
-          LOGGER.info(String.format("Error: %s\n", res.getError().getMessage()));
+          LOGGER.info(String.format("Error: %s%n", res.getError().getMessage()));
```
saturnism, line 84: we typically don't need ending newline for log messages.

functions/helloworld/src/main/java/com/example/functions/helloworld/HelloHttp.java
```diff
@@ -32,21 +32,18 @@
...
   public void service(HttpRequest request, HttpResponse response)
       throws IOException {
-    String name = "world";
-
     // Check URL parameters for "name" field
-    if (request.getFirstQueryParameter("name").isPresent()) {
-      name = request.getFirstQueryParameter("name").get();
-    }
+    // "world" is the default value
+    String name = request.getFirstQueryParameter("name").orElse("world");
```
saturnism, line 42: nice use of optional!

functions/http/src/main/java/com/example/functions/http/ParseContentType.java
```diff
@@ -26,44 +26,52 @@
...
+  private static final Gson gson = new Gson();
 
   // Responds to an HTTP request using data from the request body parsed according to the
   // "content-type" header.
   @Override
   public void service(HttpRequest request, HttpResponse response)
       throws IOException {
-    String name = null;
+    String name;
     String contentType = request.getContentType().get();
```
saturnism, line 44: maybe track for a future issue. getContentType returns an optional, so contentType may not be present.

```java
request.getContentType().ifPresent(contentType -> {
});
```

or
```java
String contentType = request.getContentType().orElseThrow(() -> new RuntimeException("content type not specified"));
```

saturnism:
overall changes lgtm. a few nits as long as they are tracked in a future pr.

functions/http/src/main/java/com/example/functions/http/ParseContentType.java
ace-n: Good catch - will refactor.

(Aside: it looks like `Content-Type` values are [_recommended_ but not _required_](https://stackoverflow.com/questions/20392345/is-content-type-http-header-always-required), at least in **reponses**.)

functions/imagemagick/src/main/java/com/example/functions/imagemagick/ImageMagick.java
lesv: +1

ace-n: Done.

functions/pom.xml
```diff
@@ -0,0 +1,47 @@
...
+    <module>helloworld</module>
+    <module>http</module>
+    <module>imagemagick</module>
+    <module>slack</module>
+    <module>concepts</module>
+    <module>logging</module>
+    <module>firebase</module>
+    <module>ocr</module>
+  </modules>
+</project>
```
lesv, line 47: Why do we have this?  How does it help users?

functions/http/src/main/java/com/example/functions/http/ParseContentType.java
ace-n: FYI: specified a _default value_ instead, and included a comment as to why. (`null` breaks the `switch...case`, and `get` on a missing `Optional` [throws a `NoSuchElementException`](https://docs.oracle.com/javase/8/docs/api/java/util/Optional.html#get--))

```java
// Default values avoid null issues (with switch/case) and exceptions from get() (optionals)
String contentType = request.getContentType().orElse("");
```

LMK if this is too much "black magic" to include in a sample.

functions/pom.xml
saturnism: this helps me to build and test all functions in one shot. it also allows me to open up functions in the IDE in one shot if i need to.  like i mentioned separately, each sub directory/module is still completely independent.

the parent in the children pom references a published pom, i feel it's good as long as a comment on it's not needed for user.

ace-n: @saturnism and I discussed this on Friday - this parent POM lets us run `mvn` builds across _all_ the samples to verify that _all of them_ work as intended.

(Otherwise, we'd have to run each one individually.)

saturnism: also discussed in the doc and the call, this causes no harm for the samples.

kurtisvg: I'll defer to @lesv for the final decision, but I'm still -1 on this (mostly because I don't like snowflakes). 

My concerns are largely mitigated since Ace has already adapted the script + left helpful comments in it and Ray made a good point about it being a one way configuration. However, I don't think we really expect users to run _all_ the tests in a single shot, which means this is only really for us. Our CI doesn't use this, which means there is nothing really stopping someone new adding a sample and not putting it here. So I guess I'd still prefer not to have it, but I'll leave it up to lesv@.

functions/concepts/pom.xml
```diff
@@ -1,5 +1,3 @@
-<?xml version="1.0" encoding="UTF-8"?>
```
kurtisvg, line 1: I don't think you want to remove this. 

@lesv had a comment on a line like this, but I think it was RE: Parent pom and not RE:xml header.

functions/pom.xml
lesv: Why not do what we do and just put a script that looks for `pom.xml` files and builds them.  For example, see [this one](https://github.com/GoogleCloudPlatform/java-docs-samples/blob/master/.kokoro/tests/run_tests.sh).

functions/concepts/src/main/java/com/example/functions/concepts/LazyFields.java
kurtisvg: @ace-n  I supper appreciate you taking the time to comment these well. These are great.

functions/pom.xml
kurtisvg: I don't really prefer we commit a new script either - parent pom for _personal_ convince makes sense, I just would prefer we didn't commit it in the repo.

functions/pom.xml
saturnism: it's not all that difficult. this is _standard_, and mostly autogenerated. xml presents issues w/o this.

functions/concepts/pom.xml
saturnism: See https://github.com/GoogleCloudPlatform/java-docs-samples/pull/2630#discussion_r407596003

i'd say - let's please not remove standard stuff! it's unnecessarily extra work. every pom.xml pretty much has this.

functions/pom.xml
saturnism: i think... this comment was in the context of the `<?xml version="1.0" encoding="UTF-8"?>` line, which was then removed from the poms.

https://github.com/GoogleCloudPlatform/java-docs-samples/blob/9be4e4d0de95f74a5b237d6de498cb579fc3f762/functions/imagemagick/pom.xml

to be clear, my comment was to make sure we _don't remove_ the xml header, since it's very standard.

functions/pom.xml
saturnism: pls send data on the csat comment. the top level pom.xml doesn't need to be included in the doc.  i'm curious to see where the confusions were and whether it was because of other issues such as children pom inheriting this, which causes build failures. that in itself will be very frustrating.

lesv: It was over 18m ago.

functions/concepts/pom.xml
lesv: We want the `pom.xml` file removed, not the line.

lesv:
Note - the robot has done it's thing and you now have conflicts.

functions/concepts/pom.xml
ace-n: Got it - reverted this change for all POMs.

functions/pom.xml
ace-n: BTW - it's possible to do this in `bash`, but it's not exactly the most straightforward thing (mainly because of array-vs-string gotchas):

```
find */pom.xml | xargs -I {} echo $(pwd)/{} | xargs -I {} dirname {} | xargs -I {} sh -c "cd {} && mvn clean verify"
```

Kurtis / Les - would you be open to keeping this parent POM if we added a comment to it like this?

```
This pom.xml file is kept for testing/Maven compatibility purposes only.

Please DO NOT add dependencies and/or common settings here. The child folders inherit from our `shared-configuration` stored on GitHub instead of this file; cross-POM changes should be added there.
```

eamonnmcmanus: FWIW I think this is simpler:

```
for pom in $(find * -name pom.xml); do (cd $(dirname $pom) && mvn clean verify); done
```

lesv: @ace-n The problem is we don't expect the engineers who follow / maintain to continue to maintain this.  We used to do it.  We stopped because it got overwhelming as the # of samples increased.

lesv:
And the 1 conflict.

functions/pom.xml
ace-n: Les, thanks for clarifying.

Since the modules are explicitly specified, I'm willing to drop this in favor of a programmatic (e.g. `bash`) solution.

lesv:
@ace-n Note the conflict -- needs to be fixed.

ace-n:
Fixed by force-push. (GitHub doesn't let you merge if there's a conflict.)

**Aside:** perhaps it's worth configuring `renovate` to only run periodically (e.g. on weekends). That might reduce the pain of these regular updates.

PR #2642, opened 2020-04-14
healthcare API: update datasets/ samples from v1beta1 to v1

noerog:
NOTE: This PR is fundamentally a branch of https://github.com/GoogleCloudPlatform/java-docs-samples/tree/master/healthcare/v1beta1 to v1. The code has been previously reviewed in this repository.
Updates for fhir/, dicom/, and hl7v2/ are coming but I wanted to keep the PRs separate.

After all healthcare/v1 samples are approved, I'll delete the existing v1beta1/ directory.

### DIFF
--- a/healthcare/v1/pom.xml
+++ b/healthcare/v1/pom.xml
@@ -1,5 +1,6 @@
+<?xml version="1.0" encoding="UTF-8"?>
 <!--
-  Copyright 2020 Google LLC
+  Copyright 2019-2020 Google LLC
 
   Licensed under the Apache License, Version 2.0 (the "License");
   you may not use this file except in compliance with the License.
@@ -13,7 +14,6 @@
   See the License for the specific language governing permissions and
   limitations under the License.
 -->
-<?xml version="1.0" encoding="UTF-8"?>
 <project xmlns="http://maven.apache.org/POM/4.0.0"
   xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
   xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
--- a/healthcare/v1/pom.xml
+++ b/healthcare/v1/pom.xml
@@ -1,0 +1,97 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+  xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+  xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+  <groupId>com.google.samples</groupId>
+  <artifactId>healthcare-samples</artifactId>
+  <version>1.0-SNAPSHOT</version>
+  <build>
+    <plugins>
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-compiler-plugin</artifactId>
+        <configuration>
+          <source>11</source>
+          <target>11</target>
+        </configuration>
+      </plugin>
+    </plugins>
+  </build>
+  <packaging>jar</packaging>
+
+  <!--
+    The parent pom defines common style checks and testing strategies for our samples.
+    Removing or replacing it should not affect the execution of the samples in anyway.
+  -->
+  <parent>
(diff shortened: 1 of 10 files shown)


healthcare/v1/pom.xml
```diff
@@ -0,0 +1,97 @@
...
+    </dependency>
+    <dependency>
+      <groupId>junit</groupId>
+      <artifactId>junit</artifactId>
+      <version>4.13</version>
+    </dependency>
+    <dependency>
+      <groupId>org.apache.httpcomponents</groupId>
+      <artifactId>httpmime</artifactId>
+      <version>4.5.12</version>
```
lesv, line 92:
```suggestion
      <version>4.5.12</version>
      <scope>test</scope>
```

healthcare/v1/pom.xml
```diff
@@ -0,0 +1,97 @@
...
+    <dependency>
+      <groupId>com.google.truth</groupId>
+      <artifactId>truth</artifactId>
+      <version>1.0.1</version>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>junit</groupId>
+      <artifactId>junit</artifactId>
+      <version>4.13</version>
```
lesv, line 87:
```suggestion
      <version>4.13</version>
      <scope>test</scope>
```

healthcare/v1/src/main/java/snippets/healthcare/datasets/DatasetCreate.java
```diff
@@ -0,0 +1,101 @@
...
+                .datasets()
+                .operations()
+                .get(operation.getName())
+                .execute();
+      }
+      System.out.println("Dataset created. Response content: " + operation.getResponse());
+    } catch (Exception ex) {
+      System.out.printf("Error during request execution: %s\n", ex.toString());
+      ex.printStackTrace(System.out);
+    }
```
lesv, line 77: Can you be more specific?
Is there any mitigation?
Do you need this at all, why not just let it propagate and mention that it might happen?

healthcare/v1/src/main/java/snippets/healthcare/datasets/DatasetDeIdentify.java
```diff
@@ -0,0 +1,110 @@
...
+                .operations()
+                .get(operation.getName())
+                .execute();
+      }
+      System.out.println(
+          "De-identified Dataset created. Response content: " + operation.getResponse());
+    } catch (Exception ex) {
+      System.out.printf("Error during request execution: %s", ex.toString());
+      ex.printStackTrace(System.out);
+    }
```
lesv, line 86: Can you be more specific?
Is there any mitigation?
Do you need this at all, why not just let it propagate and mention that it might happen?

noerog: Same as above.

healthcare/v1/src/main/java/snippets/healthcare/datasets/DatasetCreate.java
noerog: This is sort of a generic CRUD request to the service so the exception could be a number of things. I don't think it would be possible to be more specific. The mitigation should be clear from the error message returned (e.g. if auth is the issue). I see some samples catch errors/exceptions and others don't. Do you have a preference either way?

healthcare/v1/pom.xml
noerog: Done.

healthcare/v1/pom.xml
```diff
@@ -0,0 +1,97 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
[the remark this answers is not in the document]
noerog: Done.

healthcare/v1/src/main/java/snippets/healthcare/datasets/DatasetCreate.java
lesv: Java prefers that we don't catch and ignore errors.  If we are going to ignore it, then let it be caught by a parent.  That said, if it might happen, it's good to mention it.
https://google.github.io/styleguide/javaguide.html#s6.2-caught-exceptions

healthcare/v1/src/main/java/snippets/healthcare/datasets/DatasetDeIdentify.java
lesv: https://google.github.io/styleguide/javaguide.html#s6.2-caught-exceptions

PR #2673, opened 2020-04-16
Remove old GAE to Cloud SQL samples and point to new ones instead.

kurtisvg:
Remove older appengine-java8 samples and point to the new one instead.


kurtisvg:
> Have you checked the docs for references to this sample?

Yes - I wasn't able to find any references to "appengine-java8/cloudsql" or "appengine-java8/cloudsql-postgres".

averikitsch:
Great thanks!

PR #2714, opened 2020-04-21
fix flaky tests of jobs v4

### LINKED ISSUE #2681: CommuteSearchJobsTest: testCommuteSearchJobs failed
This test failed!

To configure my behavior, see [the Build Cop Bot documentation](https://github.com/googleapis/repo-automation-bots/tree/master/packages/buildcop).

If I'm commenting on this issue too often, add the `buildcop: quiet` label and
I will stop commenting.

---

commit: e6fcdaaa71a01071f8f17be0144798ce9659919c
buildURL: [Build Status](https://source.cloud.google.com/results/invocations/39c13075-d9a8-4778-9c38-9a93d1702dcf), [Sponge](http://sponge2/39c13075-d9a8-4778-9c38-9a93d1702dcf)
status: failed

munkhuushmgl:
Fixes #2681 #2684 #2682 #2683

### DIFF
--- a/jobs/v4/src/test/java/JobSearchGetJobTest.java
+++ b/jobs/v4/src/test/java/JobSearchGetJobTest.java
@@ -28,7 +28,7 @@ import org.junit.Test;
 public class JobSearchGetJobTest {
   private static final String PROJECT_ID = System.getenv("GOOGLE_CLOUD_PROJECT");
   private static final String TENANT_ID = "50c14f00-dc38-4812-989b-d9b59c7fdf07";
-  private static final String JOB_ID = "122880074199245510";
+  private static final String JOB_ID = "76779818718241478";
 
   private ByteArrayOutputStream bout;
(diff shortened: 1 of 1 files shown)


lesv:
Why does changing the JobID fix things?

munkhuushmgl:
> Why does changing the JobID fix things?

Someone deleted the Job I created for test even though I added DO_NOT_DELETE label on name and id.

kurtisvg:
Can we either create the job as part of the setup, or else hand it in through an env var? That would be compliant with the sample guidelines.

munkhuushmgl:
> Can we either create the job as part of the setup, or else hand it in through an env var? That would be compliant with the sample guidelines.

@kurtisvg I think having job ID as env var is great idea, and I will make that changes this week. Initially there was setup and teardown for creating and deleting job, but reviewer suggested to have predefined vars to reduce complexity.

Issue #2715, opened 2020-04-21
tests in asset/cloud-client are inherently flaky


lesv:
Reviewing QuickstartIT, it:
* creates a bucket
* Writes a list of assets to the bucket
* wants to see that bucket in the list of assets
* deletes the bucket.

It should instead:
* Write to a well known bucket w/ gs:/well-known-bucket/UUID/my-assets.txt
* See if well-known-bucket is in my-assets.txt

This way there isn't a hysteresis problem on creating and knowing about buckets.

Related to #2551

lesv:
FYI @peter-zheng-g @dzlier-gcp @kurtisvg

elharo:
I don't follow why this test is flaky. The current flow seems like it should work?

lesv:
It was creating a bucket, then immediatly dealing w/ it.  What was happening was that occasionally, the bucket wasn't yet available.  That's fixed now.

PR #2716, opened 2020-04-21
Update QuickStartIT.java

lesv:
Fixes: #2715
Reviewing QuickstartIT, it:

* creates a bucket
* Writes a list of assets to the bucket
* wants to see that bucket in the list of assets
* deletes the bucket.

It should instead:

* Write to a well known bucket w/ gs:/well-known-bucket/UUID/my-assets.txt
* See if well-known-bucket is in my-assets.txt
This way there isn't a hysteresis problem on creating and knowing about buckets.

- [x] Tests pass (`mvn -P lint clean verify`)

### DIFF
--- a/asset/cloud-client/src/test/java/com/example/asset/QuickStartIT.java
+++ b/asset/cloud-client/src/test/java/com/example/asset/QuickStartIT.java
@@ -19,8 +19,6 @@ package com.example.asset;
 import static com.google.common.truth.Truth.assertThat;
 
 import com.google.cloud.storage.BlobInfo;
-import com.google.cloud.storage.Bucket;
-import com.google.cloud.storage.BucketInfo;
 import com.google.cloud.storage.Storage;
 import com.google.cloud.storage.Storage.BlobListOption;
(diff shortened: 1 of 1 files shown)


lesv:
Looks like someone tried to fix things a different way.  Need to figure out what's going on.

lesv:
I've resolved the merge conflict.

asset/cloud-client/src/test/java/com/example/asset/QuickStartIT.java
```diff
@@ -45,33 +43,29 @@
 @RunWith(JUnit4.class)
 @SuppressWarnings("checkstyle:abbreviationaswordinname")
 public class QuickStartIT {
-  private static final String bucketName = UUID.randomUUID().toString();
+  private static final String bucketName = "java-docs-samples-testing";
```
kurtisvg, line 46: Can we use an environment var for this?

elharo: Can anyone who is not a googler still run this integration test?

PR #2717, opened 2020-04-21
Update Secret Manager samples to be canonical

sethvargo:
This updates Secret Manager to use the more canonical helpers and also updates the layout to use the new preferred layout.


lesv:
@sethvargo You can now merge, or ask us to.

sethvargo:
Thanks!

PR #2719, opened 2020-04-21
automl: explictly update timeoutes due to default library changes

nnegrey:
Fixes PR: https://github.com/GoogleCloudPlatform/java-docs-samples/pull/2704


automl/cloud-client/src/main/java/com/example/automl/ImportDataset.java
```diff
@@ -55,8 +59,22 @@ static void importDataset(String projectId, String datasetId, String path)
...
+      // If you want to wait for the operation to finish, adjust the timeout appropriately. The
+      // operation will still run if you choose not to wait for it to complete. You can check the
+      // status of your operation using the operation's name.
+      Empty response = operation.get(30, TimeUnit.MINUTES);
+      System.out.format("Dataset imported. %s%n", response);
+    } catch (TimeoutException e) {
+      System.out.println("The operation's polling period was not long enough.");
+      System.out.println("You can use the Operation's name to get the current status.");
+      System.out.println("The import job will still complete as expected.");
+      throw e;
```
nnegrey, line 77: @lesv, I wanted to make a note for the timeout that even though there is a timeout exception, the operation is still running in the background. But I still end up throwing the error. 

What's the best practice here?

lesv: I'm good w/ this.  I'd prefer to see comments, but this is ok.

That said, you might be very clear by changing to:
`The import job is still running and will complete as expected.`

nnegrey: Yea, I'll switch to that.

PR #2691, opened 2020-04-19
GCF: add http system/integration test sample

ace-n:
**N.B:** this uses the Node.js testing project for part of the tests

PR #2726, opened 2020-04-22
Add pubsub tests

ace-n:


functions/helloworld/hello-pub-sub/src/test/java/functions/ExampleIntegrationTest.java
```diff
@@ -0,0 +1,88 @@
...
+    postRequest.setEntity(new StringEntity(jsonStr));
+
+    // The Functions Framework Maven plugin process takes time to start up
+    // Use resilience4j to retry the test HTTP request until the plugin responds
+    RetryRegistry registry = RetryRegistry.of(RetryConfig.custom()
+        .maxAttempts(8)
+        .retryExceptions(HttpHostConnectException.class)
+        .intervalFunction(IntervalFunction.ofExponentialBackoff(200, 2))
+        .build());
+    Retry retry = registry.retry("my");
```
kurtisvg, line 71: I don't think the appropriate strategy is to retry for this - users may miss flaky behavior in tests because of that. We should ensure that the Process has started successfully (probably by waiting and/or verifying with stdout), and run the test once.

functions/helloworld/hello-pub-sub/src/test/java/functions/ExampleSystemTest.java
```diff
@@ -0,0 +1,99 @@
...
+        ProjectTopicName.of(PROJECT_ID, TOPIC_NAME)).build();
+  }
+
+  private static String getLogEntriesAsString(String startTimestamp) {
+    // Construct Stackdriver logging filter
+    // See this page for more info: https://cloud.google.com/logging/docs/view/advanced-queries
+    String filter = "resource.type=\"cloud_function\"";
+    filter += " AND severity=INFO";
+    filter += " AND resource.labels.function_name=" + FUNCTION_DEPLOYED_NAME;
+    filter += String.format(" AND timestamp>=\"%s\"", startTimestamp);
```
kurtisvg, line 54:
```suggestion
    String filter = "resource.type=\"cloud_function\""
        + " AND severity=INFO"
        + " AND resource.labels.function_name=" + FUNCTION_DEPLOYED_NAME
        +  String.format(" AND timestamp>=\"%s\"", startTimestamp);
```

functions/helloworld/hello-pub-sub/src/test/java/functions/ExampleSystemTest.java
```diff
@@ -0,0 +1,99 @@
...
+import java.time.format.DateTimeFormatter;
+import java.util.UUID;
+import java.util.stream.Collectors;
+import java.util.stream.StreamSupport;
+import org.junit.BeforeClass;
+import org.junit.Test;
+
+public class ExampleSystemTest {
+
+  // TODO<developer>: set these values (as environment variables, or hard-coded into your test)
```
kurtisvg, line 30: Let's not tell users to hard code configuration values into their code.

If you want to point them to something helpful, consider linking to SecretManager.

functions/helloworld/hello-pub-sub/src/test/java/functions/ExampleIntegrationTest.java
ace-n: The emulator takes time to start up. @grant and I decided to go with retry/backoff rather than constant-timed `sleep()`s

functions/helloworld/hello-pub-sub/src/test/java/functions/ExampleIntegrationTest.java
```diff
@@ -0,0 +1,88 @@
...
+  // The Functions Framework Maven plugin lets us run a function locally
+  private static final String BASE_URL = "http://localhost:8080";
+
+  private static Process emulatorProcess = null;
+  private static HttpClient client = HttpClientBuilder.create().build();
+
+  @BeforeClass
+  public static void setUp() throws IOException {
+    // Emulate the function locally by running the Functions Framework Maven plugin
+    emulatorProcess = (new ProcessBuilder()).command("mvn", "function:run").start();
```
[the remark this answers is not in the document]
ace-n: We could likely move the retry logic here instead, and wait until the process has "warmed up".

That said, the HTTP PR (#2691) was already merged using this approach. IMO we should merge as-is (to avoid blocking TW) and push a fast-follow fix.

functions/helloworld/hello-pub-sub/src/test/java/functions/ExampleSystemTest.java
ace-n: Will drop the mention of hard-coding here. Based on our internal practices, environment variables should be an acceptable answer for this.

functions/helloworld/hello-pub-sub/src/test/java/functions/ExampleSystemTest.java
ace-n: Done.

functions/helloworld/hello-pub-sub/src/test/java/functions/ExampleIntegrationTest.java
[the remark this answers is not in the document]
kurtisvg: That PR was _just_ merged 🙃. I'll let this slide provided you open an issue on java-docs-samples and assign it to yourself. FWIW pushing a bad sample to avoid blocking a TW is not a good reason. 

I don't think retry logic is needed at all. You can start up the process and wait for it to print "ready to serve" or whatever signal it gives that it's ready to start up.

We should save retries for legitimately flakey services. If starting the emulator startup is flakey, that's a different problem.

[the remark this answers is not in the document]
ace-n: See #2763 - would you mind chiming in and making a suggestion on which solution to use there?

PR #2733, opened 2020-04-22
Moved XMLDecl to first line to prevent parsing err

woerzberger:
Jetty fails to start with
 mvn jetty:run
with a 
 org.xml.sax.SAXParseException: The processing instruction target matching "[xX][mM][lL]" is not allowed.

This is because jetty-web.xml starts with a comment instead of a XMLDecl which is syntactically wrong according to
https://www.w3.org/TR/REC-xml/#sec-prolog-dtd and
https://www.w3.org/TR/2008/REC-xml-20081126/#sec-guessing-no-ext-info

Fixes #issue


lesv:
Unsigned CLA and no updates for 2 weeks; I'm going to close.

PR #2734, opened 2020-04-22
sample: add Spanner sample for Cloud Functions

olavloite:
Sample for querying a Cloud Spanner database through a Java Cloud Function.

### DIFF
--- a/functions/spanner/src/main/java/functions/HelloSpanner.java
+++ b/functions/spanner/src/main/java/functions/HelloSpanner.java
@@ -1,0 +1,123 @@
+/*
+ * Copyright 2020 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package functions;
+
+// [START spanner_functions_quickstart]
+import com.google.api.client.http.HttpStatusCodes;
+import com.google.cloud.functions.HttpFunction;
+import com.google.cloud.functions.HttpRequest;
+import com.google.cloud.functions.HttpResponse;
+import com.google.cloud.spanner.DatabaseClient;
+import com.google.cloud.spanner.DatabaseId;
+import com.google.cloud.spanner.ResultSet;
+import com.google.cloud.spanner.Spanner;
+import com.google.cloud.spanner.SpannerException;
+import com.google.cloud.spanner.SpannerOptions;
+import com.google.cloud.spanner.Statement;
+import com.google.common.annotations.VisibleForTesting;
+import com.google.common.base.MoreObjects;
+import java.io.BufferedWriter;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+
+// HelloSpanner is an example of querying Spanner from a Cloud Function.
+public class HelloSpanner implements HttpFunction {
+  private static final Logger LOGGER = Logger.getLogger(HelloSpanner.class.getName());
+
+  @VisibleForTesting
+  static Spanner createSpanner() {
+    return SpannerOptions.newBuilder().build().getService();
+  }
+
+  // SpannerHolder is a holder class for a Spanner instance that is initialized lazily.
+  private static final class SpannerHolder {
+    private final Object lock = new Object();
+    private volatile boolean initialized;
+    private volatile DatabaseClient client;
+    private volatile Throwable error;
+
+    private SpannerHolder() {}
+
+    // Initialize the {@link Spanner} instance in a method and not as a static variable, as it
+    // might throw an error, and we want to catch and log that specific error.
+    DatabaseClient get() throws Throwable {
+      if (!initialized) {
+        synchronized (lock) {
+          if (!initialized) {
+            try {
+              DatabaseId db =
+                  DatabaseId.of(
+                      SpannerOptions.getDefaultProjectId(),
+                      SPANNER_INSTANCE_ID,
+                      SPANNER_DATABASE_ID);
+              client = createSpanner().getDatabaseClient(db);
+            } catch (Throwable t) {
+              error = t;
+            }
+            initialized = true;
+          }
+        }
+      }
+      if (error != null) {
+        throw error;
+      }
+      return client;
+    }
+  }
+
+  // The SpannerHolder instance is shared across all instances of the HelloSpanner class.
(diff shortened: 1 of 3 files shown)


functions/spanner/src/main/java/functions/HelloSpanner.java
```diff
@@ -0,0 +1,123 @@
...
+    private final Object lock = new Object();
+    private volatile boolean initialized;
+    private volatile DatabaseClient client;
+    private volatile Throwable error;
+
+    private SpannerHolder() {}
+
+    // Initialize the {@link Spanner} instance in a method and not as a static variable, as it
+    // might throw an error, and we want to catch and log that specific error.
+    DatabaseClient get() throws Throwable {
```
ace-n, line 57: I take it there's no way to simplify this (e.g. with `vavr.io`)?

functions/spanner/src/main/java/functions/HelloSpanner.java
```diff
@@ -0,0 +1,123 @@
...
+      MoreObjects.firstNonNull(System.getenv("SPANNER_INSTANCE"), "my-instance");
+  private static final String SPANNER_DATABASE_ID =
+      MoreObjects.firstNonNull(System.getenv("SPANNER_DATABASE"), "example-db");
+
+  @Override
+  public void service(HttpRequest request, HttpResponse response) throws Exception {
+    BufferedWriter w = response.getWriter();
+    try {
+      DatabaseClient client = getClient();
+      try (ResultSet rs =
```
ace-n, line 101: Nit: would `results` be clearer?

functions/spanner/src/main/java/functions/HelloSpanner.java
```diff
@@ -0,0 +1,123 @@
...
+          client
+              .singleUse()
+              .executeQuery(Statement.of("SELECT SingerId, AlbumId, AlbumTitle FROM Albums"))) {
+        w.write("Albums:\n");
```
ace-n, line 105: Ditto - `writer`?

functions/spanner/src/test/java/functions/HelloSpannerTest.java
```diff
@@ -0,0 +1,156 @@
...
+
+  @Before
+  public void beforeTest() throws IOException {
+    Mockito.mockitoSession().initMocks(this);
+
+    request = PowerMockito.mock(HttpRequest.class);
+    response = PowerMockito.mock(HttpResponse.class);
+    client = PowerMockito.mock(DatabaseClient.class);
+
+    BufferedReader reader = new BufferedReader(new StringReader("{}"));
```
ace-n, line 80: Nit: do you need these two lines?

functions/spanner/src/test/java/functions/HelloSpannerTest.java
```diff
@@ -0,0 +1,156 @@
...
+        .thenThrow(
+            SpannerExceptionFactory.newSpannerException(
+                ErrorCode.NOT_FOUND, "Table `Albums` not found"));
+    PowerMockito.when(client.singleUse()).thenReturn(readContext);
+  }
+
+  @Test
+  public void functionsHelloSpanner_shouldShowQueryError() throws Exception {
+    setupFailedMockQuery();
+    new HelloSpanner() {
```
ace-n, line 131: Should this be moved to a helper function, or perhaps its own class?

ace-n:
Left some comments - feel free to ignore them if @lesv thinks they're not relevant. 🙂

functions/spanner/src/test/java/functions/HelloSpannerTest.java
lesv: I'm ok w/ this in a test.  In the sample, I'd want to see this using GSON or JsonObject rather than as a constant.

functions/spanner/src/test/java/functions/HelloSpannerTest.java
lesv: No - this is a very common idiom for Java 7.  In an ideal world, we be using Java 8 and lambda's.

functions/spanner/src/main/java/functions/HelloSpanner.java
lesv: We tend to not bring in a lot of stuff for our samples, that said, when we do, it's usually from Google written & supported libraries like Guava.

In general, we prefer the most specific error when throwing things in samples rather than the least, but this appears appropriate in this instance.

That said, the comment might be extended to discuss what to do if an error is thrown as this object is effectively unusable should that occur.

lesv:
@olavloite Consider the comments, otherwise ok to merge.

functions/spanner/src/test/java/functions/HelloSpannerTest.java
ace-n: Functions is Java 8 and 11 only - so do we need to worry about Java 7?

functions/spanner/src/test/java/functions/HelloSpannerTest.java
ace-n: Sorry for not being clear - I think these two lines don't do anything useful.

(If you're copying this from one of my samples - I ended up removing them in quite a few of mine.)

functions/spanner/src/main/java/functions/HelloSpanner.java
olavloite: I did look for a standard implementation for this, and the best fit that I could find was `https://commons.apache.org/proper/commons-lang/apidocs/org/apache/commons/lang3/concurrent/LazyInitializer.html`, but I didn't want to introduce an external dependency. That being said, I was surprised that I wasn't able to find anything similar in the Google client libraries, as this is something that I would expect to be a common pattern to use for Cloud Functions. (It could also be that I have overlooked an existing class somewhere, so please feel free to correct me if there is one in the Google libraries.)

olavloite: I've extended the comment to indicate what action should be taken if the initialization fails.

functions/spanner/src/test/java/functions/HelloSpannerTest.java
olavloite: Yep, that was shameless copy-pasting, and it's not needed. Removed.

functions/spanner/src/main/java/functions/HelloSpanner.java
olavloite: `rs` is a very common abbreviation for `ResultSet`. See for example https://docs.oracle.com/javase/tutorial/jdbc/basics/processingsqlstatements.html.

functions/spanner/src/main/java/functions/HelloSpanner.java
olavloite: Agree, changed to `writer`.

functions/spanner/src/test/java/functions/HelloSpannerTest.java
olavloite: I don't think re-writing this into using a lambda would make it more readable. That would mean adding a `Supplier<DatabaseClient>` field to `HelloSpanner` that can be set from the constructor. So I feel it would clutter the `HelloSpanner` class with even more code only needed for testing than the current solution.

PR #2738, opened 2020-04-22
healthcare API: fix typo in message name

noerog:


healthcare/v1/src/main/java/snippets/healthcare/hl7v2/messages/HL7v2MessageGet.java
```diff
@@ -32,7 +32,7 @@
 
 public class HL7v2MessageGet {
   private static final String MESSAGE_NAME =
-      "projects/%s/locations/%s/datasets/%s/hl7C2Store/%s/messages/%s";
+      "projects/%s/locations/%s/datasets/%s/hl7V2Store/%s/messages/%s";
```
lesv, line 35: @chingor13 Does `protoc` expose these things?  Should they?   Having to type weird things correctly to make an API work seems like a bad experience.  Ideally, they wouldn't ever have to create these strings, but since they do, shouldn't there be constants they can just build up?

@kurtisvg FYI

PR #2765, opened 2020-04-23
Simplify the `http-form-data` sample and use Java 11 features.

eamonnmcmanus:
* Remove unused dependencies from `pom.xml`.
* Remove `@MultipartConfig`, an attempted workaround for a bug in multipart handling that unfortunately doesn't help.
* Use conventional spelling for `logger` field. It is not a constant so it should not be spelled `LOGGER`.
* Use `Map.forEach` to iterate over a map.
* Use `List.of(...)` and `Map.of(...)` instead of constructing lists and maps with statements.
* Remove unnecessary `throws IOException`.

Multipart handling is currently broken on GCF. A fix has been prepared but will not be active until it is rolled out. This sample won't work on GCF until that happens.

### DIFF
--- a/functions/http/http-form-data/pom.xml
+++ b/functions/http/http-form-data/pom.xml
@@ -57,20 +57,10 @@
 
   <dependencies>
     <dependency>
-      <groupId>com.google.code.gson</groupId>
-      <artifactId>gson</artifactId>
-      <version>2.8.6</version>
-    </dependency>
-    <dependency>
-      <groupId>javax.servlet</groupId>
-      <artifactId>javax.servlet-api</artifactId>
-      <version>4.0.1</version>
-      <scope>provided</scope>
-    </dependency>
-    <dependency>
       <groupId>com.google.cloud.functions</groupId>
       <artifactId>functions-framework-api</artifactId>
       <version>1.0.1</version>
+      <scope>provided</scope>
     </dependency>
 
     <!-- The following dependencies are only required for testing -->
@@ -111,12 +101,6 @@
       <version>4.13</version>
       <scope>test</scope>
     </dependency>
-    <dependency>
-      <groupId>com.google.guava</groupId>
-      <artifactId>guava-testlib</artifactId>
-      <version>29.0-jre</version>
-      <scope>compile</scope>
-    </dependency>
   </dependencies>
 
   <build>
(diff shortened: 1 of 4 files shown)


functions/http/http-form-data/pom.xml
```diff
@@ -56,21 +56,11 @@
...
-      <groupId>javax.servlet</groupId>
-      <artifactId>javax.servlet-api</artifactId>
-      <version>4.0.1</version>
-      <scope>provided</scope>
-    </dependency>
     <dependency>
       <groupId>com.google.cloud.functions</groupId>
       <artifactId>functions-framework-api</artifactId>
       <version>1.0.1</version>
+      <scope>provided</scope>
```
ace-n, line 63: If this change doesn't break CI, can you file an issue about it and assign it to me? We should update the other samples to be consistent.

functions/http/http-form-data/pom.xml
```diff
@@ -148,31 +132,6 @@
           <trimStackTrace>false</trimStackTrace>
         </configuration>
       </plugin>
-      <plugin> <!-- Required for Java 8 (Alpha) functions in the inline editor -->
```
ace-n, line 151: Ditto - file an issue if it doesn't break tests?

(Can be all in the same issue, or in different ones - as long as I know to come back and fix it. 🙂)

functions/http/http-form-data/src/main/java/functions/HttpFormData.java
```diff
@@ -28,11 +28,9 @@
 import java.nio.file.Paths;
 import java.nio.file.StandardCopyOption;
 import java.util.logging.Logger;
-import javax.servlet.annotation.MultipartConfig;
 
-@MultipartConfig
 public class HttpFormData implements HttpFunction {
-  private static final Logger LOGGER = Logger.getLogger(HttpFormData.class.getName());
+  private static final Logger logger = Logger.getLogger(HttpFormData.class.getName());
```
ace-n, line 33: Ditto - issue, provided @saturnism approves?

functions/http/http-form-data/src/test/java/functions/HttpFormDataTest.java
```diff
@@ -113,10 +108,9 @@ public void functionsHttpFormData_shouldSaveFiles() throws IOException {
   @Test
   public void functionsHttpFormData_shouldProcessFields() throws IOException {
     when(request.getMethod()).thenReturn("POST");
-    when(request.getParts()).thenReturn(new HashMap<>());
+    when(request.getParts()).thenReturn(Map.of());
```
ace-n, line 111: @lesv @kurtisvg @saturnism is there a `checkstyle` config that checks for "outdated" syntax like this?

functions/http/http-form-data/src/main/java/functions/HttpFormData.java
ace-n: Edit: saw @lesv's comment, marking as resolved.

functions/http/http-form-data/src/test/java/functions/HttpFormDataTest.java
kurtisvg: Not as far as I know.

PR #2782, opened 2020-04-24
deps: use libraries bom for dependencies

olavloite:
As noted in #2748, the Spanner samples should use the `libraries-bom:5.2.0` to import dependencies.

### DIFF
--- a/spanner/cloud-client/pom.xml
+++ b/spanner/cloud-client/pom.xml
@@ -42,7 +42,7 @@ limitations under the License.
       <dependency>
         <groupId>com.google.cloud</groupId>
         <artifactId>libraries-bom</artifactId>
-        <version>5.1.0</version>
+        <version>5.2.0</version>
         <type>pom</type>
         <scope>import</scope>
       </dependency>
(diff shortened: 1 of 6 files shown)


olavloite:
The build failure is caused by https://github.com/googleapis/java-spanner/issues/179

lesv:
So we need to wait for the next spanner update?

olavloite:
> So we need to wait for the next spanner update?

Not necessarily. This is a flaky test failure that is not really related to the samples.

This failure occurs in the case that:
1. A non-idempotent admin RPC, such as `CreateDatabase` or `CreateBackup`, fails with an error that would normally be retryable, such as `UNAVAILABLE`.
1. The Spanner client library will in the above case query the backend for any corresponding long-running operation to check whether the RPC was successfully executed or not. If the operation is found, the Spanner client will start tracking the progress of that operation. If it is not found, a new RPC will be executed.

The query in step 2 above has a problem that the query can be specified in the wrong order, and will in that case be rejected by the backend. But for that to happen, the random error in step 1 needs to happen first. I'll start the build again and see if the error happens again.

PR #2784, opened 2020-04-24
kms: add attestation and rotation schedule sample

sethvargo:
Add two new samples: get attestation on an hsm key, remove rotation schedule from an existing key.

### DIFF
--- a/kms/src/main/java/kms/GetKeyVersionAttestation.java
+++ b/kms/src/main/java/kms/GetKeyVersionAttestation.java
@@ -1,0 +1,70 @@
+/*
+ * Copyright 2020 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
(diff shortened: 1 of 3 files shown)


kms/src/main/java/kms/UpdateKeyRemoveLabels.java
```diff
@@ -50,7 +50,7 @@ public void updateKeyRemoveLabels(
       CryptoKey key = CryptoKey.newBuilder().setName(cryptoKeyName.toString()).build();
 
       // Construct the field mask.
-      FieldMask fieldMask = FieldMaskUtil.fromString("rotation_period,next_rotation_time");
+      FieldMask fieldMask = FieldMaskUtil.fromString("labels");
```
sethvargo, line 53: oops

sethvargo: Hope this is okay to fix in this PR

PR #2785, opened 2020-04-24
Create stale.yml

lesv:
Automated close of stale issues / PR's

From: https://probot.github.io/apps/stale/

### DIFF
--- a/.github/stale.yml
+++ b/.github/stale.yml
@@ -1,0 +1,18 @@
+# Number of days of inactivity before an issue becomes stale
+daysUntilStale: 30
+# Number of days of inactivity before a stale issue is closed
+daysUntilClose: 7
+# Issues with these labels will never be considered stale
+exemptLabels:
+  - pinned
+  - security
+
+# Label to use when marking an issue as stale
+staleLabel: stale-wontfix
+# Comment to post when marking an issue as stale. Set to `false` to disable
+markComment: >
+  This issue has been automatically marked as stale because it has not had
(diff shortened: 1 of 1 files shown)


.github/stale.yml
```diff
@@ -0,0 +1,18 @@
+# Number of days of inactivity before an issue becomes stale
+daysUntilStale: 30
+# Number of days of inactivity before a stale issue is closed
```
kurtisvg, line 3: Is this _issues_ or _PRs_? (Or both?) 

I'm not sure if we should be closing issues, but I agree we should be closing PRs

lesv: Both, I agree w/ you on issues, but expect it will force us to comment.

lesv: I think I have a few changes to this.  Hold off for a bit.

lesv: PTAL

lesv:
PTAL

.github/stale.yml
```diff
@@ -0,0 +1,26 @@
...
+daysUntilClose: 7
+# Issues with these labels will never be considered stale
+exemptLabels:
```
kurtisvg, line 6: What about DO NOT MERGE?

lesv: We shouldn't have them as PR's, if they are being used by docs, they should be on official branches we review.  If a DNM isn't touched for 30 days, someone should ping.

PR #2800, opened 2020-04-27
Fix region tags

ace-n:


ace-n:
@kurtisvg please merge if LGTM - thanks!

functions/helloworld/scala-helloworld/pom.xml
```diff
@@ -43,8 +48,8 @@
       <groupId>org.scala-lang</groupId>
       <artifactId>scala-library</artifactId>
       <version>2.13.2</version>
-    </dependency>
-
+    </dependency>=
```
kurtisvg, line 51: I think this is your test failure
```suggestion
    </dependency>
```

kurtisvg:
@ace-n LGTM once tests pass

functions/helloworld/scala-helloworld/pom.xml
ace-n: Oops - thanks!

PR #2805, opened 2020-04-28
More stylistic fixes for functions samples

ace-n:
Fixes issues @eamonnmcmanus raised:
- consistent-case for `logger` var [b/154869523]
- remove `maven-compiler-plugin` dependency [b/154868919]
- add `<scope>provided</scope>` to FF API dependency [also b/154868919]

**Note:** I split these changes up into commits to make them easier to review. (Per-directory PRs didn't make as much sense here IMO, since reviewing a bunch of 100-percent-identical changes in 1 PR is often easier.)

PR #2807, opened 2020-04-28
Handle undefined environment variables.

eamonnmcmanus:
Avoid getting an exception in certain samples when environment variables
are not defined. Exceptions end up being translated into HTTP 500 status
codes, which in turn causes deployment of these functions to fail the
post-deployment health check.

### DIFF
--- a/functions/ocr/ocr-translate-text/src/main/java/functions/OcrTranslateText.java
+++ b/functions/ocr/ocr-translate-text/src/main/java/functions/OcrTranslateText.java
@@ -36,10 +36,11 @@ import java.util.logging.Level;
 import java.util.logging.Logger;
 
 public class OcrTranslateText implements BackgroundFunction<PubSubMessage> {
-  // TODO<developer> set these environment variables
-  private static final String PROJECT_ID = System.getenv("GCP_PROJECT");
-  private static final String RESULTS_TOPIC_NAME = System.getenv("RESULT_TOPIC");
   private static final Logger LOGGER = Logger.getLogger(OcrTranslateText.class.getName());
+
+  // TODO<developer> set these environment variables
+  private static final String PROJECT_ID = getenv("GCP_PROJECT");
+  private static final String RESULTS_TOPIC_NAME = getenv("RESULT_TOPIC");
   private static final String LOCATION_NAME = LocationName.of(PROJECT_ID, "global").toString();
 
   private Publisher publisher;
@@ -96,6 +97,15 @@ public class OcrTranslateText implements BackgroundFunction<PubSubMessage> {
       // Log error (since these exception types cannot be thrown by a function)
       LOGGER.log(Level.SEVERE, "Error publishing translation save request: " + e.getMessage(), e);
     }
+  }
+
+  private static String getenv(String name) {
+    String value = System.getenv(name);
+    if (value == null) {
+      logger.warning("Environment variable " + name + " is undefined");
+      value = "MISSING";
+    }
+    return value;
   }
 }
 // [END functions_ocr_translate]
--- a/functions/slack/src/main/java/functions/SlackSlashCommand.java
+++ b/functions/slack/src/main/java/functions/SlackSlashCommand.java
@@ -32,13 +32,15 @@ import java.net.HttpURLConnection;
 import java.security.GeneralSecurityException;
 import java.util.HashMap;
 import java.util.List;
+import java.util.logging.Logger;
 import java.util.stream.Collectors;
 
 public class SlackSlashCommand implements HttpFunction {
 
   // [START functions_slack_setup]
-  private static final String API_KEY = System.getenv("KG_API_KEY");
-  private static final String SLACK_SECRET = System.getenv("SLACK_SECRET");
+  private static final Logger logger = Logger.getLogger("functions.SlackSlashCommand");
+  private static final String API_KEY = getenv("KG_API_KEY");
+  private static final String SLACK_SECRET = getenv("SLACK_SECRET");
   private static final Gson gson = new Gson();
 
   private Kgsearch kgClient;
@@ -49,6 +51,15 @@ public class SlackSlashCommand implements HttpFunction {
         GoogleNetHttpTransport.newTrustedTransport(), new JacksonFactory(), null).build();
 
     verifier = new SlackSignature.Verifier(new SlackSignature.Generator(SLACK_SECRET));
+  }
+
+  private static String getenv(String name) {
+    String value = System.getenv(name);
+    if (value == null) {
+      logger.warning("Environment variable " + name + " is undefined");
+      value = "MISSING";
+    }
+    return value;
   }
   // [END functions_slack_setup]
 


functions/ocr/ocr-translate-text/src/main/java/functions/OcrTranslateText.java
```diff
@@ -97,5 +98,14 @@ public void accept(PubSubMessage pubSubMessage, Context context) {
       LOGGER.log(Level.SEVERE, "Error publishing translation save request: " + e.getMessage(), e);
     }
   }
+
+  private static String getenv(String name) {
```
ace-n, line 102: IMO we should still throw an error here - failing the deployment seems OK to me, as long as the user can find ouy why.

functions/ocr/ocr-translate-text/src/main/java/functions/OcrTranslateText.java
```diff
@@ -97,5 +98,14 @@ public void accept(PubSubMessage pubSubMessage, Context context) {
...
+    String value = System.getenv(name);
+    if (value == null) {
+      logger.warning("Environment variable " + name + " is undefined");
```
ace-n, line 105: Nit: s/`is undefined`/`was not set`. (`undefined` is a reserved keyword in some other CF languages, e.g. Node.)

Also, can we use `String.format()` here instead of concatenation?

lesv:
# Java 11
```
- testing functions/ocr/ocr-translate-text
[ERROR] COMPILATION ERROR :
[ERROR] /tmpfs/src/github/java-docs-samples/functions/ocr/ocr-translate-text/src/main/java/functions/OcrTranslateText.java:[105,7] cannot find symbol
  symbol:   variable logger
  location: class functions.OcrTranslateText
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.8.1:compile (default-compile) on project functions-ocr-translate-text: Compilation failure
[ERROR] /tmpfs/src/github/java-docs-samples/functions/ocr/ocr-translate-text/src/main/java/functions/OcrTranslateText.java:[105,7] cannot find symbol
[ERROR]   symbol:   variable logger
[ERROR]   location: class functions.OcrTranslateText
```

functions/ocr/ocr-translate-text/src/main/java/functions/OcrTranslateText.java
eamonnmcmanus: I agree in principle, but in practice the current experience is pretty bad. The deployment fails with a cryptic message `Error setting up the execution environment for your function`. If that gets changed so it says something like `Deployment succeeded but a test execution failed`, or if we modify the API so that a background function can return an HTTP status other than 500, then we could reconsider.

functions/ocr/ocr-translate-text/src/main/java/functions/OcrTranslateText.java
eamonnmcmanus: We're not using String.format in other log messages in the function where we might.

I've made the suggested change to the wording.

functions/ocr/ocr-translate-text/src/main/java/functions/OcrTranslateText.java
ace-n: So - are you saying we should throw an Exception at invocation-time instead?

(IIUC we **have** to throw the exception somewhere - it doesn't make sense to use "default values" here, because [IMO] there _are_ no sensible "default values" for things like Pub/Sub topic names.)

functions/ocr/ocr-translate-text/src/main/java/functions/OcrTranslateText.java
ace-n: OK for now, assuming this is addressed [by a future `checkstyle` config update](https://github.com/GoogleCloudPlatform/java-repo-tools/issues/136).

functions/ocr/ocr-translate-text/src/main/java/functions/OcrTranslateText.java
```diff
@@ -36,10 +36,11 @@
 import java.util.logging.Logger;
 
 public class OcrTranslateText implements BackgroundFunction<PubSubMessage> {
+  private static final Logger logger = Logger.getLogger(OcrTranslateText.class.getName());
```
ace-n, line 39: I think this change was already handled [here](https://github.com/GoogleCloudPlatform/java-docs-samples/pull/2805)?

(Maybe we should revert this line to reduce the risk of merge conflicts.)

eamonnmcmanus: I had to merge manually, since this declaration needs to be before the calls to `getenv` (since those can issue log messages).

functions/ocr/ocr-translate-text/src/main/java/functions/OcrTranslateText.java
eamonnmcmanus: The problem is that if you deploy this sample without defining the required environment variables, currently the deployment will fail with the cryptic message I mentioned. That's because of the "health check" that gets run as the last step of deployment. I'm suggesting that a logged message is a better experience. If we were able to fail with an HTTP 400 that would be better still, since it wouldn't cause the health check to fail but would still show the problem clearly. Unfortunately that's not possible with the current API.

ace-n: I believe we _can_ log the error message _and then_ throw the exception though, right? (Or will that "swallow" the error message?)

eamonnmcmanus: If we log a message and throw an exception, the message should indeed get logged, but the exception will result in a 500 status which will cause the health check to fail and the deployment to end with `Error setting up the execution environment for your function`. Users are not necessarily going to think to look in the logs at that point.

In fact, the current situation is worse than that because the null values will cause a NullPointerException in the static initializer, which means that the function class can't be loaded, and we'll give you a really obscure error message in the logs. But even fixing that NPE still leads to a poor user experience if you don't have the env vars.

ace-n: > If we log a message and throw an exception, the message should indeed get logged, but the exception will result in a 500 status which will cause the health check to fail and the deployment to end with Error setting up the execution environment for your function. **Users are not necessarily going to think to look in the logs at that point.**

Fair point - if @lesv is cool with this approach, we can merge as-is. We should still file an issue to revise this sample once we have proper error logging, though.

lesv: You can add info about this in the README and the docs.  They will want to know these things for their own apps.

lesv: A "What might go wrong" section is always helpful.  Explaining common errors is always a good thing.  If eng complains, you can explain how common the error is and ask them to fix it.

ace-n: Les - this is likely a temporary problem, so it's probably not worth adding specific information to the docs.

(There's probably an internal bug tracking this somewhere, but I can't seem to find it. Will post it here or ping you when I do.)

lesv:
ping

lesv:
@ace-n @eamonnmcmanus PING?

eamonnmcmanus:
I think the ball is in @ace-n's court.

PR #2808, opened 2020-04-28
Use a PrintWriter to make printf available.

eamonnmcmanus:
Otherwise we have to call String.format. This also fixes a place where
we weren't in fact calling String.format even though we were using the
%n format specifier.

### DIFF
--- a/functions/concepts/file-system/src/main/java/functions/FileSystem.java
+++ b/functions/concepts/file-system/src/main/java/functions/FileSystem.java
@@ -21,9 +21,9 @@ package functions;
 import com.google.cloud.functions.HttpFunction;
 import com.google.cloud.functions.HttpRequest;
 import com.google.cloud.functions.HttpResponse;
-import java.io.BufferedWriter;
 import java.io.File;
 import java.io.IOException;
+import java.io.PrintWriter;
 
 public class FileSystem implements HttpFunction {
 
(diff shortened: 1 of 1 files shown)


functions/concepts/file-system/src/main/java/functions/FileSystem.java
```diff
@@ -33,10 +33,10 @@ public void service(HttpRequest request, HttpResponse response)
       throws IOException {
     File currentDirectory = new File(".");
     File[] files = currentDirectory.listFiles();
-    BufferedWriter writer = response.getWriter();
-    writer.write("Files: %n");
+    PrintWriter writer = new PrintWriter(response.getWriter());
+    writer.println("Files:");
     for (File f : files) {
-      writer.write(String.format("\t%s%n", f.getName()));
+      writer.printf("\t%s%n", f.getName());
```
lesv, line 39: drop the %n

ace-n:
@eamonnmcmanus looks like this pattern is used elsewhere. Can you fix those here too?

```
$ grep write\(String **/*.java

concepts/file-system/src/main/java/functions/FileSystem.java:      writer.write(String.format("\t%s%n", f.getName()));
concepts/lazy-fields/src/main/java/functions/LazyFields.java:    writer.write(String.format("Lazy global: %s; non-lazy global: %s",
concepts/scopes/src/main/java/functions/Scopes.java:    writer.write(String.format("Instance: %s; function: %s", InstanceVar, functionVar));
helloworld/hello-http/src/main/java/functions/HelloHttp.java:    writer.write(String.format("Hello %s!", name));
http/parse-content-type/src/main/java/functions/ParseContentType.java:      writer.write(String.format("Hello %s!", name));
http/send-http-request/src/main/java/functions/SendHttpRequest.java:    writer.write(String.format("Received code '%s' from url '%s'.", getResponse.statusCode(), url));
logging/retrieve-logs/src/main/java/functions/RetrieveLogs.java:      writer.write(String.format("%s: %s%n", entry.getLogName(), entry.getTextPayload()));
spanner/src/main/java/functions/HelloSpanner.java:        writer.write(String.format("Error querying database: %s\n", e.getMessage()));
spanner/src/main/java/functions/HelloSpanner.java:      writer.write(String.format("Error setting up Spanner: %s\n", t.getMessage()));
```

eamonnmcmanus:
I'll send another PR to use PrintWriter everywhere we do anything printf-like.

PR #2810, opened 2020-04-28
Wrap BufferedWriter in PrintWriter to simplify some code.

eamonnmcmanus:
This enables us to replace `writer.write(String.format(...))` with
`writer.printf(...)`.

Also use the `var` keyword to simplify a couple of places where a type
had to be fully-qualified because of a naming conflict.


functions/http/send-http-request/src/main/java/functions/SendHttpRequest.java
```diff
@@ -41,17 +41,15 @@ public void service(HttpRequest request, HttpResponse response)
       throws IOException, InterruptedException {
     // Create a GET sendHttpRequest to "http://example.com"
     String url = "http://example.com";
-    java.net.http.HttpRequest getRequest =
-        java.net.http.HttpRequest.newBuilder().uri(URI.create(url)).GET().build();
+    var getRequest = java.net.http.HttpRequest.newBuilder().uri(URI.create(url)).GET().build();
```
ace-n, line 44: I have mixed feelings on `var` - it's simpler, but (to me) feels less idiomatic for Java.

@kurtisvg @lesv WDYT?

functions/spanner/src/main/java/functions/HelloSpanner.java
```diff
@@ -97,27 +97,26 @@ DatabaseClient getClient() throws Throwable {
...
               .executeQuery(Statement.of("SELECT SingerId, AlbumId, AlbumTitle FROM Albums"))) {
-        writer.write("Albums:\n");
+        writer.printf("Albums:\n");
         while (rs.next()) {
-          writer.write(
-              String.format(
-                  "%d %d %s\n",
-                  rs.getLong("SingerId"), rs.getLong("AlbumId"), rs.getString("AlbumTitle")));
+          writer.printf(
+              "%d %d %s\n",
```
ace-n, line 110: Nit: `spotbugs` does not like `\n`, and prefers `%n` instead.

Apply that comment throughout this file.

eamonnmcmanus: Will do, though I'll note that I didn't change these strings, which already had `\n` in them.

functions/http/send-http-request/src/main/java/functions/SendHttpRequest.java
eamonnmcmanus: It hasn't been in the language all that long, but I think everyone should use it in appropriate places, of which this is definitely one since it allows us to replace three fully-qualified class names with just one.

lesv: It should be fine to use in Java 11 apps.

lesv: It's J11, it's often clearer and less error prone.

PR #2814, opened 2020-04-29
Background function: add return statement

ace-n:


functions/helloworld/hello-pubsub/src/main/java/functions/HelloPubSub.java
```diff
@@ -39,6 +39,7 @@ public void accept(PubSubMessage message, Context context) {
           StandardCharsets.UTF_8);
     }
     logger.info(String.format("Hello %s!", name));
+    return;
```
eamonnmcmanus, line 42: I don't understand the purpose of this. The method is about to return anyway.

ace-n: This is a temporary fix for the docs until b/155338265 is resolved. 🙂

eamonnmcmanus: I guess I don't really see how it helps.

PR #2845, opened 2020-05-06
functions: shrink example pom sample

ace-n:
@saturnism FYI

### DIFF
--- a/functions/helloworld/helloworld/pom.xml
+++ b/functions/helloworld/helloworld/pom.xml
@@ -16,6 +16,7 @@
   limitations under the License.
 -->
 
+<!-- [START functions_example_pom] -->
 <project xmlns="http://maven.apache.org/POM/4.0.0"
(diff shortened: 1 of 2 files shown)


functions/helloworld/helloworld/pom.xml
```diff
@@ -16,22 +16,27 @@
...
   </parent>
 
+  <!-- [START functions_example_pom] -->
   <properties>
+    <!-- [END functions_example_pom] -->
     <powermock.version>2.0.7</powermock.version>
+    <!-- [START functions_example_pom] -->
     <maven.compiler.target>11</maven.compiler.target>
     <maven.compiler.source>11</maven.compiler.source>
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
```
saturnism, line 42: this is usually not required, can we remove this from the tag too?

ace-n: See #2851

PR #2846, opened 2020-05-06
functions: shrink Kotlin example POM

ace-n:
@saturnism FYI - hopefully I got everything here?

(The removed section seemed a bit small. 🙂 )

### DIFF
--- a/functions/helloworld/kotlin-helloworld/pom.xml
+++ b/functions/helloworld/kotlin-helloworld/pom.xml
@@ -162,14 +162,14 @@
             <goals>
               <goal>compile</goal>
             </goals>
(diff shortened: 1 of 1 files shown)


saturnism:
i think that block can be removed completely, i can give it a try.

saturnism:
i see why we needed it - the function is written in Kotlin, but test is in Java. We can't default the source/test directory to kotlin.

In the mean time, I feel we can remove the just the line `<source>target/generated-sources/annotations</source>`

the rest of the block can remain in tact and do not need to be excluded from the region.

saturnism:
rather than excluding the block, can we remove the line completely? I.e., the line isn't needed at all to compile this code.

PR #2847, opened 2020-05-06
deps: upgrade beam version to 2.20.0 in bigtable and dataflow samples

kolea2:
Opening in favor of #2740


kolea2:
CC @billyjacobson @lesv - beam 2.20.0 was bringing in a new version of guava that conflicted with our truth dep, so excluded that in the test dependency. Also cleaned up unnecessary declared deps.

PR #2850, opened 2020-05-06
Update functions/spanner to use plain Mockito rather than PowerMockito.

eamonnmcmanus:
This is an example of the approach that I propose to follow for the 27 other samples that currently use PowerMockito.


eamonnmcmanus:
Looks like I'm not authorized to merge this so if someone who is can that would be appreciated.

PR #2853, opened 2020-05-07
Remove PowerMock from nearly all samples, in favour of plain Mockito.

eamonnmcmanus:
Also use the more usual `MockitoAnnotations.initMocks(this)` rather than
`Mockito.mockitoSession().initMocks(this)`. Remove some unnecessary
`throws` clauses. Pass a value through a second constructor rather than
bashing a field with PowerMock's Whitebox.

### DIFF
--- a/functions/concepts/env-vars/pom.xml
+++ b/functions/concepts/env-vars/pom.xml
@@ -31,7 +31,6 @@
   </parent>
 
   <properties>
-    <powermock.version>2.0.7</powermock.version>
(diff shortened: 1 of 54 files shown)


eamonnmcmanus:
> is there anywhere it's still used?

It's still used in `SlackSlashCommand` for the `Whitebox` functionality (which pokes test values into class and instance fields of the code under test). I think we can rewrite the test to avoid that.

PR #2946, opened 2020-05-20
Functions: clarify cloudbuild.yaml

ace-n:
cc @labtopia @saturnism


ace-n:
@labtopia would it be better if we called this out in the docs?

labtopia:
Adding a comment to the docs to clarify sounds like a good idea to me.

PR #2985, opened 2020-05-28
docs(samples): Fix code style for line more than 120 character and add explanation for setting up environment variable

irvifa:
There's several files that contains code with more than 120 characters
within single row. I think it will be more easier to read if we separate
those long lines into multiple lines. Also I add how to export a PROJECT_ID
since it will be easier to do so if one's opening the google shell.

Fixes #issue

- [x] Tests pass (`mvn -P lint clean verify`)


lesv:
I'll keep this open for someone on the Firestore team to comment.  I think your changes are making the sample a bit less friendly.  In the future it might be better to open an issue to discuss what you want to do.

stephaniewang526:
@BenWhitehead owns firestore

lesv:
@BenWhitehead PTAL

firestore/src/main/java/com/example/firestore/snippets/QueryDataSnippets.java
```diff
@@ -38,10 +38,12 @@
 /** Snippets to support firestore querying data documentation. */
 class QueryDataSnippets {
 
-  private final Firestore db;
+  private static final String COLLECTION_REFERENCE_NAME = "cities";
```
BenWhitehead, line 41: In normal application code this would likely be a desired refactoring, but due to the fact that each of the users of this new variable are individual code snippets that are read on their own we prefer to keep the value inline.

firestore/src/main/java/com/example/firestore/snippets/RetrieveDataSnippets.java
```diff
@@ -45,16 +45,51 @@ void prepareExamples() throws Exception {
...
+                            Arrays.asList("kanto", "honshu"))));
+    futures.add(
+            cities.document("BJ")
+                    .set(new City(
+                            "Beijing",
+                            null,
+                            "China",
+                            true,
+                            21500000L,
+                            Arrays.asList("jingjinji", "hebei"))));
```
BenWhitehead, line 92: Applying this formatting would cause this code snipped to be longer than a single screen when viewed in the documentation. We want to keep the whole context visible on a single screen.

firestore/README.md
```diff
@@ -48,6 +48,7 @@ Run all tests:
 
 Set the `FIRESTORE_PROJECT_ID` environment variable and run the command:
 ```
+   export FIRESTORE_PROJECT_ID=<YOUR-PROJECT-ID>
```
BenWhitehead, line 51: This line of code only works for unix operating systems, not for those users on windows which is why the above line leaves it up to the developer to know how to do it for their workstation.

BenWhitehead:
Thank you for the interest in contributing to the Firestore code snippets. 

However, this PR unfortunately does not include changes that benefit the display of the code snippets on https://cloud.google.com/firestore/docs or https://firebase.google.com/docs/firestore .

PR #3042, opened 2020-06-03
Include locations/global in parent field to fix samples

sushicw:
Adds "locations/global" to the parent field specified when calling DLP api. This makes the job name produced consistent with the internal usage and, importantly to these samples, consistent with the job name used in the pubsub message indicating that the job is complete. This should fix certain tests timing out after 15 minutes waiting for the wrong pubsub message.

fixes internal bug b/158100365 (partially: only for Java samples)
Also see internal bug b/158026949 for more context.

- [X] ~~API's need to be enabled to test (tell us)~~ (*Nothing new*)
- [X] ~~Environment Variables need to be set (ask us to set them)~~ (*Nothing new*)

### DIFF
--- a/dlp/src/main/java/dlp/snippets/DeIdentifyWithDateShift.java
+++ b/dlp/src/main/java/dlp/snippets/DeIdentifyWithDateShift.java
@@ -27,8 +27,8 @@ import com.google.privacy.dlp.v2.DeidentifyContentRequest;
 import com.google.privacy.dlp.v2.DeidentifyContentResponse;
 import com.google.privacy.dlp.v2.FieldId;
 import com.google.privacy.dlp.v2.FieldTransformation;
+import com.google.privacy.dlp.v2.LocationName;
 import com.google.privacy.dlp.v2.PrimitiveTransformation;
-import com.google.privacy.dlp.v2.ProjectName;
 import com.google.privacy.dlp.v2.RecordTransformations;
 import com.google.privacy.dlp.v2.Table;
 import com.google.privacy.dlp.v2.Value;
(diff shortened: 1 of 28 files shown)


dlp/src/main/java/dlp/snippets/DeIdentifyWithDateShift.java
```diff
@@ -98,7 +98,7 @@ public static void deIdentifyWithDateShift(
       // Combine configurations into a request for the service.
       DeidentifyContentRequest request =
           DeidentifyContentRequest.newBuilder()
-              .setParent(ProjectName.of(projectId).toString())
+              .setParent(String.format("%s/locations/global", ProjectName.of(projectId).toString()))
```
lesv, line 101: Really seems like the client library should be changed, like adding a `toGlobalString()` or `asParent()` methods.

sushicw: Unfortunately the client library is generated, so we don't currently have a straightforward way to add arbitrary helper functions. 

I agree that it's not ideal, but I'm unaware of better available options.

tmatsuo: FYI, the Python library has a class method `location_path`. Worth checking if there's a similar thing in Java client.

sushicw: Thanks Takashi, you're right! In Java it's LocationName and it's definitely the right way to go. Updating now.

lesv: @sushicw Please ping me once you've updated the PR.

lesv:
DO NOT MERGE - @sushicw is looking into something first.

sushicw:
> @sushicw is looking into something first.

Finished checking, we're good to go. Thanks for the quick hold.

PR #3049, opened 2020-06-04
Fix DLP inspect tests to fail on timeout

sushicw:
Update DLP inspect tests to fail properly, instead of timing out after 15 minutes and passing. 
Also remove some broken code that tried to cancel early. 

Fixes internal bug b/158026949

- [X] API's need to be enabled to test (tell us)
- [X] Environment Variables need to be set (ask us to set them)

### DIFF
--- a/dlp/src/test/java/dlp/snippets/InspectTests.java
+++ b/dlp/src/test/java/dlp/snippets/InspectTests.java
@@ -24,10 +24,8 @@ import static org.junit.Assert.assertNotNull;
 import com.google.cloud.dlp.v2.DlpServiceClient;
 import com.google.privacy.dlp.v2.CancelDlpJobRequest;
 import java.io.ByteArrayOutputStream;
-import java.io.IOException;
 import java.io.PrintStream;
 import java.util.Arrays;
-import java.util.concurrent.ExecutionException;
 import java.util.concurrent.TimeUnit;
 import org.junit.After;
 import org.junit.Before;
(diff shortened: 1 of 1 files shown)


dlp/src/test/java/dlp/snippets/InspectTests.java
```diff
@@ -124,60 +122,30 @@ public void testInspectImageFile() {
   }
 
   @Test
-  public void testInspectGcsFile() throws InterruptedException, ExecutionException, IOException {
-
+  public void testInspectGcsFile() throws Exception {
```
lesv, line 125: It's actually best to throw the most specific exceptions so that users understand and can mitigate what's going on.  That said, in a test, this is less important.

dlp/src/test/java/dlp/snippets/InspectTests.java
```diff
@@ -124,60 +122,30 @@ public void testInspectImageFile() {
...
-    try (DlpServiceClient client = DlpServiceClient.create()) {
-      client.cancelDlpJob(request);
-    }
+    assertThat(output, containsString("Job status: DONE"));
   }
 
   @Test
-  public void testInspectDatastoreEntity()
-      throws InterruptedException, ExecutionException, IOException {
+  public void testInspectDatastoreEntity() throws Exception {
```
lesv, line 133: Why did you reduce the throws here?

dlp/src/test/java/dlp/snippets/InspectTests.java
sushicw: See below.

PR #3051, opened 2020-06-04
Add Java code sample for redaction

jakubrauch:
To be linked from https://cloud.google.com/dlp/docs/transformations-reference

Fixes [N/A]

- [yes, except where consistency made more sense] I have followed [Sample Format Guide](https://github.com/GoogleCloudPlatform/java-docs-samples/blob/master/SAMPLE_FORMAT.md)
- [yes] `pom.xml` parent set to latest `shared-configuration`
- [nothing new] Appropriate changes to README are included in PR
- [nothing new] API's need to be enabled to test (tell us)
- [nothing new] Environment Variables need to be set (ask us to set them)
- [yes] Tests pass (`mvn -P lint clean verify`)
- [yes] Please **merge** this PR for me once it is approved.

### DIFF
--- a/dlp/src/main/java/dlp/snippets/DeIdentifyWithRedaction.java
+++ b/dlp/src/main/java/dlp/snippets/DeIdentifyWithRedaction.java
@@ -1,0 +1,92 @@
+/*
+ * Copyright 2019 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package dlp.snippets;
+
+// [START dlp_deidentify_redact]
+
+import com.google.cloud.dlp.v2.DlpServiceClient;
+import com.google.privacy.dlp.v2.ContentItem;
+import com.google.privacy.dlp.v2.DeidentifyConfig;
+import com.google.privacy.dlp.v2.DeidentifyContentRequest;
+import com.google.privacy.dlp.v2.DeidentifyContentResponse;
+import com.google.privacy.dlp.v2.InfoType;
+import com.google.privacy.dlp.v2.InfoTypeTransformations;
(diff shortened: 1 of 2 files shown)


dlp/src/main/java/dlp/snippets/DeIdentifyWithRedaction.java
```diff
@@ -0,0 +1,92 @@
...
+import com.google.privacy.dlp.v2.DeidentifyContentResponse;
+import com.google.privacy.dlp.v2.InfoType;
+import com.google.privacy.dlp.v2.InfoTypeTransformations;
+import com.google.privacy.dlp.v2.InfoTypeTransformations.InfoTypeTransformation;
+import com.google.privacy.dlp.v2.InspectConfig;
+import com.google.privacy.dlp.v2.LocationName;
+import com.google.privacy.dlp.v2.PrimitiveTransformation;
+import com.google.privacy.dlp.v2.RedactConfig;
+
+public class DeIdentifyWithRedaction {
```
kolea2, line 34: I would change any class or method names containing `DeIdentify` to have a lowercase 'i' to be consistent with the client,  for example `DeidentifyConfig`.

stephaniewang526: Actually samples in java-docs-samples repo all have capitalized class names -- so should be okay.

jakubrauch: Yes, it's using this format for consistency with other examples. We could consider migrating all DeIdentify to Deidentify in a separate item. That would require keeping old names around until all references to them have been migrated to a lowercase name.

dlp/src/main/java/dlp/snippets/DeIdentifyWithRedaction.java
```diff
@@ -0,0 +1,92 @@
+/*
+ * Copyright 2019 Google LLC
```
lesv, line 2: 2020 ??

dlp/src/main/java/dlp/snippets/DeIdentifyWithRedaction.java
```diff
@@ -0,0 +1,92 @@
...
+              .setDeidentifyConfig(redactConfig)
+              .setInspectConfig(inspectConfig)
+              .build();
+
+      // Use the client to send the API request.
+      DeidentifyContentResponse response = dlp.deidentifyContent(request);
+
+      // Parse the response and process results
+      System.out.println("Text after redaction: " + response.getItem().getValue());
+    } catch (Exception e) {
```
lesv, line 87: Really should explain a bit more here, also can you be more specific about the kind of errors that may occur

PR #3055, opened 2020-06-04
Create Java code sample for omit finding with substring using custom detector

sushicw:
Fixes internal bug b/156967687

- [X] API's need to be enabled to test (tell us) (**Nothing New**)
- [X] Environment Variables need to be set (ask us to set them) (**Nothing New**)

### DIFF
--- a/dlp/src/main/java/dlp/snippets/InspectStringCustomExcludingSubstring.java
+++ b/dlp/src/main/java/dlp/snippets/InspectStringCustomExcludingSubstring.java
@@ -16,7 +16,7 @@
 
 package dlp.snippets;
 
-// [START dlp_inspect_string_with_exclusion_dict_substring]
+// [START dlp_inspect_string_custom_excluding_substring]
 
 import com.google.cloud.dlp.v2.DlpServiceClient;
 import com.google.privacy.dlp.v2.ByteContentItem;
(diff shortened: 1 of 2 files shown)


dlp/src/main/java/dlp/snippets/InspectStringCustomExcludingSubstring.java
```diff
@@ -0,0 +1,120 @@
...
+      InspectContentResponse response = dlp.inspectContent(request);
+
+      // Parse the response and process results
+      System.out.println("Findings: " + response.getResult().getFindingsCount());
+      for (Finding f : response.getResult().getFindingsList()) {
+        System.out.println("\tQuote: " + f.getQuote());
+        System.out.println("\tInfo type: " + f.getInfoType().getName());
+        System.out.println("\tLikelihood: " + f.getLikelihood());
+      }
+    } catch (Exception e) {
```
lesv, line 115: It would be nice if you were more specific here and say what might happen and how they might mitigate it.
https://google.github.io/styleguide/javaguide.html#s6.2-caught-exceptions

sushicw: This is copied from existing parallel sample code. I agree that it's pretty crappy, and we're also not really consistent: sometimes we do this, sometimes we just stick a "throws IOException" on the function. 

Of the two I think I prefer the latter, it makes it clear that this code is explicitly not handling that. Can I handle these in a  follow-up for all our DLP Java samples? (see internal bug b/158239329)

lesv: SGTM

PR #3112, opened 2020-06-09
Create Java code sample for multiple rules

sushicw:
Fixes internal bug b/156974340

- [X] API's need to be enabled to test (tell us) (**Nothing new**)
- [X] Environment Variables need to be set (ask us to set them) (**Nothing new**)

### DIFF
--- a/dlp/src/main/java/dlp/snippets/InspectStringMultipleRules.java
+++ b/dlp/src/main/java/dlp/snippets/InspectStringMultipleRules.java
@@ -1,0 +1,143 @@
+/*
+ * Copyright 2020 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package dlp.snippets;
+
+// [START dlp_inspect_string_multiple_rules]
+
+import com.google.cloud.dlp.v2.DlpServiceClient;
+import com.google.privacy.dlp.v2.ByteContentItem;
+import com.google.privacy.dlp.v2.ByteContentItem.BytesType;
+import com.google.privacy.dlp.v2.ContentItem;
+import com.google.privacy.dlp.v2.CustomInfoType.DetectionRule.HotwordRule;
+import com.google.privacy.dlp.v2.CustomInfoType.DetectionRule.LikelihoodAdjustment;
+import com.google.privacy.dlp.v2.CustomInfoType.DetectionRule.Proximity;
+import com.google.privacy.dlp.v2.CustomInfoType.Dictionary;
+import com.google.privacy.dlp.v2.CustomInfoType.Dictionary.WordList;
+import com.google.privacy.dlp.v2.CustomInfoType.Regex;
+import com.google.privacy.dlp.v2.ExclusionRule;
+import com.google.privacy.dlp.v2.Finding;
+import com.google.privacy.dlp.v2.InfoType;
+import com.google.privacy.dlp.v2.InspectConfig;
+import com.google.privacy.dlp.v2.InspectContentRequest;
+import com.google.privacy.dlp.v2.InspectContentResponse;
+import com.google.privacy.dlp.v2.InspectionRule;
+import com.google.privacy.dlp.v2.InspectionRuleSet;
+import com.google.privacy.dlp.v2.Likelihood;
+import com.google.privacy.dlp.v2.LocationName;
+import com.google.privacy.dlp.v2.MatchingType;
+import com.google.protobuf.ByteString;
+import java.io.IOException;
+
+public class InspectStringMultipleRules {
+
+  public static void inspectStringMultipleRules() throws IOException {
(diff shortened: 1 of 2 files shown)


dlp/src/test/java/dlp/snippets/InspectTests.java
```diff
@@ -251,6 +251,15 @@ public void testInspectStringCustomHotwordNegativeExample() throws Exception {
     assertThat(output, not(containsString("John Doe")));
   }
 
+  @Test
```
dzlier-gcp, line 254: Can you add tests for the "doctor", "Quasimodo", and "Redacted" rules used in the sample as well?

sushicw: Yes, can do.

dlp/src/test/java/dlp/snippets/InspectTests.java
```diff
@@ -251,6 +251,42 @@ public void testInspectStringCustomHotwordNegativeExample() throws Exception {
...
+    assertThat(output, containsString("VERY_LIKELY"));
+  }
+
+  @Test
+  public void testInspectStringMultipleRulesDoctorRule() throws Exception {
+    InspectStringMultipleRules.inspectStringMultipleRules(PROJECT_ID,
+        "doctor: Jane Doe");
+
+    String output = bout.toString();
+    assertThat(output, containsString("Findings: 0"));
```
dzlier-gcp, line 269: Shouldn't there be a finding here?

sushicw: No, because the custom rule sets it as UNLIKELY which is below the default threshold for results to return.

dzlier-gcp: What are the other potential likelihoods? Unlikely to me sounds like "probably not but possibly," not "ignore anything found".

sushicw: https://cloud.google.com/dlp/docs/likelihood

The minimum likelihood to return is configurable (minLikelihood property on InspectConfig) and defaults to POSSIBLE.

dzlier-gcp: Oh I see, so does "setFixedLikelihood" mean that all results that match the rule will have that as a likelihood, instead of whatever likelihood it would have assigned?

dlp/src/main/java/dlp/snippets/InspectStringMultipleRules.java
```diff
@@ -0,0 +1,143 @@
...
+
+      // Construct hotword rules
+      HotwordRule patientRule = HotwordRule.newBuilder()
+          .setHotwordRegex(Regex.newBuilder()
+              .setPattern("patient"))
+          .setProximity(Proximity.newBuilder()
+              .setWindowBefore(10))
+          .setLikelihoodAdjustment(LikelihoodAdjustment.newBuilder()
+              .setFixedLikelihood(Likelihood.VERY_LIKELY))
+          .build();
```
lesv, line 77: Is this formatted correctly?  Shouldn't they be aligned to make reading easier?

dlp/src/main/java/dlp/snippets/InspectStringMultipleRules.java
```diff
@@ -0,0 +1,143 @@
...
+
+      HotwordRule doctorRule = HotwordRule.newBuilder()
+          .setHotwordRegex(Regex.newBuilder()
+              .setPattern("doctor"))
+          .setProximity(Proximity.newBuilder()
+              .setWindowBefore(10))
+          .setLikelihoodAdjustment(LikelihoodAdjustment.newBuilder()
+              .setFixedLikelihood(Likelihood.UNLIKELY))
+          .build();
```
lesv, line 86: ditto

dlp/src/main/java/dlp/snippets/InspectStringMultipleRules.java
lesv: Better yet, a single line sections
```
      HotwordRule patientRule = HotwordRule.newBuilder()
          .setHotwordRegex(Regex.newBuilder().setPattern("patient"))
          .setProximity(Proximity.newBuilder().setWindowBefore(10))
          .setLikelihoodAdjustment(LikelihoodAdjustment.newBuilder().setFixedLikelihood(Likelihood.VERY_LIKELY))
          .build();
```
Or is that too wide?

lesv:
@dzlier-gcp Just a reminder, you need to add the `kokoro-run` label for non-members.

dlp/src/test/java/dlp/snippets/InspectTests.java
sushicw: Correct. It's adding a rule that says "when a match has this hotword nearby, set the likelihood to X".

dlp/src/main/java/dlp/snippets/InspectStringMultipleRules.java
sushicw: The autoformat breaks the setLikelihoodAdjustment line because it's too long:

```
      HotwordRule patientRule = HotwordRule.newBuilder()
          .setHotwordRegex(Regex.newBuilder().setPattern("patient"))
          .setProximity(Proximity.newBuilder().setWindowBefore(10))
          .setLikelihoodAdjustment(
              LikelihoodAdjustment.newBuilder().setFixedLikelihood(Likelihood.VERY_LIKELY))
          .build();

```

I found it easier to read the structure when the other lines were also broken.

The linter / autoformatter seems happy either way, and I didn't see anything in the style guide that was strongly opinionated about it, so I went with what seemed easier to read. 

LMK if you still want this changed, I don't feel strongly about it and don't mind switching if my instincts on what's easier to read seem off.

PR #3146, opened 2020-06-11
Inspect gcs sample b 157050959

sushicw:
Fixes internal bug b/157050959

- [X] API's need to be enabled to test (tell us) (**Nothing new**)
- [X] Environment Variables need to be set (ask us to set them) (**Nothing new**)

### DIFF
--- a/dlp/src/main/java/dlp/snippets/InspectGcsFileWithSampling.java
+++ b/dlp/src/main/java/dlp/snippets/InspectGcsFileWithSampling.java
@@ -1,0 +1,165 @@
+/*
+ * Copyright 2019 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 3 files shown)


dlp/src/main/java/dlp/snippets/InspectGcsFileWithSampling.java
```diff
@@ -0,0 +1,174 @@
...
+            handleMessage(dlpJob, done, pubsubMessage, ackReplyConsumer);
+          };
+      Subscriber subscriber = Subscriber.newBuilder(subscriptionName, messageHandler).build();
+      subscriber.startAsync();
+
+      // Wait for the original job to complete
+      try {
+        done.get(15, TimeUnit.MINUTES);
+      } catch (TimeoutException e) {
+        System.out.println("Job was not completed after 15 minutes.");
```
lesv, line 137: Is there something a user should do here?

sushicw: For now I'm keeping this consistent with the other existing samples. Obviously "println and return" is probably not the right answer for most prod systems, but it does keep the sample simple. Hopefully it's easy for users to tune to their own needs. 

As a broader effort, it would be great to have a consistent approach to recommend across samples, across APIs, and (ideally) across languages as well.

PR #3176, opened 2020-06-16
add Java snippets for creating, listing, and deleting queues

aarjo:
Environment Variables need to be set:

- GOOGLE_CLOUD_PROJECT to the project id

### DIFF
--- a/tasks/src/main/java/com/example/task/CreateQueue.java
+++ b/tasks/src/main/java/com/example/task/CreateQueue.java
@@ -19,6 +19,7 @@ import com.google.cloud.tasks.v2.CloudTasksClient;
 import com.google.cloud.tasks.v2.LocationName;
 import com.google.cloud.tasks.v2.Queue;
 import com.google.cloud.tasks.v2.QueueName;
+import java.io.IOException;
 
 public class CreateQueue {
   /**
@@ -30,7 +31,7 @@ public class CreateQueue {
    * @throws Exception on Cloud Tasks Client errors.
    */
   public static void createQueue(String projectId, String locationId, String queueId)
-      throws Exception {
+      throws IOException {
 
     // Instantiates a client.
     try (CloudTasksClient client = CloudTasksClient.create()) {
--- a/tasks/src/main/java/com/example/task/CreateQueue.java
+++ b/tasks/src/main/java/com/example/task/CreateQueue.java
@@ -17,8 +17,8 @@ package com.example.task;
 // [START cloud_tasks_create_queue]
 import com.google.cloud.tasks.v2.CloudTasksClient;
 import com.google.cloud.tasks.v2.LocationName;
-import com.google.cloud.tasks.v2.QueueName;
 import com.google.cloud.tasks.v2.Queue;
+import com.google.cloud.tasks.v2.QueueName;
 
 public class CreateQueue {
   /**
--- a/tasks/src/main/java/com/example/task/CreateQueue.java
+++ b/tasks/src/main/java/com/example/task/CreateQueue.java
@@ -22,14 +22,16 @@ import com.google.cloud.tasks.v2.QueueName;
 import java.io.IOException;
 
 public class CreateQueue {
-  /**
-   * Create a queue using the Cloud Tasks client.
-   *
-   * @param projectId the Id of the project.
-   * @param queueId the name of your Queue.
-   * @param locationId the GCP region of your queue.
-   * @throws Exception on Cloud Tasks Client errors.
-   */
+
+  public static void main(String[] args) throws IOException {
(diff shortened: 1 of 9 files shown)


tasks/src/test/java/com/example/task/CreateQueueIT.java
```diff
@@ -0,0 +1,77 @@
...
+import java.io.PrintStream;
+import java.util.UUID;
+import org.junit.After;
+import org.junit.Before;
+import org.junit.BeforeClass;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.junit.runners.JUnit4;
+
+/** Tests for creating Tasks with HTTP targets. */
```
jennstubley, line 32: Update test comments

tasks/src/test/java/com/example/task/CreateQueueIT.java
```diff
@@ -0,0 +1,77 @@
...
+  private static void requireEnvVar(String varName) {
+    assertNotNull(
+        String.format("Environment variable '%s' must be set to perform these tests.", varName),
+        System.getProperty(varName));
+  }
+
+  @BeforeClass
+  public static void checkRequirements() {
+    requireEnvVar("GOOGLE_CLOUD_PROJECT");
+  }
```
jennstubley, line 51: Is there a reason for doing this rather than using the sample project like the create task test does?

aarjo: It makes it easier to run the tests in your own project without having to modify the project id in the code. It seems that tests added to the java-docs recently have been using this approach (ex. https://github.com/GoogleCloudPlatform/java-docs-samples/blob/c3bccc301b14b01924ed05f422160190dcd94646/dataproc/src/test/java/InstantiateInlineWorkflowTemplateTest.java) so I just followed that since that's how Python did it as well.

jennstubley: Sounds good, thanks for the explanation. 

If you want you can update CreateTask tests while you're here to keep consistency within the /tasks directory.

tasks/src/main/java/com/example/task/CreateQueue.java
```diff
@@ -0,0 +1,51 @@
...
+  /**
+   * Create a queue using the Cloud Tasks client.
+   *
+   * @param projectId the Id of the project.
+   * @param queueId the name of your Queue.
+   * @param locationId the GCP region of your queue.
+   * @throws Exception on Cloud Tasks Client errors.
+   */
+  public static void createQueue(String projectId, String locationId, String queueId)
+      throws Exception {
```
lesv, line 33: Any chance you can be more specific?

tasks/src/main/java/com/example/task/DeleteQueue.java
```diff
@@ -0,0 +1,46 @@
...
+  /**
+   * Delete a queue using the Cloud Tasks client.
+   *
+   * @param projectId the Id of the project.
+   * @param queueId the name of your Queue.
+   * @param locationId the GCP region of your queue.
+   * @throws Exception on Cloud Tasks Client errors.
+   */
+  public static void deleteQueue(String projectId, String locationId, String queueId)
+      throws Exception {
```
lesv, line 31: Ideally we are as specific as possible with exceptions based on what developers might see.

tasks/src/main/java/com/example/task/CreateQueue.java
```diff
@@ -0,0 +1,51 @@
...
+import com.google.cloud.tasks.v2.Queue;
+import com.google.cloud.tasks.v2.QueueName;
+
+public class CreateQueue {
+  /**
+   * Create a queue using the Cloud Tasks client.
+   *
+   * @param projectId the Id of the project.
+   * @param queueId the name of your Queue.
+   * @param locationId the GCP region of your queue.
```
lesv, line 29: Typically, we have a main method that sets these values that users can modify to run this on their own.  Take a look at both our style guide and other smaples

aarjo: Added!

aarjo:
> Do you need to update the README.md to mention these?
> 
> Also, I don't see any RegionTags -- You'll eventually want to add those.

I'll update the README, and I believe there are region tags starting above the imports, no?

Also, what would be the best way to resolve the static analysis error? Making the snippet methods private means they can't be called in the testing classes.

lesv:
It's not required (or sometimes possible) to get all the static analysis errors.  You can google them spotBugs to see if any of them are easy to fix.  It's to get you to think about writing better code.

Yea - I just checked two files and didn't see them in the PR, but they were there.  Just tell me when you'd like me to merge.

PR #3208, opened 2020-06-18
samples: added UUID to prevent concurreny errors

munkhuushmgl:
Fixes the concurrency error which mentioned in #3206


munkhuushmgl:
@lesv Not sure how to fix these static analysis issues.

kurtisvg:
@munkhuushmgl The static analysis is advisory only - it's not needed to pass

PR #2949, opened 2020-05-20
feat: adding hadoopfs sample

bradmiro:
This sample submits a Hadoop FS command to a Dataproc cluster (-ls, -du, etc.)

PR #3262, opened 2020-06-25
feat: adding hadoopfs and autoscaling samples

bradmiro:
Adding a sample + test for submitting a HadoopFS command using the Java client library.

Adding a sample + test for creating a Dataproc cluster with autoscaling configured.

Follow-up to #2949


bradmiro:
Apologies for the delay / mess @lesv , this should be squared away now.

bradmiro:
Hmmm, still lint errors. Fixing...

PR #3283, opened 2020-06-30
samples: refactor dialogflow for 2.1.0

### LINKED ISSUE #3207: dialogflow/cloud-client should be rewritten to support v2
In #3130 upgrading to `com.google.cloud:google-cloud-dialogflow:2.0.0` there are many changes that break, so the sample should be rewritten.

@stephaniewang526 Could you find someone to work on this?

munkhuushmgl:
Fixes #3207 

I will refactor IntentMgmt.java and break them down into individual files in different PR.

### DIFF
--- a/dialogflow/cloud-client/src/main/java/com/example/dialogflow/DetectIntentAudio.java
+++ b/dialogflow/cloud-client/src/main/java/com/example/dialogflow/DetectIntentAudio.java
@@ -16,7 +16,7 @@
 
 package com.example.dialogflow;
 
-// Imports the Google Cloud client library
+// [START dialogflow_detect_intent_audio]
 
 import com.google.cloud.dialogflow.v2.AudioEncoding;
 import com.google.cloud.dialogflow.v2.DetectIntentRequest;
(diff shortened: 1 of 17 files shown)


dialogflow/cloud-client/src/main/java/com/example/dialogflow/DetectIntentAudio.java
```diff
@@ -27,33 +27,19 @@
...
   public static QueryResult detectIntentAudio(
-      String projectId,
-      String audioFilePath,
-      String sessionId,
-      String languageCode)
-      throws Exception {
+          String projectId,
+          String audioFilePath,
+          String sessionId,
+          String languageCode) throws IOException {
```
lesv, line 42: Looking at the code in googleapis/java-dialogflow searching for `throws` and `catch`, I see lots of specific enhanced exceptions - our samples should reflect the richness of those.  For example, [`ApiException`](http://googleapis.github.io/gax-java/1.7.1/apidocs/com/google/api/gax/rpc/ApiException.html) has 16 known subclasses that Gax throws.  gRPC also throws `io.grpc.StatusException` which has additional info to help developers understand the cause of their errors.  There is also `io.grpc.StatusRuntimeException` and `io.grpc.ManagedChannelProvider.ProviderNotFoundException`.

dialogflow/cloud-client/src/main/java/com/example/dialogflow/IntentManagement.java
```diff
@@ -18,6 +18,7 @@
 
 // Imports the Google Cloud client library
 
+import com.google.cloud.dialogflow.v2.AgentName;
```
lesv, line 21: I don't see any tags in this file - is that an issue?

dialogflow/cloud-client/src/test/java/com/example/dialogflow/DetectIntentKnowledgeTest.java
```diff
@@ -16,29 +16,30 @@
...
+import java.util.UUID;
+
+import static com.google.common.truth.Truth.assertThat;
+import static org.junit.Assert.assertEquals;
+
 
 @RunWith(JUnit4.class)
 @SuppressWarnings("checkstyle:abbreviationaswordinname")
-public class DetectIntentKnowledgeTest {
+public class qDetectIntentKnowledgeTest {
```
lesv, line 42: Is this intentional?

munkhuushmgl: Nope

dialogflow/cloud-client/src/main/java/com/example/dialogflow/IntentManagement.java
munkhuushmgl: There are multiple tags.
I will need to create CL before I make changes to IntentManagement.java.

I will create different PR which includes splitting IntentMgmt class into seperate files, updating to latest format and etc..

munkhuushmgl:
> Consider my comments on Exceptions, ok to ignore, but it would be nice if we were more specific so users knew what to expect.

Will this be sufficient?
```
  public static KnowledgeBase createKnowledgeBase(String projectId, String displayName)
      throws ApiException, IOException 
```

PR #3362, opened 2020-07-15
Add a simple example of a function built using Gradle.

eamonnmcmanus:


functions/helloworld/hello-http-gradle/gradle/wrapper/gradle-wrapper.properties
```diff
@@ -0,0 +1,5 @@
+distributionBase=GRADLE_USER_HOME
+distributionPath=wrapper/dists
+distributionUrl=https\://services.gradle.org/distributions/gradle-5.5.1-bin.zip
```
glaforge, line 3: What about going with a recent version of Gradle like the latest 6.5.1?

eamonnmcmanus:
> _I assume there's no practical way to combine this with an existing Maven-based sample?_

Not that I can see. We use Maven if there's a `pom.xml` in the root directory of the function, and Gradle if there's a `build.gradle`.

Can someone merge this for me?

PR #3432, opened 2020-07-29
Update renovate.json for Apache Beam packages

lesv:
May fix #3423 & #3424


lesv:
Probably won't fix, but it would be good if it's there anyway.

PR #3445, opened 2020-07-31
docs: add sample for setting timeout and retry settings

olavloite:
Adds an example for how to set custom timeout and retry settings for the Spanner client.

### DIFF
--- a/spanner/cloud-client/src/main/java/com/example/spanner/CustomTimeoutAndRetrySettingsExample.java
+++ b/spanner/cloud-client/src/main/java/com/example/spanner/CustomTimeoutAndRetrySettingsExample.java
@@ -1,0 +1,86 @@
+/*
+ * Copyright 2020 Google Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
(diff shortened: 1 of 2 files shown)


spanner/cloud-client/src/main/java/com/example/spanner/CustomTimeoutAndRetrySettingsExample.java
```diff
@@ -0,0 +1,86 @@
...
+                .setInitialRetryDelay(Duration.ofMillis(500))
+                .setMaxRetryDelay(Duration.ofSeconds(64))
+                .setRetryDelayMultiplier(1.5)
+
+                // Configure RPC and total timeout settings.
+                .setInitialRpcTimeout(Duration.ofSeconds(60))
+                .setMaxRpcTimeout(Duration.ofSeconds(60))
+                .setRpcTimeoutMultiplier(1.0)
+                .setTotalTimeout(Duration.ofSeconds(60))
+                .build());
```
lesv, line 64: Isn't something supposed to be set here?

lesv:
Odd - I changed the question before I sent the review and it sent the old question.  Please see in the PR.

spanner/cloud-client/src/main/java/com/example/spanner/CustomTimeoutAndRetrySettingsExample.java
olavloite: The problem here is that the `setRetryableCodes(..)` and `setRetrySettings(..)` methods return a `UnaryCallSettings.Builder` instance, and not a `SpannerOptions.Builder`, and that cannot (easily) be changed as it is part of the generated gapic client. What we are really doing here is exposing a part of the generated gapic client (builder) for the end user to be able to specify timeout and retry settings. That means that it is not possible to create the `SpannerOptions` instance in one single chain when this option is used, but that it must have at least one separate chain that sets the custom retry and/or timeout settings.

Adding the `.setProjectId(projectId)` to this group is possible, but I don't feel that it makes it more readable. I added a comment to clarify why the retry settings are set in a separate chain.

PR #3480, opened 2020-08-06
samples: Pub/Sub Spring

anguillanneuf:
Add code snippets for a Pub/Sub Spring user guide (similar to how they are done for [Spanner Spring](https://github.com/GoogleCloudPlatform/java-docs-samples/tree/master/spanner/spring-data)). 

Things worth mentioning:
- ~~`shared-configuration` is not added as a parent in `pom.xml` because `spring-boot-starter-parent`need to exist as a parent.~~
- No test are added because these will be used as code snippets. I also don't see tests in the Spanner samples.


pubsub/spring/src/main/java/demo/PubSubApplication.java
```diff
@@ -0,0 +1,149 @@
...
+        Flux.fromStream(
+                // Generate a stream that sends a numbered message every 10 seconds.
+                Stream.generate(
+                    new Supplier<Message<String>>() {
+                      @Override
+                      public Message<String> get() {
+                        try {
+                          Thread.sleep(10000);
+                        } finally {
+                          Random rand = new Random();
```
lesv, line 135: Ok, but shouldn't this be somewhere earlier - possibly as a classvar or static so it gets reused?

anguillanneuf: Sounds good. Updated.

anguillanneuf:
Thanks @lesv for the review! I modified the sample as suggested.

pubsub/spring/src/main/java/demo/PubSubApplication.java
```diff
@@ -0,0 +1,151 @@
...
+@SpringBootApplication
+public class PubSubApplication {
+
+  private static final Log LOGGER = LogFactory.getLog(PubSubApplication.class);
+
+  public static Message<String> generateOneMessagePeriodically(int millis) {
+    try {
+      Thread.sleep(millis);
+    } finally {
+      Random rand = new Random();
```
lesv, line 55: LOL - I was meaning consider making `rand` static or a class variable.  https://docs.oracle.com/javase/8/docs/api/java/util/Random.html

It's a bit heavyweight.

anguillanneuf: Oh sorry I misunderstood!

anguillanneuf:
@lesv I misunderstood your earlier comment. I'm also not authorized to merge.

PR #3562, opened 2020-08-18
fix(cloud-sql): Updated sample to include full JDBC url in comments.

shubha-rajan:
Fixes internal bug b/165099138.

### DIFF
--- a/cloud-sql/mysql/servlet/src/main/java/com/example/cloudsql/ConnectionPoolContextListener.java
+++ b/cloud-sql/mysql/servlet/src/main/java/com/example/cloudsql/ConnectionPoolContextListener.java
@@ -50,6 +50,12 @@ public class ConnectionPoolContextListener implements ServletContextListener {
     // The configuration object specifies behaviors for the connection pool.
     HikariConfig config = new HikariConfig();
 
+    // The following URL is equivalent to setting the config options below:
+    // jdbc:postgresql:///<DB_NAME>?cloudSqlInstance=<CLOUD_SQL_CONNECTION_NAME>&
+    // socketFactory=com.google.cloud.sql.postgres.SocketFactory&user=<DB_USER>&password=<DB_PASS>
+    // See the link below for more info on building a JDBC URL for the Cloud SQL JDBC Socket Factory
+    // https://github.com/GoogleCloudPlatform/cloud-sql-jdbc-socket-factory#creating-the-jdbc-url
+
     // Configure which instance and what database user to connect with.
     config.setJdbcUrl(String.format("jdbc:mysql:///%s", DB_NAME));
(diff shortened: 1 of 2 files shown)


cloud-sql/mysql/servlet/src/main/java/com/example/cloudsql/ConnectionPoolContextListener.java
```diff
@@ -51,6 +51,8 @@ private DataSource createConnectionPool() {
     HikariConfig config = new HikariConfig();
 
     // Configure which instance and what database user to connect with.
+    // Equivalent URL:
+    // jdbc:mysql:///<DATABASE_NAME>?cloudSqlInstance=<INSTANCE_CONNECTION_NAME>&socketFactory=com.google.cloud.sql.mysql.SocketFactory&user=<MYSQL_USER_NAME>&password=<MYSQL_USER_PASSWORD>
```
lesv, line 55: that's a long line, and doesn't actually represent what you have below.

cloud-sql/postgres/servlet/src/main/java/com/example/cloudsql/ConnectionPoolContextListener.java
```diff
@@ -51,6 +51,8 @@ private DataSource createConnectionPool() {
     HikariConfig config = new HikariConfig();
 
     // Configure which instance and what database user to connect with.
+    // Equivalent URL:
+    // jdbc:postgresql:///<DATABASE_NAME>?cloudSqlInstance=<INSTANCE_CONNECTION_NAME>&socketFactory=com.google.cloud.sql.postgres.SocketFactory&user=<POSTGRESQL_USER_NAME>&password=<POSTGRESQL_USER_PASSWORD>
```
lesv, line 55: ditto

shubha-rajan:
I moved the URL to after the data source property for socket factory was set. It's still really long though. The issue is that we're constructing that URL step by step, but I'm not sure if users realize that the partial URL we have in the sample is for connections that use the socket factory.

kurtisvg:
I would actually prefer the line to be above. The reason I asked for this change was because someone _didn't_ read the code and tried to copy L54 without the additional parameters, and filed feedback when it didn't work. I wanted to give a sneak peak of the full JDBC url in advance for the inpatient folks.

I'm also okay if we opt to just link to this section of the README for the full JDBC configurations: https://github.com/GoogleCloudPlatform/cloud-sql-jdbc-socket-factory#creating-the-jdbc-url

As long as it's next to the base url I'm happy.

lesv:
I'll leave it to your judgement.

shubha-rajan:
@kurtisvg moved it back to before the base URL. I'll merge after you approve

PR #3677, opened 2020-09-09
chore(functions/firebase): Add infinite-loop guard + test

ace-n:
As requested by @jskeet

(N.B: this isn't _actually_ broken on Cloud Functions itself, but it depends on undocumented behavior.)

### DIFF
--- a/functions/firebase/firestore-reactive/src/main/java/functions/FirebaseFirestoreReactive.java
+++ b/functions/firebase/firestore-reactive/src/main/java/functions/FirebaseFirestoreReactive.java
@@ -79,11 +79,17 @@ public class FirebaseFirestoreReactive implements RawBackgroundFunction {
 
     String affectedDoc = context.resource().split("/documents/")[1].replace("\"", "");
 
-    logger.info(String.format("Replacing value: %s --> %s", currentValue, newValue));
-    try {
-      FIRESTORE.document(affectedDoc).set(newFields, SetOptions.merge()).get();
-    } catch (ExecutionException | InterruptedException e) {
-      logger.log(Level.SEVERE, "Error updating Firestore document: " + e.getMessage(), e);
+    if (!currentValue.equals(newValue)) {
+      logger.info(String.format("Replacing value: %s --> %s", currentValue, newValue));
+      try {
+        FIRESTORE.document(affectedDoc).set(newFields, SetOptions.merge()).get();
+      } catch (ExecutionException | InterruptedException e) {
+        logger.log(Level.SEVERE, "Error updating Firestore document: " + e.getMessage(), e);
+      }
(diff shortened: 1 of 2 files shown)


functions/firebase/firestore-reactive/src/test/java/functions/FirebaseFirestoreReactiveTest.java
```diff
@@ -92,6 +92,25 @@ public void functionsFirebaseReactive_shouldCapitalizeOriginalValue()  {
...
+  @Test
+  public void functionsFirebaseReactive_shouldIgnoreCapitalizedValues()  {
+
+    String jsonStr = gson.toJson(Map.of("value",
+        Map.of("fields",
+            Map.of("original",
+                Map.of("stringValue", "FOO")))));
+
+    MockContext context = new MockContext();
+    context.resource = "projects/_/databases/(default)/documents/messages/ABCDE12345";
```
lesv, line 104: It would be good to explain this a bit.  I know it's in a test and not published, but it's not obvious.

functions/firebase/firestore-reactive/src/main/java/functions/FirebaseFirestoreReactive.java
```diff
@@ -79,11 +79,17 @@ public void accept(String json, Context context) {
 
     String affectedDoc = context.resource().split("/documents/")[1].replace("\"", "");
 
-    logger.info(String.format("Replacing value: %s --> %s", currentValue, newValue));
-    try {
-      FIRESTORE.document(affectedDoc).set(newFields, SetOptions.merge()).get();
-    } catch (ExecutionException | InterruptedException e) {
-      logger.log(Level.SEVERE, "Error updating Firestore document: " + e.getMessage(), e);
+    if (!currentValue.equals(newValue)) {
```
lesv, line 82: Consider a comment about why you are doing this.  Something as simple as "performance optimization" will do.

lesv (approved):
LGTM - I asked for some comments, ok not to.

functions/firebase/firestore-reactive/src/main/java/functions/FirebaseFirestoreReactive.java
ace-n: Is the comment in the `else` clause enough?

lesv: Since it's kind of important, it makes sense to say it early so they understand why the if is there.

ace-n: Comments reworded, @lesv PTAL?

PR #3689, opened 2020-09-10
chore(functions/slack): update Slack sample

ace-n:
- Use URL encoding instead of JSON for incoming (from Slack) requests
- Fix improperly formatted outgoing (to Slack) JSON response
- Move to official (and largely identical) Slack API package

cc @jskeet as FYI

### DIFF
--- a/functions/slack/src/main/java/functions/SlackSlashCommand.java
+++ b/functions/slack/src/main/java/functions/SlackSlashCommand.java
@@ -16,7 +16,6 @@
 
 package functions;
 
-import com.github.seratch.jslack.app_backend.SlackSignature;
 import com.google.api.client.googleapis.javanet.GoogleNetHttpTransport;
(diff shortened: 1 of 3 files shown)


functions/slack/src/main/java/functions/SlackSlashCommand.java
```diff
@@ -146,7 +153,10 @@ String formatSlackMessage(JsonObject kgResponse, String query) {
       addPropertyIfPresent(attachmentJson, "image_url", imageJson, "contentUrl");
     }
 
-    responseJson.add("attachments", attachmentJson);
+    JsonArray attachmentList = new JsonArray();
+    attachmentList.add(attachmentJson);
+
+    responseJson.add("attachments", attachmentList);
 
     return gson.toJson(responseJson);
```
lesv, line 161: didn't you get rid of this?

lesv (approved):

PR #3692, opened 2020-09-10
fix incorrect usage of String.format

elharo:
Not sure how many places this incorrect code is copied and pasted, but this is what I saw where I was working. 

Guess what? I just saw a second major bug in these lines. Message and not null object are swapped!


elharo:
going to try to fix this in one go as #3693

PR #3707, opened 2020-09-11
samples: add standalone numeric samples for Spanner

olavloite:
Adds standalone samples for the new `NUMERIC` data type for Spanner.

### DIFF
--- a/spanner/cloud-client/src/main/java/com/example/spanner/AddNumericColumnSample.java
+++ b/spanner/cloud-client/src/main/java/com/example/spanner/AddNumericColumnSample.java
@@ -1,0 +1,61 @@
+/*
+ * Copyright 2020 Google Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.example.spanner;
+
+//[START spanner_add_numeric_column]
+import com.google.api.gax.longrunning.OperationFuture;
+import com.google.cloud.spanner.DatabaseAdminClient;
+import com.google.cloud.spanner.Spanner;
+import com.google.cloud.spanner.SpannerOptions;
+import com.google.common.collect.ImmutableList;
+import com.google.spanner.admin.database.v1.UpdateDatabaseDdlMetadata;
+import java.util.concurrent.ExecutionException;
+
+class AddNumericColumnSample {
+
+  static void addNumericColumn() {
+    // TODO(developer): Replace these variables before running the sample.
+    String projectId = "my-project";
+    String instanceId = "my-instance";
+    String databaseId = "my-database";
+
+    try (Spanner spanner =
+        SpannerOptions.newBuilder().setProjectId(projectId).build().getService()) {
+      DatabaseAdminClient adminClient = spanner.getDatabaseAdminClient();
+      addNumericColumn(adminClient, instanceId, databaseId);
+    }
+  }
+
+  static void addNumericColumn(
+      DatabaseAdminClient adminClient, String instanceId, String databaseId) {
+    OperationFuture<Void, UpdateDatabaseDdlMetadata> operation =
+        adminClient.updateDatabaseDdl(
+            instanceId,
+            databaseId,
+            ImmutableList.of("ALTER TABLE Venues ADD COLUMN Revenue NUMERIC"),
+            null);
+    try {
+      operation.get();
+      System.out.printf("Successfully added column `Revenue`%n");
+    } catch (ExecutionException e) {
+      System.out.printf("Adding column `Revenue` failed: %s%n", e.getCause().getMessage());
+    } catch (InterruptedException e) {
+      System.out.printf("Adding column `Revenue` was interrupted%n");
+    }
+  }
+}
+// [END spanner_add_numeric_column]
--- a/spanner/cloud-client/src/main/java/com/example/spanner/AddNumericColumnSample.java
+++ b/spanner/cloud-client/src/main/java/com/example/spanner/AddNumericColumnSample.java
@@ -27,7 +27,7 @@ import java.util.concurrent.ExecutionException;
 
 class AddNumericColumnSample {
 
-  static void addNumericColumn() {
+  static void addNumericColumn() throws InterruptedException, ExecutionException {
     // TODO(developer): Replace these variables before running the sample.
     String projectId = "my-project";
     String instanceId = "my-instance";
@@ -41,21 +41,17 @@ class AddNumericColumnSample {
   }
 
   static void addNumericColumn(
-      DatabaseAdminClient adminClient, String instanceId, String databaseId) {
+      DatabaseAdminClient adminClient, String instanceId, String databaseId) throws InterruptedException, ExecutionException {
     OperationFuture<Void, UpdateDatabaseDdlMetadata> operation =
         adminClient.updateDatabaseDdl(
             instanceId,
             databaseId,
             ImmutableList.of("ALTER TABLE Venues ADD COLUMN Revenue NUMERIC"),
             null);
-    try {
-      operation.get();
-      System.out.printf("Successfully added column `Revenue`%n");
-    } catch (ExecutionException e) {
-      System.out.printf("Adding column `Revenue` failed: %s%n", e.getCause().getMessage());
-    } catch (InterruptedException e) {
-      System.out.printf("Adding column `Revenue` was interrupted%n");
-    }
+    // Wait for the operation to finish.
+    // This will throw an ExecutionException if the operation fails. 
+    operation.get();
+    System.out.printf("Successfully added column `Revenue`%n");
   }
 }
 // [END spanner_add_numeric_column]
--- a/spanner/cloud-client/src/main/java/com/example/spanner/AddNumericColumnSample.java
+++ b/spanner/cloud-client/src/main/java/com/example/spanner/AddNumericColumnSample.java
@@ -16,7 +16,7 @@
 
 package com.example.spanner;
 
-//[START spanner_add_numeric_column]
+// [START spanner_add_numeric_column]
 import com.google.api.gax.longrunning.OperationFuture;
 import com.google.cloud.spanner.DatabaseAdminClient;
 import com.google.cloud.spanner.Spanner;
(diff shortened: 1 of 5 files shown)


spanner/cloud-client/src/main/java/com/example/spanner/AddNumericColumnSample.java
```diff
@@ -0,0 +1,61 @@
...
+            ImmutableList.of("ALTER TABLE Venues ADD COLUMN Revenue NUMERIC"),
+            null);
+    try {
+      operation.get();
+      System.out.printf("Successfully added column `Revenue`%n");
+    } catch (ExecutionException e) {
+      System.out.printf("Adding column `Revenue` failed: %s%n", e.getCause().getMessage());
+    } catch (InterruptedException e) {
+      System.out.printf("Adding column `Revenue` was interrupted%n");
+    }
```
lesv, line 58: If you aren't handling exceptions, there is no reason to catch them.
See [Sample Style Guide](https://github.com/GoogleCloudPlatform/java-docs-samples/blob/master/SAMPLE_FORMAT.md#exception-handling).

lesv (approved):
Consider removing the catch in AddNumericColumnSample.java

spanner/cloud-client/src/main/java/com/example/spanner/AddNumericColumnSample.java
lesv: It might still be a good idea to mention they might occur in a comment.

skuruppu (approved):
Looks good. I also want @larkee to verify as well.

spanner/cloud-client/src/main/java/com/example/spanner/InsertNumericDataSample.java
```diff
@@ -0,0 +1,68 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.example.spanner;
+
+// [START spanner_update_data_with_numeric]
```
larkee, line 19: Updated tag to match the timestamp sample.

```suggestion
// [START spanner_update_data_with_numeric_column]
```

spanner/cloud-client/src/main/java/com/example/spanner/InsertNumericDataSample.java
```diff
@@ -0,0 +1,68 @@
...
+      insertNumericData(client);
+    }
+  }
+
+  static void insertNumericData(DatabaseClient client) {
+    client.write(
+        ImmutableList.of(
+            Mutation.newInsertOrUpdateBuilder("SingerRevenues")
+                .set("SingerId")
+                .to(1L)
```
larkee, line 49:
```suggestion
                .to(4L)
```

spanner/cloud-client/src/main/java/com/example/spanner/InsertNumericDataSample.java
```diff
@@ -0,0 +1,68 @@
...
+                .set("Revenue")
+                .to(new BigDecimal("35000"))
+                .build(),
+            Mutation.newInsertOrUpdateBuilder("SingerRevenues")
+                .set("SingerId")
+                .to(6L)
```
larkee, line 55:
```suggestion
                .to(19L)
```

spanner/cloud-client/src/main/java/com/example/spanner/InsertNumericDataSample.java
```diff
@@ -0,0 +1,68 @@
...
+                .set("Revenue")
+                .to(new BigDecimal("104500"))
+                .build(),
+            Mutation.newInsertOrUpdateBuilder("SingerRevenues")
+                .set("SingerId")
+                .to(14L)
```
larkee, line 61:
```suggestion
                .to(42L)
```

spanner/cloud-client/src/main/java/com/example/spanner/InsertNumericDataSample.java
```diff
@@ -0,0 +1,68 @@
...
+// [START spanner_update_data_with_numeric]
+import com.google.cloud.spanner.DatabaseClient;
+import com.google.cloud.spanner.DatabaseId;
+import com.google.cloud.spanner.Mutation;
+import com.google.cloud.spanner.Spanner;
+import com.google.cloud.spanner.SpannerOptions;
+import com.google.common.collect.ImmutableList;
+import java.math.BigDecimal;
+
+class InsertNumericDataSample {
```
larkee, line 28: References to Insert should be updated to Update to match the sample tag. The VenueId values should also be updated (as per the suggestions) to match existing rows in the table.

spanner/cloud-client/src/test/java/com/example/spanner/SpannerStandaloneExamplesIT.java
```diff
@@ -100,6 +119,57 @@ public void executeSqlWithCustomTimeoutAndRetrySettings_shouldWriteData() {
...
+    assertThat(out).contains("Records successfully inserted");
+  }
+
+  @Test
+  public void queryWithNumericParameter_shouldReturnResults() {
+    String projectId = spanner.getOptions().getProjectId();
+    DatabaseClient client =
+        spanner.getDatabaseClient(DatabaseId.of(projectId, instanceId, databaseId));
+    client.write(
+        ImmutableList.of(
```
larkee, line 149: Would the table not already having this entries from the previous test?

spanner/cloud-client/src/main/java/com/example/spanner/InsertNumericDataSample.java
```diff
@@ -0,0 +1,68 @@
...
+            Mutation.newInsertOrUpdateBuilder("SingerRevenues")
+                .set("SingerId")
+                .to(14L)
+                .set("Revenue")
+                .to(new BigDecimal("99999999999999999999999999999.99"))
+                .build()));
+    System.out.println("Records successfully inserted");
+  }
+}
+// [END spanner_update_data_with_numeric]
```
larkee, line 68:
```suggestion
// [END spanner_update_data_with_numeric_column]
```

larkee (changes requested):
LGTM overall. There were just a few updates to the spec that need to be reflected here.

spanner/cloud-client/src/main/java/com/example/spanner/AddNumericColumnSample.java
olavloite: Done.

spanner/cloud-client/src/main/java/com/example/spanner/InsertNumericDataSample.java
olavloite: Done.

spanner/cloud-client/src/main/java/com/example/spanner/InsertNumericDataSample.java
olavloite: Done

spanner/cloud-client/src/main/java/com/example/spanner/InsertNumericDataSample.java
olavloite: Done.

spanner/cloud-client/src/test/java/com/example/spanner/SpannerStandaloneExamplesIT.java
olavloite: No, JUnit (the Java test framework) does not guarantee the order in which tests are executed, and it is strongly recommended for this and other reasons to create test cases that are completely independent of all other test cases. So this test case cannot rely on any data already being present in the table, as there is no guarantee that this test case is executed after any of the other test cases.

(An exception to the above rule is normally made for heavy weight operations like creating the data model. That is created only once for the entire test file.)

spanner/cloud-client/src/main/java/com/example/spanner/InsertNumericDataSample.java
olavloite: Done.

spanner/cloud-client/src/main/java/com/example/spanner/InsertNumericDataSample.java
olavloite: Done.

spanner/cloud-client/src/main/java/com/example/spanner/InsertNumericDataSample.java
olavloite: Done.

lesv (approved):

spanner/cloud-client/src/main/java/com/example/spanner/UpdateNumericDataSample.java
```diff
@@ -0,0 +1,68 @@
...
+                .set("Revenue")
+                .to(new BigDecimal("104500"))
+                .build(),
+            Mutation.newInsertOrUpdateBuilder("Venues")
+                .set("VenueId")
+                .to(42L)
+                .set("Revenue")
+                .to(new BigDecimal("99999999999999999999999999999.99"))
+                .build()));
+    System.out.println("Records successfully updated");
```
larkee, line 65:
```suggestion
    System.out.println("Venues successfully updated");
```

spanner/cloud-client/src/test/java/com/example/spanner/SpannerStandaloneExamplesIT.java
```diff
@@ -100,6 +121,75 @@ public void executeSqlWithCustomTimeoutAndRetrySettings_shouldWriteData() {
...
+
+  @Test
+  public void updateNumericData_shouldWriteData() {
+    String projectId = spanner.getOptions().getProjectId();
+    String out =
+        runExample(
+            () ->
+                UpdateNumericDataSample.updateNumericData(
+                    spanner.getDatabaseClient(DatabaseId.of(projectId, instanceId, databaseId))));
+    assertThat(out).contains("Records successfully updated");
```
larkee, line 160:
```suggestion
    assertThat(out).contains("Venues successfully updated");
```

larkee (approved):
LGTM once the table name in the update output strings are updated :+1:

skuruppu:
It seems that the Static analysis failures are always happening due to the way the samples are written. @lesv would it be ok to merge in this PR?

Also, going forwards, @olavloite are there fixes that we could make such that we don't fail this Kokoro job?

lesv:
Static analysis often reports unfixable things.  I'll be trying to fix this (reduce the noise) tomorrow.  That said, it's not a required check, so feel free to submit.

lesv:
I submitted for you.  Just realized you never do it.

olavloite:
> Also, going forwards, @olavloite are there fixes that we could make such that we don't fail this Kokoro job?

@skuruppu Some are not really fixable, while others probably are. But what in general would be needed is to split the current `SpannerSamples` into individual files according to the most recent style guide, and then (once the unnecessary noise has been removed) fix the warnings that are relevant.

PR #3714, opened 2020-09-12
feat: add samples for numeric

olavloite:
Adds samples for the `NUMERIC` data type to the existing JDBC samples.

### DIFF
--- a/spanner/jdbc/src/main/java/com/example/spanner/jdbc/AutocommitUpdateDataExample.java
+++ b/spanner/jdbc/src/main/java/com/example/spanner/jdbc/AutocommitUpdateDataExample.java
@@ -45,8 +45,8 @@ class AutocommitUpdateDataExample {
       // The following statement will automatically be committed by the connection.
       int updateCount =
           statement.executeUpdate(
-              "INSERT INTO Singers (SingerId, FirstName, LastName)\n"
-                  + "VALUES (9999, 'Lloyd', 'Pineda')");
+              "INSERT INTO Singers (SingerId, FirstName, LastName, Revenues)\n"
+                  + "VALUES (9999, 'Lloyd', 'Pineda', 80100.29)");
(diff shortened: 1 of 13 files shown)


spanner/jdbc/src/main/java/com/example/spanner/jdbc/AutocommitUpdateDataExample.java
```diff
@@ -45,8 +45,8 @@ static void update(String projectId, String instanceId, String databaseId) throw
       // The following statement will automatically be committed by the connection.
       int updateCount =
           statement.executeUpdate(
-              "INSERT INTO Singers (SingerId, FirstName, LastName)\n"
-                  + "VALUES (9999, 'Lloyd', 'Pineda')");
+              "INSERT INTO Singers (SingerId, FirstName, LastName, Revenues)\n"
```
lesv, line 48: Why do you add the \n ?

lesv (approved):
LGTM - I don't understand why you are adding \n everywhere at the end of strings you are concatenating.  It seems superfluous.

skuruppu (approved):

spanner/jdbc/src/main/java/com/example/spanner/jdbc/AutocommitUpdateDataExample.java
olavloite: (It's not specifically added in this change, so I assume that it is more of a general question for all these samples)

No really specific reason, other than that it is quite customary to have the `VALUES (...)` clause on a new line. See for example the examples here: https://cloud.google.com/spanner/docs/dml-syntax#insert_examples

olavloite:
@lesv @skuruppu I'm apparently not authorized to merge these changes (anymore?). Would one of you do that?

lesv:
I merged, but I still don't understand why so many `\n`'s are present in the SQL

PR #3746, opened 2020-09-17
Fix pom.xml

MaxRossGoogle:
I was unable to get this sample to build using gcloud builds submit. Adding this Dockerfile didn't solve the whole problem, but it did solve part of it.

PR #3745, opened 2020-09-17
Create missing Dockerfile

MaxRossGoogle:
I was unable to get this sample to build using gcloud builds submit. Adding this Dockerfile didn't solve the whole problem, but it did solve part of it.

PR #3747, opened 2020-09-17
Add Dockerfile and fix pom.xml so that the sample builds with Cloud Build

MaxRossGoogle:
This sample was not building with Cloud Build. The two changes I've made allow it to build. I have limited Docker and Maven expertise. Please don't assume I know what I'm doing.

### DIFF
--- a/run/events-storage/Dockerfile
+++ b/run/events-storage/Dockerfile
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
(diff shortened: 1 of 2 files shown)


averikitsch:
Please see notes on #3746 and #3745

averikitsch:
More info: this sample is for the quickstart so we will use the Dockerfile. Reviewing this now.

averikitsch (approved):

PR #3769, opened 2020-09-21
Update static analysis image to include java 8 and 11 runtimes.

dzlier-gcp:
Fixes #issue

### DIFF
--- a/.kokoro/static_analysis/common.cfg
+++ b/.kokoro/static_analysis/common.cfg
@@ -19,11 +19,17 @@ gfile_resources: "/bigstore/cloud-devrel-kokoro-resources/trampoline"
 build_file: "java-docs-samples/.kokoro/trampoline.sh"
 env_vars: {
     key: "TRAMPOLINE_IMAGE"
-    value: "gcr.io/cloud-devrel-kokoro-resources/java8"
+    value: "gcr.io/cloud-devrel-kokoro-resources/java11"
 }
 env_vars: {
     key: "TRAMPOLINE_BUILD_FILE"
     value: "github/java-docs-samples/.kokoro/tests/run_static_analysis.sh"
+}
+
+# Set the JAVA VERSION env var.
+env_vars: {
+    key: "JAVA_VERSION"
+    value: "1.8,11"
 }
 
 # Access btlr binaries used in the tests
--- a/.kokoro/static_analysis/common.cfg
+++ b/.kokoro/static_analysis/common.cfg
@@ -26,12 +26,6 @@ env_vars: {
     value: "github/java-docs-samples/.kokoro/tests/run_static_analysis.sh"
 }
 
-# Set the JAVA VERSION env var.
-env_vars: {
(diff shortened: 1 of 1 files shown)


.kokoro/static_analysis/common.cfg
```diff
@@ -19,13 +19,19 @@ gfile_resources: "/bigstore/cloud-devrel-kokoro-resources/trampoline"
...
     key: "TRAMPOLINE_BUILD_FILE"
     value: "github/java-docs-samples/.kokoro/tests/run_static_analysis.sh"
 }
 
+# Set the JAVA VERSION env var.
+env_vars: {
+    key: "JAVA_VERSION"
+    value: "1.8,11"
+}
+
```
kurtisvg, line 34: The static analysis check doesn't use this: https://github.com/GoogleCloudPlatform/java-docs-samples/blob/master/.kokoro/tests/run_static_analysis.sh

kurtisvg (changes requested):

lesv (commented):
@dzlier-gcp This should have been discussed at the Java meeting today.

dzlier-gcp:
> @dzlier-gcp This should have been discussed at the Java meeting today.

I didn't think I'd be working in this repo, I just noticed that static analysis has been failing while I was resolving a Healthcare Datasets bug.

Should I close this for now for further discussion, or just remove the unused env var?

.kokoro/static_analysis/common.cfg
dzlier-gcp: Maven uses this though doesn't it?

dzlier-gcp: Or I guess static analysis doesn't run any mvn commands.

kurtisvg: Maven doesn't, there is something in the other script we used to control which versions are run:
https://github.com/GoogleCloudPlatform/java-docs-samples/blob/a683a732cff3d6833815461c2d59bf3f64c5c0a6/.kokoro/tests/run_tests.sh#L159-L164

kurtisvg (approved):

kurtisvg:
This is only controlling which JVM version the static analysis check is being run with. It seems fine to run it on Java 11 over Java 8, since some samples only support Java 11 or higher. This seems okay to merge unless I'm missing smething. 

It's likely still going to fail for a majority of the samples, as we discussed in the Java maintenance meeting the checks are overly aggressive (which Les is looking into).

dzlier-gcp:
> This is only controlling which JVM version the static analysis check is being run with. It seems fine to run it on Java 11 over Java 8, since some samples only support Java 11 or higher. This seems okay to merge unless I'm missing smething.
> 
> It's likely still going to fail for a majority of the samples, as we discussed in the Java maintenance meeting the checks are overly aggressive (which Les is looking into).

All right, I need a new approval after removing that env var though.

dzlier-gcp:
Oh nvm

PR #3770, opened 2020-09-21
Determine pubsub event timestamp correctly in retry-timeout.

### LINKED ISSUE #3734: Java retry timeout should use context.timestamp to test event age
## In which file did you encounter the issue?
https://github.com/GoogleCloudPlatform/java-docs-samples/blob/master/functions/concepts/retry-timeout/src/main/java/functions/RetryTimeout.java

### Did you change the file? If so, how?
This is how I'm checking the event age.
```
@Override
  public void accept(PubSubMessage message, Context context) {
    ZonedDateTime utcNow = ZonedDateTime.now(ZoneOffset.UTC);
    ZonedDateTime timestamp = ZonedDateTime.parse(context.timestamp());
    long eventAge = Duration.between(timestamp, utcNow).toMillis();

    // Ignore events that are too old
    if (eventAge > MAX_EVENT_AGE) {
      logger.info(String.format("Dropping event %s.", context.eventId()));
      return;
    }

    // Process events that are recent enough
    // To retry this invocation, throw an exception here
    logger.info(String.format("Processing event %s.", context.eventId()));
  }
```

## Describe the issue
Using Java, I've recently tried to implement a background [Cloud Function retry timeout](https://github.com/GoogleCloudPlatform/java-docs-samples/blob/master/functions/concepts/retry-timeout/src/main/java/functions/RetryTimeout.java) to avoid infinite retries. I've implemented the same check in Go and Javascript in the past. I noticed that the Java sample differs from the Go and Javascript samples. Go and Javascript both base the event duration on the `timestamp` property of the `context` object. I believe this is the correct way to determine the age of the event pubsub. The Java sample attempts to determine the event age using `message.getData()` which holds user defined data. There is no timestamp in the message data unless the users adds it.

eamonnmcmanus:
In the retry-timeout sample for GCF java11, we should determine
the event timestamp from the Context parameter, not from properties
of the passed-in PubSub object.

Fixes #3734.

### DIFF
--- a/functions/concepts/retry-timeout/src/main/java/functions/RetryTimeout.java
+++ b/functions/concepts/retry-timeout/src/main/java/functions/RetryTimeout.java
@@ -21,7 +21,6 @@ package functions;
 import com.google.cloud.functions.BackgroundFunction;
 import com.google.cloud.functions.Context;
 import com.google.gson.Gson;
-import com.google.gson.JsonObject;
 import functions.eventpojos.PubSubMessage;
 import java.time.Duration;
 import java.time.ZoneOffset;
@@ -42,25 +41,19 @@ public class RetryTimeout implements BackgroundFunction<PubSubMessage> {
   @Override
   public void accept(PubSubMessage message, Context context) {
     ZonedDateTime utcNow = ZonedDateTime.now(ZoneOffset.UTC);
-    ZonedDateTime timestamp = utcNow;
+    ZonedDateTime timestamp = ZonedDateTime.parse(context.timestamp());
 
-    String data = message.getData();
-    JsonObject body = gson.fromJson(data, JsonObject.class);
-    if (body != null && body.has("timestamp")) {
-      String tz = body.get("timestamp").getAsString();
-      timestamp = ZonedDateTime.parse(tz);
-    }
     long eventAge = Duration.between(timestamp, utcNow).toMillis();
 
     // Ignore events that are too old
     if (eventAge > MAX_EVENT_AGE) {
-      logger.info(String.format("Dropping event %s.", data));
+      logger.info(String.format("Dropping event with timestamp %s.", timestamp));
       return;
     }
 
     // Process events that are recent enough
     // To retry this invocation, throw an exception here
-    logger.info(String.format("Processing event %s.", data));
+    logger.info(String.format("Processing event with timestamp %s.", timestamp));
   }
 }
 // [END functions_tips_infinite_retries]
--- a/functions/concepts/retry-timeout/src/test/java/functions/RetryTimeoutTest.java
+++ b/functions/concepts/retry-timeout/src/test/java/functions/RetryTimeoutTest.java
@@ -17,14 +17,16 @@
 package functions;
 
 import static com.google.common.truth.Truth.assertThat;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
 
+import com.google.cloud.functions.Context;
 import com.google.common.testing.TestLogHandler;
 import com.google.gson.Gson;
 import functions.eventpojos.PubSubMessage;
 import java.time.Instant;
 import java.time.ZoneOffset;
 import java.time.ZonedDateTime;
-import java.util.Map;
 import java.util.logging.Logger;
 import org.junit.After;
 import org.junit.Before;
@@ -32,7 +34,6 @@ import org.junit.BeforeClass;
 import org.junit.Test;
 import org.junit.runner.RunWith;
 import org.junit.runners.JUnit4;
-import org.mockito.MockitoAnnotations;
 
 @RunWith(JUnit4.class)
 public class RetryTimeoutTest {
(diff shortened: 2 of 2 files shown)


functions/concepts/retry-timeout/src/main/java/functions/RetryTimeout.java
```diff
@@ -42,25 +41,19 @@
   @Override
   public void accept(PubSubMessage message, Context context) {
     ZonedDateTime utcNow = ZonedDateTime.now(ZoneOffset.UTC);
-    ZonedDateTime timestamp = utcNow;
+    ZonedDateTime timestamp = ZonedDateTime.parse(context.timestamp());
```
lesv, line 44: Shouldn't there be some kind of comment discussing these two and why they are different and when you want to use one and not the other?

lesv (changes requested):
Great - shouldn't we explain it to the user?

functions/concepts/retry-timeout/src/main/java/functions/RetryTimeout.java
ace-n: @lesv I suspect @eamonnmcmanus meant to delete the `utcNow` declaration?

functions/concepts/retry-timeout/src/main/java/functions/RetryTimeout.java
```diff
@@ -42,25 +41,19 @@
   @Override
   public void accept(PubSubMessage message, Context context) {
     ZonedDateTime utcNow = ZonedDateTime.now(ZoneOffset.UTC);
```
ace-n, line 43: @eamonnmcmanus did you mean to delete this?

functions/concepts/retry-timeout/src/test/java/functions/RetryTimeoutTest.java
```diff
@@ -65,41 +64,29 @@ public void afterTest() {
...
-    new RetryTimeout().accept(pubsubMessage, null);
+    new RetryTimeout().accept(pubsubMessage, mockContext);
 
     String logMessage = LOG_HANDLER.getStoredLogRecords().get(0).getMessage();
-    assertThat(String.format("Processing event %s.", timestampData)).isEqualTo(logMessage);
+    assertThat(logMessage).contains("Processing event with timestamp " + timestamp);
   }
 
   @Test
   public void retryTimeout_handlesStopMsg() {
```
ace-n, line 80: Should this test contain an assertion? (I don't think it does right now.)

ace-n (changes requested):

functions/concepts/retry-timeout/src/main/java/functions/RetryTimeout.java
lesv: Still, there are two sources of truth, and we got it wrong the first time, so that deserves an explanation, since other users are likely to make the same error unless they are told why.

functions/concepts/retry-timeout/src/main/java/functions/RetryTimeout.java
ace-n: Edit: nvm, I see you used it below.

functions/concepts/retry-timeout/src/main/java/functions/RetryTimeout.java
ace-n: Ah - I misread, he's comparing `utcNow` to the parsed timestamp. (`utcNow` was previously the _default timestamp value_ in case one wasn't included in the Pub/Sub message.)

This LGTM. ✅

eamonnmcmanus: I think the situation is pretty confusing here. I don't know under what circumstances, if any, the `messageId` and `publishTime` properties of the payload are set. I just deployed a pub/sub function and sent a message to it, and both these fields were null. We may have a much bigger cleanup to do to get this straight, though hopefully the migration to CloudEvents should make things more consistent.

Anyway, let's get this fix in for now, and try to improve things further afterwards.

functions/concepts/retry-timeout/src/test/java/functions/RetryTimeoutTest.java
eamonnmcmanus: If you're looking at the diffs it can be confusing, since I deleted the empty-payload test. The two remaining test methods are basically identical apart from the timestamp, and the assertion is on the last line of each.

functions/concepts/retry-timeout/src/main/java/functions/RetryTimeout.java
lesv: k - I'd still rather you commented about it in the source, but it's up to you.

lesv (approved):

lesv:
I'm going to dismiss @ace-n review based on @eamonnmcmanus comments.  I'd like to see comments, hopefully that will get fixed in the next PR.

PR #3831, opened 2020-09-25
docs: update set custom timeout sample

### LINKED ISSUE #3805: spanner: update the code sample to set timeout at method level
The current code sample of setting the custom timeout is based on the client level. We recently have added the ability to set the timeout at the method level: https://github.com/googleapis/java-spanner/pull/379. Are we able to update the existing code sample?

olavloite:
Updates the 'Set custom timeout and retry settings' sample to one that uses the possibility of setting a specific timeout for one method call. This does however not support setting retry settings.

Fixes #3805

### DIFF
--- a/spanner/cloud-client/pom.xml
+++ b/spanner/cloud-client/pom.xml
@@ -53,6 +53,7 @@ limitations under the License.
     <dependency>
       <groupId>com.google.cloud</groupId>
       <artifactId>google-cloud-spanner</artifactId>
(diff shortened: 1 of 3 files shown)


lesv:
@skuruppu Should we abandon this PR?

skuruppu:
It seems that way. @olavloite if you can confirm and close this PR, that would be great.

olavloite:
Yep, agreed.

PR #3832, opened 2020-09-25
fix: remove static analysis job

elharo:
Let's see if this will pass.


kurtisvg (changes requested):
The decision in the last Java samples maintenance meeting was that we were going to leave these checks and @lesv was going to look into making them more useful. 

We shouldn't be unilaterally making changes like this without giving a chance for discussion from the  wider group.

lesv:
Hopefully this will be improved by #3833 -- it probably won't easily pass.

elharo:
This isn't a unilateral change. It's a PR sent for review so the wider group can see it. The decision at the meeting was that the problem was going to be fixed on Tuesday. That didn't happen. I'm going to repush this PR since it got closed before I could get the information I needed from it.

PR #3834, opened 2020-09-25
Turn off static analysis CI

elharo:
One more time...


lesv:
This wasn't what we agreed on for Monday - wait for 1.0.20 shared-configuration to pass in #3833

PR #3884, opened 2020-10-04
Wordcount on Dataproc WORKS! (No more NPEs)

jaceklaskowski:
Fixes NPE when submitting Wordcount to Dataproc

/cc @kolea2 @igorbernstein2

### DIFF
--- a/bigtable/spark/build.sbt
+++ b/bigtable/spark/build.sbt
@@ -18,6 +18,8 @@ name := "bigtable-spark-samples"
 
 version := "0.1"
 
+resolvers += Resolver.mavenLocal
+
 // Versions to match Dataproc 1.4
 // https://cloud.google.com/dataproc/docs/concepts/versioning/dataproc-release-1.4
 scalaVersion := "2.11.12"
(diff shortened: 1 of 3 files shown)


bigtable/spark/build.sbt
```diff
@@ -18,6 +18,8 @@ name := "bigtable-spark-samples"
 
 version := "0.1"
 
+resolvers += Resolver.mavenLocal
```
lesv, line 21: Why are you asking for Maven Local?  is there another step that loads things into it?

lesv (approved):
Please answer the question before merging.

bigtable/spark/build.sbt
jaceklaskowski: Nice catch! It's to speed up dependency resolution when a user uses maven (most likely) or gradle as well as sbt.

The main reason was to let me debug an issue with a dependency that I built locally.

I thought I'd remove it, but decided not to since it may speed things up while resolving dependencies. I'm fine either way.

lesv: You probably want to remove it, as it will look there anyway.

jaceklaskowski: > it will look there anyway.

Nope. It will not (but can) as described in https://www.scala-sbt.org/1.x/docs/Library-Dependencies.html#Resolvers:

> sbt can search your local Maven repository if you add it as a repository

PR #4038, opened 2020-10-20
chore(testing): remove static analysis checks

kurtisvg:


lesv (approved):

lesv:
But won't it still be invoked?

PR #4048, opened 2020-10-21
clean up grammar and spelling

elharo:
@lesv  Most notably, Java has "method arguments" instead of "function parameters", but overall good advice.


averikitsch (approved):
Thanks!

PR #4197, opened 2020-11-08
feat: use LazySpannerInitializer instead of DCL

### LINKED ISSUE #2862: Functions - Java Spanner example uses DCL
## Describe the issue

GCF Java 11 [Spanner sample]( https://github.com/GoogleCloudPlatform/java-docs-samples/blob/master/functions/spanner/src/main/java/functions/HelloSpanner.java#L59) uses Double checked locking.  The pattern is known to be broken, should follow the [lazy field init example](https://github.com/GoogleCloudPlatform/java-docs-samples/blob/master/functions/concepts/lazy-fields/src/main/java/functions/LazyFields.java)

olavloite:
The Spanner client library now includes a lazy initializer for Spanner instances that can be used with Google Cloud Functions and other libraries that want to create an instance the first time one is needed and reuse this instance for subsequent requests.

Fixes #2862

### DIFF
--- a/functions/spanner/src/main/java/functions/HelloSpanner.java
+++ b/functions/spanner/src/main/java/functions/HelloSpanner.java
@@ -23,8 +23,8 @@ import com.google.cloud.functions.HttpRequest;
 import com.google.cloud.functions.HttpResponse;
 import com.google.cloud.spanner.DatabaseClient;
 import com.google.cloud.spanner.DatabaseId;
+import com.google.cloud.spanner.LazySpannerInitializer;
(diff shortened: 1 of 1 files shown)


olavloite:
cc @saturnism

thiagotnunes (approved):
LGTM

saturnism:
thanks! /cc @ace-n

ace-n (approved):
I like how much cleaner this looks 👍 

(I don't believe I own the `spanner` GCF samples though, since they're not "strictly for GCF".)

lesv (approved):

PR #4352, opened 2020-12-07
chore: updates to Cloud SQL samples

shubha-rajan:
- Added example connection URL to SQL Server sample
- Added instructions for forcing connection with Private IP to MySQL and Postgres samples
- Updated renovate.json to only update driver if new jre8 versions are released.


renovate.json
```diff
@@ -42,6 +42,10 @@
     {
       "packageNames": ["javax.servlet:javax.servlet-api"],
       "rangeStrategy": "pin"
+    },
+    {
+      "packageNames": ["com.microsoft.sqlserver:mssql-jdbc"],
```
kurtisvg, line 47: I think a PR with this was already merged?

kurtisvg (approved):

renovate.json
shubha-rajan: Ah, my branch is stale. I'll rebase it.

PR #4448, opened 2020-12-15
Pub/Sub: add --region to command to start dataflow job

anguillanneuf:
Fixes internal b/164138878.

### DIFF
--- a/pubsub/streaming-analytics/README.md
+++ b/pubsub/streaming-analytics/README.md
@@ -64,7 +64,7 @@ Sample(s) showing how to use [Google Cloud Pub/Sub] with [Google Cloud Dataflow]
 1. Set `REGION` to a Dataflow [regional endpoint].
 
    ```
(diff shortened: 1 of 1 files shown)


pubsub/streaming-analytics/README.md
```diff
@@ -102,6 +103,7 @@ mvn compile exec:java \
   -Dexec.cleanupDaemonThreads=false \
   -Dexec.args="\
     --project=$PROJECT_NAME \
+    --region=us-central1 \
```
kolea2, line 106: what are your thoughts on instead of hardcoding a value here, make it `$REGION`? Since you are linking to the list of regions, it may be easier to allow the user to choose from there?

anguillanneuf: Good idea! Fixed.

stephaniewang526 (approved):

kolea2 (approved):

PR #4551, opened 2021-01-14
Use spring-boot-starter-data-r2dbc and version manage as little dependencies as possible

breun:
The PR replaces the dependencies on `spring-data-r2dbc` and `r2dbc-pool` with `spring-boot-starter-data-r2dbc` (already version managed by Spring Boot), which includes those dependencies.

This PR also removes other version numbers for dependencies which are already version managed via Spring Boot to simplify and future-proof the `pom.xml` file.

### DIFF
--- a/cloud-sql/r2dbc/pom.xml
+++ b/cloud-sql/r2dbc/pom.xml
@@ -36,21 +36,14 @@
       <artifactId>spring-boot-starter-thymeleaf</artifactId>
     </dependency>
     <dependency>
-      <groupId>org.springframework.data</groupId>
-      <artifactId>spring-data-r2dbc</artifactId>
-      <version>1.2.3</version>
-    </dependency>
-    <dependency>
-      <groupId>io.r2dbc</groupId>
-      <artifactId>r2dbc-pool</artifactId>
-      <version>0.8.5.RELEASE</version>
+      <groupId>org.springframework.boot</groupId>
+      <artifactId>spring-boot-starter-data-r2dbc</artifactId>
     </dependency>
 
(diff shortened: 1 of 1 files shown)


cloud-sql/r2dbc/pom.xml
```diff
@@ -36,21 +36,14 @@
...
-      <groupId>org.springframework.data</groupId>
-      <artifactId>spring-data-r2dbc</artifactId>
-      <version>1.2.3</version>
-    </dependency>
-    <dependency>
-      <groupId>io.r2dbc</groupId>
-      <artifactId>r2dbc-pool</artifactId>
-      <version>0.8.5.RELEASE</version>
+      <groupId>org.springframework.boot</groupId>
+      <artifactId>spring-boot-starter-data-r2dbc</artifactId>
```
shubha-rajan, line 40: should we have a version here for `spring-boot-starter-data-r2dbc`?

shubha-rajan (commented):
Thanks for the PR. This will help prevent any possible incompatibilities between Spring Boot and our r2dbc package versions. Please see my question about versioning for `spring-boot-starter-data-r2dbc`

cloud-sql/r2dbc/pom.xml
breun: Spring Boot provides dependency management for Spring Data artifacts, so unless there is a specific reason to deviate from the version provided by Spring Boot it is recommended to not manage that version yourself in your project.

shubha-rajan: sounds good.

shubha-rajan (approved):

PR #4821, opened 2021-03-03
feat: add Cloud SQL samples for Client-side Encryption

shubha-rajan:
Add samples for using Tink and Cloud KMS to encrypt data client-side

- [x ] I have followed [Sample Format Guide](https://github.com/GoogleCloudPlatform/java-docs-samples/blob/master/SAMPLE_FORMAT.md)

### DIFF
--- a/cloud-sql/mysql/client-side-encryption/README.md
+++ b/cloud-sql/mysql/client-side-encryption/README.md
@@ -6,9 +6,6 @@
 maven utilities) by following the [java setup guide](https://cloud.google.com/java/docs/setup) and
 [create a project](https://cloud.google.com/resource-manager/docs/creating-managing-projects#creating_a_project).
 
-1. If you haven't already, set up a Python Development Environment by following the [python setup guide](https://cloud.google.com/python/setup) and 
-[create a project](https://cloud.google.com/resource-manager/docs/creating-managing-projects#creating_a_project).
-
 1. Create a 2nd Gen Cloud SQL Instance by following these 
 [instructions](https://cloud.google.com/sql/docs/mysql/create-instance). Note the connection string,
 database user, and database password that you create.
--- a/cloud-sql/mysql/client-side-encryption/README.md
+++ b/cloud-sql/mysql/client-side-encryption/README.md
@@ -1,0 +1,43 @@
+# Encrypting fields in Cloud SQL - MySQL with Tink
+
+## Before you begin
+
+1. If you haven't already, set up a Java Development Environment (including google-cloud-sdk and
+maven utilities) by following the [java setup guide](https://cloud.google.com/java/docs/setup) and
+[create a project](https://cloud.google.com/resource-manager/docs/creating-managing-projects#creating_a_project).
+
+1. If you haven't already, set up a Python Development Environment by following the [python setup guide](https://cloud.google.com/python/setup) and 
+[create a project](https://cloud.google.com/resource-manager/docs/creating-managing-projects#creating_a_project).
+
+1. Create a 2nd Gen Cloud SQL Instance by following these 
+[instructions](https://cloud.google.com/sql/docs/mysql/create-instance). Note the connection string,
+database user, and database password that you create.
+
+1. Create a database for your application by following these 
+[instructions](https://cloud.google.com/sql/docs/mysql/create-manage-databases). Note the database
+name.
+
+1. Create a KMS key for your application by following these
+[instructions](https://cloud.google.com/kms/docs/creating-keys). Copy the resource name of your
+created key.
+
+1. Create a service account with the 'Cloud SQL Client' permissions by following these 
+[instructions](https://cloud.google.com/sql/docs/mysql/connect-external-app#4_if_required_by_your_authentication_method_create_a_service_account).
+Then, add the 'Cloud KMS CryptoKey Encrypter/Decrypter' permission for the key to your service account 
+by following these [instructions](https://cloud.google.com/kms/docs/iam).
+
+## Running Locally
+
+Before running, copy the `example.envrc` file to `.envrc` and replace the values for 
+`GOOGLE_APPLICATION_CREDENTIALS`, `DB_USER`, `DB_PASS`, `DB_NAME`, `CLOUD_SQL_CONNECTION_NAME`,
+and `CLOUD_KMS_URI` with the values from your project. Then run `source .envrc` or optionally use 
+[direnv](https://direnv.net/).
+
+Once the environment variables have been set, run:
+```
+mvn exec:java -Dexec.mainClass=cloudsql.tink.EncryptAndInsertData
+```
+and 
+```
+mvn exec:java -Dexec.mainClass=cloudsql.tink.QueryAndDecryptData
+```
--- a/cloud-sql/mysql/client-side-encryption/pom.xml
+++ b/cloud-sql/mysql/client-side-encryption/pom.xml
@@ -1,0 +1,115 @@
+<!--
+  Copyright 2021 Google LLC
+
+  Licensed under the Apache License, Version 2.0 (the "License");
+  you may not use this file except in compliance with the License.
+  You may obtain a copy of the License at
+
+       http://www.apache.org/licenses/LICENSE-2.0
+
+  Unless required by applicable law or agreed to in writing, software
+  distributed under the License is distributed on an "AS IS" BASIS,
+  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  See the License for the specific language governing permissions and
+  limitations under the License.
+-->
+<project>
+  <modelVersion>4.0.0</modelVersion>
+  <packaging>jar</packaging>
+  <version>1.0-SNAPSHOT</version>
+  <groupId>com.example.cloudsql</groupId>
+  <artifactId>cloud-sql-tink-mysql</artifactId>
+  <name>Cloud SQL Client Side Encryption Samples</name>
+
+  <!--
+    The parent pom defines common style checks and testing strategies for our samples.
+    Removing or replacing it should not affect the execution of the samples in anyway.
+  -->
+  <parent>
+    <groupId>com.google.cloud.samples</groupId>
+    <artifactId>shared-configuration</artifactId>
+    <version>1.0.21</version>
+  </parent>
+
+  <properties>
+    <maven.compiler.target>1.8</maven.compiler.target>
+    <maven.compiler.source>1.8</maven.compiler.source>
+    <failOnMissingWebXml>false</failOnMissingWebXml>
+  </properties>
+
+  <dependencyManagement>
+    <dependencies>
+      <dependency>
+        <groupId>com.google.api-client</groupId>
+        <artifactId>google-api-client</artifactId>
+        <version>1.31.3</version>
+      </dependency>
+      <dependency>
+        <groupId>com.google.guava</groupId>
+        <artifactId>guava</artifactId>
+        <version>30.1-android</version>
+      </dependency>
+      <dependency>
+        <groupId>com.google.http-client</groupId>
+        <artifactId>google-http-client</artifactId>
+        <version>1.39.0</version>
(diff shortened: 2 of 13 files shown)


cloud-sql/mysql/client-side-encryption/README.md
```diff
@@ -0,0 +1,43 @@
+# Encrypting fields in Cloud SQL - MySQL with Tink
+
+## Before you begin
+
+1. If you haven't already, set up a Java Development Environment (including google-cloud-sdk and
+maven utilities) by following the [java setup guide](https://cloud.google.com/java/docs/setup) and
+[create a project](https://cloud.google.com/resource-manager/docs/creating-managing-projects#creating_a_project).
+
+1. If you haven't already, set up a Python Development Environment by following the [python setup guide](https://cloud.google.com/python/setup) and 
+[create a project](https://cloud.google.com/resource-manager/docs/creating-managing-projects#creating_a_project).
```
kurtisvg, line 10: 👀

cloud-sql/mysql/client-side-encryption/pom.xml
```diff
@@ -0,0 +1,102 @@
...
+  <parent>
+    <groupId>com.google.cloud.samples</groupId>
+    <artifactId>shared-configuration</artifactId>
+    <version>1.0.21</version>
+  </parent>
+
+  <properties>
+    <maven.compiler.target>1.8</maven.compiler.target>
+    <maven.compiler.source>1.8</maven.compiler.source>
+    <failOnMissingWebXml>false</failOnMissingWebXml>
```
kurtisvg, line 37: nit: Does this need to be here?

cloud-sql/mysql/client-side-encryption/pom.xml
```diff
@@ -0,0 +1,102 @@
...
+        <artifactId>google-api-client</artifactId>
+        <version>1.31.3</version>
+      </dependency>
+    </dependencies>
+  </dependencyManagement>
+
+  <dependencies>
+    <dependency>
+      <groupId>com.google.http-client</groupId>
+      <artifactId>google-http-client-jackson2</artifactId>
```
kurtisvg, line 60: nit: What is this dependency used for?

cloud-sql/mysql/client-side-encryption/src/main/java/cloudsql/tink/EncryptAndInsertData.java
```diff
@@ -0,0 +1,104 @@
...
+public class EncryptAndInsertData {
+
+  public static void main(String[] args) throws GeneralSecurityException, SQLException {
+    // Saving credentials in environment variables is convenient, but not secure - consider a more
+    // secure solution such as Cloud Secret Manager to help keep secrets safe.
+    String dbUser = System.getenv("DB_USER"); // e.g. "root", "mysql"
+    String dbPass = System.getenv("DB_PASS");
+    String dbName = System.getenv("DB_NAME");
+    String cloudSqlConnectionName = System.getenv("CLOUD_SQL_CONNECTION_NAME");
+    String kmsUri = System.getenv("CLOUD_KMS_URI");
```
kurtisvg, line 41: nit: Can we add comment examples for all of these?

cloud-sql/mysql/client-side-encryption/src/main/java/cloudsql/tink/CloudKmsEnvelopeAead.java
```diff
@@ -0,0 +1,53 @@
...
+import java.security.GeneralSecurityException;
+
+public class CloudKmsEnvelopeAead {
+
+  public Aead envAead;
+
+  public CloudKmsEnvelopeAead(String kmsUri) throws GeneralSecurityException {
+    AeadConfig.register();
+    envAead = getEnvelopeAead(kmsUri);
+  }
```
kurtisvg, line 37: I'm not sure I understand this part - any reason not to just use the static function?

cloud-sql/mysql/client-side-encryption/src/main/java/cloudsql/tink/EncryptAndInsertData.java
```diff
@@ -0,0 +1,104 @@
...
+    String team = "TABS";
+    String tableName = "votes";
+    String email = "hello@example.com";
+
+    // Initialize database connection pool and create table
+    DataSource pool = createConnectionPool(dbUser, dbPass, dbName, cloudSqlConnectionName);
+    createTable(pool, tableName);
+
+    // Initialize envelope AEAD
+    Aead envAead = new CloudKmsEnvelopeAead(kmsUri).envAead;
```
kurtisvg, line 52: This is a super intriguing way to set this up. My one concern is that this pretty much assumes that the user is only going to run one (the last one?). 

I think perhaps since they are all on the same page, it's not unreasonable to expect the user to run them in order. What do you think about another snippet for setting up the database connection/schema, and then maybe a comment that points users to the previous set up?

kurtisvg (changes requested):

cloud-sql/mysql/client-side-encryption/README.md
shubha-rajan: Oh no.. my copy-pasting has been exposed! Fixed in latest commit.

cloud-sql/mysql/client-side-encryption/pom.xml
shubha-rajan: Nope! Another copy-paste remnant I forgot to delete. Removed.

cloud-sql/mysql/client-side-encryption/pom.xml
shubha-rajan: So this is because of a ClassNotFound Exception that shows up due to what I think is a dependency conflict between Tink and the SQLAdmin API with `google-api-client`. The first error I got was that SQLAdmin needed at least version 1.31.1, and Tink used 1.26 or something. After pinning the later version in the dependencies I got
```
java.lang.ClassNotFoundException: com.google.api.client.json.jackson2.JacksonFactory
```
Directly adding jackson2 to the dependencies seems to fix it.

cloud-sql/mysql/client-side-encryption/src/main/java/cloudsql/tink/EncryptAndInsertData.java
shubha-rajan: Done.

cloud-sql/mysql/client-side-encryption/src/main/java/cloudsql/tink/CloudKmsEnvelopeAead.java
shubha-rajan: Using the static function works. Updated it in the last set of commits.

I think it's just like this because I originally had the KMS URI as an attribute of the class as well, but then I took it out.

cloud-sql/mysql/client-side-encryption/src/main/java/cloudsql/tink/EncryptAndInsertData.java
shubha-rajan: I moved the database setup code to its own class and think it looks a lot cleaner without the duplicated code in both classes. Would we want another region tag for that code, or would we just point to the connectivity samples there?

cloud-sql/mysql/client-side-encryption/src/main/java/cloudsql/tink/CloudKmsEnvelopeAead.java
```diff
@@ -0,0 +1,47 @@
...
+import com.google.crypto.tink.KmsClients;
+import com.google.crypto.tink.aead.AeadConfig;
+import com.google.crypto.tink.aead.AeadKeyTemplates;
+import com.google.crypto.tink.integration.gcpkms.GcpKmsClient;
+import com.google.crypto.tink.proto.KeyTemplate;
+import java.security.GeneralSecurityException;
+
+public class CloudKmsEnvelopeAead {
+
+  public static Aead getEnvelopeAead(String kmsUri) throws GeneralSecurityException {
```
kurtisvg, line 32: nit: You could probably just change to this to `get(` to avoid stutter in the names.

kurtisvg (approved):
Look great. I think you probably want a snippet for the database stuff, since the connectivity samples don't show the table creation. This also feel more maintainable - we don't have to worry about modifying it in the other. 

Also - I'd suggest trying to make the region tags have a similar prefix so they are easier to find together. Something like:

- cloud_sql_mysql_cse_key
- cloud_sql_mysql_cse_db
- cloud_sql_mysql_cse_insert
- cloud_sql_mysql_cse_query

cloud-sql/mysql/client-side-encryption/src/main/java/com/example/cloudsql/EncryptAndInsertData.java
```diff
@@ -0,0 +1,122 @@
...
+import java.sql.Timestamp;
+import java.util.Date;
+import javax.sql.DataSource;
+
+class EncryptAndInsertData {
+
+  public static void main(String[] args) throws GeneralSecurityException, SQLException {
+    // TODO(developer): Replace these variables before running the sample.
+    String dbUser = "your-username"; // e.g. "root", "mysql"
+    String dbPass = "your-password";
```
thaidn: Not sure it's a good idea to store the password in source code. Maybe replace this with a command-line argument?

cloud-sql/mysql/client-side-encryption/src/main/java/com/example/cloudsql/EncryptAndInsertData.java
```diff
@@ -0,0 +1,122 @@
...
+import java.sql.Connection;
+import java.sql.PreparedStatement;
+import java.sql.SQLException;
+import java.sql.Timestamp;
+import java.util.Date;
+import javax.sql.DataSource;
+
+class EncryptAndInsertData {
+
+  public static void main(String[] args) throws GeneralSecurityException, SQLException {
```
thaidn: This means developers have to compile and run this as a command-line tool.

I thought we want to show developers how to encrypt before the data is written to Cloud SQL, no?

cloud-sql/mysql/client-side-encryption/src/main/java/com/example/cloudsql/EncryptAndInsertData.java
```diff
@@ -0,0 +1,122 @@
...
+        voteStmt.execute();
+        System.out.println(String.format("Successfully inserted row into table %s", tableName));
+      }
+    }
+  }
+
+  public static Aead getEnvelopeAead(String kmsUri) throws GeneralSecurityException {
+    AeadConfig.register();
+
+    // Generate a new envelope key template, then generate key material.
```
thaidn: We actually have a plan to deprecate this envelope encryption API.

Here's the new way (which is identical to Python):

KmsClient client = new GcpKmsClient().withDefaultCredentials();

Aead remote = client.getAead(kmsUri);

return new KmsEnvelopeAead(AeadKeyTemplates.AES128_GCM, remote);

cloud-sql/mysql/client-side-encryption/src/main/java/com/example/cloudsql/EncryptAndInsertData.java
shubha-rajan: I think you're looking at an outdated version. The latest version reads this information in from environment variables.
https://github.com/GoogleCloudPlatform/java-docs-samples/pull/4821/files#diff-aab83b5ceecd4d6e83b84401f33a99c50b07b092e62493a6e31534a1019778adR35

cloud-sql/mysql/client-side-encryption/src/main/java/com/example/cloudsql/EncryptAndInsertData.java
shubha-rajan: Yeah, in the latest version of this, the method in this file encrypts the data before inserting it into the table.

PR #4847, opened 2021-03-10
fix: add missing dependency

gcseh:


iot/api-client/end-to-end-example/pom.xml
```diff
@@ -85,6 +85,11 @@
       <artifactId>commons-cli</artifactId>
       <version>1.4</version>
     </dependency>
+    <dependency>
+      <groupId>com.google.api-client</groupId>
+      <artifactId>google-api-client-jackson2</artifactId>
+      <version>1.28.1</version>
```
averikitsch, line 91: Looks like there is a new version: 1.31.3

mwlebour (approved):

iot/api-client/end-to-end-example/pom.xml
```diff
@@ -85,6 +85,11 @@
       <artifactId>commons-cli</artifactId>
       <version>1.4</version>
     </dependency>
+    <dependency>
```
elharo, line 88: Gson should be used instead, per security team advice.

see go/json-madness

gcseh: ok, thanks!
will do later.

PR #4849, opened 2021-03-10
fix(deps): fix scalaVersion and pin jackson-module-scala

kolea2:
fix renovate config and roll back deps


billyjacobson (approved):

lesv (approved):
You can also pin inside RenovateBot if that's what you are really trying to do.

PR #4869, opened 2021-03-11
chore(functions): remove custom Pub/Sub POJOs

ace-n:
The GCS event types don't quite match up yet (per @grant), so I'm holding off on migrating those for now.

@eamonnmcmanus please **merge for me** once you've approved. Thanks!


grant:
Note, events delivered to Cloud Functions are not CloudEvents (yet). 

I don't think Java converts the events to CloudEvents. As such, using the CloudEvent type library is at your own risk and I'm not sure of the compatibility with non-CloudEvents

grant (approved):
Pub/Sub events looks fine.

Note failures like:

```
src/test/java/functions/SubscribeToTopicTest.java:[24,28] package functions.eventpojos does not exist
```

functions/concepts/retry-pubsub/src/main/java/functions/RetryPubSub.java
```diff
@@ -20,22 +20,22 @@
 
 import com.google.cloud.functions.BackgroundFunction;
 import com.google.cloud.functions.Context;
+import com.google.events.cloud.pubsub.v1.Message;
```
eamonnmcmanus, line 23: If it's straightforward to change this, I think we should. The common name `Message` is likely to collide with types of the same name in other packages, for example `com.google.protobuf.Message`. Since Java doesn't have a way to rename imported types, a class that needs two different `Message` types will have to fully-qualify one of them everywhere. I think the old name `PubSubMessage` was better, even if it seems a bit redundant with the package name.

If this is a huge pain to fix, I don't think it's all _that_ important.

ace-n: I'll defer to @grant on this, since I believe he wrote the `com.google.events` package.

grant: Right now we don't have a way to change this in the generator – it generates this name using the proto field name.

I've [filed an issue](https://github.com/googleapis/google-cloudevents-java/issues/69) where we can improve this with a future version of the library with the type name.

eamonnmcmanus: OK, I don't think this is important enough to block on.

eamonnmcmanus (approved):

lesv (approved):

PR #5090, opened 2021-04-20
Add comment about authenticated push JWT token validation.

fayssalmartanigcp:
This snippet is used in the Cloud Pub/Sub docs (https://cloud.google.com/pubsub/docs/push#validating_tokens) and many users are not aware that signature verification of the token is not enough, the claim needs to be validated also.

See similar change in the nodejs repo: https://github.com/GoogleCloudPlatform/nodejs-docs-samples/pull/2222.


kurtisvg (approved):

kurtisvg:
LGTM, but added a couple of folks that might own the samples to double check.

anguillanneuf (approved):

PR #5114, opened 2021-04-24
test: update IT to run Pub/Sub Lite I/O sample on GCP

### LINKED ISSUE #5113: Pub/Sub Lite I/O connector IT
Need to update the test in `/pubsublite/streaming-analytics` to use GCP instead of Beam's test suite. 

- ~need a permanent test topic with some messages published~ `pubsublite-streaming-analytics` in `us-east1-b`
- need to create a test subscription to the topic
- ~need a permanent bucket for output~ `pubsublite-it`
- start the pipeline
- check if files are written to the bucket
- for cleanup, delete the files for this test, delete the Dataflow pipeline (not sure how to find its name), delete the test subscription. 

A tricky thing is we may need to wait 5-10 minutes for the pipeline to start up and start writing files to Cloud Storage.

anguillanneuf:
Fixes #5113 

This test will add another 10 minutes to the tests 😖   

The tricky bit is canceling the Dataflow job using the Discovery API (apiary client) for Dataflow, which uses a different auth method: https://github.com/googleapis/google-auth-library-java#using-credentials-with-google-http-client

#### How does the test work?
A permanent Pub/Sub Lite topic has been created for this test. This Lite topic has some messages already published to it. Every time the test runs, a new Lite subscription is created for the topic and is configured to read all the messages in the topic. The Dataflow job, when run correctly, shall window the messages based on some logic defined by the windowing function and then write three files to Cloud Storage. The test checks for those files and passes when their content match the desired output.

### DIFF
--- a/pubsublite/streaming-analytics/src/test/java/PubsubliteToGcsIT.java
+++ b/pubsublite/streaming-analytics/src/test/java/PubsubliteToGcsIT.java
@@ -59,8 +59,8 @@ public class PubsubliteToGcsIT {
   @Rule public final TestPipeline testPipeline = TestPipeline.create();
 
   private static final String projectId = System.getenv("GOOGLE_CLOUD_PROJECT");
-  private String cloudRegion = "us-east1";
-  private final char zoneId = 'b';
+  private static final String cloudRegion = "us-east1";
+  private static final char zoneId = 'b';
   private static final String suffix = UUID.randomUUID().toString();
   private static final String topicId = "pubsublite-streaming-analytics";
   private static final String subscriptionId = "pubsublite-streaming-analytics-" + suffix;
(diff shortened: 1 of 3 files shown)


pubsublite/streaming-analytics/src/test/java/PubsubliteToGcsIT.java
```diff
@@ -13,79 +13,209 @@
...
+    Subscription subscription =
+        Subscription.newBuilder()
+            .setDeliveryConfig(
+                DeliveryConfig.newBuilder()
+                    .setDeliveryRequirement(DeliveryRequirement.DELIVER_IMMEDIATELY))
+            .setName(subscriptionPath.toString())
+            .setTopic(topicPath.toString())
+            .build();
+
+    AdminClientSettings adminClientSettings =
```
eaball35, line 111: Does this need to be created before each test and created to tear down later?

pubsublite/streaming-analytics/src/test/java/PubsubliteToGcsIT.java
```diff
@@ -13,79 +13,209 @@
...
 
 public class PubsubliteToGcsIT {
   @Rule public final TestPipeline testPipeline = TestPipeline.create();
 
-  private Instant baseTime = new Instant(0);
+  private static final String projectId = System.getenv("GOOGLE_CLOUD_PROJECT");
+  private String cloudRegion = "us-east1";
+  private final char zoneId = 'b';
+  private static final String suffix = UUID.randomUUID().toString();
+  private static final String topicId = "pubsublite-streaming-analytics";
```
lesv, line 65: If this is run by several processes will it work?   Shouldn't this have a UUID suffix?

lesv (approved):
LGTM - My concerns:
* if this crashes during a test, you'll slowly grow the bucket.
Many apps will look for a timestamp older than 24h and cleanup in setup.
* I don't see publishing to the topic, That's ok if the data will hang out for all time.

pubsublite/streaming-analytics/src/test/java/PubsubliteToGcsIT.java
anguillanneuf: Good point, no, not really. I can move them out of the `@Before` and `@After`.

pubsublite/streaming-analytics/src/test/java/PubsubliteToGcsIT.java
anguillanneuf: Yep, this is a permanent topic that allows many subscriptions to be attached to it.

anguillanneuf:
@lesv Thanks for the review!

- The GCS bucket used by the test has a rule that deletes files older than 1 day.
- Yep, the data will hang out for all time.

eaball35 (approved):

averikitsch:
This great! Thanks Tianzi 👏

PR #5146, opened 2021-04-28
feat: add instructions for flex templates

anguillanneuf:
Add instructions for turning a Dataflow pipeline into a [Dataflow flex template](https://cloud.google.com/dataflow/docs/guides/templates/using-flex-templates), which can be run with different input parameters via gcloud as well as in Cloud Console.

- updated README to include instructions.
- `metadata.json` is a required file for creating a flex template.
- the Dataflow Discovery API dependencies have to be given the scope `runtime` for flex templates.
- Flex templates don't allow `pipeline.run().waitUntilFinish()`. Removed `waitUntilFisnish()` from sample.
- used a blocking call to wait 10 minutes in test instead.

### DIFF
--- a/pubsublite/streaming-analytics/metadata.json
+++ b/pubsublite/streaming-analytics/metadata.json
@@ -1,27 +1,27 @@
 {
   "name": "Pub/Sub Lite to Clous Storage",
-  "description": "An Apache Beam streaming pipeline that reads messages from Pub/Sub Lite, applies fixed windowing on the messages, and writes the results to files on Cloud Storage",
+  "description": "An Apache Beam streaming pipeline that reads messages from Pub/Sub Lite, applies fixed windowing on the messages, and writes the results to files on Cloud Storage.",
   "parameters": [
     {
       "name": "subscription",
-      "label": "Pub/Sub Lite subscription.",
-      "helpText": "Pub/Sub subscription to read from."
+      "label": "Pub/Sub Lite subscription",
+      "helpText": "Pub/Sub subscription to read from, e.g. projects/my-project/locations/us-central1-b/subscriptions/my-subscription."
     },
     {
       "name": "windowSize",
       "label": "Window size in minutes",
-      "helpText": "Window size of output files in minutes, default to 1.",
+      "helpText": "Window size of output files in minutes.",
       "regexes": [
-        "^[1-9][0-9]*"
+        "^[1-9][0-9]+$"
       ]
     },
     {
       "name": "output",
-      "label": "Filename prefix of output files, e.g. gs://my-bucket/my-filename",
-      "helpText": "Filename prefix of output files including the filepath.",
+      "label": "Filename prefix of output files including the file path",
+      "helpText": "Filename prefix of output files including the file path, e.g. gs://my-bucket/my-filename-prefix.",
       "isOptional": true,
       "regexes": [
-        "^gs:\/(\/[a-zA-Z0-9-]+){2,}"
+        "^gs:\\/\\/.*$"
       ]
     }
   ]
--- a/pubsublite/streaming-analytics/metadata.json
+++ b/pubsublite/streaming-analytics/metadata.json
@@ -1,0 +1,28 @@
+{
+  "name": "Pub/Sub Lite to Clous Storage",
+  "description": "An Apache Beam streaming pipeline that reads messages from Pub/Sub Lite, applies fixed windowing on the messages, and writes the results to files on Cloud Storage",
+  "parameters": [
+    {
+      "name": "subscription",
+      "label": "Pub/Sub Lite subscription.",
+      "helpText": "Pub/Sub subscription to read from."
+    },
+    {
+      "name": "windowSize",
+      "label": "Window size in minutes",
+      "helpText": "Window size of output files in minutes, default to 1.",
+      "regexes": [
+        "^[1-9][0-9]*"
+      ]
+    },
+    {
+      "name": "output",
+      "label": "Filename prefix of output files, e.g. gs://my-bucket/my-filename",
+      "helpText": "Filename prefix of output files including the filepath.",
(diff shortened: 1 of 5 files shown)


pubsublite/streaming-analytics/src/main/java/examples/PubsubliteToGcs.java
```diff
@@ -114,9 +114,8 @@ public static void main(String[] args) throws InterruptedException {
                 .accumulatingFiredPanes())
         .apply("Write elements to GCS", new WriteOneFilePerWindow(options.getOutput(), numShards));
 
-    // Execute the pipeline with an optional timeout for your local program.
-    // This will not cancel the remote job.
-    pipeline.run().waitUntilFinish(Duration.standardMinutes(10));
+    // Execute the pipeline.
+    pipeline.run();
```
kurtisvg, line 118: Seems like leaving a timeout here might be a good idea?

kurtisvg (changes requested):

pubsublite/streaming-analytics/src/main/java/examples/PubsubliteToGcs.java
anguillanneuf: When experimented it yesterday, all jobs with `waitUntilFinish()` failed for flex templates. [Here](https://pantheon.corp.google.com/logs/query?dateRangeEnd=2021-04-27T22:53:15.534Z&dateRangeStart=2021-04-27T22:48:19.009Z&expandAll=false&interval=CUSTOM&project=java-docs-samples-testing&timestamp=2021-04-27T22:53:15.534Z&query=resource.type%3D%22dataflow_step%22%0Aresource.labels.job_id%3D%222021-04-27_15_48_18-9752491772384961993%22%0AlogName%3D(%22projects%2Fjava-docs-samples-testing%2Flogs%2Fdataflow.googleapis.com%252Fjob-message%22%20OR%20%22projects%2Fjava-docs-samples-testing%2Flogs%2Fdataflow.googleapis.com%252Flauncher%22)%0A) is an example.

@davidcavazos has raised this issue with the team building flex template before about supporting waiting, but right now it's not.

kurtisvg (approved):

pubsublite/streaming-analytics/src/main/java/examples/PubsubliteToGcs.java
kurtisvg: SGTM, thanks for the context.

pubsublite/streaming-analytics/metadata.json
```diff
@@ -0,0 +1,28 @@
+{
+  "name": "Pub/Sub Lite to Clous Storage",
+  "description": "An Apache Beam streaming pipeline that reads messages from Pub/Sub Lite, applies fixed windowing on the messages, and writes the results to files on Cloud Storage",
+  "parameters": [
+    {
+      "name": "subscription",
+      "label": "Pub/Sub Lite subscription.",
+      "helpText": "Pub/Sub subscription to read from."
```
davidcavazos, line 8: Can we specify the format (i.e. just the name or the full subscription path?) of the subscription here? Having a validation regex would also be nice.

pubsublite/streaming-analytics/metadata.json
```diff
@@ -0,0 +1,28 @@
...
+        "^[1-9][0-9]*"
+      ]
+    },
+    {
+      "name": "output",
+      "label": "Filename prefix of output files, e.g. gs://my-bucket/my-filename",
+      "helpText": "Filename prefix of output files including the filepath.",
+      "isOptional": true,
+      "regexes": [
+        "^gs:\/(\/[a-zA-Z0-9-]+){2,}"
```
davidcavazos, line 24: Nit: this excludes some valid names like including `.` or `_` or other characters.

I would be okay with having a more relaxed regex like: `^gs:\/\/.*$`

https://cloud.google.com/storage/docs/naming-buckets

pubsublite/streaming-analytics/metadata.json
```diff
@@ -0,0 +1,28 @@
...
+      "name": "subscription",
+      "label": "Pub/Sub Lite subscription.",
+      "helpText": "Pub/Sub subscription to read from."
+    },
+    {
+      "name": "windowSize",
+      "label": "Window size in minutes",
+      "helpText": "Window size of output files in minutes, default to 1.",
+      "regexes": [
+        "^[1-9][0-9]*"
```
davidcavazos, line 15: Nit: Just like the `^` matches the beginning of the text, `$` matches the end. How about something like `^[0-9]+$`?

pubsublite/streaming-analytics/src/main/java/examples/PubsubliteToGcs.java
davidcavazos, line 118: It would be nice to have a comment mentioning that a Flex Template expects to only run, and not waitUntilFinish.

davidcavazos (approved):
Overall looks great! Added some minor notes.

pubsublite/streaming-analytics/metadata.json
anguillanneuf: Done.

pubsublite/streaming-analytics/metadata.json
anguillanneuf: Done.

pubsublite/streaming-analytics/metadata.json
anguillanneuf: Done. I needed to do `^[1-9][0-9]=$` to exclude zero.

pubsublite/streaming-analytics/src/main/java/examples/PubsubliteToGcs.java
anguillanneuf: Done.

PR #5189, opened 2021-05-09
Set up application to be packaged as an uber-jar via application.properties

bartprokop:
I have noticed that Quarkus GAE "hello world" example doesn't work, as uber JAR is not produced. To rectify this, it is necessary to add 

```
quarkus.package.type=uber-jar
```

to ```applications.properties```, as per [documentation](https://quarkus.io/guides/deploying-to-google-cloud)


bartprokop:
@googlebot I signed it!

averikitsch (approved):
Thank you

PR #5240, opened 2021-05-14
run tests concurrently and with timeouts

kurtisvg:
This PR does the following:
1. Changes "continuous" tests (on merge) to only test vs the previous commit 
2. Uses `btlr`, a binary to run all the tests concurrently (concurrency based on the size of the VM)
3. Per-folder timeout of 30 minutes


.kokoro/tests/run_test_java.sh
```diff
@@ -0,0 +1,76 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Fail the tests if no Java version was found.
+POM_JAVA=$(grep -oP '(?<=<maven.compiler.target>).*?(?=</maven.compiler.target>)' pom.xml)
+ALLOWED_VERSIONS=("1.8" "11")
```
lesv, line 18: We should add 17 at some point.

lesv (approved):

.kokoro/java11/continuous.cfg
```diff
@@ -19,3 +19,9 @@ env_vars: {
     key: "TRAMPOLINE_BUILD_FILE"
     value: "github/java-docs-samples/.kokoro/tests/run_tests.sh"
 }
+
+# Only diff from previous commit
```
lesv, line 23: So your changing the meaning?

.kokoro/tests/run_tests.sh
```diff
@@ -103,109 +93,24 @@ if [[ ",$JAVA_VERSION," =~ "11" ]]; then
...
-      if [[ $EXIT -ne 0 ]]; then
-        RTN=1
-        echo -e "\n Cloud Run build/deploy failed: gcloud returned a non-zero exit code. \n"
-      else
-        echo -e "\n Cloud Run build/deploy completed.\n"
-      fi
-    fi
+btlr_args=(
+    "run"
+    "--max-cmd-duration=20m"
```
lesv, line 98: I wonder if this is true.

lesv (approved):

.kokoro/tests/run_tests.sh
kurtisvg: this is a limitation enforced by btlr - it'll *should* kill the process once this time-limit expires. I left this deliberately low for now to try and help spot troublesome projects that are taking too long to run tests on.

.kokoro/java11/continuous.cfg
kurtisvg: This is changing the scope of the continuous tests from "test everything every commit" to "test everything changed in the last commit". In python-docs-samples, they switched to this model to prevent every RenovateBot PR from triggering a full test suite. There are a lot less conflicts if there isn't a 3-4 hour suite run everytime a commit is made.

PR #5249, opened 2021-05-17
Update SAMPLE_FORMAT.md

lesv:
* New samples should consider using Java 11
* I updated the languages to allow Java 17 or alternate JVM's if required - we'll have to fix our infrastructure if it's used.
* discouraged the use of `var`.  There isn't a `let` keyword in Java 11


averikitsch (approved):
Thanks!

PR #5254, opened 2021-05-18
Update renovate.json

lesv:
reject `-SP.xxx` suffixes in Versions

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -17,6 +17,9 @@
       "groupName": "AppEngine packages"
     },
     {
+      "allowedVersions": "!/.+\\-SP\\.[0-9]+$/"
+    },
+    {
       "packagePatterns": [
         "^org.apache.beam:"
       ],
--- a/renovate.json
+++ b/renovate.json
@@ -17,7 +17,7 @@
       "groupName": "AppEngine packages"
     },
     {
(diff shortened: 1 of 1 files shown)


renovate.json
```diff
@@ -16,6 +16,9 @@
       ],
       "groupName": "AppEngine packages"
     },
+    {
```
kurtisvg, line 19: Do we need to limit to our packages? e.i. "com.google.cloud" or whatever?

renovate.json
```diff
@@ -16,6 +16,9 @@
...
+      "allowedVersions": "!/.+\\-SP\\.[0-9]+$/"
```
kurtisvg, line 20: Should this be:
```suggestion
      "allowedVersions": "!/.+\-SP\.[0-9]+$/"
```

kurtisvg (changes requested):

renovate.json
lesv: I wish, but there is com.google.api, com.google.api-client, as well as com.google.cloud.  (maybe a few more)

kurtisvg (approved):
LGTM, but still unsure on the regex.

shubha-rajan (approved):

kurtisvg:
Changed my mind - regex makes sense at second glance.

PR #5263, opened 2021-05-20
chore(spanner): remove samples that have moved to the client library

olavloite:
The samples for the Google Cloud Spanner client library have been moved to the source repository of the client library. These samples are no longer in use for the documentation, and are also no longer kept up to date and should
be removed.


thiagotnunes (approved):
That is a nice one `+1, -5032` =)

kurtisvg (approved):

lesv (approved):

PR #5306, opened 2021-05-25
Add Load Csv Data Example to JDBC examples

kaylaleung:
Updated to be in jdbc directory rather than cloud-client.


spanner/jdbc/src/main/java/com/example/spanner/jdbc/LoadCsvExample.java
```diff
@@ -0,0 +1,290 @@
+/*
+ * Copyright 2021 Google Inc.
```
averikitsch, line 2:
```suggestion
 * Copyright 2021 Google LLC
```

spanner/jdbc/src/main/java/com/example/spanner/jdbc/LoadCsvExample.java
```diff
@@ -0,0 +1,290 @@
...
+import org.apache.commons.cli.Options;
+import org.apache.commons.csv.CSVFormat;
+import org.apache.commons.csv.CSVParser;
+import org.apache.commons.csv.CSVRecord;
+
+/** Sample showing how to load CSV file data into Spanner */
+public class LoadCsvExample {
+  public static final String EXCEL = "EXCEL";
+  public static final String POSTGRESQL_CSV = "POSTGRESQL_CSV";
+  public static final String POSTGRESQL_TEXT = "POSTGRESQL_TEXT}";
```
averikitsch, line 49: Is the "}" intentional?

spanner/jdbc/src/main/java/com/example/spanner/jdbc/LoadCsvExample.java
```diff
@@ -0,0 +1,290 @@
...
+
+  enum SpannerDataType {
```
averikitsch, line 51: Should these be imported from the library, `com.google.spanner.v1.TypeCode`?

spanner/jdbc/src/main/java/com/example/spanner/jdbc/LoadCsvExample.java
```diff
@@ -0,0 +1,290 @@
...
+    DATE,
+    TIMESTAMP
+  }
+
+  public static Boolean hasHeader = false;
+  public static Connection connection;
+  public static Map<String, SpannerDataType> tableColumns = new LinkedHashMap<>();
+
+  /** Return the data type of the column type **/
+  public static SpannerDataType parseSpannerDataType(String columnType) {
```
averikitsch, line 69: It seems odd we expect users to implement this theirselves. Is there not a library method?

spanner/jdbc/src/main/java/com/example/spanner/jdbc/LoadCsvExample.java
```diff
@@ -0,0 +1,290 @@
...
+      mutations.add(builder.build());
+    }
+    CloudSpannerJdbcConnection spannerConnection = connection
+        .unwrap(CloudSpannerJdbcConnection.class);
+    spannerConnection.write(mutations);
+    spannerConnection.close();
+    System.out.println("Data successfully written into table.");
+  }
+
+  static void loadCsv(String projectId, String instanceId, String databaseId, String tableName,
```
averikitsch, line 244: How will a user run this sample? Please also see [Method Structure](https://github.com/GoogleCloudPlatform/java-docs-samples/blob/master/SAMPLE_FORMAT.md#method-structure)

spanner/jdbc/src/main/java/com/example/spanner/jdbc/LoadCsvExample.java
```diff
@@ -0,0 +1,290 @@
...
+    opt.addOption("n", true, "String Representing Null Value");
+    opt.addOption("d", true, "Character Separating Columns");
+    opt.addOption("e", true, "Character To Escape");
+    CommandLineParser clParser = new DefaultParser();
+    CommandLine cmd = clParser.parse(opt, optFlags);
+
+    SpannerOptions options = SpannerOptions.newBuilder().build();
+    Spanner spanner = options.getService();
+
+    try {
```
averikitsch, line 261: Can you use try with resources?

spanner/jdbc/src/main/java/com/example/spanner/jdbc/LoadCsvExample.java
```diff
@@ -0,0 +1,290 @@
...
+      // If file has header, verify that header fields are valid
+      if (hasHeader && !isValidHeader(parser)) {
+        return;
+      }
+
+      // Write CSV record data to Cloud Spanner
+      try {
+        writeToSpanner(parser, tableName, cmd);
+      } catch (SQLException e) {
+        System.out.println(e.getMessage());
```
averikitsch, line 282: What should a user do when there is a SQLException?

averikitsch (changes requested):

spanner/jdbc/src/main/java/com/example/spanner/jdbc/LoadCsvExample.java
kaylaleung: I was under the impression that the example of how users would run the sample would be just to follow the same setup as provided in the test I wrote for this function, but let me know if you mean something else.

Also I'm not exactly sure what kind of main function the Method Structure is referring to as I haven't seen any of the other example files have main functions?

spanner/jdbc/src/main/java/com/example/spanner/jdbc/LoadCsvExample.java
dttran: Not sure what you mean by "users implement this themselves"? Could you elaborate?

spanner/jdbc/src/main/java/com/example/spanner/jdbc/LoadCsvExample.java
dttran: @averikitsch  I don't think Method Structure applies here, as this is an example similar to other existing examples in the same directory.

dttran (approved):

spanner/jdbc/src/main/java/com/example/spanner/jdbc/LoadCsvExample.java
averikitsch: You expect the user to copy and paste 20+ lines of regex if statements just to set the data type. `parseSpannerDataType` seems like it should be provided by a client library

spanner/jdbc/src/main/java/com/example/spanner/jdbc/LoadCsvExample.java
averikitsch: We usually provided an overloaded method, "Any declared function arguments should include a no-arg, main method with examples for how the user can initialize the method arguments and call the entrypoint for the snippet. ". Also does the [README](https://github.com/GoogleCloudPlatform/java-docs-samples/tree/ab0e5d10706468c27dabf27c7b2b20e2c684d8f4/spanner/jdbc) properly describe how to run the snippet?

kaylaleung: Is there an example that you could point me to where this overloaded method is implemented for the other example files? 

Also will update the README 👍🏼

spanner/jdbc/src/main/java/com/example/spanner/jdbc/LoadCsvExample.java
kaylaleung: I brought this up as a concern with the client team they confirmed that no client library exists that does this and this current method would seem to be the best way.

averikitsch: sg

spanner/jdbc/src/main/java/com/example/spanner/jdbc/LoadCsvExample.java
averikitsch: [Example 1 using an overloaded method](https://github.com/GoogleCloudPlatform/java-docs-samples/blob/master/secretmanager/src/main/java/secretmanager/CreateSecret.java#L28) and [Example 2 using a main method](https://github.com/googleapis/java-bigquery/blob/master/samples/snippets/src/main/java/com/example/bigquery/CopyTable.java#L30)

spanner/jdbc/src/main/java/com/example/spanner/jdbc/LoadCsvExample.java
```diff
@@ -0,0 +1,283 @@
...
+          writeToSpanner(parser, tableName);
+        } catch (SQLException e) {
+          /* SQLExceptions are thrown when the table name cannot be queried for in the database
+          or the connection established does not have permissions to write data into the table */
+          System.out.println(e.getMessage());
+        }
+      }
+    } finally {
+      spanner.close();
+      connection.close();
```
averikitsch, line 279: If you use try with resources you shouldn't need to close this. It seems like it would be cleaner to move connection out of the try with resources

spanner/jdbc/src/main/java/com/example/spanner/jdbc/LoadCsvExample.java
```diff
@@ -0,0 +1,283 @@
...
+
+    SpannerOptions options = SpannerOptions.newBuilder().build();
+    Spanner spanner = options.getService();
+
+    // Initialize connection to Cloud Spanner
+    try (
+        Connection conn = DriverManager.getConnection(
+            String.format("jdbc:cloudspanner:/projects/%s/instances/%s/databases/%s",
+            projectId, instanceId, databaseId));
+        Reader in = new FileReader(filePath)
```
averikitsch, line 256: It may be cleaner not to separately initialize this outside of the CVSParser. The nested tries are harder to follow

spanner/jdbc/src/main/java/com/example/spanner/jdbc/LoadCsvExample.java
```diff
@@ -241,8 +219,20 @@ public static void writeToSpanner(Iterable<CSVRecord> records, String tableName,
     System.out.println("Data successfully written into table.");
   }
 
-  static void loadCsv(String projectId, String instanceId, String databaseId, String tableName,
-      String filePath, String[] optFlags) throws Exception {
+  static void loadCsv() throws Exception {
```
averikitsch, line 222: You can move this to the top of the file to make it more discoverable

spanner/jdbc/src/main/java/com/example/spanner/jdbc/LoadCsvExample.java
kaylaleung: I tried to simplify to 2 tries instead of 3, let me know if that seems cleaner based on my understanding of your comment?

averikitsch (approved):
Sorry for all the back and forth. This looks great!

kaylaleung:
> Sorry for all the back and forth. This looks great!

Awesome, thank you also for your help :) 

I don't think I have merge access, so I'll leave it to you to merge?

PR #5360, opened 2021-06-03
Update and rename stale.yml to flakybot.yaml

lesv:
* Add default issuePriority: p2
* rename stale.yml => flakeybot.yaml


kurtisvg (commented):
Flakybot is a different than the stale bot we were using before. Has it been added to this repo? Does it support `daysUntilstale`?

lesv:
I see Flakeybot create issues, but am checking to find out if they aren't related.  :(

lesv:
Pilot Error.

PR #5421, opened 2021-06-11
Google cloud client compute samples

sitalakshmis:
> Requesting to merge Compute Engine Diregapic samples


compute/cloud-client/pom.xml
```diff
@@ -1,9 +1,8 @@
 <?xml version="1.0" encoding="UTF-8"?>
```
lesv, line 1: Copyright & License?

compute/cloud-client/pom.xml
```diff
@@ -29,6 +28,11 @@
       <artifactId>google-cloud-compute</artifactId>
       <version>1.2.0-alpha</version>
     </dependency>
+    <dependency>
+      <groupId>com.google.cloud</groupId>
+      <artifactId>google-cloud-storage</artifactId>
+      <version>1.115.0</version>
+    </dependency>
```
lesv, line 35: Do you wish to use the Libraries-BOM?

compute/cloud-client/src/main/java/compute/CreateInstance.java
```diff
@@ -28,78 +28,79 @@
...
 
       if (response.hasError()) {
         System.out.println("Instance creation failed ! ! " + response.getError());
         return;
       }
       System.out.println("####### Instance creation complete #######");
+
+    } catch (com.google.api.gax.rpc.UnknownException e) {
+      System.out.println("####### Instance creation complete #######");
     }
```
lesv, line 103: You might wish to explain why this error is ok.  If I got an unknown exception, I'd normally be concerned.

compute/cloud-client/src/main/java/compute/DefaultValue.java
```diff
@@ -0,0 +1,77 @@
...
+    // TODO(developer): Create a Google Cloud Storage bucket.
+    // bucketName: to run this snippet, an existing Google Cloud Storage bucket is required.
+    String project = "your-project-id";
+    String bucketName = "your-bucket-name";
+    defaultValues(project, bucketName);
+  }
+
+  // Check whether an optional param was set to the default value 
+  // when not provided an explicit value.
+  public static void defaultValues(String project, String bucketName)
```
lesv, line 43: `defaultValues` seems like an odd name for this method.

compute/cloud-client/src/main/java/compute/DefaultValue.java
```diff
@@ -0,0 +1,77 @@
...
+      throws IOException, InterruptedException {
+    // Initialize client that will be used to send requests. This client only needs to be created
```
lesv, line 45: maybe some whitespace

compute/cloud-client/src/main/java/compute/DeleteInstance.java
```diff
@@ -23,50 +23,48 @@
...
 
       if (response.hasError()) {
         System.out.println("Instance deletion failed ! ! " + response.getError());
         return;
       }
       System.out.println("####### Instance deletion complete #######");
+
+    } catch (com.google.api.gax.rpc.UnknownException e) {
+      System.out.println("####### Instance deletion complete #######");
     }
```
lesv, line 67: Why do we believe this is ok?

lesv (changes requested):
PTAL - fix the requested things.

@eaball35 should still review.

compute/cloud-client/pom.xml
sitalakshmis: The samples in other languages haven't used BOM as well. Hence, to maintain consistency, I would prefer to use individual dependencies.

compute/cloud-client/src/main/java/compute/CreateInstance.java
sitalakshmis: This is a temporary workaround for LRO. SocketTimeoutException is masked as UnknownException while creating an instance. Hence, even when the instance creation process is running at the background UnknownException is being thrown.

compute/cloud-client/src/main/java/compute/DeleteInstance.java
sitalakshmis: This is a temporary workaround for LRO. SocketTimeoutException is masked as UnknownException while deleting an instance. Hence, even when the delete instance process is running at the background UnknownException is being thrown. Please suggest alternate ways if you think it'll be more appropriate.

compute/cloud-client/pom.xml
lesv: Ok, though, do other languages provide a BOM?  It's supper important for Java users.

compute/cloud-client/src/main/java/compute/CreateInstance.java
lesv: You might wish to add a comment about that - especially since it's a work around.

compute/cloud-client/src/main/java/compute/DeleteInstance.java
lesv: ditto

compute/cloud-client/src/main/java/compute/CreateInstance.java
```diff
@@ -28,78 +28,81 @@
...
       throws IOException, InterruptedException {
     // Below are sample values that can be replaced.
-    // machineType: should be of the format zones/zone-name/machineTypes/machine-type. Refer: https://cloud.google.com/compute/docs/machine-types
-    // sourceImage: path to the image that is to be mounted should be specified. Refer: https://cloud.google.com/compute/docs/images
-    // diskSizeGb: the storage size of the disk that will be attached to the instance
+    // machineType: machine type of the VM being created. This value uses the format zones/{zone}/machineTypes/{type_name}. For a list of machine types, see https://cloud.google.com/compute/docs/machine-types
+    // sourceImage: path to the operating system image to mount. For details about images you can mount, see https://cloud.google.com/compute/docs/images
+    // diskSizeGb: storage size of the boot disk to attach to the instance.
+    // networkName: network interface to associate with the instance.
     String machineType = "zones/us-central1-a/machineTypes/n1-standard-1";
```
lesv, line 51: take a look at our [guide](https://github.com/GoogleCloudPlatform/java-docs-samples/blob/master/SAMPLE_FORMAT.md)

These should be created using `String.format(...` and use the zone  that is passed in.

compute/cloud-client/src/main/java/compute/CreateInstance.java
```diff
@@ -28,78 +28,81 @@
...
 
       if (response.hasError()) {
         System.out.println("Instance creation failed ! ! " + response.getError());
         return;
       }
       System.out.println("####### Instance creation complete #######");
+
+    } catch (com.google.api.gax.rpc.UnknownException e) {
+      // Handle SocketTimeoutException which is being thrown as UnknownException.
+      // (Instance creation process will run to completion in the background)
```
lesv, line 103: Is there metadata on the exception to tell me it's a timeout?  If so, we should check it.

lesv (changes requested):

compute/cloud-client/src/main/java/compute/CreateInstance.java
sitalakshmis: There is metadata to tell that if there was an internal error or if the remote call failed, then the snippet would throw and APIException instead of the SocketTimeoutException which occurs when the default deadline is reached.
Please see, https://github.com/googleapis/java-compute/blob/diregapic/google-cloud-compute/src/main/java/com/google/cloud/compute/v1/ZoneOperationsClient.java#L506

sitalakshmis:
@eaball35 Requesting for review

compute/cloud-client/src/main/java/compute/SetUsageExportBucket.java
```diff
@@ -0,0 +1,154 @@
...
+    // bucketName: Cloud Storage Bucket used to store Compute Engine usage reports.
+    // An existing Google Cloud Storage bucket is required.
+    // reportNamePrefix: Prefix of the name of the usage report that would
+    // store Google Compute Engine data.
+    try (ProjectsClient projectsClient = ProjectsClient.create()) {
+
+      // Initialize UsageExportLocation object with provided bucket name and no report name prefix.
+      UsageExportLocation usageExportLocation = UsageExportLocation.newBuilder()
+          .setBucketName(bucketName)
+          .setReportNamePrefix(reportNamePrefix).build();
```
eaball35, line 70: Comment says "with no report name prefix"  - Are you not initializing with report name prefix here?

compute/cloud-client/src/main/java/compute/SetUsageExportBucket.java
```diff
@@ -0,0 +1,154 @@
...
+          String reportNamePrefix = usageExportLocation.getReportNamePrefix();
+
+          if (reportNamePrefix.length() == 0) {
+            // Although the server explicitly sent the empty string value,
+            // the next usage report generated with these settings still has the default
+            // prefix value "usage_gce".
+            // (see, https://cloud.google.com/compute/docs/reference/rest/v1/projects/get)
+            reportNamePrefix = "usage_gce";
+            System.out.println(
+                "Report name prefix not set, replacing with default value of `usage_gce`.");
```
eaball35, line 117: Similar to `setUsageExportBucket` will reportNamePrefix default to 'usage_gce' or do you need to explicitly define as you have here?

compute/cloud-client/src/main/java/compute/SetUsageExportBucket.java
```diff
@@ -0,0 +1,154 @@
...
+    try (ProjectsClient projectsClient = ProjectsClient.create()) {
+
+      // Initialize UsageExportLocation object with provided bucket name and no report name prefix.
+      UsageExportLocation usageExportLocation = UsageExportLocation.newBuilder()
+          .setBucketName(bucketName)
+          .setReportNamePrefix(reportNamePrefix).build();
+
+      if (reportNamePrefix.length() == 0) {
+        // Sending an empty value for reportNamePrefix results in the
+        // next usage report being generated with the default prefix value "usage_gce".
```
eaball35, line 74: Do you provided suggests "If not supplied, defaults to usage" rather than usage_gce

compute/cloud-client/src/main/java/compute/SetUsageExportBucket.java
```diff
@@ -0,0 +1,154 @@
...
+      // Initialize UsageExportLocation object with empty builder to disable usage reports.
+      UsageExportLocation usageExportLocation = UsageExportLocation.newBuilder().build();
+
+      // Disable the usage export location.
+      projectsClient.setUsageExportBucket(SetUsageExportBucketProjectRequest.newBuilder()
+          .setProject(project)
+          .setUsageExportLocationResource(usageExportLocation)
+          .build());
+
+      TimeUnit.SECONDS.sleep(5);
```
eaball35, line 147: Why are we sleeping here? May help to add comment?

eaball35 (approved):
Left a few comments, otherwise LGTM

compute/cloud-client/src/main/java/compute/SetUsageExportBucket.java
sitalakshmis: The snippets in other languages use this format. Hence sticking with "usage_gce".

compute/cloud-client/src/main/java/compute/SetUsageExportBucket.java
sitalakshmis: Yes, and "usage_gce" is intentional to follow the same format in all languages.

lesv (approved):

compute/cloud-client/src/main/java/compute/DeleteInstance.java
lesv: Argh - I think my ditto remark was for something else.

lesv: ditto was to say add a comment

PR #5561, opened 2021-07-06
Added "union" to accepted types in the avro schema

### LINKED ISSUE #5560: Not accepting unions in avro schema
## Describe the issue

Im trying to convert an avro file to csv using this github which I ran directly in the cloud shell.
The build failed with this error: "An exception occured while executing the Java class. Field type union is not supported"

My avro schema indeed has unions, for example: {"default": null, "name": "tech_filename", "type": ["null", "string"]}.

I tried removing them but the dataflow job fails due to "java.io.IOException: Failed to start reading from source" which comes from "Caused by: org.apache.avro.AvroTypeException: Found null, expecting int".

So I need the unions "type":["null", "int"] to take care of these exceptions but it isn't accepted by the build.

Is there any chance to get the build to accept unions in the avro schema?

Julien-Bernardo:
I opened an issue on this topic #5560 and finally found the solution myself. I wanted to share this because it allows the user to read avro files with nulls in them. This change worked for me and the pipeline returns the intended csv.

Fixes #5560

- [ x] I have followed [Sample Format Guide](https://github.com/GoogleCloudPlatform/java-docs-samples/blob/master/SAMPLE_FORMAT.md)
- [ x] `pom.xml` parent set to latest `shared-configuration`
- [ x] Appropriate changes to README are included in PR
- [ x] API's need to be enabled to test (tell us)
- [ x] Environment Variables need to be set (ask us to set them)
- [ x] **Tests** pass:   `mvn clean verify` **required**
- [ x] **Lint**  passes: `mvn -P lint checkstyle:check` **required**
- [x ] **Static Analysis**:  `mvn -P lint clean compile pmd:cpd-check spotbugs:check` **advisory only**
- [x ] Please **merge** this PR for me once it is approved.

### DIFF
--- a/dataflow/transforms/src/main/java/com/example/AvroToCsv.java
+++ b/dataflow/transforms/src/main/java/com/example/AvroToCsv.java
@@ -38,7 +38,7 @@ public class AvroToCsv {
 
   private static final Logger LOG = LoggerFactory.getLogger(AvroToCsv.class);
   private static final List<String> acceptedTypes = Arrays.asList(
-      new String[]{"string", "boolean", "int", "long", "float", "double"});
+      new String[]{"string", "boolean", "int", "long", "float", "double", "union"});
 
   private static String getSchema(String schemaPath) throws IOException {
     ReadableByteChannel channel = FileSystems.open(FileSystems.matchNewResource(


Julien-Bernardo:
@googlebot I signed it!

shubha-rajan (approved):

shubha-rajan:
Thanks for the patch! LGTM as long as tests pass.

PR #5634, opened 2021-07-19
docs: Bigtable data generator

billyjacobson:
A sample showing how to generate a large write load onto Cloud Bigtable

Fixes #issue

### DIFF
--- a/bigtable/beam/bulk-data-generator/src/main/java/BulkWrite.java
+++ b/bigtable/beam/bulk-data-generator/src/main/java/BulkWrite.java
@@ -51,12 +51,12 @@ public class BulkWrite {
   static final long MB_PER_SEC = 120;
 
   static final String COLUMN_FAMILY = "cf";
+  public static final String TABLE_PREFIX = "data-";
 
   public static void main(String[] args) {
     BigtableOptions options =
         PipelineOptionsFactory.fromArgs(args).withValidation().as(BigtableOptions.class);
     Pipeline p = Pipeline.create(options);
-    System.out.println(options.getBigtableInstanceId());
 
     BigtableTableAdminClient adminClient = null;
     try {
@@ -84,11 +84,19 @@ public class BulkWrite {
     long numRows = (long) ((TB_PER_TABLE * ONE_TB) / (MB_PER_ROW * ONE_MB));
     long rate = clusterNodeCount * MB_PER_SEC / newTableIds.size();
 
+    String generateLabel = String
+        .format("Generate %d rows at %dMB per second for %d tables", numRows, rate,
+            newTableIds.size());
+    String mutationLabel = String
+        .format("Create mutations that write %d MB to each row", MB_PER_ROW);
+
+    System.out.println(generateLabel);
+    System.out.println(mutationLabel);
+
     PCollection<Mutation> mutations = p
-        .apply(String.format("Generate %d rows at %dMB per second for %d tables", numRows, rate,
-            newTableIds.size()), GenerateSequence.from(0).to(numRows)
+        .apply(generateLabel, GenerateSequence.from(0).to(numRows)
             .withRate(rate, Duration.standardSeconds(1)))
-        .apply(String.format("Create mutations that write %d MB to each row", MB_PER_ROW),
+        .apply(mutationLabel,
             ParDo.of(new CreateMutationFn()));
 
     for (String tableId : newTableIds) {
@@ -125,7 +133,7 @@ public class BulkWrite {
       List<CloudBigtableTableConfiguration> bigtableTableConfigs = new ArrayList<>();
 
       for (int i = 0; i < numTablesToCreate; i++) {
-        String tableId = "data-" + UUID.randomUUID().toString().substring(0, 20);
+        String tableId = TABLE_PREFIX + UUID.randomUUID().toString().substring(0, 20);
         CreateTableRequest createTableRequest = CreateTableRequest.of(tableId)
             .addFamily(COLUMN_FAMILY);
         adminClient.createTable(createTableRequest);
@@ -146,7 +154,7 @@ public class BulkWrite {
       String clusterId = new BigtableClusterName(cluster.getName()).getClusterId();
       String zoneId = BigtableClusterUtilities.getZoneId(cluster);
       int clusterNodeCount = clusterUtility.getClusterNodeCount(clusterId, zoneId);
-      System.out.println("cluster size " + clusterNodeCount);
+      System.out.println("Cluster size " + clusterNodeCount);
       return clusterNodeCount;
     } catch (Exception e) {
       e.printStackTrace();
--- a/bigtable/beam/bulk-data-generator/src/main/java/BulkWrite.java
+++ b/bigtable/beam/bulk-data-generator/src/main/java/BulkWrite.java
@@ -1,0 +1,197 @@
+/*
+ * Copyright 2021 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import com.google.bigtable.admin.v2.Cluster;
+import com.google.cloud.bigtable.admin.v2.BigtableTableAdminClient;
+import com.google.cloud.bigtable.admin.v2.BigtableTableAdminSettings;
+import com.google.cloud.bigtable.admin.v2.models.CreateTableRequest;
+import com.google.cloud.bigtable.beam.CloudBigtableIO;
+import com.google.cloud.bigtable.beam.CloudBigtableTableConfiguration;
+import com.google.cloud.bigtable.grpc.BigtableClusterName;
+import com.google.cloud.bigtable.grpc.BigtableClusterUtilities;
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Random;
+import java.util.UUID;
+import org.apache.beam.runners.dataflow.options.DataflowPipelineOptions;
+import org.apache.beam.sdk.Pipeline;
+import org.apache.beam.sdk.io.GenerateSequence;
+import org.apache.beam.sdk.options.Default;
+import org.apache.beam.sdk.options.Description;
+import org.apache.beam.sdk.options.PipelineOptionsFactory;
+import org.apache.beam.sdk.transforms.DoFn;
+import org.apache.beam.sdk.transforms.ParDo;
+import org.apache.beam.sdk.values.PCollection;
+import org.apache.hadoop.hbase.client.Mutation;
+import org.apache.hadoop.hbase.client.Put;
+import org.apache.hadoop.hbase.util.Bytes;
+import org.joda.time.Duration;
+
+public class BulkWrite {
+
+  static final double TB_PER_TABLE = .5;
+  static final long MB_PER_ROW = 1;
+  static final long ONE_MB = 1000 * 1000;
+  static final long ONE_GB = 1000 * ONE_MB;
+  static final long ONE_TB = 1000 * ONE_GB;
+  static final long MB_PER_SEC = 120;
+
+  static final String COLUMN_FAMILY = "cf";
+
+  public static void main(String[] args) {
+    BigtableOptions options =
+        PipelineOptionsFactory.fromArgs(args).withValidation().as(BigtableOptions.class);
+    Pipeline p = Pipeline.create(options);
+    System.out.println(options.getBigtableInstanceId());
+
+    BigtableTableAdminClient adminClient = null;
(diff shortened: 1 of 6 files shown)


lesv (changes requested):
Looks like there is a lint issue and there doesn't appear to be any tests?

bigtable/beam/bulk-data-generator/src/main/java/BulkWrite.java
```diff
@@ -0,0 +1,197 @@
...
+  static final String COLUMN_FAMILY = "cf";
+
+  public static void main(String[] args) {
+    BigtableOptions options =
+        PipelineOptionsFactory.fromArgs(args).withValidation().as(BigtableOptions.class);
+    Pipeline p = Pipeline.create(options);
+    System.out.println(options.getBigtableInstanceId());
+
+    BigtableTableAdminClient adminClient = null;
+    try {
```
telpirion, line 62: Should the IOException be allowed to bubble up to the next level?

telpirion: Or the pattern I see a lot is something like:

```
try (adminClient = BigtableTableAdminClient.create(adminSettings)) {
// more sample here
}
```

billyjacobson:
> Looks like there is a lint issue and there doesn't appear to be any tests?

Added test and fixed lint

bigtable/beam/bulk-data-generator/src/main/java/BulkWrite.java
telpirion: I think this is still an issue in this code. 

We do something like this in the Vertex AI Java samples:
https://github.com/googleapis/java-aiplatform/blob/master/samples/snippets/src/main/java/aiplatform/CreateBatchPredictionJobTextClassificationSample.java#L61

bigtable/beam/bulk-data-generator/src/main/java/bigtable/BulkWrite.java
```diff
@@ -0,0 +1,201 @@
...
+  static final long ONE_MB = 1000 * 1000;
+  static final long ONE_GB = 1000 * ONE_MB;
+  static final long ONE_TB = 1000 * ONE_GB;
+  static final long MB_PER_SEC = 120;
+
+  static final String COLUMN_FAMILY = "cf";
+  public static final String TABLE_PREFIX = "data-";
+  static final SecureRandom random = new SecureRandom();
+
+  public static void main(String[] args) {
```
telpirion, line 60: Add `throws IOException` here.

bigtable/beam/bulk-data-generator/src/main/java/bigtable/BulkWrite.java
```diff
@@ -0,0 +1,201 @@
...
+    BigtableOptions options =
+        PipelineOptionsFactory.fromArgs(args).withValidation().as(BigtableOptions.class);
+    Pipeline p = Pipeline.create(options);
+
+    try {
```
telpirion, line 65: Please remove `try/catch`

bigtable/beam/bulk-data-generator/src/main/java/bigtable/BulkWrite.java
telpirion: Another pattern you might consider, something that makes testing easier:

1. Break this file into two methods: `main()` and `bulkWrite()`. 
2. Define the inputs that customers need to provide for this sample in the method signature for `bulkWrite()`.
3. In your integration test, call `bulkWrite()` while passing in your test variables.

bigtable/beam/bulk-data-generator/src/main/java/bigtable/BulkWrite.java
```diff
@@ -0,0 +1,201 @@
...
+public class BulkWrite {
+
+  static final double TB_PER_TABLE = .5;
+  static final long MB_PER_ROW = 1;
+  static final long ONE_MB = 1000 * 1000;
+  static final long ONE_GB = 1000 * ONE_MB;
+  static final long ONE_TB = 1000 * ONE_GB;
+  static final long MB_PER_SEC = 120;
+
+  static final String COLUMN_FAMILY = "cf";
```
telpirion, line 56: Is this a value that a developer/customer might want to customize, depending on their use case?

bigtable/beam/bulk-data-generator/src/main/java/bigtable/BulkWrite.java
```diff
@@ -0,0 +1,201 @@
...
+          .setInstanceId(options.getBigtableInstanceId())
+          .build();
+
+      BigtableTableAdminClient adminClient = BigtableTableAdminClient.create(adminSettings);
+
+      int clusterNodeCount = getClusterNodeCount(options.getProject(),
+          options.getBigtableInstanceId());
+      List<String> newTableIds = getNewTableIds(adminClient, options.getBigtableSize());
+
+      if (newTableIds.isEmpty()) {
```
telpirion, line 77: NIT: Add a comment here that explains why you want the method to bail if this check fails. I would also describe the circumstances _why_ this check would fail too.

bigtable/beam/bulk-data-generator/src/main/java/bigtable/BulkWrite.java
```diff
@@ -0,0 +1,201 @@
...
+        return;
+      }
+
+      long numRows = (long) ((TB_PER_TABLE * ONE_TB) / (MB_PER_ROW * ONE_MB));
+      long rate = clusterNodeCount * MB_PER_SEC / newTableIds.size();
+
+      String generateLabel = String
```
telpirion, line 84: NIT: I would move the code that instantiates `generateLabel` and `mutationLabel` inside of the call to `System.out.println`, unless you're going to reuse these two strings.

bigtable/beam/bulk-data-generator/src/main/java/bigtable/BulkWrite.java
```diff
@@ -0,0 +1,201 @@
...
+  static final long MB_PER_SEC = 120;
+
+  static final String COLUMN_FAMILY = "cf";
+  public static final String TABLE_PREFIX = "data-";
+  static final SecureRandom random = new SecureRandom();
+
+  public static void main(String[] args) {
+    BigtableOptions options =
+        PipelineOptionsFactory.fromArgs(args).withValidation().as(BigtableOptions.class);
+    Pipeline p = Pipeline.create(options);
```
telpirion, line 63: Use a more descriptive variable name, e.g. `pipeline`.

bigtable/beam/bulk-data-generator/src/main/java/bigtable/BulkWrite.java
```diff
@@ -0,0 +1,201 @@
...
+                .build()));
+      }
+
+      p.run();
+    } catch (IOException e) {
+      System.out.println("Unable to connect to Bigtable instance.");
+    }
+  }
+
+  private static List<String> getNewTableIds(BigtableTableAdminClient adminClient,
```
telpirion, line 114: Since this code isn't provided inline with the sample, please add a `//` comment describing what it does.

bigtable/beam/bulk-data-generator/src/main/java/bigtable/BulkWrite.java
```diff
@@ -0,0 +1,201 @@
...
+        newTableIds.add(tableId);
+
+        System.out.println(tableId);
+      }
+    }
+
+    return newTableIds;
+  }
+
+  private static int getClusterNodeCount(String projectId, String instanceId) {
```
telpirion, line 145: Same as above, please provide a `//` comment describing what this method does.

bigtable/beam/bulk-data-generator/src/main/java/bigtable/BulkWrite.java
```diff
@@ -0,0 +1,201 @@
...
+    try {
+      BigtableClusterUtilities clusterUtility = BigtableClusterUtilities
+          .forInstance(projectId, instanceId);
+      Cluster cluster = clusterUtility.getSingleCluster();
+      String clusterId = new BigtableClusterName(cluster.getName()).getClusterId();
+      String zoneId = BigtableClusterUtilities.getZoneId(cluster);
+      int clusterNodeCount = clusterUtility.getClusterNodeCount(clusterId, zoneId);
+      System.out.println("Cluster size " + clusterNodeCount);
+      return clusterNodeCount;
+    } catch (IOException | GeneralSecurityException e) {
```
telpirion, line 155: Please move the `IOException` and `GeneralSecurityException` up to the method signature.

bigtable/beam/bulk-data-generator/src/main/java/bigtable/BulkWrite.java
```diff
@@ -0,0 +1,201 @@
...
+
+        System.out.println(tableId);
+      }
+    }
+
+    return newTableIds;
+  }
+
+  private static int getClusterNodeCount(String projectId, String instanceId) {
+    try {
```
telpirion, line 146: Remove `try/catch`.

bigtable/beam/bulk-data-generator/src/main/java/bigtable/BulkWrite.java
```diff
@@ -0,0 +1,201 @@
...
+      int clusterNodeCount = clusterUtility.getClusterNodeCount(clusterId, zoneId);
+      System.out.println("Cluster size " + clusterNodeCount);
+      return clusterNodeCount;
+    } catch (IOException | GeneralSecurityException e) {
+      System.out.println("Unable to get cluster size. Treating as single-node cluster.");
+      return 1;
+    }
+  }
+
+  static class CreateMutationFn extends DoFn<Long, Mutation> {
```
telpirion, line 161: NIT: I would add a comment here that describes why/how you need to create this subclass of `DoFn`.

bigtable/beam/bulk-data-generator/src/test/java/bigtable/BulkWriteTest.java
```diff
@@ -0,0 +1,115 @@
...
+        BigtableInstanceAdminClient.create(projectId)) {
+      instanceAdmin.deleteInstance(INSTANCE_ID);
+    } catch (IOException e) {
+      System.out.println("Error during AfterClass while deleting instance: \n" + e.toString());
+    }
+  }
+
+  @Test
+  public void testBulkWrite() {
+    BulkWrite.main(
```
telpirion, line 91: As I mentioned in a previous comment, I would consider rewriting this test so that you don't need to pass the test variables into your test as args to your `main()` method.

Here's an example of a test from the Vertex AI samples:
https://github.com/googleapis/java-aiplatform/blob/master/samples/snippets/src/test/java/aiplatform/CreateTrainingPipelineImageObjectDetectionSampleTest.java#L99-L101

bigtable/beam/bulk-data-generator/src/main/java/BulkWrite.java
billyjacobson: I can't do that because building the bigtable admin settings can throw an IOException

billyjacobson: The only error create throws is IOException, so I don't need the try if I have the throws IOException to Main

bigtable/beam/bulk-data-generator/src/main/java/bigtable/BulkWrite.java
billyjacobson: Breaking the file won't really work since I need to run the code on dataflow during the test or it will run for a very long time.

Re: column family: This isn't really something developers are going to run since it just puts random data into Bigtable, it's more of a tool for demos. This also creates the tables so it will always have that family.

bigtable/beam/bulk-data-generator/src/main/java/bigtable/BulkWrite.java
billyjacobson: p is the standard name for apache beam pipelines

bigtable/beam/bulk-data-generator/src/main/java/bigtable/BulkWrite.java
billyjacobson: I need to print them out for the test

bigtable/beam/bulk-data-generator/src/main/java/bigtable/BulkWrite.java
billyjacobson: how am I supposed to default the value without catching the errors?

billyjacobson: figured out

bigtable/beam/bulk-data-generator/src/main/java/bigtable/BulkWrite.java
billyjacobson: this is a pretty standard apache beam pattern

bigtable/beam/bulk-data-generator/src/test/java/bigtable/BulkWriteTest.java
billyjacobson: done

bigtable/beam/bulk-data-generator/src/main/java/bigtable/BulkWrite.java
telpirion: Okay

telpirion (approved):

lesv (approved):

PR #5645, opened 2021-07-20
chore: auto-approve.yml

lesv:
- [x ] Please **merge** this PR for me once it is approved.
@averikitsch FYI
@sofisl FYI
@kurtisvg FYI
@eaball35 FYI

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -12,4 +12,4 @@
 /iot/                           @gcseh @GoogleCloudPlatform/api-iot @GoogleCloudPlatform/java-samples-reviewers
 /logging/                       @GoogleCloudPlatform/observability-devx @GoogleCloudPlatform/java-samples-reviewers
 /storage/**/*.java              @GoogleCloudPlatform/cloud-storage-dpes @GoogleCloudPlatform/java-samples-reviewers
-.github/auto-approve.yml        @googleapis/github-automation
+.github/auto-approve.yml        @sofisl
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -12,4 +12,4 @@
 /iot/                           @gcseh @GoogleCloudPlatform/api-iot @GoogleCloudPlatform/java-samples-reviewers
 /logging/                       @GoogleCloudPlatform/observability-devx @GoogleCloudPlatform/java-samples-reviewers
 /storage/**/*.java              @GoogleCloudPlatform/cloud-storage-dpes @GoogleCloudPlatform/java-samples-reviewers
-.github/auto-approve.yml        @sofisl
+.github/auto-approve.yml        @googleapis/github-automation/ @sofisl
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -12,3 +12,4 @@
 /iot/                           @gcseh @GoogleCloudPlatform/api-iot @GoogleCloudPlatform/java-samples-reviewers
 /logging/                       @GoogleCloudPlatform/observability-devx @GoogleCloudPlatform/java-samples-reviewers
 /storage/**/*.java              @GoogleCloudPlatform/cloud-storage-dpes @GoogleCloudPlatform/java-samples-reviewers
+.github/auto-approve.yml        @googleapis/github-automation
--- a/.github/auto-approve.yml
+++ b/.github/auto-approve.yml
@@ -20,7 +20,7 @@ rules:
 # checking the file paths. 
   changedFiles:
   - "/pom\\.xml$"
-  - "/gradle\\.build$"
+#  - "/gradle\\.build$"
   
 # (Optional) The max amount of files changed in a PR. 
 # maxFiles: 
--- a/.github/auto-approve.yml
+++ b/.github/auto-approve.yml
@@ -13,7 +13,7 @@ rules:
 - author: "renovate-bot"
 
 # Title for PRs from renovate-bot MUST start with `fix(deps):` or `chore(deps)"`.
-  title: "^((fix)|(chore))\\(deps\\):"
+  title: "^(fix\\(deps\\)|chore\\(deps\\)):"
 
 # (Optional) The only changed file paths in an incoming PR MUST end with `pom.xml`. These are listed in regex. If this property 
 # is omitted, auto-approve bot will allow PRs that match the author and title to merge, without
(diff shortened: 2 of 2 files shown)


lesv:
Actually - hold off on merging until Sofi reviews.

sofisl (approved):

.github/auto-approve.yml
```diff
@@ -0,0 +1,25 @@
...
+  title: "^(fix\\(deps\\)|chore\\(deps\\)):"
+
+# (Optional) The only changed file paths in an incoming PR MUST end with `package.json`. These are listed in regex. If this property 
+# is omitted, auto-approve bot will allow PRs that match the author and title to merge, without
+# checking the file paths. 
+  changedFiles:
+  - "/pom\\.xml$"
+
+# (Optional) The max amount of files changed in a PR. 
+#  maxFiles: 75
```
sofisl, line 25: Eventually, this shouldn't be commented out. You can also leave it blank if you don't want to impose a limit.

.github/CODEOWNERS
```diff
@@ -12,3 +12,4 @@
 /iot/                           @gcseh @GoogleCloudPlatform/api-iot @GoogleCloudPlatform/java-samples-reviewers
 /logging/                       @GoogleCloudPlatform/observability-devx @GoogleCloudPlatform/java-samples-reviewers
 /storage/**/*.java              @GoogleCloudPlatform/cloud-storage-dpes @GoogleCloudPlatform/java-samples-reviewers
+.github/auto-approve.yml        @googleapis/github-automation
```
kurtisvg, line 15: This line doesn't work because @googleapis is a cross-org team. You'll need to create one in the GCP org. (otherwise, maybe it makes more sense to add repo owners here)

eaball35 (approved):

.github/auto-approve.yml
```diff
@@ -0,0 +1,25 @@
...
+# changedFiles nor maxFiles are required, but they will be enforced on a PR if they are set.
+# Each rule only has the allowed values listed below; there is no mixing of values between rules,
+# nor can any values be replaced. To understand the full set of rules for the JSON file, check out
+# src/valid-pr-schema.json
+
+# Option 5: incoming PRs from renovate-bot
+- author: "renovate-bot"
+
+# Title for PRs from renovate-bot MUST start with `fix(deps):` or `chore(deps)"`.
+  title: "^(fix\\(deps\\)|chore\\(deps\\)):"
```
kurtisvg, line 16:
```suggestion
  title: "^((fix)|(chore))\\(deps\\):"
```

.github/auto-approve.yml
```diff
@@ -0,0 +1,25 @@
...
+
+# (Optional) The only changed file paths in an incoming PR MUST end with `package.json`. These are listed in regex. If this property
```
kurtisvg, line 18: [](url)
```suggestion
# (Optional) The only changed file paths in an incoming PR MUST end with `pom.xml`. These are listed in regex. If this property 
```

.github/auto-approve.yml
kurtisvg: Also: Is this regex specifically for _prefix_? Or should it have a `.*` at the end?

.github/auto-approve.yml
```diff
@@ -0,0 +1,25 @@
...
+- author: "renovate-bot"
+
+# Title for PRs from renovate-bot MUST start with `fix(deps):` or `chore(deps)"`.
+  title: "^(fix\\(deps\\)|chore\\(deps\\)):"
+
+# (Optional) The only changed file paths in an incoming PR MUST end with `package.json`. These are listed in regex. If this property 
+# is omitted, auto-approve bot will allow PRs that match the author and title to merge, without
+# checking the file paths. 
+  changedFiles:
+  - "/pom\\.xml$"
```
kurtisvg, line 22: This is specifically suffix? or should it be `.*/pom.xml$`?

Tangential - this "changedFiles" inclusive or exclusive or the rest of the items in the list?

kurtisvg (changes requested):

.github/auto-approve.yml
sofisl: Yes, it is specifically a prefix!

.github/auto-approve.yml
sofisl: This is specifically a suffix, as (I assume) we don't know how many directory levels deep pom.xml should be. 

For renovate-bot PRs, changedFiles is inclusive - no file in the PR can violate the regex rules of the naming.

.github/auto-approve.yml
kurtisvg: Great -- thanks! As a fly by API critique, it would be nice if that was obvious from the name  (e.g. `titlePrefix`) or if it's too late for breaking changes, that the comment in the example makes that more clear.

.github/CODEOWNERS
sofisl: Oh, I didn't realize that. @lesv, would you mind adding a line below for `.github/auto-approve.yml` for @sofisl? That way, I'll be notified if for changes to auto-approve.yml. However, you should leave that line for now otherwise the auto-approve.yml file will fail, and we need that restriction for googleapis libraries.

[Tracking issue](https://github.com/googleapis/repo-automation-bots/issues/2290)

.github/auto-approve.yml
kurtisvg: To clarify - if we put in:
```
- "/pom\\.xml@"
- "/gradle\\.build$"
```

Would it accept PRs that have both pom _and_ gradle files OR would it be both pom _or_ gradle files?

lesv: Should be AND

lesv:
PTAL

lesv:
@sofisl The hack of adding just you doesn't' actually work.  I'll see how ridged it is.

.github/auto-approve.yml
kurtisvg: If it's AND do we want to list then as two separate regexs? Does that mean it'll only approve PRs with gradle AND poms? What happens if it's just poms?

lesv: @sofisl At some point we should add gradle.build, but that's not updating ATM.
@kurtisvg The app doesn't like gradle.build

lesv:
@kurtisvg PTAL

kurtisvg (approved):

PR #5695, opened 2021-07-24
Compute pagination samples

Sita04:
Fixes #issue

> Adding samples to demonstrate pagination in GCE.

### DIFF
--- a/compute/cloud-client/src/main/java/compute/ListAllInstances.java
+++ b/compute/cloud-client/src/main/java/compute/ListAllInstances.java
@@ -40,10 +40,8 @@ public class ListAllInstances {
     // the `instancesClient.close()` method on the client to 
     // safely clean up any remaining background resources.
     try (InstancesClient instancesClient = InstancesClient.create()) {
-      // Listing all instances for the project
 
-      // Requesting a list of Instances, aggregated by the zone they are located in.
-      // The max_results parameter tells the API to return no more than 5 results per response page.
+      // Use the `setMaxResults` parameter to limit the number of results that the API returns per response page.
       AggregatedListInstancesRequest aggregatedListInstancesRequest = AggregatedListInstancesRequest
           .newBuilder()
           .setProject(project)
@@ -53,9 +51,9 @@ public class ListAllInstances {
       InstancesClient.AggregatedListPagedResponse response = instancesClient
           .aggregatedList(aggregatedListInstancesRequest);
 
-      // Despite telling the API to return only 5 instances per page, you don't need to handle the
-      // pagination system yourself. The returned AggregatedListPager object handles it
-      // automatically for you, requesting next pages as you iterate over the results.
+      // Despite using the `setMaxResults` parameter, you don't need to handle the pagination
+      // yourself. The returned `AggregatedListPager` object handles pagination
+      // automatically, requesting next pages as you iterate over the results.
       for (Map.Entry<String, InstancesScopedList> zoneInstances : response.iterateAll()) {
         // Instances scoped by each zone
         String zone = zoneInstances.getKey();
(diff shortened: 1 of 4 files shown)


compute/cloud-client/src/main/java/compute/ListImages.java
```diff
@@ -0,0 +1,102 @@
...
+    // Initialize client that will be used to send requests. This client only needs to be created
+    // once, and can be reused for multiple requests. After completing all of your requests, call
+    // the `instancesClient.close()` method on the client to
+    // safely clean up any remaining background resources.
+    try (ImagesClient imagesClient = ImagesClient.create()) {
+
+      // Listing only non-deprecated images to reduce the size of the reply.
+      ListImagesRequest imagesRequest = ListImagesRequest.newBuilder()
+          .setProject(project)
+          .setMaxResults(2)
```
kolea2, line 55: do we need to set this? If this example is showing the default, should we be setting parameters like this?

compute/cloud-client/src/main/java/compute/ListImages.java
```diff
@@ -0,0 +1,102 @@
...
+          .setMaxResults(pageSize)
+          .setFilter("deprecated.state != DEPRECATED")
+          .build();
+
+      // Use the `iteratePages` attribute of returned iterable to have more granular control of
+      // iteration over paginated results from the API. Each time you want to access the
+      // next page, the library retrieves that page from the API.
+      int pageNumber = 1;
+      for (ListPage page : imagesClient.list(imagesRequest).iteratePages()) {
+        System.out.println("Page Number: " + pageNumber++);
```
kolea2, line 94: Shouldn't we be incrementing after this?

compute/cloud-client/src/main/java/compute/ListAllInstances.java
```diff
@@ -39,9 +40,21 @@ public static void listAllInstances(String project) throws IOException {
     // the `instancesClient.close()` method on the client to 
     // safely clean up any remaining background resources.
     try (InstancesClient instancesClient = InstancesClient.create()) {
-      // Listing all instances for the project
```
kolea2, line 42: to confirm, are we replacing this example? Or should there be two?

compute/cloud-client/src/main/java/compute/ListImages.java
Sita04: Yes, I am incrementing the pageNumber as part of the sysout statement!

compute/cloud-client/src/main/java/compute/ListAllInstances.java
Sita04: No, we are not replacing it. I felt that the comment was redundant and hence removed it.

compute/cloud-client/src/main/java/compute/ListImages.java
Sita04: The idea of this snippet is to demo the abstraction involved in fetching images even though the maxResults is set. Even though, the maxResult here is set to 2 and 2 images are fetched per call, this is abstracted when the images are listed.

kolea2: How many images are fetched per call without setting this? Would it be valuable to show both? I am just a little confused having one example explicitly setting `maxResults`, and one arbitrarily picking 2.

Sita04: If the method is not set, there would be a single abstracted call, which will fetch the images.
By setting setMaxResults, we are explicitly mentioning the number of images to be fetched for each page and inorder to get additional images, we can use the next page token.

But here, that complexity is abstracted in the code sample, even though the underlying mechanism will involve calling multiple next page tokens. 

The first method abstracts in entirety the pagination concept, whereas the second method shows how to iterate with the returned pages.

kolea2 (approved):

PR #5703, opened 2021-07-28
fix(compute): instance_delete modified operation status monitoring

Sita04:

### DIFF
--- a/compute/cloud-client/src/main/java/compute/DeleteInstance.java
+++ b/compute/cloud-client/src/main/java/compute/DeleteInstance.java
@@ -19,10 +19,13 @@ package compute;
 // [START compute_instances_delete]
 
 import com.google.cloud.compute.v1.InstancesClient;
+// [START compute_instances_operation_check]
 import com.google.cloud.compute.v1.Operation;
 import com.google.cloud.compute.v1.Operation.Status;
+import com.google.cloud.compute.v1.WaitZoneOperationRequest;
 import com.google.cloud.compute.v1.ZoneOperationsClient;
 import java.io.IOException;
+// [END compute_instances_operation_check]
 
 public class DeleteInstance {
 
@@ -42,22 +45,32 @@ public class DeleteInstance {
     // once, and can be reused for multiple requests. After completing all of your requests, call
     // the `instancesClient.close()` method on the client to safely
     // clean up any remaining background resources.
-    try (InstancesClient instancesClient = InstancesClient.create();
-        ZoneOperationsClient zoneOperationsClient = ZoneOperationsClient.create()) {
+    try (InstancesClient instancesClient = InstancesClient.create()) {
 
       System.out.println(String.format("Deleting instance: %s ", instanceName));
       // Describe which instance is to be deleted.
-      Operation response = instancesClient.delete(project, zone, instanceName);
+      Operation operation = instancesClient.delete(project, zone, instanceName);
 
       // [START compute_instances_operation_check]
-      if (response.getStatus() == Status.RUNNING) {
+      // 'operation' refers to an instance of 'Operation'.
+      if (operation.getStatus() == Status.RUNNING) {
+        // Create a zone operations client.
+        ZoneOperationsClient zoneOperationsClient = ZoneOperationsClient.create();
+
+        // Create the wait request.
+        WaitZoneOperationRequest waitZoneOperationRequest = WaitZoneOperationRequest.newBuilder()
+            .setProject(project)
+            .setZone(zone)
+            .setOperation(String.valueOf(operation.getId())).build();
+
         // Wait for the delete operation to complete; default timeout is 2 mins
-        response = zoneOperationsClient.wait(project, zone, String.valueOf(response.getId()));
+        operation = zoneOperationsClient.wait(waitZoneOperationRequest);
+        zoneOperationsClient.close();
       }
       // [END compute_instances_operation_check]
 
-      if (response.hasError()) {
-        System.out.println("Instance deletion failed ! ! " + response.getError());
+      if (operation.hasError()) {
+        System.out.println("Instance deletion failed ! ! " + operation.getError());
         return;
       }
       System.out.println("####### Instance deletion complete #######");
--- a/compute/cloud-client/src/main/java/compute/DeleteInstance.java
+++ b/compute/cloud-client/src/main/java/compute/DeleteInstance.java
@@ -19,13 +19,11 @@ package compute;
 // [START compute_instances_delete]
 
 import com.google.cloud.compute.v1.InstancesClient;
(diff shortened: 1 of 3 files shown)


compute/cloud-client/src/main/java/compute/DeleteInstance.java
```diff
@@ -19,10 +19,13 @@
 // [START compute_instances_delete]
 
 import com.google.cloud.compute.v1.InstancesClient;
+// [START compute_instances_operation_check]
```
kurtisvg, line 22: Do we need to include the imports separately? Ideally this should be part of the whole snippet: go/cloud-sql-dpe-terraform

compute/cloud-client/src/main/java/compute/DeleteInstance.java
```diff
@@ -19,10 +19,13 @@
...
 import com.google.cloud.compute.v1.Operation;
 import com.google.cloud.compute.v1.Operation.Status;
+import com.google.cloud.compute.v1.WaitZoneOperationRequest;
 import com.google.cloud.compute.v1.ZoneOperationsClient;
 import java.io.IOException;
+// [END compute_instances_operation_check]
```
kurtisvg, line 28: this looks like a different region tag than delete? Should we just make it a new sample?

kurtisvg (changes requested):

compute/cloud-client/src/main/java/compute/DeleteInstance.java
Sita04: In order to provide more context (on import packages) to the customers using the sample, this approach has been adopted, across languages for this particular region tag.
Please refer to: https://github.com/GoogleCloudPlatform/php-docs-samples/blob/master/compute/cloud-client/instances/src/delete_instance.php

Making the "compute_instances_operation_check" as a standalone snippet proved to be complex to demonstrate and wasn't intuitive, hence went with the above approach.

compute/cloud-client/src/main/java/compute/DeleteInstance.java
Sita04: ditto as above. Thanks!

compute/cloud-client/src/main/java/compute/DeleteInstance.java
kurtisvg: Do you have a one pager or write up somewhere describing why this is necessary? This is contradictory to go/code-snippets-style, which says "A code snippet demonstrates one task and only one task."

Sita04: The operation_check snippet is not a standalone snippet and would always have to be used with another API, say create or delete.
If we make this "operation check" an independent snippet, we'll anyway have to reference it from other snippets, which would inturn be a contradiction to the snippets-style.

I can think of a solution to this:
Could we conform to the snippets-style by increasing a little bit of redundancy? 
1. We can keep the "operation check" in the create and delete snippets as such, but remove the region tags.
2. Create an new snippet for "operation check" and that can have the region tags.

wdyt?

kurtisvg: That SGTM. Feel free to ping me when you need me to review.

compute/cloud-client/src/main/java/compute/WaitForOperation.java
```diff
@@ -0,0 +1,76 @@
...
+          System.out.println("Error in executing the operation ! ! " + operation.getError());
+          return;
+        }
+      }
+      System.out.println("Operation executed successfully !");
+    } catch (SocketTimeoutException e) {
+      // Handle SocketTimeoutException which is being thrown as UnknownException.
+      // (Operation will run to completion in the background)
+      System.out.println("Operation executed successfully !");
+    }
```
kurtisvg, line 73: This doesn't seem right to me. `SocketTimeoutException` extends IOException - do we really need to catch it here? 

It also doesn't seem like we have any guarantee the Operation has actually executed successfully at this point, just that we're no longer waiting for the end of the operation.

kurtisvg (changes requested):

compute/cloud-client/src/main/java/compute/WaitForOperation.java
Sita04: This is a temporary workaround for LRO. SocketTimeoutException is masked as UnknownException while the operation is running at the background.

kurtisvg: I don't understand what this means. Why is this a temporary workaround? What's causing the `SocketTimeoutException`?

Sita04: The default wait timeout is 2 mins. But often, the operation takes more time (> 2mins) to execute. Hence, it throws a SocketTimeout exception.

To account for increased operation times, we can either increase the default timeout (to something >2mins) or resort to LRO. (as a solution to the above problem).
But, both these options are currently not supported by the library.

kurtisvg: To clarify - the `SocketTimeoutException` is thrown when `wait` exceeds 2 minutes? Or is it thrown for other reasons?

Checking the API docs for [zoneOperation.wait](https://cloud.google.com/compute/docs/reference/rest/v1/zoneOperations/wait), it says that "it waits for no more than the default deadline (2 minutes) and then returns the current state of the operation, which might be DONE or still in progress".

So this raises to me the following concerns: 
1. The comment says "// Wait for the operation to complete; default timeout is 2 mins." - however, it doesn't seem like you can wait anything but 2 minutes? 
2. So where is `SocketTimeoutException` coming from? Shouldn't it be returning an operation object?
3. Why do we say "Operation executed successfully" if the operation hasn't actually executed? Shouldn't we use a loop to hit the endpoint until it's successful? 

@lesv FYI

kurtisvg (changes requested):

compute/cloud-client/src/main/java/compute/WaitForOperation.java
lesv: Wow - this seems wrong on so many levels.  An exception should never be the success case.  It's in the very word - Exception - something out of the ordinary.  If this is the ordinary for just a short bit, you should comment that it will change, and we should try not to publish since it's a really bad thing.

lesv: Re-reading the thread, we are trying to say *THIS API IS NOT STABLE AND IS GOING TO CHANGE - THIS CODE IS UNSTABLE BUT IS BEING SHOWN TO ALLOW YOU TO ACCESS THE NEW API*.  As I write this, and realize that it's for the Compute API, I realize that we should just fix the API before publishing the samples.

PR #5771, opened 2021-08-11
fix(compute): modified LRO wrappers

Sita04:
1. Modified LRO wrappers to wait for the operation completion
2. Included wait for operation as a separate sample

### DIFF
--- a/compute/cloud-client/src/main/java/compute/CreateInstance.java
+++ b/compute/cloud-client/src/main/java/compute/CreateInstance.java
@@ -88,20 +88,19 @@ public class CreateInstance {
       // Insert the instance in the specified project and zone.
       Operation response = instancesClient.insert(project, zone, instanceResource);
 
-      if (response.getStatus() == Status.RUNNING) {
-        // Wait for the create operation to complete; default timeout is 2 mins
-        response = zoneOperationsClient.wait(project, zone, String.valueOf(response.getId()));
+      // Wait for the create operation to complete.
+      // Timeout is set at 180000ms or 3 minutes.
+      long startTime = System.currentTimeMillis();
+      while (response.getStatus() != Status.DONE
+          && System.currentTimeMillis() - startTime < 180000) {
+        // The default wait timeout is 2 mins.
+        response = zoneOperationsClient.get(project, zone, String.valueOf(response.getId()));
       }
 
-      if (response.hasError()) {
-        System.out.println("Instance creation failed ! ! " + response.getError());
+      if (response.getStatus() != Status.DONE || response.hasError()) {
+        System.out.println("Instance creation failed ! ! " + response);
         return;
       }
-      System.out.println("####### Instance creation complete #######");
-
-    } catch (com.google.api.gax.rpc.UnknownException e) {
-      // Handle SocketTimeoutException which is being thrown as UnknownException.
-      // (Instance creation process will run to completion in the background)
       System.out.println("####### Instance creation complete #######");
     }
   }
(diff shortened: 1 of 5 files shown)


compute/cloud-client/src/main/java/compute/CreateInstance.java
```diff
@@ -88,21 +89,21 @@ public static void createInstance(String project, String zone, String instanceNa
...
-      if (response.getStatus() == Status.RUNNING) {
-        // Wait for the create operation to complete; default timeout is 2 mins
-        response = zoneOperationsClient.wait(project, zone, String.valueOf(response.getId()));
+      // Wait for the create operation to complete.
+      // Timeout is set at 180000ms or 3 minutes.
+      long startTime = System.currentTimeMillis();
+      while (response.getStatus() != Status.DONE
+          && System.currentTimeMillis() - startTime < 180000) {
+        // The default wait timeout is 2 mins.
+        response = zoneOperationsClient.get(project, zone, String.valueOf(response.getId()));
```
kurtisvg, line 98: This is using `.get` instead of `.wait`. [Should we be using wait instead](https://cloud.google.com/compute/docs/reference/rest/v1/zoneOperations/wait)? 

If we are using wait and it has a 2 minute timeout, we can probably just try a max of 2-3 time or something and get the same effect.

compute/cloud-client/src/main/java/compute/CreateInstance.java
```diff
@@ -88,21 +89,21 @@ public static void createInstance(String project, String zone, String instanceNa
...
       Operation response = instancesClient.insert(project, zone, instanceResource);
 
-      if (response.getStatus() == Status.RUNNING) {
-        // Wait for the create operation to complete; default timeout is 2 mins
-        response = zoneOperationsClient.wait(project, zone, String.valueOf(response.getId()));
+      // Wait for the create operation to complete.
+      // Timeout is set at 180000ms or 3 minutes.
+      long startTime = System.currentTimeMillis();
+      while (response.getStatus() != Status.DONE
+          && System.currentTimeMillis() - startTime < 180000) {
```
kurtisvg, line 96: [Please use the `java.time` package here.](https://github.com/GoogleCloudPlatform/java-docs-samples/blob/master/SAMPLE_FORMAT.md#time)
```suggestion
      LocalTime endTime = LocalTime.now().plusMinutes(3);
      while (response.getStatus() != Status.DONE && LocalTime.now().isBefore(endTime) {
```

kurtisvg: same elsewhere

compute/cloud-client/src/main/java/compute/CreateInstance.java
kurtisvg: I don't have a strong preference here, but it does seem like `wait` would be less quota usage?

kurtisvg (changes requested):

compute/cloud-client/src/main/java/compute/CreateInstance.java
Sita04: Changed it. Thanks!

compute/cloud-client/src/main/java/compute/CreateInstance.java
Sita04: 'wait' method in Java leads to SocketTimeout exception. This issue will be fixed and released as part of LRO.
In the meantime, to demo the concept, 'get' is being used here.

kurtisvg (approved):

PR #5883, opened 2021-08-27
feat: JSON samples for Spanner JDBC

olavloite:
Adds samples for the new JSON data type for Spanner JDBC.

Do not merge yet: Depends on changes in the Spanner JDBC driver that have not yet been included in the libraries bom.

### DIFF
--- a/spanner/jdbc/src/main/java/com/example/spanner/jdbc/JsonInsertDataExample.java
+++ b/spanner/jdbc/src/main/java/com/example/spanner/jdbc/JsonInsertDataExample.java
@@ -1,0 +1,84 @@
+/*
+ * Copyright 2021 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.example.spanner.jdbc;
+
+// [START spanner_jdbc_json_insert_data]
+
+import com.google.cloud.spanner.Value;
+
+import java.sql.Connection;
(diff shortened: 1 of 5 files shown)


zoercai (approved):

spanner/jdbc/src/main/java/com/example/spanner/jdbc/JsonInsertDataExample.java
```diff
@@ -0,0 +1,86 @@
...
+      try (PreparedStatement ps =
+          connection.prepareStatement("INSERT INTO Venues\n"
+                      + "(VenueId, VenueDetails)\n"
+                      + "VALUES\n"
+                      + "(?, ?)")) {
+        for (Venue venue : VENUES) {
+          ps.setLong(1, venue.venueId);
+          // Tell the JDBC driver that we want to set a JSON value and not a STRING value
+          // by specifying the JsonType SQL type.
+          ps.setObject(2, venue.venueDetails, JsonType.INSTANCE);
```
thiagotnunes, line 77: Do you think this is a more natural way to specify the type other than `ps.setObject(2, Value.json(venue.venueDetails))`?

thiagotnunes (approved):

spanner/jdbc/src/main/java/com/example/spanner/jdbc/JsonInsertDataExample.java
olavloite: Not necessarily, and I was a little bit in doubt here what is the best to show:
- `ps.setObject(2, Value.json(venue.venueDetails))` is the easiest and prettiest, and would be what I would recommend that users that have compile time access to the Spanner JDBC driver should use.
- Users who do not have access to the JDBC driver on compile time (that is; the application only works with the generic JDBC interfaces and the driver is loaded dynamically by the environment) must use one of the other options.

Maybe we should create two different samples for this to show both. That is what we decided to do for creating a retry loop for aborted transactions:
- JDBC only: https://github.com/GoogleCloudPlatform/java-docs-samples/blob/master/spanner/jdbc/src/main/java/com/example/spanner/jdbc/TransactionWithRetryLoopUsingOnlyJdbcExample.java
- With compile time access to the JDBC driver: https://github.com/GoogleCloudPlatform/java-docs-samples/blob/master/spanner/jdbc/src/main/java/com/example/spanner/jdbc/TransactionWithRetryLoopExample.java

PR #5964, opened 2021-09-08
chore(functions/logging): include trace code for HTTP requests

ace-n:
Revision to #5908 that includes the trace code.

(Java [does not support](https://stackoverflow.com/questions/5113374) checking for undefined variables [as it throws a compiler error], so instead I've added a comment telling end-users to delete the erroring code.)

### DIFF
--- a/run/logging-manual/src/main/java/com/example/cloudrun/App.java
+++ b/run/logging-manual/src/main/java/com/example/cloudrun/App.java
@@ -42,11 +42,12 @@ public class App {
         (req, res) -> {
           // [START cloudrun_manual_logging]
           // [START run_manual_logging]
-          // [START functions_manual_logging]
           // Build structured log messages as an object.
           Object globalLogFields = null;
-          // [END functions_manual_logging]
+
           // Add log correlation to nest all log messages beneath request log in Log Viewer.
+          // TODO(developer): delete this code if you're creating a Cloud
+          //                  Function and it is *NOT* triggered by HTTP.
           String traceHeader = req.headers("x-cloud-trace-context");
           if (traceHeader != null && project != null) {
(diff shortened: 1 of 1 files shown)


eaball35 (approved):

ace-n:
@grayside please merge for me once you LGTM - thanks! 🙂

run/logging-manual/src/main/java/com/example/cloudrun/App.java
```diff
@@ -42,11 +42,12 @@ public static void main(String[] args) {
...
-          // [START functions_manual_logging]
           // Build structured log messages as an object.
           Object globalLogFields = null;
-          // [END functions_manual_logging]
+
           // Add log correlation to nest all log messages beneath request log in Log Viewer.
+          // TODO(developer): delete this code if you're creating a Cloud
+          //                  Function and it is *NOT* triggered by HTTP.
           String traceHeader = req.headers("x-cloud-trace-context");
           if (traceHeader != null && project != null) {
```
averikitsch, line 52: Does `traceHeader != null` not prevent the code to be deleted from executing?

ace-n: It does, but if `req` is not defined I'm thinking this would throw a compiler error.

lesv (approved):

run/logging-manual/src/main/java/com/example/cloudrun/App.java
eamonnmcmanus: `req` _is_ declared, on line 42. The issue is that _`get`_ is not declared, so there's no way to know what the `(req, res) -> {...}` lambda is supposed to mean.

PR #5971, opened 2021-09-09
Update GCFv2 PubSub sample

averikitsch:
Fixes #issue

### DIFF
--- a/functions/v2/pubsub/subscribe-to-topic/src/main/java/functions/SubscribeToTopic.java
+++ b/functions/v2/pubsub/subscribe-to-topic/src/main/java/functions/SubscribeToTopic.java
@@ -17,15 +17,10 @@
 package functions;
 
 // [START functions_cloudevent_pubsub]
-import static io.cloudevents.core.CloudEventUtils.mapData;
-
-import com.fasterxml.jackson.databind.DeserializationFeature;
-import com.fasterxml.jackson.databind.ObjectMapper;
 import com.google.cloud.functions.CloudEventsFunction;
+import com.google.gson.Gson;
 import functions.eventpojos.PubSubBody;
 import io.cloudevents.CloudEvent;
-import io.cloudevents.core.data.PojoCloudEventData;
-import io.cloudevents.jackson.PojoCloudEventDataMapper;
 import java.nio.charset.StandardCharsets;
 import java.util.Base64;
 import java.util.logging.Logger;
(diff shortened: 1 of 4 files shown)


averikitsch:
@lesv I am curious what you think about this change from a idiomatic Java perspective. I like the idea of using the CloudEvent sdk and "staying within that ecosystem", but for a simpler sample it may add more complexity than the more common GSON library. I also have to make sure I specify ignore unknown fields so the unmarshalling doesn't break, but I don't think I have to do that for GSON.

lesv:
We generally prefer GSon

functions/v2/pubsub/subscribe-to-topic/src/main/java/functions/SubscribeToTopic.java
```diff
@@ -31,13 +30,19 @@
...
-    Gson gson = new Gson();
-    PubSubBody pubSubBody = gson.fromJson(new String(event.getData().toBytes()), PubSubBody.class);
-    String messageString =
-        new String(
-            Base64.getDecoder().decode(pubSubBody.getMessage().getData()), StandardCharsets.UTF_8);
+    logger.info("Event: " + event.getId());
+    logger.info("Event Type: " + event.getType());
 
-    logger.info(messageString);
+    if (event.getData() != null) {
```
ace-n, line 36: Would it be better to use "guards" here? i.e:

```
if (event.getData() == null) {
  return;
}

...
```

Then you don't have to wrap this in an `if` statement.

ace-n (changes requested):
Minor nit that might shrink the diff a bit. 🙂

functions/v2/pubsub/subscribe-to-topic/src/main/java/functions/SubscribeToTopic.java
averikitsch: I have no opinion here. To me, these are essentially the same

ace-n: Gotcha - I'm still in favor of the guard approach, but I'll let @lesv break the tie.

lesv: Easier to understand inside the if, random returns, especially when wrapped (google style) are confusing.
Google doesn't allow `if (event.getData() == null) return`.  

FYI: Java 8 has the [Guards](https://docs.oracle.com/javase/8/docs/api/java/security/Guard.html) interface, but it's not applicable here.

That said, i'd really prefer `if (event.hasData()) { ... }` but that's not defined, so go with as is.

lesv (approved):

PR #5973, opened 2021-09-09
fix: update workflows quickstart to use env vars for config

grant:
Workflows Quickstart: enable env vars so docs can use this format.

---

Tested locally. Examples:

```
# Project config
GOOGLE_CLOUD_PROJECT=serverless-com-demo mvn compile exec:java -Dexec.mainClass=com.example.workflows.WorkflowsQuickstart

# Full config
GOOGLE_CLOUD_PROJECT=serverless-com-demo LOCATION=us-central1 WORKFLOW=myFirstWorkflow mvn compile exec:java -Dexec.mainClass=com.example.workflows.WorkflowsQuickstart

# Test
GOOGLE_CLOUD_PROJECT=serverless-com-demo mvn clean verify
```

### DIFF
--- a/workflows/cloud-client/src/main/java/com/example/workflows/WorkflowsQuickstart.java
+++ b/workflows/cloud-client/src/main/java/com/example/workflows/WorkflowsQuickstart.java
@@ -29,13 +29,17 @@ import java.util.concurrent.ExecutionException;
 
 public class WorkflowsQuickstart {
 
+  private static final String PROJECT = System.getenv("GOOGLE_CLOUD_PROJECT");
+  private static final String LOCATION = System.getenv().getOrDefault("LOCATION", "us-central1");
+  private static final String WORKFLOW = System.getenv().getOrDefault("WORKFLOW", "myFirstWorkflow");
+
   public static void main(String... args)
       throws IOException, InterruptedException, ExecutionException {
-    // TODO(developer): Replace these variables before running the sample.
-    String projectId = "your-project-id";
-    String location = "us-central1";
-    String workflow = "myFirstWorkflow";
-    workflowsQuickstart(projectId, location, workflow);
+    if (PROJECT == null) {
+      throw new IllegalArgumentException(
+        "Environment variable 'GOOGLE_CLOUD_PROJECT' is required to run this quickstart.");
+    }
+    workflowsQuickstart(PROJECT, LOCATION, WORKFLOW);
   }
 
   private static volatile boolean finished;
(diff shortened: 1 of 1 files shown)


averikitsch:
This goes against our sample guidelines for Java

lesv:
EnvVar's make it easier to run, but your removing the pedagogical tool that actually helps folks understand how to use the sample.  I've added @kurtisvg since he participated in the UXR work that got this as our standard.

Once you've thought about that and if Kurtis comments, I'm ok with this if you insist.

grant:
> This goes against our sample guidelines for Java

Hey Averi, thanks for the comment. Can you please specify the guideline? This PR mirrors using env vars rather than have a user try the code and gett . Env vars are used by Spark, BigTable, IoT and other samples this way.

For example:
- https://github.com/GoogleCloudPlatform/java-docs-samples/blob/master/appengine-java8/tasks/quickstart/src/test/java/com/example/task/CreateTaskIT.java
- https://cloud.google.com/tasks/docs/quickstart#add_a_task_to_the_cloud_tasks_queue

The same does not run using the CloudSite instructions otherwise.

---

I can't tell if you're requesting changes or have a different solution.

averikitsch:
There are older samples that don't fit the sample guidelines found at [SAMPLE_FORMAT.md](https://github.com/GoogleCloudPlatform/java-docs-samples/blob/master/SAMPLE_FORMAT.md). We don't recommend using the CLI (exec:exec) or env vars in samples. Therefore the solution would be to revert these changes, so samples stay consistent. I'll defer to @kurtisvg though.

grant:
OK. Running this program as-is produces a strange error unlike other quickstarts.

Specifically, users running this will see the quickstart not working. This is due to `"your-project-id"` as not a valid project. Other quickstarts don't have this required code change.

For context, the doc is here:
https://cloud.google.com/workflows/docs/quickstart-client-libraries

Should we have the docs have a new step just for Java? I'm okay with any suggestion, but the [current instructions](https://github.com/GoogleCloudPlatform/java-docs-samples/blob/master/workflows/cloud-client/README.md) for running this sample are incomplete.

grant:
Looks like Kurtis is gone for a month more, is there an alternate solution we want to allow this sample to work with instructions?

averikitsch (approved):
Since this is a quickstart, I will allow the env vars + CLI. For other samples please see go/code-snippets-style#running-from-the-command-line

grant:
> Since this is a quickstart, I will allow the env vars + CLI. For other samples please see go/code-snippets-style#running-from-the-command-line

Understood, thanks!

PR #6120, opened 2021-10-01
(feat) Storage Transfer Service: add gapic samples and refactor test suite

JesseLovelace:
This adds the gapic samples for transfer service, replacing the old apiary samples as the "main samples", while marking the apiary samples as apiary. The new gapic samples will be on the main documentation page, while the apiary samples will be in a migration guide. 

This is also refactors the test suite. Previously, the tests relied on hard-coded buckets that have the correct permissions set up. This is potentially flaky, and also makes checking out the samples and running the tests difficult. Now, buckets are simply spun up, given the correct permissions, and then deleted at the end of the tests. 

Do not merge label: this shouldn't be merged until all artifacts related to the migration to gapic (i.e.: the python samples, the migration guide, the quickstart guides, the CLs publishing all these), are also ready to go


storage/storage-transfer/src/main/java/com/google/cloud/storage/storagetransfer/samples/TransferFromAws.java
```diff
@@ -69,56 +67,55 @@ public static void transferFromAws(String projectId,
...
         .setScheduleStartDate(startDate)
         .setScheduleEndDate(startDate)
-        .setStartTimeOfDay(startTime);
+        .setStartTimeOfDay(startTime)
+        .build();
 
     // Set up the transfer job
     TransferJob transferJob =
-        new TransferJob()
+        TransferJob.newBuilder()
```
frankyn, line 102: this is so nice and a lot smaller delta

storage/storage-transfer/pom.xml
```diff
@@ -51,6 +51,11 @@
         </exclusion>
       </exclusions>
     </dependency>
+    <dependency>
```
frankyn, line 54: One question on directory structure, wdyt about moving these files one directory up outside of the `storage/` directory because it's a different API?

frankyn (changes requested):
LGTM, just have one question on directory structure, the code changes to the samples make this a very appealing migration for existing developers because the delta isn't as heavy. 
It still does have a list of data models to worry about but the main thing that's awesome is the removal of auth init.

storage-transfer/src/main/java/com/google/cloud/storage/storagetransfer/samples/QuickstartSample.java
```diff
@@ -0,0 +1,66 @@
...
+            .build();
+
+    StorageTransferServiceClient storageTransfer = StorageTransferServiceClient.create();
+
+    TransferJob response =
+        storageTransfer.createTransferJob(
+            CreateTransferJobRequest.newBuilder().setTransferJob(transferJob).build());
+
+    storageTransfer
+        .runTransferJobCallable()
```
frankyn, line 56: Is this the only way to start the job? Wondering if there's an equivalent method such as `storageTransfer.createTransferJob()` -> `storageTransfer.runTransferJob()`?

JesseLovelace: The only other way would be `runTransferJobAsync()`. There is no `runTransferJob()`.

frankyn: Thanks for checking! Disregard this comment.

JesseLovelace: Would look like

```
storageTransfer.runTransferJobAsync(
            TransferProto.RunTransferJobRequest.newBuilder()
            .setProjectId(projectId)
            .setJobName(response.getName())
            .build())
            .get();
```

frankyn: Well it's nice that we have async, but let's leave that for our readers to find.

storage-transfer/src/main/java/com/google/cloud/storage/storagetransfer/samples/TransferFromAws.java
```diff
@@ -0,0 +1,129 @@
+/*
+ * Copyright 2020 Google Inc.
```
frankyn, line 2: 2021*

storage-transfer/src/test/java/com/google/cloud/storage/storagetransfer/samples/test/ITStoragetransferSamplesTest.java
```diff
@@ -0,0 +1,388 @@
+package com.google.cloud.storage.storagetransfer.samples.test;
```
frankyn, line 1: Missing copyright

storage-transfer/src/main/java/com/google/cloud/storage/storagetransfer/samples/CheckLatestTransferOperation.java
```diff
@@ -0,0 +1,64 @@
+/*
+ * Copyright 2020 Google Inc.
```
frankyn, line 2: 2021*

storage-transfer/src/main/java/com/google/cloud/storage/storagetransfer/samples/TransferFromAws.java
```diff
@@ -0,0 +1,129 @@
...
+        TransferJob.newBuilder()
+            .setDescription(jobDescription)
+            .setProjectId(projectId)
+            .setTransferSpec(transferSpec)
+            .setSchedule(schedule)
+            .setStatus(Status.ENABLED)
+            .build();
+
+    // Create a Transfer Service client
+    StorageTransferServiceClient storageTransfer = StorageTransferServiceClient.create();
```
frankyn, line 118: Do you want to move client creation before model construction?

frankyn (changes requested):
Few Nits

storage-transfer/src/main/java/com/google/cloud/storage/storagetransfer/samples/TransferFromAws.java
JesseLovelace: I can go either way on it. My idea was just having client creation right before the client is actually used, but I can move it up if you prefer

frankyn: I was thinking consistency, but I think your right in this case. You're only creating a client when you actually need it.

frankyn (approved):

storage-transfer/src/main/java/com/google/cloud/storage/storagetransfer/samples/QuickstartSample.java
```diff
@@ -0,0 +1,40 @@
+package com.google.cloud.storage.storagetransfer.samples;
+
+import com.google.storagetransfer.v1.proto.StorageTransferServiceClient;
+import com.google.storagetransfer.v1.proto.TransferProto.CreateTransferJobRequest;
+import com.google.storagetransfer.v1.proto.TransferTypes.GcsData;
+import com.google.storagetransfer.v1.proto.TransferTypes.TransferJob;
+import com.google.storagetransfer.v1.proto.TransferTypes.TransferSpec;
+
+public class QuickstartSample {
+  /** Quickstart sample using transfer service to transfer from one GCS bucket to another. */
```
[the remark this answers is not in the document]
danielbankhead: Hey, just now seeing this - I think either is fine, but `runTransferJob` might be a bit more complex for the customer (a user would have to know to disable to job). While more verbose, I think scheduling would make more sense (although it is a somewhat of a misnomer if we want to run this one time). For this reason, I think a one-time helper function would be nice to have across the GAPIC clients.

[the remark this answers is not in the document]
JesseLovelace: What do you mean by "a user would have to know to disable to job"?

Jobs that are created and then run this way, only run once. There's no need to disable it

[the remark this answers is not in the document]
danielbankhead: You're correct that there isn't a reason to explicitly disable it; I'm thinking someone new to the product may be concerned because 'Frequency' isn't set and the documentation mentions ['Schedule one-time transfer operations or recurring transfer operations'](https://cloud.google.com/storage-transfer/docs/overview#what-is-storage-transfer), making it a bit ambiguous if this transfer job will run again in the future.

Setting the schedule to the same date sets the Frequency to 'Once' in the UI - making it clearer to the user that this won't run again in the future.

Screen Shot 2021-11-12 at 8 04 47 AM

I think this would be valuable to capture as feedback and share upstream

PR #6423, opened 2021-11-22
docs(iap-samples): added bash sample for iap authentication 

Sita04:
Added sample for IAP authentication to obtain an OIDC token using a local service account file.

### DIFF
--- a/iap/bash/auth-oidc-service-account.sh
+++ b/iap/bash/auth-oidc-service-account.sh
@@ -1,0 +1,43 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+get_token() {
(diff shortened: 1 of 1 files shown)


eaball35 (changes requested):
Testing?

iap/bash/auth-oidc-service-account.sh
```diff
@@ -0,0 +1,43 @@
+#!/usr/bin/env bash
```
lesv, line 1: Copyright - if your goal is to publish this.  A README would also be helpful

lesv (changes requested):

PR #6434, opened 2021-11-24
restructure gcf/v2 directory

averikitsch:
I am not a fan of the parent directories with only 1 project, therefore each function now is in a project directly under v2/


ace-n (approved):
LGTM, but don't merge until @labtopia signs off.

ace-n:
No objections from @labtopia; merging.

PR #6446, opened 2021-12-02
Pub/Sub: update Spring tutorial with test + other nits

anguillanneuf:
- add test (suggestions welcome)
- update deprecated classes

@elefeint


elefeint (approved):

pubsub/spring/src/main/java/demo/PubSubApplication.java
```diff
@@ -91,18 +90,8 @@ public void messageReceiver(
...
-          public void onFailure(Throwable throwable) {
-            LOGGER.info("There was an error sending the message.");
-          }
-
-          @Override
-          public void onSuccess(String result) {
-            LOGGER.info("Message was sent via the outbound channel adapter to topic-two!");
-          }
-        });
+    adapter.setFailureCallback(
```
elefeint, line 93: Do you want to also add a success callback to match the behavior of the old code?

anguillanneuf: Oh yes. Thanks for noticing that. Done.

PR #6599, opened 2021-12-21
dataflow: avoid importing log4j at all

davidcavazos:
Remove any direct dependencies of log4j and use slf4j instead, as recommended by the [documentation](https://cloud.google.com/dataflow/docs/guides/logging#java_1), and update to Beam 2.34 which doesn't depend on log4j.

Confirmed with `mvn dependency:tree | grep log4j-api` that there are no direct or indirect dependency instances of log4j on any sample by doing this.

### DIFF
--- a/dataflow/encryption-keys/pom.xml
+++ b/dataflow/encryption-keys/pom.xml
@@ -34,13 +34,13 @@
     <maven.compiler.target>11</maven.compiler.target>
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
 
-    <beam.version>2.31.0</beam.version>
+    <beam.version>2.34.0</beam.version>
 
     <maven-compiler-plugin.version>3.8.1</maven-compiler-plugin.version>
     <maven-exec-plugin.version>3.0.0</maven-exec-plugin.version>
     <maven-jar-plugin.version>3.2.0</maven-jar-plugin.version>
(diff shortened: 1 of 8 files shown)


davidcavazos:
Hopefully fixes #5875

lesv (approved):

anguillanneuf (approved):

anguillanneuf:
You may need to drop the beam version for some of these samples for tests to pass.

davidcavazos:
Fixed lint issues. For the spanner sample, I'm trying Beam 2.32, if that passes I'll try 2.33 since we know something's broken with 2.34.

davidcavazos:
I noticed that the Spanner sample was using Java 8, which [ends its support in 3 months](https://endoflife.date/java). Updating to Java 11, maybe that'll fix the failing tests as well.

davidcavazos:
Java 11 didn't fix the tests, but downgrading to Beam 2.33 works for now. We still need to investigate why Beam 2.34 breaks SpannerIO.

cc: @jsimonweb @skuruppu @ansh0l -- tests for the SpannerIO sample with Beam 2.34 are still a blocker to upgrade versions (#5875)

ansh0l:
@davidcavazos I will investigate separately on beam 2.34 issue. I'm assuming that for now, things are working with beam 2.33 in place.

PR #6607, opened 2022-01-05
Make script copy and pasteable

averikitsch:
Fixes #issue

### DIFF
--- a/run/filesystem/gcsfuse_run.sh
+++ b/run/filesystem/gcsfuse_run.sh
@@ -13,6 +13,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 # [START cloudrun_fuse_script]
+#!/usr/bin/env bash
 set -eo pipefail
 
 # Create mount directory for service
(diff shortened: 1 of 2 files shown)


iennae (approved):
LGTM

run/filesystem/gcsfuse_run.sh
```diff
@@ -13,6 +13,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 # [START cloudrun_fuse_script]
+#!/usr/bin/env bash
```
lesv, line 16: This is just for illusion only.  The special flag that this is doesn't work in this situation.  As long as your ok w/ that I'll approve.

lesv (approved):

run/filesystem/gcsfuse_run.sh
averikitsch: Yes, it is still at the top of the file. This is so it shows up on devsite.

PR #6692, opened 2022-01-26
fix(pubsublite): remove processing time trigger

anguillanneuf:
By default, `FixedWindows` are triggered by event time (if no trigger is specified), which is the publish timestamp of a message. 

The example was using the processing time trigger because the PubsubLiteIO at one time had a bug. 

The bug has since been fixed (b/186174854), so we no longer need to demonstrate how to use processing time triggers for the sample to work. 

The `pom.xml` changes are required for running the sample via command line - I had previously thought they were required by the test only. Without them as compile dependencies, the following error will be thrown:

```
An exception occured while executing the Java class. com/google/api/services/dataflow/Dataflow: com.google.api.services.dataflow.Dataflow
```

### DIFF
--- a/pubsublite/streaming-analytics/pom.xml
+++ b/pubsublite/streaming-analytics/pom.xml
@@ -128,6 +128,18 @@
       <version>${beam.version}</version>
     </dependency>
 
+    <dependency>
+      <groupId>com.google.api-client</groupId>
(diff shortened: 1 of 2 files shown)


davidcavazos (approved):
Thanks for updating this. LGTM!

eaball35 (approved):

PR #6702, opened 2022-02-03
fix(functions/v2): Fix typo

ace-n:
Fixes b/217616726


lesv:
A quick google suggests that prin·ci·ple is the correct spelling.

ace-n:
The API itself (and [AWS](https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_policies_elements_principal.html), for what that's worth) uses `principal`.

averikitsch (approved):

PR #6703, opened 2022-02-04
feat: add cloud spanner change streams samples

haikuo-google:
The followings are included:
* Sample code for creating change stream.
* Sample code for querying change stream, basically it does an change stream initial query to get a list of partition tokens, and then use the partition tokens to query the data change records for each partition.
* Sample code for dropping change stream.
* ChangeStreamRecordMapper and models, these are the files that help change stream users understand how to model change stream records after receiving them from API, they are copied from org.apache.beam.sdk.io.gcp.spanner.changestreams.* with unused code being removed.
* README for how to run the sample code and tests.

### DIFF
--- a/spanner/changestreams/README.md
+++ b/spanner/changestreams/README.md
@@ -1,0 +1,24 @@
+# Cloud Spanner Change Stream Sample
+
+## Setup
+
+This sample requires [Java](https://www.java.com/en/download/) and [Maven](http://maven.apache.org/) to run the integration test.
+
+1.  **Follow the set-up instructions in [the documentation](https://cloud.google.com/java/docs/setup).**
+
+2.  Enable APIs for your project.
+    [Click here](https://console.cloud.google.com/flows/enableapi?apiid=spanner.googleapis.com&showconfirmation=true)
+    to visit Cloud Platform Console and enable the Google Cloud Spanner API.
+
+3.  Create a Cloud Spanner instance and database via the Cloud Plaform Console's
+    [Cloud Spanner section](http://console.cloud.google.com/spanner).
+
+4.  Enable application default credentials by running the command `gcloud auth application-default login`.
+
+## Run integration test
+
+Run the following Maven command to run integration test:
+
+```
+mvn test -Dspanner.test.instance=my-instance -Dspanner.test.database=my-db -Dtest=com.example.spanner.changestreams.ChangeStreamSampleIT
+```
--- a/spanner/changestreams/pom.xml
+++ b/spanner/changestreams/pom.xml
@@ -1,0 +1,77 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+ Copyright 2022 Google LLC.
+
+ Licensed under the Apache License, Version 2.0 (the "License");
+ you may not use this file except in compliance with the License.
+ You may obtain a copy of the License at
+
+       http://www.apache.org/licenses/LICENSE-2.0
+
+ Unless required by applicable law or agreed to in writing, software
+ distributed under the License is distributed on an "AS IS" BASIS,
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ See the License for the specific language governing permissions and
+ limitations under the License.
+-->
+
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+
+  <modelVersion>4.0.0</modelVersion>
+  <properties>
+    <maven.compiler.target>1.8</maven.compiler.target>
+    <maven.compiler.source>1.8</maven.compiler.source>
+  </properties>
+
+  <!--
+    The parent pom defines common style checks and testing strategies for our samples.
+    Removing or replacing it should not affect the execution of the samples in anyway.
+  -->
+  <parent>
+    <groupId>com.google.cloud.samples</groupId>
+    <artifactId>shared-configuration</artifactId>
+    <version>1.2.0</version>
+  </parent>
+
+  <groupId>com.example.spanner</groupId>
+  <artifactId>changestreams</artifactId>
+  <name>Cloud Spanner Change Streams Sample</name>
+  <dependencyManagement>
+    <dependencies>
+      <dependency>
+        <groupId>com.google.cloud</groupId>
+        <artifactId>libraries-bom</artifactId>
+        <version>24.2.0</version>
+        <type>pom</type>
+        <scope>import</scope>
+      </dependency>
+    </dependencies>
+  </dependencyManagement>
+
+  <dependencies>
+      <dependency>
+          <groupId>com.google.guava</groupId>
+          <artifactId>guava</artifactId>
+          <version>31.0.1-jre</version>
+      </dependency>
+      <dependency>
+          <groupId>org.apache.avro</groupId>
+          <artifactId>avro</artifactId>
+          <version>1.10.1</version>
+      </dependency>
+      <dependency>
+          <groupId>com.google.cloud</groupId>
+          <artifactId>google-cloud-spanner</artifactId>
+      </dependency>
+
+    <!-- Test dependencies -->
+    <dependency>
+      <groupId>junit</groupId>
+      <artifactId>junit</artifactId>
+      <version>4.13.2</version>
+      <scope>test</scope>
+    </dependency>
+  </dependencies>
+</project>
--- a/spanner/changestreams/src/main/java/com/example/spanner/changestreams/ChangeStreamRecordMapper.java
+++ b/spanner/changestreams/src/main/java/com/example/spanner/changestreams/ChangeStreamRecordMapper.java
@@ -1,0 +1,153 @@
+/*
+ * Copyright 2022 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.example.spanner.changestreams;
+
+import com.example.spanner.changestreams.model.ChangeStreamRecord;
+import com.example.spanner.changestreams.model.ChildPartition;
+import com.example.spanner.changestreams.model.ChildPartitionsRecord;
+import com.example.spanner.changestreams.model.ColumnType;
+import com.example.spanner.changestreams.model.DataChangeRecord;
+import com.example.spanner.changestreams.model.HeartbeatRecord;
+import com.example.spanner.changestreams.model.Mod;
+import com.example.spanner.changestreams.model.ModType;
+import com.example.spanner.changestreams.model.TypeCode;
+import com.example.spanner.changestreams.model.ValueCaptureType;
+import com.google.cloud.Timestamp;
+import com.google.cloud.spanner.Struct;
+import com.google.cloud.spanner.Type;
+import com.google.common.collect.Sets;
+import java.util.HashSet;
+import java.util.List;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+/**
+ * ChangeStreamRecordMapper converts a Struct returned from Change Streams API into a well-defined
+ * model, which could be one of DataChangeRecord, ChildPartitionsRecord or HeartbeatRecord.
+ */
+public class ChangeStreamRecordMapper {
+
+  public List<ChangeStreamRecord> toChangeStreamRecords(Struct row) {
+    return row.getStructList(0).stream()
+      .flatMap(struct -> toChangeStreamRecord(struct))
+      .collect(Collectors.toList());
+  }
+
+  private Stream<ChangeStreamRecord> toChangeStreamRecord(Struct row) {
+    final Stream<DataChangeRecord> dataChangeRecords =
+        row.getStructList("data_change_record").stream()
+        .filter(this::isNonNullDataChangeRecord)
+        .map(struct -> toDataChangeRecord(struct));
+
+    final Stream<HeartbeatRecord> heartbeatRecords =
+        row.getStructList("heartbeat_record").stream()
+        .filter(this::isNonNullHeartbeatRecord)
+        .map(struct -> toHeartbeatRecord(struct));
+
+    final Stream<ChildPartitionsRecord> childPartitionsRecords =
+        row.getStructList("child_partitions_record").stream()
+          .filter(this::isNonNullChildPartitionsRecord)
+          .map(struct -> toChildPartitionsRecord(struct));
+
+    return Stream.concat(
+      Stream.concat(dataChangeRecords, heartbeatRecords), childPartitionsRecords);
+  }
+
+  private boolean isNonNullDataChangeRecord(Struct row) {
+    return !row.isNull("commit_timestamp");
+  }
+
+  private boolean isNonNullHeartbeatRecord(Struct row) {
+    return !row.isNull("timestamp");
+  }
+
+  private boolean isNonNullChildPartitionsRecord(Struct row) {
+    return !row.isNull("start_timestamp");
+  }
+
+  private DataChangeRecord toDataChangeRecord(Struct row) {
+    final Timestamp commitTimestamp = row.getTimestamp("commit_timestamp");
+    return new DataChangeRecord(
+      commitTimestamp,
+      row.getString("server_transaction_id"),
+      row.getBoolean("is_last_record_in_transaction_in_partition"),
+      row.getString("record_sequence"),
+      row.getString("table_name"),
+      row.getStructList("column_types").stream()
+        .map(this::columnTypeFrom)
+        .collect(Collectors.toList()),
+      row.getStructList("mods").stream().map(this::modFrom).collect(Collectors.toList()),
+      ModType.valueOf(row.getString("mod_type")),
+      ValueCaptureType.valueOf(row.getString("value_capture_type")),
+      row.getLong("number_of_records_in_transaction"),
+      row.getLong("number_of_partitions_in_transaction"));
+  }
+
+  private HeartbeatRecord toHeartbeatRecord(Struct row) {
+    final Timestamp timestamp = row.getTimestamp("timestamp");
+    return new HeartbeatRecord(timestamp);
+  }
+
+  private ChildPartitionsRecord toChildPartitionsRecord(Struct row) {
+    final Timestamp startTimestamp = row.getTimestamp("start_timestamp");
+    return new ChildPartitionsRecord(
+      startTimestamp,
+      row.getString("record_sequence"),
+      row.getStructList("child_partitions").stream()
+        .map(struct -> childPartitionFrom(struct))
+        .collect(Collectors.toList()));
+  }
+
+  private ColumnType columnTypeFrom(Struct struct) {
+    // TODO: Move to type struct.getJson when backend is fully migrated
+    final String type = getJsonString(struct, "type");
+    return new ColumnType(
+      struct.getString("name"),
+        new TypeCode(type),
+      struct.getBoolean("is_primary_key"),
+      struct.getLong("ordinal_position"));
+  }
+
+  private Mod modFrom(Struct struct) {
+    // TODO: Move to keys struct.getJson when backend is fully migrated
+    final String keys = getJsonString(struct, "keys");
+    // TODO: Move to oldValues struct.getJson when backend is fully migrated
+    final String oldValues =
+        struct.isNull("old_values") ? null : getJsonString(struct, "old_values");
+    // TODO: Move to newValues struct.getJson when backend is fully migrated
+    final String newValues =
+        struct.isNull("new_values") ? null : getJsonString(struct, "new_values");
+    return new Mod(keys, oldValues, newValues);
+  }
+
+  private ChildPartition childPartitionFrom(Struct struct) {
+    final HashSet<String> parentTokens =
+        Sets.newHashSet(struct.getStringList("parent_partition_tokens"));
+    return new ChildPartition(struct.getString("token"), parentTokens);
+  }
+
+  // TODO: Remove when backend is fully migrated to JSON.
+  private String getJsonString(Struct struct, String columnName) {
+    if (struct.getColumnType(columnName).equals(Type.json())) {
+      return struct.getJson(columnName);
+    } else if (struct.getColumnType(columnName).equals(Type.string())) {
+      return struct.getString(columnName);
+    } else {
+      throw new IllegalArgumentException("Can not extract string from value " + columnName);
+    }
+  }
+}
--- a/spanner/changestreams/src/main/java/com/example/spanner/changestreams/ChangeStreamSample.java
+++ b/spanner/changestreams/src/main/java/com/example/spanner/changestreams/ChangeStreamSample.java
@@ -1,0 +1,240 @@
+/*
+ * Copyright 2022 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.example.spanner.changestreams;
+
+// TODO(haikuo@google.com): we should remove the models and the mapper once the change stream
+//  connector code that includes these models are released into Beam repo.
+import com.example.spanner.changestreams.model.ChangeStreamRecord;
+import com.example.spanner.changestreams.model.ChildPartition;
+import com.example.spanner.changestreams.model.ChildPartitionsRecord;
+import com.example.spanner.changestreams.model.DataChangeRecord;
+import com.example.spanner.changestreams.model.HeartbeatRecord;
+import com.google.api.gax.longrunning.OperationFuture;
+import com.google.cloud.Timestamp;
+import com.google.cloud.spanner.DatabaseAdminClient;
+import com.google.cloud.spanner.DatabaseClient;
+import com.google.cloud.spanner.DatabaseId;
+import com.google.cloud.spanner.Mutation;
+import com.google.cloud.spanner.ResultSet;
+import com.google.cloud.spanner.Spanner;
+import com.google.cloud.spanner.SpannerOptions;
+import com.google.cloud.spanner.Statement;
+import com.google.common.collect.ImmutableList;
+import com.google.spanner.admin.database.v1.UpdateDatabaseDdlMetadata;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+import java.util.concurrent.TimeUnit;
+
+/**
+ * Sample code for querying change stream, it:
+ * 1. Creates a table with simple schema and a change stream that watches the table.
+ * 2. Inserts test data into the table.
+ * 3. Executes a change stream initial query to get change stream partition tokens.
+ * 4. Executes a change stream partition query to get data change records of the inserted rows.
+ * 5. Drops the created table and change stream.
+ */
+public class ChangeStreamSample {
+  private static final long TIMEOUT_MINUTES = 2;
+
+  public static void run(String instanceId, String databaseId, String prefix) {
+    final String tableName = prefix + "_Singers";
+    final String changeStreamName = prefix + "_ChangeStreamSingers";
+    final SpannerOptions options = SpannerOptions.newBuilder().build();
+    final Spanner spanner = options.getService();
+    final DatabaseAdminClient dbAdminClient = spanner.getDatabaseAdminClient();
+    final DatabaseClient dbClient = spanner.getDatabaseClient(
+        DatabaseId.of(options.getProjectId(), instanceId, databaseId));
+
+    try {
+      createChangeStream(dbAdminClient, instanceId, databaseId, tableName, changeStreamName);
+      queryChangeStream(dbClient, tableName, changeStreamName);
+      dropChangeStream(dbAdminClient, instanceId, databaseId, tableName, changeStreamName);
+    } catch (Exception e) {
+      e.printStackTrace();
+    } finally {
+      spanner.close();
+    }
+  }
+
+  // [START spanner_change_streams_sample_create_change_streams]
+  static void createChangeStream(
+      DatabaseAdminClient dbAdminClient, String instanceId, String databaseId, String tableName,
+      String changeStreamName) throws Exception {
+    System.out.println(
+        String.format("Updating database DDL to create table %s and change stream %s.",
+          tableName, changeStreamName));
+
+    OperationFuture<Void, UpdateDatabaseDdlMetadata> op = dbAdminClient
+        .updateDatabaseDdl(
+        instanceId,
+        databaseId,
+        Arrays.asList(
+          String.format("CREATE TABLE %s ("
+            + "  SingerId   INT64 NOT NULL,"
+            + "  FirstName  STRING(1024),"
+            + "  LastName   STRING(1024)"
+            + ") PRIMARY KEY (SingerId)", tableName),
+          String.format("CREATE CHANGE STREAM %s FOR %s", changeStreamName, tableName)),
+        null
+      );
+
+    op.get(TIMEOUT_MINUTES, TimeUnit.MINUTES);
+  }
+  // [END spanner_change_streams_sample_create_change_streams]
+
+  // [START spanner_change_streams_sample_query_change_streams]
+  public static void queryChangeStream(DatabaseClient dbClient, String tableName,
+                                       String changeStreamName) {
+    // Insert test data into the table.
+    System.out.println(
+        String.format("Inserting rows "
+          + "(1, singer_1_first_name, singer_1_last_name) and "
+          + "(2, singer_2_first_name, singer_2_last_name) "
+          + "into %s table.", tableName));
+
+    insertRows(dbClient, tableName);
+
+    final Timestamp startTimestamp = Timestamp.now();
+    // end = start + 30 seconds.
+    final Timestamp endTimestamp = Timestamp.ofTimeSecondsAndNanos(
+        startTimestamp.getSeconds() + 30, startTimestamp.getNanos());
+
+    final ChangeStreamRecordMapper changeStreamRecordMapper =
+        new ChangeStreamRecordMapper();
+
+    // Execute an initial query to get partition tokens.
+    System.out.println("Executing change stream initial query.");
+    // For initial query the partition token is null.
+    List<ChangeStreamRecord> initialQueryRecords = executeChangeStreamQueryAndPrint(
+        dbClient, changeStreamName, startTimestamp, endTimestamp, null,
+        changeStreamRecordMapper);
+
+    System.out.println("Executing change stream partition queries.");
+    for (ChangeStreamRecord record : initialQueryRecords) {
+      // Executes a partition query to print data records that we just inserted.
+      if (record instanceof ChildPartitionsRecord) {
+        ChildPartitionsRecord childPartitionsRecord = (ChildPartitionsRecord) record;
+        for (ChildPartition childPartition : childPartitionsRecord.getChildPartitions()) {
+          executeChangeStreamQueryAndPrint(
+              dbClient, changeStreamName, childPartitionsRecord.getStartTimestamp(), endTimestamp,
+              childPartition.getToken(), changeStreamRecordMapper);
+        }
+      } else if (record instanceof DataChangeRecord) {
+        throw new IllegalArgumentException("Got unexpected DataChangeRecord from Change Streams "
+          + "initial query");
+      }
+    }
+  }
+
+  // Insert two rows into Singers table.
+  static void insertRows(DatabaseClient client, String tableName) {
+    client.write(
+        ImmutableList.of(
+        Mutation.newInsertOrUpdateBuilder(tableName)
+          .set("SingerId").to(1)
+          .set("FirstName").to("singer_1_first_name")
+          .set("LastName").to("singer_1_last_name")
+          .build(),
+        Mutation.newInsertOrUpdateBuilder(tableName)
+          .set("SingerId").to(2)
+          .set("FirstName").to("singer_2_first_name")
+          .set("LastName").to("singer_2_last_name")
+          .build()
+      )
+    );
+  }
+
+  // Execute a change stream query, return and print out the result records.
+  // For initial query, partitionToken is expected to be null.
+  public static List<ChangeStreamRecord> executeChangeStreamQueryAndPrint(
+      DatabaseClient dbClient, String changeStreamName, Timestamp startTimestamp,
+      Timestamp endTimestamp, String partitionToken,
+      ChangeStreamRecordMapper changeStreamRecordMapper) {
+    System.out.println("Executing a change stream query with: "
+        + "start_timestamp => " + startTimestamp
+        + ", end_timestamp => " + endTimestamp
+        + ", partition_token => " + partitionToken
+        + ", heartbeat_milliseconds => 5000");
+
+    final String query =
+        String.format("SELECT * FROM READ_%s ("
+          + "start_timestamp => @startTimestamp,"
+          + "end_timestamp => @endTimestamp,"
+          + "partition_token => @partitionToken,"
+          + "heartbeat_milliseconds => @heartbeatMillis"
+          + ")", changeStreamName);
+
+    final ResultSet resultSet =
+        dbClient
+          .singleUse()
+          .executeQuery(
+            Statement.newBuilder(query)
+              .bind("startTimestamp").to(startTimestamp)
+              .bind("endTimestamp").to(endTimestamp)
+              .bind("partitionToken").to(partitionToken)
+              .bind("heartbeatMillis").to(5000)
+              .build());
+
+    List<ChangeStreamRecord> result = new ArrayList<>();
+    while (resultSet.next()) {
+      // Parses result set into change stream result format.
+      final List<ChangeStreamRecord> records =
+          changeStreamRecordMapper.toChangeStreamRecords(resultSet.getCurrentRowAsStruct());
+
+      // Prints out all the query results.
+      for (final ChangeStreamRecord record : records) {
+        if (record instanceof DataChangeRecord) {
+          System.out.println("Received a DataChangeRecord: " + record);
+        } else if (record instanceof HeartbeatRecord) {
+          System.out.println("Received a HeartbeatRecord: " + record);
+        } else if (record instanceof ChildPartitionsRecord) {
+          System.out.println("Received a ChildPartitionsRecord: " + record);
+        } else {
+          // We should never reach here.
+          throw new IllegalArgumentException("Unknown record type " + record.getClass());
+        }
+      }
+      result.addAll(records);
+    }
+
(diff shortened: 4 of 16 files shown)


haikuo-google:
@thiagotnunes 

Could you take a look?

I was wrong earlier, I need to update model to support JSON type, I will update this PR once https://github.com/apache/beam/pull/16726 is merged.

spanner/changestreams/pom.xml
```diff
@@ -0,0 +1,63 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0"
```
averikitsch, line 2: Missing license header

spanner/changestreams/pom.xml
```diff
@@ -0,0 +1,63 @@
...
+  -->
+  <parent>
+    <groupId>com.google.cloud.samples</groupId>
+    <artifactId>shared-configuration</artifactId>
+    <version>1.2.0</version>
+  </parent>
+
+  <groupId>com.example.spanner</groupId>
+  <artifactId>changestreams</artifactId>
+  <name>Cloud Spanner Change Streams Sample</name>
```
averikitsch, line 25: nit: group this near the modelVersion in order to retain consistency across samples

spanner/changestreams/pom.xml
```diff
@@ -0,0 +1,63 @@
...
+  <dependencyManagement>
+    <dependencies>
+      <dependency>
+        <groupId>com.google.cloud</groupId>
+        <artifactId>libraries-bom</artifactId>
+        <version>24.0.0</version>
```
averikitsch, line 31: use most up to date version

spanner/changestreams/pom.xml
```diff
@@ -0,0 +1,63 @@
...
+        <scope>import</scope>
+      </dependency>
+    </dependencies>
+  </dependencyManagement>
+
+  <dependencies>
+      <dependency>
+          <groupId>com.google.cloud</groupId>
+          <artifactId>google-cloud-spanner</artifactId>
+          <version>6.16.0</version>
```
averikitsch, line 42: remove if using bom

spanner/changestreams/src/main/java/com/example/spanner/changestreams/ChangeStreamRecordMapper.java
```diff
@@ -0,0 +1,134 @@
+/*
+ * Copyright 2021 Google Inc.
```
averikitsch, line 2:
```suggestion
 * Copyright 2022 Google LLC
```

averikitsch (changes requested):
We aren't currently creating samples that run from the CLI: go/code-snippets-style#running-from-the-command-line and go/code-snippets-style#running-from-the-command-line

spanner/changestreams/src/main/java/com/example/spanner/changestreams/Utils.java
```diff
@@ -0,0 +1,53 @@
...
+    System.err.println("Usage: " + className + " "
+        + "<instance_id> "
+        + "<database_id> "
+        + "<prefix>");
+  }
+
+  public static void updateDdl(String instanceId,
+                               String databaseId,
+                               List<String> statements) throws Exception {
+    Spanner spanner = SpannerOptions.newBuilder().build().getService();
```
thiagotnunes, line 37: Could you move this into the constructor and make the methods non-static?

If you do so, make sure to give the constructor the project id and then you can construct the SpannerOptions with that as well:

```
SpannerOptions.newBuilder().setProject(projectId).build().getService();
```

You can close spanner afterwards in another method (like `close()`).

This is useful, because if you have not set a project id in your gcloud configuration, we would still use the one that was given through the parameters.

spanner/changestreams/src/main/java/com/example/spanner/changestreams/Utils.java
```diff
@@ -0,0 +1,53 @@
...
+    try {
+      DatabaseAdminClient dbAdminClient = spanner.getDatabaseAdminClient();
+      OperationFuture<Void, UpdateDatabaseDdlMetadata> op = dbAdminClient
+          .updateDatabaseDdl(
+            instanceId,
+            databaseId,
+            statements,
+            null
+        );
+      op.get();
```
thiagotnunes, line 48: Could we provide a timeout here?

spanner/changestreams/src/main/java/com/example/spanner/changestreams/QueryChangeStreamSample.java
```diff
@@ -0,0 +1,175 @@
...
+          .singleUse()
+          .executeQuery(
+            Statement.newBuilder(query)
+              .bind("startTimestamp").to(Timestamp.parseTimestamp(startTimestamp))
+              .bind("endTimestamp").to(Timestamp.parseTimestamp(endTimestamp))
+              .bind("partitionToken").to(partitionToken)
+              .bind("heartbeatMillis").to(5000)
+              .build());
+
+    ChangeStreamRecordMapper changeStreamRecordMapper = new ChangeStreamRecordMapper();
```
thiagotnunes, line 151: I still think that we don't need to map the records. We could just return raw structs from here. This would eliminate all the classes we created and the mapper from the PR as well.

I think we discussed this before, sorry I don't remember if we reached a decision. Could you remind me?

Thanks!

spanner/changestreams/src/main/java/com/example/spanner/changestreams/QueryChangeStreamSample.java
```diff
@@ -0,0 +1,175 @@
...
+    List<ChangeStreamRecord> res = new LinkedList<>();
```
thiagotnunes, line 152: It is a bit peculiar we are using a linked list here. Any reason you don't want to use an ArrayList?

spanner/changestreams/src/main/java/com/example/spanner/changestreams/QueryChangeStreamSample.java
```diff
@@ -0,0 +1,175 @@
...
+      // Prints out all the query results.
+      for (final ChangeStreamRecord record : records) {
+        if (record instanceof DataChangeRecord) {
+          System.out.println("Received a DataChangeRecord: " + record);
+        } else if (record instanceof HeartbeatRecord) {
+          System.out.println("Received a HeartbeatRecord: " + record);
+        } else if (record instanceof ChildPartitionsRecord) {
+          System.out.println("Received a ChildPartitionsRecord: " + record);
+        } else {
+          throw new IllegalArgumentException("Unknown record type " + record.getClass());
```
thiagotnunes, line 167: Could we add a comment here, saying this should not happen?

spanner/changestreams/src/main/java/com/example/spanner/changestreams/QueryChangeStreamSample.java
```diff
@@ -0,0 +1,175 @@
...
+      return;
+    }
+
+    final String instanceId = args[0];
+    final String databaseId = args[1];
+    final String prefix = args[2];
+    final String tableName = prefix + "_Singers";
+    final String changeStreamName = prefix + "_ChangeStreamSingers";
+
+    SpannerOptions options = SpannerOptions.newBuilder().build();
```
thiagotnunes, line 55: We should not build the options multiple times (in this class and in the Utils class). I think we should concentrate the building and getting the service in one class, which performs all the operations we would like.

spanner/changestreams/README.md
```diff
@@ -0,0 +1,48 @@
...
+2.  Enable APIs for your project.
+    [Click here](https://console.cloud.google.com/flows/enableapi?apiid=spanner.googleapis.com&showconfirmation=true)
+    to visit Cloud Platform Console and enable the Google Cloud Spanner API.
+
+3.  Create a Cloud Spanner instance and database via the Cloud Plaform Console's
+    [Cloud Spanner section](http://console.cloud.google.com/spanner).
+
+4.  Enable application default credentials by running the command `gcloud auth application-default login`.
+
+## Build and Run
```
thiagotnunes, line 18: Is there anyway we could make the sample self-contained, meaning that in a single execution it creates a table / change streams, inserts data, queries the change stream, prints out the records and finally drop the table and change stream?

spanner/changestreams/src/main/java/com/example/spanner/changestreams/QueryChangeStreamSample.java
haikuo-google: Sorry for leaving the PR for a long time! So I remembered the last time we agreed that the it's helpful to have some models which can teach customer how to decode these structs, and I recall that was the reason that we moved the PR from java client to here. 

I agreed that it's cumbersome to have these models, but imagine if I am a customer who is interested in trying change stream, I need to think a lot about how to process the results once receiving those raw structs, and try to decode them based data change record structure on the user guide, it would be easier to get people start if they can just access the models.

haikuo-google: Found the comments from original PR: https://github.com/googleapis/java-spanner/pull/1565#discussion_r762654054

spanner/changestreams/README.md
haikuo-google: They were all together in the original PR: https://github.com/googleapis/java-spanner/pull/1565, I separated them as per our discussion which was long time ago. I don't mind either way, I actually think do them all together is better - let me know your final thoughts, I can change accordingly

spanner/changestreams/src/main/java/com/example/spanner/changestreams/QueryChangeStreamSample.java
thiagotnunes: Ah ok, let's keep it then!

spanner/changestreams/README.md
thiagotnunes: Ah ok, thanks for reminding me. I think that we separated them so that it would be easier to show them in separate sample sections. We could also tag sections within the same file if you feel that is better.

spanner/changestreams/src/main/java/com/example/spanner/changestreams/Utils.java
haikuo-google: > If you do so, make sure to give the constructor the project id and then you can construct the SpannerOptions with that as well:

Are you suggesting that we need the user to input project id as well?

> You can close spanner afterwards in another method (like close()).
You mean keep the Spanner object as a member of Utils, and provide a separate method just for closing Spanner object? When does the caller calls the close method though? At the end of main method?

spanner/changestreams/README.md
haikuo-google: Sorry by tagging sections you meant this? https://idratherbewriting.com/java-javadoc-tags/

spanner/changestreams/src/main/java/com/example/spanner/changestreams/QueryChangeStreamSample.java
haikuo-google: Considering we have merged change stream connector code into Beam repo, I think I can just import Beam and use the model.

spanner/changestreams/README.md
thiagotnunes: Oh sorry, I should had explained it better.

We use comments to denote when a sample starts / ends in the Java file. This is what I mean by tags. These are arbitrary tags, which are given to the docs team so that they know when a sample starts and ends. See an example here: [start tag](https://github.com/googleapis/java-spanner/blob/main/samples/snippets/src/main/java/com/example/spanner/QueryWithJsonParameterSample.java#L19), [end tag](https://github.com/googleapis/java-spanner/blob/main/samples/snippets/src/main/java/com/example/spanner/QueryWithJsonParameterSample.java#L64)

spanner/changestreams/src/main/java/com/example/spanner/changestreams/Utils.java
thiagotnunes: I am suggesting that - if the project, instance and database are always the same for the lifecycle of the `Utils` class - we could create a Utils constructor that receives those (or at least the project, which should not change). Then you can build the options in the constructor as well and get the service. Something like:


```java
public class Utils {
  public Utils(String projectId, String instanceId, String databaseId) {
    final SpannerOptions options = SpannerOptions.newBuilder().setProject(projectId).build();
    final DatabaseId id = DatabaseId.of(projectId, instanceId, databaseId);
    this.spanner = options.getService();
    this.databaseClient = spanner.getDatabaseClient(id); // this only works if the database is already created (I think that is the case)
    this.databaseAdminClient = spanner.getDatabaseAdminClient();
  }

  public void close() {
    spanner.close();
  }
}
```

spanner/changestreams/pom.xml
haikuo-google: Hmm how? If I remove this, `import com.google.cloud.spanner.*;` doesn't work anymore.

spanner/changestreams/README.md
haikuo-google: Thank you! Done

spanner/changestreams/src/main/java/com/example/spanner/changestreams/QueryChangeStreamSample.java
haikuo-google: Done.

spanner/changestreams/src/main/java/com/example/spanner/changestreams/QueryChangeStreamSample.java
haikuo-google: Done.

spanner/changestreams/src/main/java/com/example/spanner/changestreams/QueryChangeStreamSample.java
haikuo-google: Changed to ArrayList

spanner/changestreams/src/main/java/com/example/spanner/changestreams/Utils.java
haikuo-google: Done

spanner/changestreams/src/main/java/com/example/spanner/changestreams/Utils.java
haikuo-google: Thanks, removed the util since we do everything in one file.

spanner/changestreams/src/main/java/com/example/spanner/changestreams/ChangeStreamRecordMapper.java
haikuo-google: Done.

spanner/changestreams/pom.xml
haikuo-google: Done.

spanner/changestreams/pom.xml
haikuo-google: Done.

spanner/changestreams/pom.xml
haikuo-google: Done.

haikuo-google:
@averikitsch @thiagotnunes  Thanks for taking a look! I resolved the comments, I am wondering if you could review it again?

spanner/changestreams/pom.xml
```diff
@@ -0,0 +1,78 @@
...
+      </dependency>
+      <dependency>
+          <groupId>org.apache.avro</groupId>
+          <artifactId>avro</artifactId>
+          <version>1.10.1</version>
+      </dependency>
+      <dependency>
+          <groupId>com.google.cloud</groupId>
+          <artifactId>google-cloud-spanner</artifactId>
+          <version>6.16.0</version>
```
thiagotnunes, line 67: You should not need to specify the version here, since you are using the BOM above. What error are you seeing?

haikuo-google: Oh I thought I need to remove the entire dependency, yeah just removing version works!

spanner/changestreams/README.md
```diff
@@ -0,0 +1,42 @@
...
+
+Please use the same prefix for the following commands to manage resources properly.
+
+1. Run the following Maven command to build the application:
```
thiagotnunes, line 22: It seems like we don't need to have (bash) client based samples anymore. You can remove the run section, as it's expected for customers to copy and paste the code instead of running from the command line.

spanner/changestreams/README.md
```diff
@@ -0,0 +1,42 @@
...
+   2. Insert test data to the created table and run change stream queries.
+   3. Drop the created table and change stream.
+
+    ```
+    mvn exec:java -Dexec.mainClass=com.example.spanner.changestreams.ChangeStreamSample -Dexec.args="my-instance my-db prefix"
+    ```
+
+## Build and Run
+
+Run the following Maven command to run unit test:
```
thiagotnunes, line 38: Is the following a unit test or an integration test?

spanner/changestreams/src/main/java/com/example/spanner/changestreams/ChangeStreamRecordMapper.java
```diff
@@ -0,0 +1,134 @@
...
+import com.example.spanner.changestreams.model.ValueCaptureType;
+import com.google.cloud.Timestamp;
+import com.google.cloud.spanner.Struct;
+import com.google.common.collect.Sets;
+import java.util.HashSet;
+import java.util.List;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+public class ChangeStreamRecordMapper {
```
thiagotnunes, line 37: See the updated [ChangeStreamRecordMapper](https://github.com/apache/beam/blob/master/sdks/java/io/google-cloud-platform/src/main/java/org/apache/beam/sdk/io/gcp/spanner/changestreams/mapper/ChangeStreamRecordMapper.java#L332-L341)

We need to allow for reading json and strings for the column type and mod fields.

spanner/changestreams/src/main/java/com/example/spanner/changestreams/ChangeStreamSample.java
```diff
@@ -0,0 +1,255 @@
...
+ * 1. Creates a table with simple schema and a change stream that watches the table.
+ * 2. Inserts test data into the table.
+ * 3. Executes a change stream initial query to get change stream partition tokens.
+ * 4. Executes a change stream partition query to get data change records of the inserted rows.
+ * 5. Drops the created table and change stream.
+ */
+public class ChangeStreamSample {
+  private static final long TIMEOUT_MINUTES = 2;
+
+  public static void main(String... args) {
```
thiagotnunes, line 54: We don't need a main here if we are going with a non-client based sample

spanner/changestreams/src/main/java/com/example/spanner/changestreams/ChangeStreamSample.java
```diff
@@ -0,0 +1,255 @@
...
+          + "<prefix>");
+      return;
+    }
+
+    final String instanceId = args[0];
+    final String databaseId = args[1];
+    final String prefix = args[2];
+    final String tableName = prefix + "_Singers";
+    final String changeStreamName = prefix + "_ChangeStreamSingers";
+    SpannerOptions options = SpannerOptions.newBuilder().build();
```
thiagotnunes, line 68: nit: could we make the other variables `final` as well, since you are using `final` above?

spanner/changestreams/src/main/java/com/example/spanner/changestreams/ChangeStreamSample.java
```diff
@@ -0,0 +1,255 @@
...
+
+  // [START change_stream_sample_create_change_stream]
+  static void createChangeStream(
+      DatabaseAdminClient dbAdminClient, String instanceId, String databaseId, String tableName,
+      String changeStreamName) throws Exception {
+    System.out.println(
+        String.format("Updating database DDL to create table %s and change stream %s.",
+          tableName, changeStreamName));
+
+    updateDdl(
```
thiagotnunes, line 107: nit: maybe it is just easier to inline this function so that the sample between tags is self-contained?

spanner/changestreams/src/main/java/com/example/spanner/changestreams/ChangeStreamSample.java
```diff
@@ -0,0 +1,255 @@
...
+
+  // [START change_stream_sample_drop_change_stream]
+  public static void dropChangeStream(
+      DatabaseAdminClient dbAdminClient, String instanceId, String databaseId, String tableName,
+      String changeStreamName) throws Exception {
+    System.out.println(
+        String.format("Updating database DDL to drop table %s and change stream %s.",
+          tableName, changeStreamName));
+
+    updateDdl(
```
thiagotnunes, line 244: nit: maybe it is easier to inline this function to make the sample self-contained?

spanner/changestreams/src/main/java/com/example/spanner/changestreams/ChangeStreamSample.java
```diff
@@ -0,0 +1,255 @@
...
+  }
+  // [END change_stream_sample_create_change_stream]
+
+  // [START change_stream_sample_query_change_stream]
+  public static void queryChangeStream(DatabaseClient dbClient, String tableName,
+                                       String changeStreamName) {
+    // Insert test data into the table.
+    System.out.println(
+        String.format("Inserting rows "
+          + "(1, Melissa, Garcia) and "
```
thiagotnunes, line 128: Should we use real names here or just something like "User 1 First Name", "User 1 Last Name"?

spanner/changestreams/src/main/java/com/example/spanner/changestreams/ChangeStreamSample.java
```diff
@@ -0,0 +1,255 @@
...
+        dbClient, changeStreamName, startTimestampStr, endTimestampStr, null);
+
+    System.out.println("Executing change stream partition queries.");
+    for (ChangeStreamRecord record : initialQueryRecords) {
+      // Executes a partition query to print data records that we just inserted.
+      if (record instanceof ChildPartitionsRecord) {
+        for (ChildPartition childPartition :
+            ((ChildPartitionsRecord) record).getChildPartitions()) {
+          executeChangeStreamQueryAndPrint(
+              dbClient, changeStreamName, startTimestampStr, endTimestampStr,
```
thiagotnunes, line 153: Don't we need to update the start timestamp here based on the `childPartition.getStartTimestamp`?

spanner/changestreams/src/main/java/com/example/spanner/changestreams/ChangeStreamSample.java
```diff
@@ -0,0 +1,255 @@
...
+          + "(2, Dylan, Shaw) "
+          + "into %s table.", tableName));
+
+    insertRows(dbClient, tableName);
+
+    final Timestamp startTimestamp = Timestamp.now();
+    // end = start + 30 seconds.
+    final Timestamp endTimestamp = Timestamp.ofTimeSecondsAndNanos(
+        startTimestamp.getSeconds() + 30, startTimestamp.getNanos());
+    final String startTimestampStr = startTimestamp.toString();
```
thiagotnunes, line 138: nit: We should not convert to string here, but do it as late as possible so that we remain using the `Timestamp` type

spanner/changestreams/src/main/java/com/example/spanner/changestreams/ChangeStreamSample.java
```diff
@@ -0,0 +1,255 @@
...
+    final String endTimestampStr = endTimestamp.toString();
+
+    // Execute an initial query to get partition tokens.
+    System.out.println("Executing change stream initial query.");
+    List<ChangeStreamRecord> initialQueryRecords = executeChangeStreamQueryAndPrint(
+        dbClient, changeStreamName, startTimestampStr, endTimestampStr, null);
```
thiagotnunes, line 144: nit: could you add a comment that for the initial query the partition token is null?

spanner/changestreams/src/main/java/com/example/spanner/changestreams/ChangeStreamSample.java
```diff
@@ -0,0 +1,255 @@
...
+          + "read_options => null,"
+          + "heartbeat_milliseconds => @heartbeatMillis"
+          + ")", changeStreamName);
+
+    final ResultSet resultSet =
+        dbClient
+          .singleUse()
+          .executeQuery(
+            Statement.newBuilder(query)
+              .bind("startTimestamp").to(Timestamp.parseTimestamp(startTimestamp))
```
thiagotnunes, line 203: If you give this method a `Timestamp` instance instead of a `String` you would not need to parse it here.

spanner/changestreams/src/main/java/com/example/spanner/changestreams/ChangeStreamSample.java
```diff
@@ -0,0 +1,255 @@
...
+              .bind("endTimestamp").to(Timestamp.parseTimestamp(endTimestamp))
+              .bind("partitionToken").to(partitionToken)
+              .bind("heartbeatMillis").to(5000)
+              .build());
+
+    ChangeStreamRecordMapper changeStreamRecordMapper = new ChangeStreamRecordMapper();
```
thiagotnunes, line 209: The mapper is stateless, meaning we can share it across executeQueries. Maybe create it as a field of the class (and instantiate in the constructor / initial method) and reuse it. (What I am trying to say is that we don't need to create a new instance for every query)

spanner/changestreams/src/main/java/com/example/spanner/changestreams/ChangeStreamSample.java
```diff
@@ -0,0 +1,255 @@
...
+    List<ChangeStreamRecord> res = new ArrayList<>();
```
thiagotnunes, line 210: Don't abbreviate variable names. Per convention use something like `result`

spanner/changestreams/src/main/java/com/example/spanner/changestreams/ChangeStreamSample.java
```diff
@@ -0,0 +1,255 @@
...
+      // Prints out all the query results.
+      for (final ChangeStreamRecord record : records) {
+        if (record instanceof DataChangeRecord) {
+          System.out.println("Received a DataChangeRecord: " + record);
+        } else if (record instanceof HeartbeatRecord) {
+          System.out.println("Received a HeartbeatRecord: " + record);
+        } else if (record instanceof ChildPartitionsRecord) {
+          System.out.println("Received a ChildPartitionsRecord: " + record);
+        } else {
+          // We should never receive reach here.
```
thiagotnunes, line 225: typo here. `We should never reach here`

spanner/changestreams/README.md
haikuo-google: Removed the main and updated the README.

spanner/changestreams/README.md
haikuo-google: Integration test, updated.

spanner/changestreams/src/main/java/com/example/spanner/changestreams/ChangeStreamRecordMapper.java
haikuo-google: Thanks, updated.

spanner/changestreams/src/main/java/com/example/spanner/changestreams/ChangeStreamSample.java
haikuo-google: Removed main.

spanner/changestreams/src/main/java/com/example/spanner/changestreams/ChangeStreamSample.java
haikuo-google: Done.

spanner/changestreams/src/main/java/com/example/spanner/changestreams/ChangeStreamSample.java
haikuo-google: Done.

spanner/changestreams/src/main/java/com/example/spanner/changestreams/ChangeStreamSample.java
haikuo-google: Done.

spanner/changestreams/src/main/java/com/example/spanner/changestreams/ChangeStreamSample.java
haikuo-google: Changed.

spanner/changestreams/src/main/java/com/example/spanner/changestreams/ChangeStreamSample.java
haikuo-google: Fixed.

spanner/changestreams/src/main/java/com/example/spanner/changestreams/ChangeStreamSample.java
haikuo-google: Good point, done.

spanner/changestreams/src/main/java/com/example/spanner/changestreams/ChangeStreamSample.java
haikuo-google: Done.

spanner/changestreams/src/main/java/com/example/spanner/changestreams/ChangeStreamSample.java
haikuo-google: Added as a private static member of the class.

spanner/changestreams/src/main/java/com/example/spanner/changestreams/ChangeStreamSample.java
haikuo-google: Done.

spanner/changestreams/src/main/java/com/example/spanner/changestreams/ChangeStreamSample.java
haikuo-google: Fixed.

spanner/changestreams/src/main/java/com/example/spanner/changestreams/ChangeStreamSample.java
haikuo-google: Added

haikuo-google:
@averikitsch @thiagotnunes I have addressed all the comments, could you both take a look? Thanks!

spanner/changestreams/src/main/java/com/example/spanner/changestreams/ChangeStreamRecordMapper.java
```diff
@@ -0,0 +1,151 @@
...
+import com.google.cloud.spanner.Type;
+import com.google.common.collect.Sets;
+import java.util.HashSet;
+import java.util.List;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+public class ChangeStreamRecordMapper {
+
+  ChangeStreamRecordMapper() {}
```
thiagotnunes, line 40: nit: I don't think we need the package-private constructor here (this is needed in the connector for serialization purposes)

spanner/changestreams/src/main/java/com/example/spanner/changestreams/ChangeStreamSample.java
```diff
@@ -0,0 +1,236 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.example.spanner.changestreams;
+
+// TODO(haikuo@google.com): we should remove these models once the change stream connector code that
```
thiagotnunes, line 19: I think you will be able to remove the mapper as well

spanner/changestreams/src/main/java/com/example/spanner/changestreams/ChangeStreamSample.java
```diff
@@ -0,0 +1,236 @@
...
+    final DatabaseAdminClient dbAdminClient = spanner.getDatabaseAdminClient();
+    final DatabaseClient dbClient = spanner.getDatabaseClient(
+        DatabaseId.of(options.getProjectId(), instanceId, databaseId));
+
+    try {
+      createChangeStream(dbAdminClient, instanceId, databaseId, tableName, changeStreamName);
+      queryChangeStream(dbClient, tableName, changeStreamName);
+      dropChangeStream(dbAdminClient, instanceId, databaseId, tableName, changeStreamName);
+    } catch (Exception e) {
+      System.out.println(e);
```
thiagotnunes, line 70: Instead of `System.out.println(e)`, do a `e.printStackTrace()`. Otherwise you will not get the stack trace.

spanner/changestreams/src/main/java/com/example/spanner/changestreams/ChangeStreamSample.java
```diff
@@ -0,0 +1,236 @@
...
+ * Sample code for querying change stream, it:
+ * 1. Creates a table with simple schema and a change stream that watches the table.
+ * 2. Inserts test data into the table.
+ * 3. Executes a change stream initial query to get change stream partition tokens.
+ * 4. Executes a change stream partition query to get data change records of the inserted rows.
+ * 5. Drops the created table and change stream.
+ */
+public class ChangeStreamSample {
+  private static final long TIMEOUT_MINUTES = 2;
+  private static final ChangeStreamRecordMapper changeStreamRecordMapper =
```
thiagotnunes, line 53: nit: this is fine, but it could be a parameter to the functions as the dbAdminClient

spanner/changestreams/src/main/java/com/example/spanner/changestreams/ChangeStreamSample.java
thiagotnunes: It seems we forgot to change this String?

spanner/changestreams/src/main/java/com/example/spanner/changestreams/ChangeStreamSample.java
```diff
@@ -0,0 +1,236 @@
...
+    // Execute an initial query to get partition tokens.
+    System.out.println("Executing change stream initial query.");
+    // For initial query the partition token is null.
+    List<ChangeStreamRecord> initialQueryRecords = executeChangeStreamQueryAndPrint(
+        dbClient, changeStreamName, startTimestamp, endTimestamp, null);
+
+    System.out.println("Executing change stream partition queries.");
+    for (ChangeStreamRecord record : initialQueryRecords) {
+      // Executes a partition query to print data records that we just inserted.
+      if (record instanceof ChildPartitionsRecord) {
```
thiagotnunes, line 128: nit: Should we throw something if we get DataRecords?

spanner/changestreams/src/main/java/com/example/spanner/changestreams/ChangeStreamSample.java
```diff
@@ -0,0 +1,236 @@
...
+        + ", partition_token => " + partitionToken
+        + ", heartbeat_milliseconds => 5000"
+        + ", read_options => null");
+
+    final String query =
+        String.format("SELECT * FROM READ_%s ("
+          + "start_timestamp => @startTimestamp,"
+          + "end_timestamp => @endTimestamp,"
+          + "partition_token => @partitionToken,"
+          + "read_options => null,"
```
thiagotnunes, line 174: nit: can we omit read_options?

spanner/changestreams/src/test/java/com/example/spanner/changestreams/ChangeStreamSampleIT.java
```diff
@@ -0,0 +1,76 @@
...
+ * Tests for ChangeStreamSample.
+ */
+@RunWith(JUnit4.class)
+@SuppressWarnings("checkstyle:abbreviationaswordinname")
+public class ChangeStreamSampleIT {
+  private static String instanceId = System.getProperty("spanner.test.instance");
+  private static String databaseId = System.getProperty("spanner.test.database");
+  private static String prefix = System.getProperty("spanner.test.prefix");
+
+  private ByteArrayOutputStream bout;
```
thiagotnunes, line 40: Do we need to close `bout` on `tearDown`?

spanner/changestreams/src/test/java/com/example/spanner/changestreams/ChangeStreamSampleIT.java
```diff
@@ -0,0 +1,76 @@
...
+  @Test
+  public void testChangeStreamSample() {
+    assertNotNull(instanceId);
+    assertNotNull(databaseId);
+    assertNotNull(prefix);
+    ChangeStreamSample.run(instanceId, databaseId, prefix);
+
+    String got = bout.toString();
+    Assert.assertTrue(got.contains("Received a ChildPartitionsRecord"));
+    Assert.assertTrue(got.contains("Received a HeartbeatRecord"));
```
thiagotnunes, line 65: Will we always get a heartbeat record? Or can the test be flaky because of that?

spanner/changestreams/src/main/java/com/example/spanner/changestreams/ChangeStreamSample.java
```diff
@@ -0,0 +1,236 @@
...
+      queryChangeStream(dbClient, tableName, changeStreamName);
+      dropChangeStream(dbAdminClient, instanceId, databaseId, tableName, changeStreamName);
+    } catch (Exception e) {
+      System.out.println(e);
+    } finally {
+      spanner.close();
+    }
+  }
+
+  // [START change_stream_sample_create_change_stream]
```
thiagotnunes, line 76: Update the tags to contain `spanner` (as in `spanner_change_stream_sample_create_change_stream`). This should fix the snippet-bot failure you are getting

thiagotnunes: Btw, I think we should use change_streams (plural) instead of change_stream (singular)

thiagotnunes:
@haikuo-google edit the title of the PR to be according https://www.conventionalcommits.org/en/v1.0.0/#summary

Something like `feat: add cloud spanner change streams samples` (lower case, no dot)

spanner/changestreams/src/main/java/com/example/spanner/changestreams/ChangeStreamRecordMapper.java
haikuo-google: Removed.

spanner/changestreams/src/main/java/com/example/spanner/changestreams/ChangeStreamSample.java
haikuo-google: Done.

spanner/changestreams/src/main/java/com/example/spanner/changestreams/ChangeStreamSample.java
haikuo-google: Done

spanner/changestreams/src/main/java/com/example/spanner/changestreams/ChangeStreamSample.java
haikuo-google: Added exception

spanner/changestreams/src/test/java/com/example/spanner/changestreams/ChangeStreamSampleIT.java
haikuo-google: Closed

spanner/changestreams/src/test/java/com/example/spanner/changestreams/ChangeStreamSampleIT.java
haikuo-google: Removed

spanner/changestreams/src/main/java/com/example/spanner/changestreams/ChangeStreamSample.java
haikuo-google: Updated comments, but I think to actually use the code we might need to do some trick, since the mapper requires other params like partition metadata, will try to fix it later.

spanner/changestreams/src/main/java/com/example/spanner/changestreams/ChangeStreamSample.java
haikuo-google: Changed, let me know if you prefer to change all the naming to plural, I didn't since it's a bit wired to say "read a change streams", and a lots of our code is already using change stream, "change streams" seems like something makes more sense in terms marking, I am fine either ways though.

haikuo-google:
@thiagotnunes thanks! All comments addressed, please take a look again, thank you!

spanner/changestreams/README.md
```diff
@@ -0,0 +1,24 @@
+# Cloud Spanner Change Stream Sample
+
```
averikitsch, line 2: Is this used in the docs? If so can you add the link to learn more.

spanner/changestreams/src/main/java/com/example/spanner/changestreams/ChangeStreamRecordMapper.java
```diff
@@ -0,0 +1,149 @@
...
+import com.google.cloud.Timestamp;
+import com.google.cloud.spanner.Struct;
+import com.google.cloud.spanner.Type;
+import com.google.common.collect.Sets;
+import java.util.HashSet;
+import java.util.List;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+public class ChangeStreamRecordMapper {
```
averikitsch, line 38: A comment on what this class is would be nice

averikitsch (approved):

spanner/changestreams/src/main/java/com/example/spanner/changestreams/ChangeStreamSample.java
```diff
@@ -0,0 +1,240 @@
...
+      // Executes a partition query to print data records that we just inserted.
+      if (record instanceof ChildPartitionsRecord) {
+        ChildPartitionsRecord childPartitionsRecord = (ChildPartitionsRecord) record;
+        for (ChildPartition childPartition : childPartitionsRecord.getChildPartitions()) {
+          executeChangeStreamQueryAndPrint(
+              dbClient, changeStreamName, childPartitionsRecord.getStartTimestamp(), endTimestamp,
+              childPartition.getToken(), changeStreamRecordMapper);
+        }
+      } else {
+        throw new IllegalArgumentException("Got a record that's not ChildPartitionsRecord for "
```
thiagotnunes, line 138: Can't we get Heartbeat records here? Maybe we should throw only if we get a DataChangeRecord?

spanner/changestreams/src/test/java/com/example/spanner/changestreams/ChangeStreamSampleIT.java
```diff
@@ -0,0 +1,82 @@
...
+    } catch (IOException e) {
+      e.printStackTrace();
+    }
+
+    System.setOut(stdOut);
+  }
+
+  @Test
+  public void testChangeStreamSample() {
+    assertNotNull(instanceId);
```
thiagotnunes, line 65: IT tests seem to be failing here

spanner/changestreams/README.md
haikuo-google: We don't have a link to the doc yet, since this is a feature in private preview.

spanner/changestreams/src/test/java/com/example/spanner/changestreams/ChangeStreamSampleIT.java
thiagotnunes: We need to set system properties in the POM file to specify the instance, database and prefix (per above). An example of how to do this can be seen here: https://github.com/googleapis/java-spanner/blob/main/samples/snippets/pom.xml#L149

We need to find out which projects / instance / database can be used here.

thiagotnunes: The project seems to be `java-docs-samples-testing` as specified here: https://github.com/GoogleCloudPlatform/java-docs-samples/blob/main/.kokoro/tests/run_tests.sh#L56

I don't have access to it, so I don't know which spanner instance / database we can use. 

@averikitsch do you know if there is a Spanner instance / database that we can use in this project for integration testing?
@haikuo-google we will need to allowlist this project for change streams

spanner/changestreams/src/main/java/com/example/spanner/changestreams/ChangeStreamSample.java
haikuo-google: Done.

spanner/changestreams/src/main/java/com/example/spanner/changestreams/ChangeStreamRecordMapper.java
haikuo-google: Done.

thiagotnunes (approved):

PR #6709, opened 2022-02-10
chore: add E2E tests for Cloud Run

averikitsch:
Fixes #issue

### DIFF
--- a/run/jobs/pom.xml
+++ b/run/jobs/pom.xml
@@ -46,6 +46,12 @@ limitations under the License.
       <version>1.1.3</version>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>com.google.cloud</groupId>
+      <artifactId>google-cloud-logging</artifactId>
+      <version>3.6.2</version>
+      <scope>test</scope>
+    </dependency>
   </dependencies>
   
   <build>
--- a/run/jobs/src/test/java/com/example/JobsIntegrationTests.java
+++ b/run/jobs/src/test/java/com/example/JobsIntegrationTests.java
@@ -61,11 +61,11 @@ public class JobsIntegrationTests {
     System.out.println("Start Cloud Build...");
     Process p = setup.start();
     // Set timeout
-    if (!p.waitFor(15, TimeUnit.MINUTES)) {
-      p.destroy();
-      System.out.println("Process timed out.");
-      throw new InterruptedByTimeoutException();
-    }
+//     if (!p.waitFor(15, TimeUnit.MINUTES)) {
+//       p.destroy();
+//       System.out.println("Process timed out.");
+//       throw new InterruptedByTimeoutException();
+//     }
     // Read process output
     BufferedReader in = new BufferedReader(new InputStreamReader(p.getInputStream()));
     String line;
--- a/run/jobs/src/test/java/com/example/JobsIntegrationTests.java
+++ b/run/jobs/src/test/java/com/example/JobsIntegrationTests.java
@@ -60,12 +60,7 @@ public class JobsIntegrationTests {
     setup.redirectErrorStream(true);
     System.out.println("Start Cloud Build...");
     Process p = setup.start();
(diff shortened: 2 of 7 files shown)


averikitsch:
Project needs to be allowlisted

run/jobs/pom.xml
```diff
@@ -46,6 +46,12 @@ limitations under the License.
       <version>1.1.3</version>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>com.google.cloud</groupId>
+      <artifactId>google-cloud-logging</artifactId>
+      <version>3.6.2</version>
+      <scope>test</scope>
```
lesv, line 53: Why is this just `test` scope?

run/jobs/src/test/java/com/example/JobsIntegrationTests.java
```diff
@@ -0,0 +1,126 @@
...
+
+public class JobsIntegrationTests {
+
+  private static final String project = System.getenv("GOOGLE_CLOUD_PROJECT");
+  private static final String suffix = UUID.randomUUID().toString();
+  private static String service;
+
+  @BeforeClass
+  public static void setup() throws Exception {
+    service = "job-quickstart-f374953a-498c-43b8-847a-7998ae5665d5"; // "job-quickstart-" + suffix;
```
lesv, line 49: Is this really a fixed thing?

lesv: looking at the code, perhaps this should be a real UUID?

run/jobs/src/test/java/com/example/JobsIntegrationTests.java
```diff
@@ -0,0 +1,126 @@
...
+
+    ProcessBuilder setup = new ProcessBuilder();
+    setup.command(
+        "gcloud",
+        "builds",
+        "submit",
+        "--project=" + project,
+        "--config=./src/test/java/com/example/resources/e2e_test_setup.yaml",
+        String.format("--substitutions=_SERVICE=%s,_VERSION=%s", service, suffix));
```
lesv, line 58: I always wonder why we do this?  Why we think it's a good idea?

run/jobs/src/test/java/com/example/resources/e2e_test_cleanup.yaml
```diff
@@ -0,0 +1,19 @@
+steps:
```
lesv, line 1: You could have a copyright header here.

run/jobs/src/test/java/com/example/resources/e2e_test_setup.yaml
```diff
@@ -0,0 +1,28 @@
+steps:
```
lesv, line 1: You could also have a copyright header here

lesv (approved):
* Please answer questions
* Add copyrights on human generated content.
* I'm not a fan of code that executes commands... Are you sure you wouldn't be better off w/ a script?
* Think about what happens if this is executed in 4 VM's at the same time.  Would it work?  Why / Why not?  (I'm talking about creating builders with the same name, then deleting them)

run/jobs/pom.xml
averikitsch: The tests are looking for the logs in Cloud Logging. The sample writes to stdout.

run/jobs/src/test/java/com/example/JobsIntegrationTests.java
averikitsch: What specifically? Using substitutions - I want my job and image to have UUID tags, using String.format - looks nicer than +, Cloud Build - no "easy" programmatic way to build a container or deploy to Cloud Run.

lesv: Running commands from code in our tests.

lesv (approved):

lesv (approved):

PR #6846, opened 2022-03-03
chore(secret-manager): add checksum-usage to snippets

gwhitehawk:
Adds checksum-usage snippets to SecretManager samples.

### DIFF
--- a/secretmanager/src/main/java/secretmanager/AccessSecretVersion.java
+++ b/secretmanager/src/main/java/secretmanager/AccessSecretVersion.java
@@ -21,6 +21,8 @@ import com.google.cloud.secretmanager.v1.AccessSecretVersionResponse;
 import com.google.cloud.secretmanager.v1.SecretManagerServiceClient;
 import com.google.cloud.secretmanager.v1.SecretVersionName;
 import java.io.IOException;
+import java.util.zip.CRC32C;
+import java.util.zip.Checksum;
 
 public class AccessSecretVersion {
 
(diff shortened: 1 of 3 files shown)


secretmanager/src/main/java/secretmanager/AccessSecretVersion.java
```diff
@@ -45,6 +47,15 @@ public static void accessSecretVersion(String projectId, String secretId, String
       // Access the secret version.
       AccessSecretVersionResponse response = client.accessSecretVersion(secretVersionName);
 
+      // Verify checksum.
+      byte[] data = response.getPayload().getData().toByteArray();
+      Checksum checksum = new CRC32C();
+      checksum.update(data, 0, data.length);
```
kurtisvg, line 53: nit: consider making this a single line (same with the other example)

kurtisvg (approved):

gwhitehawk:
I have a question regarding java.util.zip.Crc32c usage - this lib only exists since Java 9, so the code doesn't compile for Java 8. Google has a lib (https://cloud.google.com/appengine/docs/standard/java/javadoc/com/google/appengine/api/files/Crc32c) which can be used but it's deprecated (presumably because now an SE lib exists). Is Java8-compatibility necessary in these samples? I wouldn't want to advise users to use a deprecated library - if they still use java8, they could figure out how to replace the standard checksum lib usage with a custom one.

secretmanager/src/main/java/secretmanager/AccessSecretVersion.java
kurtisvg: Sorry, I must not have realized that the checksum was a void return type - this is totally fine.

kurtisvg:
@gwhitehawk - You have 2 options:

1. Changed the `pom.xml` compileTarget property to "11" (which will cause the j8 tests to skip it) and just include a comment to the library to use if they are <=j8
2. Use the older library and leave a comment that says for J9+ they can use the stdlib

PR #6862, opened 2022-03-08
docs(compute-samples): init add preemptible samples and tests

Sita04:
Fixes b/202503297

### DIFF
--- a/compute/cloud-client/src/main/java/compute/preemptible/CreatePreemptibleInstance.java
+++ b/compute/cloud-client/src/main/java/compute/preemptible/CreatePreemptibleInstance.java
@@ -1,0 +1,113 @@
+/*
+ * Copyright 2022 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package compute.preemptible;
+
+// [START compute_preemptible_create]
+
+import com.google.cloud.compute.v1.AttachedDisk;
+import com.google.cloud.compute.v1.AttachedDiskInitializeParams;
+import com.google.cloud.compute.v1.InsertInstanceRequest;
+import com.google.cloud.compute.v1.Instance;
+import com.google.cloud.compute.v1.InstancesClient;
+import com.google.cloud.compute.v1.NetworkInterface;
+import com.google.cloud.compute.v1.Operation;
+import com.google.cloud.compute.v1.Scheduling;
+import java.io.IOException;
+import java.util.concurrent.ExecutionException;
+
+public class CreatePreemptibleInstance {
+
+  public static void main(String[] args)
+      throws IOException, ExecutionException, InterruptedException {
+    // TODO(developer): Replace these variables before running the sample.
+    // projectId: project ID or project number of the Cloud project you want to use.
+    // zone: name of the zone you want to use. For example: “us-west3-b”
+    // instanceName: name of the new virtual machine.
+    String projectId = "your-project-id-or-number";
+    String zone = "zone-name";
+    String instanceName = "instance-name";
+
+    createPremptibleInstance(projectId, zone, instanceName);
+  }
+
+  // Send an instance creation request with preemptible settings to the Compute Engine API
+  // and wait for it to complete.
+  public static void createPremptibleInstance(String projectId, String zone, String instanceName)
+      throws IOException, ExecutionException, InterruptedException {
+
+    String machineType = String.format("zones/%s/machineTypes/e2-small", zone);
+    String sourceImage = String
+        .format("projects/debian-cloud/global/images/family/%s", "debian-11");
+    long diskSizeGb = 10L;
+    String networkName = "default";
+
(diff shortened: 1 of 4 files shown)


prash-mi (commented):
Overall it lgtm, just a couple of observations:

1. Do we want to add test case for 'ListZoneOpetations.preemptionHistory' method instead of `ListPagedResponse listZoneOperations` ?
2. I find it a bit confusing when the `ListZoneOpetations.main` calls `preemptionHistory` and not `listZoneOperations`, probably we may want to refactor/rename the code a bit

Over to @kolea2 to have another look

Sita04:
> Overall it lgtm, just a couple of observations:
> 
> 1. Do we want to add test case for 'ListZoneOpetations.preemptionHistory' method instead of `ListPagedResponse listZoneOperations` ?
> 2. I find it a bit confusing when the `ListZoneOpetations.main` calls `preemptionHistory` and not `listZoneOperations`, probably we may want to refactor/rename the code a bit
> 
> Over to @kolea2 to have another look

The ListZoneOperations snippet lists all operations performed in a particular zone. As an example, I have leveraged the preemptionHistory to list all pre-empted operations in the zone. Hence, the class name is generic.

In the testcase, I called the listZoneOperations directly because that's what we are trying to test. I didn't have to use a helped method like in the main file.

Please lmk if this clarifies.

compute/cloud-client/src/main/java/compute/preemptible/CreatePreemptibleInstance.java
```diff
@@ -0,0 +1,113 @@
...
+  }
+
+  // Send an instance creation request with preemptible settings to the Compute Engine API
+  // and wait for it to complete.
+  public static void createPremptibleInstance(String projectId, String zone, String instanceName)
+      throws IOException, ExecutionException, InterruptedException {
+
+    String machineType = String.format("zones/%s/machineTypes/e2-small", zone);
+    String sourceImage = String
+        .format("projects/debian-cloud/global/images/family/%s", "debian-11");
```
kolea2, line 54: is `debian-11` a constant? If so, why not just inline the entire string? If not, can we make it a variable?

compute/cloud-client/src/main/java/compute/preemptible/CreatePreemptibleInstance.java
```diff
@@ -0,0 +1,113 @@
...
+          .setProject(projectId)
+          .setZone(zone)
+          .setInstanceResource(instanceResource)
+          .build();
+
+      // Wait for the create operation to complete.
+      Operation response = instancesClient.insertAsync(insertInstanceRequest).get();
+
+      if (response.hasError()) {
+        System.out.println("Instance creation failed ! ! " + response);
```
kolea2, line 104: would it be better to throw an exception here?

compute/cloud-client/src/main/java/compute/preemptible/ListZoneOperations.java
```diff
@@ -0,0 +1,99 @@
...
+  // the results to instance name.
+  private static void preemptionHistory(String projectId, String zone, String instanceName)
+      throws IOException {
+
+    String filter;
+    String thisInstanceName;
+    String targetLink;
+    List<List<String>> history = new ArrayList<>();
+
+    if (instanceName.length() != 0) {
```
kolea2, line 75: should we add a null check?

compute/cloud-client/src/main/java/compute/preemptible/ListZoneOperations.java
```diff
@@ -0,0 +1,99 @@
...
+          instanceName);
+    } else {
+      filter = "operationType=\"compute.instances.preempted\"";
+    }
+
+    for (Operation operation : listZoneOperations(projectId, zone, filter).iterateAll()) {
+      targetLink = operation.getTargetLink();
+      thisInstanceName = targetLink.substring(targetLink.lastIndexOf("/") + 1);
+
+      // The filter used is not 100% accurate, it's `contains` not `equals`
```
kolea2, line 87: can we use equals?

compute/cloud-client/src/main/java/compute/preemptible/CreatePreemptibleInstance.java
Sita04: It's a variable. Since, the other samples also follow this particular way of initialization, I've stuck to the same.

compute/cloud-client/src/main/java/compute/preemptible/ListZoneOperations.java
Sita04: Contains would be a better fit here, because the results are not tightly coupled with the filter we are using. Hence, we are overcoming it by a workaround.

compute/cloud-client/src/main/java/compute/preemptible/CreatePreemptibleInstance.java
Sita04: That would be nice. But we thought of a graceful exception handling and hence followed this style across all previous samples.

compute/cloud-client/src/main/java/compute/preemptible/ListZoneOperations.java
Sita04: Will modify this. Thanks.

compute/cloud-client/src/main/java/compute/preemptible/ListZoneOperations.java
kolea2: I would update the comment here with this explanation instead ^

Sita04: Done. Thanks.

compute/cloud-client/src/main/java/compute/preemptible/CreatePreemptibleInstance.java
Sita04: Done. Thanks.

kolea2 (approved):

PR #6879, opened 2022-03-21
healthcare API: do not use client library for FHIR search requests

noerog:
Part of fix for b/221324824. Our client libraries do not allow for FHIR GET/POST search requests to be set in the HTTP request path, which is required by the FHIR specification. Due to a bug, you can send search requests by specifying a resource type in the request body, but this is an anti-pattern. Until we fix the issue with the client library, we need to use the Apache HTTP client to explicitly make GET/POST requests and not use the client library.

### DIFF
--- a/healthcare/v1/src/main/java/snippets/healthcare/fhir/resources/FhirResourceSearchGet.java
+++ b/healthcare/v1/src/main/java/snippets/healthcare/fhir/resources/FhirResourceSearchGet.java
@@ -42,7 +42,7 @@ public class FhirResourceSearchGet {
   private static final String FHIR_NAME =
       "projects/%s/locations/%s/datasets/%s/fhirStores/%s/fhir/%s";
   // The endpoint URL for the Healthcare API. Required for HttpClient.
-  private static final String ROOT_URL = "https://healthcare.googleapis.com";
+  private static final String API_ENDPOINT = "https://healthcare.googleapis.com";
   private static final JsonFactory JSON_FACTORY = new JacksonFactory();
   private static final NetHttpTransport HTTP_TRANSPORT = new NetHttpTransport();
 
--- a/healthcare/v1/src/main/java/snippets/healthcare/fhir/resources/FhirResourceSearchGet.java
+++ b/healthcare/v1/src/main/java/snippets/healthcare/fhir/resources/FhirResourceSearchGet.java
@@ -22,15 +22,21 @@ import com.google.api.client.http.javanet.NetHttpTransport;
 import com.google.api.client.json.JsonFactory;
 import com.google.api.client.json.jackson2.JacksonFactory;
 import com.google.api.services.healthcare.v1.CloudHealthcare;
-import com.google.api.services.healthcare.v1.CloudHealthcare.Projects.Locations.Datasets.FhirStores.Fhir.Search;
 import com.google.api.services.healthcare.v1.CloudHealthcareScopes;
-import com.google.api.services.healthcare.v1.model.HttpBody;
-import com.google.api.services.healthcare.v1.model.SearchResourcesRequest;
 import com.google.auth.http.HttpCredentialsAdapter;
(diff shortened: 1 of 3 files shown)


lesv:
@noerog I'm no longer the person to give this to, you should be able to unassign Averi and it will pick someone else.  That said, @kolea2 guides the reviewers.  Moving from GAPIC -> APIary, I wasn't thinking when I approved earlier, but @danoscarmike should comment on that.

healthcare/v1/src/main/java/snippets/healthcare/fhir/resources/FhirResourceSearchGet.java
```diff
@@ -22,42 +22,63 @@
...
-    // String fhirStoreName =
+    // String resourceName =
     //    String.format(
     //        FHIR_NAME, "project-id", "region-id", "dataset-id", "fhir-store-id");
     // String resourceType = "Patient";
 
-    // Initialize the client, which will be used to interact with the service.
-    CloudHealthcare client = createClient();
+    // The endpoint URL for the Healthcare API. Required for HttpClient.
+    String rootUrl = "https://healthcare.googleapis.com";
```
kolea2, line 55: I think this should be a final variable at the top

healthcare/v1/src/main/java/snippets/healthcare/fhir/resources/FhirResourceSearchPost.java
```diff
@@ -50,12 +50,18 @@ public static void fhirResourceSearchPost(String resourceName)
     //    String.format(
     //        FHIR_NAME, "project-id", "region-id", "dataset-id", "store-id", "resource-type");
 
-    // Initialize the client, which will be used to interact with the service.
-    CloudHealthcare client = createClient();
+    // The endpoint URL for the Healthcare API. Required for HttpClient.
+    String rootUrl = "https://healthcare.googleapis.com";
```
kolea2, line 54: same here

kolea2 (changes requested):

healthcare/v1/src/main/java/snippets/healthcare/fhir/resources/FhirResourceSearchGet.java
noerog: Done.

healthcare/v1/src/main/java/snippets/healthcare/fhir/resources/FhirResourceSearchPost.java
noerog: Done.

healthcare/v1/src/main/java/snippets/healthcare/fhir/resources/FhirResourceSearchGet.java
```diff
@@ -22,42 +22,62 @@
...
+import org.apache.http.client.methods.RequestBuilder;
+import org.apache.http.client.utils.URIBuilder;
+import org.apache.http.entity.StringEntity;
+import org.apache.http.impl.client.HttpClients;
 
 public class FhirResourceSearchGet {
   private static final String FHIR_NAME =
       "projects/%s/locations/%s/datasets/%s/fhirStores/%s/fhir/%s";
+  // The endpoint URL for the Healthcare API. Required for HttpClient.
+  private static final String ROOT_URL = "https://healthcare.googleapis.com";
```
kolea2, line 45: sorry, last nit - I would rename this to something like `API_ENDPOINT` or something similar.

kolea2 (approved):
LGTM after last nit

healthcare/v1/src/main/java/snippets/healthcare/fhir/resources/FhirResourceSearchGet.java
noerog: Done.

PR #7070, opened 2022-04-27
docs(compute-samples): added region tags

Sita04:


compute/cloud-client/src/main/java/compute/GetInstanceTemplate.java
```diff
@@ -16,6 +16,8 @@
 
 package compute;
 
+// [START compute_template_get]
+
```
savijatv, line 20: Can't add a comment at the unchanged line #31. Is this a 'new' template that we're retrieving here? Should it be 'name of the template to retrieve'

Sita04: Thanks. Modified.

eaball35 (approved):

PR #7097, opened 2022-05-03
chore: require Java 17 CI check

shubha-rajan:


lesv:
@shubha-rajan LGTM, but you might wish to look at recent problems in nightly before commiting this.

shubha-rajan:
Possible reasons why Java 17 is running all the tests instead of just the diff are due to git behaving differently in the images. 
The Java 17 image is using an earlier version of git (2.25) which is throwing errors including:
- `error: Could not access 'origin/main...'` (from running git diff `$GIT_DIFF`)
- `fatal: unsafe repository ('/tmpfs/src/github/java-docs-samples' is owned by someone else)` (from running git branch -a)
Upgrading the git version might fix it.

PR #7117, opened 2022-05-06
chore: add maven-war-plugin to fix java17 periodic failures

shubha-rajan:
A lot of tests are failing in Java 17 with an error that looks like: 
```
[WARNING] Error injecting: org.apache.maven.plugin.war.WarMojo
com.google.inject.ProvisionException: Unable to provision, see the following errors:

1) Error injecting constructor, java.lang.ExceptionInInitializerError: Cannot access defaults field of Properties
  at org.apache.maven.plugin.war.WarMojo.<init>(Unknown Source)
  while locating org.apache.maven.plugin.war.WarMojo
...
```
Adding maven-war-plugin fixes these errors.

### DIFF
--- a/appengine-java11-bundled-services/datastore/pom.xml
+++ b/appengine-java11-bundled-services/datastore/pom.xml
@@ -127,6 +127,11 @@
     <outputDirectory>${project.build.directory}/${project.build.finalName}/WEB-INF/classes</outputDirectory>
     <plugins>
       <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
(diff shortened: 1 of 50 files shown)


ludoch:
Can't it be done in a more centralized location? Parent pom?

shubha-rajan:
> Can't it be done in a more centralized location? Parent pom?

I considered that, but not all the samples need the plugin. Should we still add it to the parent pom?

lesv (approved):
LGTM, but I'm w/ Ludo, this probably could have been done in java-repo-tools, but this is likely to fix things quicker.

PR #7179, opened 2022-06-02
Create UpdateSecretWithAlias

nataliesalvati:
Code sample update for Access By Alias Launch

Fixes #issue


secretmanager/src/test/java/secretmanager/SnippetsIT.java
```diff
@@ -359,4 +359,12 @@ public void testUpdateSecret() throws IOException {
 
     assertThat(stdOut.toString()).contains("Updated secret");
   }
+  
+  @Test
+  public void testUpdateSecretWithAlias() throws IOException {
+    SecretName name = SecretName.parse(TEST_SECRET_WITH_VERSIONS.getName());
+    UpdateSecretWithAlias.updateSecret(name.getProject(), name.getSecret());
+
+    assertThat(stdOut.toString()).contains("test");
```
shubha-rajan, line 368: is there a way that we can set the alias to something more specific than "test"? I'm worried about possible false positives with this test passing if, for example, the secret name has "test" in it. Maybe update `updateSecret` to take in the alias string as an argument, and append a uuid to it or something?

shubha-rajan (changes requested):

secretmanager/src/test/java/secretmanager/SnippetsIT.java
nataliesalvati: I think that's a fair concern, but then this wouldn't be consistent across the other samples. The secret name shouldn't show up in the output though, since the sample prints the alias map, and not the secret itself.

shubha-rajan: in that case it should be fine, and if we start running into issues, we can change it.

shubha-rajan (approved):

shubha-rajan (approved):

PR #7186, opened 2022-06-13
fix: fix tests using wrong Pub/Sub func name

ace-n:
Fixes issue encountered in #6876

### DIFF
--- a/.kokoro/tests/build_cloud_functions.sh
+++ b/.kokoro/tests/build_cloud_functions.sh
@@ -42,6 +42,17 @@ export FUNCTIONS_GCS_FN_NAME="gcs-${SUFFIX}"
 # Set identity token (required for functions without --allow-unauthenticated)
 export FUNCTIONS_IDENTITY_TOKEN=$(gcloud auth print-identity-token)
 
+# Identify function language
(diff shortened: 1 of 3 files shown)


.kokoro/tests/build_cloud_functions.sh
```diff
@@ -57,7 +68,7 @@ elif [[ "$file" == *"hello-pubsub"* ]]; then
   gcloud functions deploy $FUNCTIONS_PUBSUB_FN_NAME \
     --region $FUNCTIONS_REGION \
     --runtime $FUNCTIONS_JAVA_RUNTIME \
-    --entry-point "functions.HelloPubSub" \
+    --entry-point "functions.${LANGUAGE}HelloPubSub" \
```
averikitsch, line 71: I didn't check all of them but [this](https://github.com/GoogleCloudPlatform/java-docs-samples/blob/2990b83f517c3948385f1610a0ead297aeb53d67/functions/helloworld/kotlin-hello-pubsub/src/main/kotlin/functions/KotlinHelloBackground.kt#L27) shows that it's the file name not the entrypoint name that has LANGUAGE prefixed

anguillanneuf (approved):
Thanks Ace!

PR #7187, opened 2022-06-14
test: add got to assertion message

olavloite:
Adds the actual change text that was received to the assertion message
so the log will contain a hint as to why the test seems to be flaky.

cc @Neenu1995 @haikuo-google 

Updates #7165


Shabirmean (approved):

olavloite:
@haikuo-google The change that we receive now has a small difference with what was (initially) expected. The test case expects `oldValuesJson=''`, while the actual value is `oldValuesJson='{}'`. (I.e. initially we expected an empty string, now we get an empty JSON object).

Is this change in behavior expected?

averikitsch (approved):

PR #7188, opened 2022-06-14
chore: update GSON dependency

averikitsch:
Fixes https://github.com/advisories/GHSA-4jrv-ppp4-jm57


lesv (approved):

lesv:
Looks like something is unhappy.

averikitsch:
Adding notes here. The monitoring sample is not deleting the uptime checks therefore there is an too many entity error for the project. This request ` client.deleteUptimeCheckConfig(UptimeCheckConfigName.of(projectId, checkName));` does not fail if entity is not found.

PR #7224, opened 2022-07-21
docs(compute-samples): added samples for moving VM instances

Sita04:
Added samples to move instances across zones (https://cloud.google.com/compute/docs/instances/moving-instance-across-zones)

Apologies for the PR size.

### DIFF
--- a/compute/cloud-client/src/main/java/compute/CreateInstanceWithExistingDisks.java
+++ b/compute/cloud-client/src/main/java/compute/CreateInstanceWithExistingDisks.java
@@ -1,0 +1,118 @@
+// Copyright 2022 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package compute;
+
+// [START compute_instances_create_with_existing_disks]
+
+import com.google.cloud.compute.v1.AttachedDisk;
+import com.google.cloud.compute.v1.Disk;
+import com.google.cloud.compute.v1.DisksClient;
+import com.google.cloud.compute.v1.InsertInstanceRequest;
+import com.google.cloud.compute.v1.Instance;
+import com.google.cloud.compute.v1.InstancesClient;
+import com.google.cloud.compute.v1.NetworkInterface;
+import com.google.cloud.compute.v1.Operation;
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
+
+public class CreateInstanceWithExistingDisks {
+
+  public static void main(String[] args)
+      throws IOException, ExecutionException, InterruptedException, TimeoutException {
+    // TODO(developer): Replace these variables before running the sample.
+
+    // Project ID or project number of the Cloud project you want to use.
+    String projectId = "YOUR_PROJECT_ID";
+
+    // Name of the zone to create the instance in. For example: "us-west3-b"
+    String zone = "europe-central2-b";
+
+    // Name of the new virtual machine (VM) instance.
+    String instanceName = "YOUR_INSTANCE_NAME";
+
+    // Array of disk names to be attached to the new virtual machine.
+    // First disk in this list will be used as the boot device.
+    List<String> diskNames = new ArrayList<>();
+
+    createInstanceWithExistingDisks(projectId, zone, instanceName, diskNames);
+  }
+
+  // Create a new VM instance using selected disks. The first disk in diskNames will be used as boot disk.
+  public static void createInstanceWithExistingDisks(String projectId, String zone,
+      String instanceName, List<String> diskNames)
+      throws IOException, ExecutionException, InterruptedException, TimeoutException {
+
+    // Initialize client that will be used to send requests. This client only needs to be created
+    // once, and can be reused for multiple requests. After completing all of your requests, call
+    // the `instancesClient.close()` method on the client to safely
+    // clean up any remaining background resources.
+    try (InstancesClient instancesClient = InstancesClient.create();
+        DisksClient disksClient = DisksClient.create()) {
+
+      if (diskNames.size() == 0) {
+        throw new Error("At least one disk should be provided");
+      }
+
+      // Create the list of attached disks to be used in instance creation.
+      List<AttachedDisk> attachedDisks = new ArrayList<>();
+      Disk disk = null;
+      for (String name : diskNames) {
+        disk = disksClient.get(projectId, zone, name);
+        attachedDisks.add(
+            AttachedDisk.newBuilder()
+                .setSource(disk.getSelfLink())
+                .build()
+        );
+      }
+
+      // Make the first disk in the list as the boot device.
+      attachedDisks.set(0,
+          AttachedDisk.newBuilder(attachedDisks.get(0)).setBoot(true).build());
+
+      // Create the instance.
+      Instance instance = Instance.newBuilder()
+          .setName(instanceName)
+          // Add the attached disks to the instance.
+          .addAllDisks(attachedDisks)
+          .setMachineType(String.format("zones/%s/machineTypes/n1-standard-1", zone))
+          .addNetworkInterfaces(
+              NetworkInterface.newBuilder().setName("global/networks/default").build())
+          .build();
+
+      // Create the insert instance request.
+      InsertInstanceRequest insertInstanceRequest = InsertInstanceRequest.newBuilder()
+          .setProject(projectId)
+          .setZone(zone)
+          .setInstanceResource(instance)
+          .build();
+
+      // Wait for the create operation to complete.
+      Operation response = instancesClient.insertAsync(insertInstanceRequest)
+          .get(3, TimeUnit.MINUTES);
+
+      if (response.hasError()) {
+        System.out.println("Instance creation failed ! ! " + response);
+        return;
+      }
+      System.out.println("Operation Status: " + response.getStatus());
+
+    }
+  }
+}
+// [END compute_instances_create_with_existing_disks]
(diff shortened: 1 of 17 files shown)


Shabirmean:
Thank you very much for the samples @Sita04 , these looks great 😃 

Just a few thoughts:
- I see that there are multiple different actions that are part of this PR. I feel like it will make it easy on the reviewers if this was broken into multiple PRs.
   - Get instance
   - ListDisks, CreateEmptyDisk, DeleteDisk 
   - CreateWithDisks, CreateDiskFromImage, CreateDiskFromSnapshot
   - ListSnapshots, CreateSnapshot, DeleteSnapshot
   - SetDiskAutodelete

- Breaking up the PR will also ensure that important details of the PR are not overlooked by the reviewers

- Also, was the expectation to write a specific sample to showcase instance migration to a new zone? I looked around and couldn't exactly find a specific sample that shows this action.

Let me know what you think.

Sita04:
Thanks for the review, @Shabirmean :)

Agree with the PR splitting strategy. Would be easier for the reviewers. I will be more careful the next time. Thanks for pointing it out.

> the expectation to write a specific sample to showcase instance migration to a new zone?

Yes, these samples represent a collection that could be used in the doc under moving VM instances page. They do not refer to a sample on how to move instances.
Please refer to this bug for more info: b/202504009

compute/cloud-client/src/main/java/compute/disks/CreateSnapshot.java
```diff
@@ -0,0 +1,134 @@
...
+    String region = "your-disk-region";
+
+    // The Cloud Storage multi-region or the Cloud Storage region where you
+    // want to store your snapshot.
+    // You can specify only one storage location. Available locations:
+    // https://cloud.google.com/storage/docs/locations#available-locations
+    String location = "europe-central2";
+
+    // project ID or project number of the Cloud project that
+    // hosts the disk you want to snapshot. If not provided, will look for
```
savijatv, line 61: Please review and change this text accordingly.

If you do not provide an ID or number for the source disk project, a disk from the project specified by the `projectId` will be taken for creating the snapshot.

compute/cloud-client/src/main/java/compute/disks/DeleteDisk.java
```diff
@@ -0,0 +1,74 @@
...
+public class DeleteDisk {
+
+  public static void main(String[] args)
+      throws IOException, ExecutionException, InterruptedException, TimeoutException {
+    // TODO(developer): Replace these variables before running the sample.
+
+    // Project ID or project number of the Cloud project you want to use.
+    String projectId = "YOUR_PROJECT_ID";
+
+    // Name of the zone in which is the disk you want to delete.
```
savijatv, line 36: The zone from where you want to delete the disk.

compute/cloud-client/src/main/java/compute/disks/ListDisks.java
```diff
@@ -0,0 +1,69 @@
...
+public class ListDisks {
+
+  public static void main(String[] args)
+      throws IOException, ExecutionException, InterruptedException, TimeoutException {
+    // TODO(developer): Replace these variables before running the sample.
+
+    // Project ID or project number of the Cloud project you want to use.
+    String projectId = "YOUR_PROJECT_ID";
+
+    // Name of the zone in which to list the disks.
```
savijatv, line 35: The zone where the disks are located.

compute/cloud-client/src/main/java/compute/disks/SetDiskAutodelete.java
```diff
@@ -0,0 +1,100 @@
...
+public class SetDiskAutodelete {
+
+  public static void main(String[] args)
+      throws IOException, ExecutionException, InterruptedException, TimeoutException {
+    // TODO(developer): Replace these variables before running the sample.
+
+    // Project ID or project number of the Cloud project you want to use.
+    String projectId = "YOUR_PROJECT_ID";
+
+    // Name of the zone in which is the disk you want to modify.
```
savijatv, line 37: The zone of the disk that you want to modify.

compute/cloud-client/src/main/java/compute/disks/SetDiskAutodelete.java
```diff
@@ -0,0 +1,100 @@
...
+    String zone = "europe-central2-b";
+
+    // Name of the instance the disk is attached to.
+    String instanceName = "YOUR_INSTANCE_NAME";
+
+    // The name of the disk which flag you want to modify.
```
savijatv, line 43: The name of the disk that you want to modify? or The name of the disk for which you want to modify the autodelete flag?

compute/cloud-client/src/main/java/compute/disks/CreateSnapshot.java
Sita04: Done, modified it to sound better. PTAL.

compute/cloud-client/src/main/java/compute/CreateInstanceWithExistingDisks.java
```diff
@@ -0,0 +1,119 @@
...
+
+    // Name of the zone to create the instance in. For example: "us-west3-b"
+    String zone = "europe-central2-b";
+
+    // Name of the new virtual machine (VM) instance.
+    String instanceName = "YOUR_INSTANCE_NAME";
+
+    // Array of disk names to be attached to the new virtual machine.
+    // First disk in this list will be used as the boot disk.
+    List<String> diskNames = new ArrayList<>();
```
Shabirmean, line 51: Should we have a list of names here?  I think we should add some examples here for the user to change. If not it looks as if an empty list is okay.

```suggestion
    List<String> diskNames = List.of("your-boot-disk", "another-disk1", "another-disk2");
```

compute/cloud-client/src/main/java/compute/CreateInstanceWithExistingDisks.java
```diff
@@ -0,0 +1,119 @@
...
+    try (InstancesClient instancesClient = InstancesClient.create();
+        DisksClient disksClient = DisksClient.create()) {
+
+      if (diskNames.size() == 0) {
+        throw new Error("At least one disk should be provided");
+      }
+
+      // Create the list of attached disks to be used in instance creation.
+      List<AttachedDisk> attachedDisks = new ArrayList<>();
+      Disk disk = null;
```
Shabirmean, line 75: Why define this outside? Can we move it into the for loop?

compute/cloud-client/src/main/java/compute/CreateInstanceWithExistingDisks.java
```diff
@@ -0,0 +1,119 @@
...
+        attachedDisks.add(
+            AttachedDisk.newBuilder()
+                .setSource(disk.getSelfLink())
+                .build()
+        );
+      }
+
+      // Make the first disk in the list as the boot disk.
+      attachedDisks.set(0,
+          AttachedDisk.newBuilder(attachedDisks.get(0)).setBoot(true).build());
```
Shabirmean, line 87:
```suggestion
      for (int i = 0; i < diskNames.size(); i++) {
        String diskName = diskNames.get(i);
        Disk disk = disksClient.get(projectId, zone, diskName);
        AttachedDisk attDisk = null;

        if (i == 0) {
            // Set the first disk in the list as boot disk.
            attDisk = AttachedDisk.newBuilder().setSource(disk.getSelfLink()).setBoot(true).build();
        } else {
            attDisk = AttachedDisk.newBuilder().setSource(disk.getSelfLink()).build();
        }
        attachedDisks.add(attDisk);
      }
```

compute/cloud-client/src/main/java/compute/disks/CreateSnapshot.java
```diff
@@ -0,0 +1,134 @@
...
+    String zone = "europe-central2-b";
+
+    // The region of the source disk from which you create the snapshot (for regional disks).
+    String region = "your-disk-region";
+
+    // The Cloud Storage multi-region or the Cloud Storage region where you
+    // want to store your snapshot.
+    // You can specify only one storage location. Available locations:
+    // https://cloud.google.com/storage/docs/locations#available-locations
+    String location = "europe-central2";
```
Shabirmean, line 58: Are the regions and locations the same? As in are they from the same list? if so can we use one variable for both here?

compute/cloud-client/src/main/java/compute/disks/ListDisks.java
```diff
@@ -0,0 +1,69 @@
...
+      ListDisksRequest listDisksRequest = ListDisksRequest.newBuilder()
+          .setProject(projectId)
+          .setZone(zone)
+          .setFilter(filter)
+          .build();
+
+      for (Disk disk : disksClient.list(listDisksRequest).iterateAll()) {
+        System.out.println(disk.getName());
+      }
+      System.out.println("Listed all disks.");
```
Shabirmean, line 65: Can we move this to be printed before the for loop and say `List of disks:`

compute/cloud-client/src/main/java/compute/disks/ListSnapshots.java
```diff
@@ -0,0 +1,61 @@
...
+      // Create the List Snapshot request.
+      ListSnapshotsRequest listSnapshotsRequest = ListSnapshotsRequest.newBuilder()
+          .setProject(projectId)
+          .setFilter(filter)
+          .build();
+
+      for (Snapshot snapshot : snapshotsClient.list(listSnapshotsRequest).iterateAll()) {
+        System.out.println(snapshot.getName());
+      }
+      System.out.println("Listed all snapshots");
```
Shabirmean, line 57: Same as ListDisk, can we move it before the for loop and say `List of snapshots:`

savijatv (approved):

compute/cloud-client/src/main/java/compute/CreateInstanceWithExistingDisks.java
Sita04: That's a great suggestion. I've added it. Thanks.

compute/cloud-client/src/main/java/compute/disks/CreateSnapshot.java
Sita04: They are different in that locations might have dual or multiple regions. The location list may also vary with time and need not always be the same as the region list. 
Hence, I would prefer to keep it as two separate variables, if that makes sense?

Sita04:
@Shabirmean Thanks for the review. I've incorporated your suggestions. PTAL.

compute/cloud-client/src/main/java/compute/disks/CreateSnapshot.java
Shabirmean: Sounds good to me

Shabirmean (approved):
LGTM! Thank you for taking in all the suggestions!

PR #7310, opened 2022-08-26
docs(compute-samples): add sample and test for local SSD disk

Sita04:
Sample to create disk with local SSD

### DIFF
--- a/compute/cloud-client/src/main/java/compute/CreateWithLocalSsd.java
+++ b/compute/cloud-client/src/main/java/compute/CreateWithLocalSsd.java
@@ -38,31 +38,35 @@ public class CreateWithLocalSsd {
   public static void main(String[] args)
       throws IOException, ExecutionException, InterruptedException, TimeoutException {
     // TODO(developer): Replace these variables before running the sample.
-    // project: project ID or project number of the Cloud project you want to use.
-    String project = "your-project-id";
+    // projectId: project ID or project number of the Cloud project you want to use.
+    String projectId = "your-project-id";
     // zone: name of the zone to create the instance in. For example: "us-west3-b"
     String zone = "zone-name";
     // instanceName: name of the new virtual machine (VM) instance.
     String instanceName = "instance-name";
 
-    createWithLocalSsd(project, zone, instanceName);
+    createWithLocalSsd(projectId, zone, instanceName);
   }
 
   // Create a new VM instance with Debian 10 operating system and SSD local disk.
-  public static void createWithLocalSsd(String project, String zone, String instanceName)
+  public static void createWithLocalSsd(String projectId, String zone, String instanceName)
       throws IOException, ExecutionException, InterruptedException, TimeoutException {
 
+    int diskSizeGb = 10;
+    boolean boot = true;
+    boolean autoDelete = true;
+    String diskType = String.format("zones/%s/diskTypes/pd-standard", zone);
     // Get the latest debian image.
     Image newestDebian = getImageFromFamily("debian-cloud", "debian-10");
+    List<AttachedDisk> disks = new ArrayList<>();
 
     // Create the disks to be included in the instance.
-    List<AttachedDisk> disks = new ArrayList<>();
-    String diskType = String.format("zones/%s/diskTypes/pd-standard", zone);
-    disks.add(diskFromImage(diskType, 10, true, newestDebian.getSelfLink(), true));
-    disks.add(localSsdDisk(zone));
+    disks.add(
+        createDiskFromImage(diskType, diskSizeGb, boot, newestDebian.getSelfLink(), autoDelete));
+    disks.add(createLocalSsdDisk(zone));
 
     // Create the instance.
-    Instance instance = createInstance(project, zone, instanceName, disks);
+    Instance instance = createInstance(projectId, zone, instanceName, disks);
 
     if (instance != null) {
       System.out.printf("Instance created with local SSD: %s", instance.getName());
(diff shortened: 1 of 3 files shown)


prash-mi:
LGTM! @kolea2 Do you want to take a quick glance and let us know your thoughts?

compute/cloud-client/src/main/java/compute/CreateWithLocalSsd.java
```diff
@@ -0,0 +1,197 @@
...
+  public static void createWithLocalSsd(String project, String zone, String instanceName)
+      throws IOException, ExecutionException, InterruptedException, TimeoutException {
+
+    // Get the latest debian image.
+    Image newestDebian = getImageFromFamily("debian-cloud", "debian-10");
+
+    // Create the disks to be included in the instance.
+    List<AttachedDisk> disks = new ArrayList<>();
+    String diskType = String.format("zones/%s/diskTypes/pd-standard", zone);
+    disks.add(diskFromImage(diskType, 10, true, newestDebian.getSelfLink(), true));
```
kolea2, line 61: what are these parameters? `10`, `true`, etc seem like magic numbers. Can we replace them with variable names or refactor this to be clearer?

compute/cloud-client/src/main/java/compute/CreateWithLocalSsd.java
```diff
@@ -0,0 +1,197 @@
...
+  //    boot disk of an instance.
+  //
+  //    sourceImage: source image to use when creating this disk.
+  //    You must have read access to this disk. This can be one of the publicly available images
+  //    or an image from one of your projects.
+  //    This value uses the following format: "projects/{project_name}/global/images/{image_name}"
+  //
+  //    autoDelete: boolean flag indicating whether this disk should be deleted
+  //    with the VM that uses it.
+  private static AttachedDisk diskFromImage(String diskType, int diskSizeGb, boolean boot,
```
kolea2, line 108: `createDiskFromImage` or `getDiskFromImage` would be a little clearer IMO

compute/cloud-client/src/main/java/compute/CreateWithLocalSsd.java
```diff
@@ -0,0 +1,197 @@
...
+        .build();
+
+    return bootDisk;
+  }
+
+  // Create an AttachedDisk object to be used in VM instance creation. The created disk contains
+  // no data and requires formatting before it can be used.
+  // Args:
+  //    zone: The zone in which the local SSD drive will be attached.
+  private static AttachedDisk localSsdDisk(String zone) {
```
kolea2, line 133: same comment here, consider `create` or `get` depending on which is more appropriate.

compute/cloud-client/src/main/java/compute/CreateWithLocalSsd.java
```diff
@@ -0,0 +1,197 @@
...
+    // Initialize client that will be used to send requests. This client only needs to be created
+    // once, and can be reused for multiple requests. After completing all of your requests, call
+    // the `instancesClient.close()` method on the client to safely
+    // clean up any remaining background resources.
+    try (InstancesClient instancesClient = InstancesClient.create()) {
+
+      // machineType: machine type of the VM being created. This value uses the
+      // following format: "zones/{zone}/machineTypes/{type_name}".
+      // For example: "zones/europe-west3-c/machineTypes/f1-micro"
+      String machineType = String.format("zones/%s/machineTypes/%s", zone, "n1-standard-1");
```
kolea2, line 168: `n1-standard-1` should be replaced as a variable or inlined into the string itself.

compute/cloud-client/src/main/java/compute/CreateWithLocalSsd.java
```diff
@@ -0,0 +1,197 @@
...
+          .setMachineType(machineType)
+          .addNetworkInterfaces(NetworkInterface.newBuilder().setName(networkLink).build())
+          .addAllDisks(disks)
+          .build();
+
+      Operation response = instancesClient.insertAsync(project, zone, instance)
+          .get(3, TimeUnit.MINUTES);
+
+      if (response.hasError()) {
+        System.out.println("Instance creation failed ! ! " + response);
```
kolea2, line 187: should we throw here instead?

kolea2 (changes requested):

compute/cloud-client/src/main/java/compute/CreateWithLocalSsd.java
```diff
@@ -0,0 +1,197 @@
...
+      System.out.printf("Instance created with local SSD: %s", instance.getName());
+    }
+
+  }
+
+  // Retrieve the newest image that is part of a given family in a project.
+  // Args:
+  //    project: project ID or project number of the Cloud project you want to get image from.
+  //    family: name of the image family you want to get image from.
+  private static Image getImageFromFamily(String project, String family) throws IOException {
```
kolea2, line 77: I would rename this explicitly to `projectId` or `projectNumber` (and anywhere else)

kolea2 (changes requested):

compute/cloud-client/src/main/java/compute/CreateWithLocalSsd.java
Sita04: Done. Added variables to make the params value clearer. Thanks.

compute/cloud-client/src/main/java/compute/CreateWithLocalSsd.java
Sita04: Replaced with createDiskFromImage()

compute/cloud-client/src/main/java/compute/CreateWithLocalSsd.java
Sita04: Replaced with getLocalSsdDisk()

Sita04:
@kolea2 Thanks for your review! Made all the changes. PTAL

compute/cloud-client/src/main/java/compute/CreateWithLocalSsd.java
kolea2: I think this still needs to be renamed

kolea2 (approved):
lgtm after last variable change made. Thanks!

PR #7328, opened 2022-09-30
fix(samples): Fix Video Stitcher region tag prefix

irataxy:
b:249884966

### DIFF
--- a/media/stitcher/src/main/java/com/example/stitcher/CreateCdnKey.java
+++ b/media/stitcher/src/main/java/com/example/stitcher/CreateCdnKey.java
@@ -16,7 +16,7 @@
 
 package com.example.stitcher;
 
-// [START video_stitcher_create_cdn_key]
+// [START videostitcher_create_cdn_key]
 
 import com.google.cloud.video.stitcher.v1.AkamaiCdnKey;
 import com.google.cloud.video.stitcher.v1.CdnKey;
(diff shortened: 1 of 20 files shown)


alicejli (commented):
Hi! Thanks for putting this together - I'm actually working on similar fixes for the autogenerated snippet region tags (https://github.com/googleapis/gapic-generator-java/pull/1047).

Since these samples are the handwritten ones instead of the autogenerated ones, I'm not sure if you need to follow the same convention. 

Here is the canonical C# example that my PR is aiming to follow: **https://github.com/googleapis/google-cloud-dotnet/blob/main/apis/Google.Cloud.Video.Stitcher.V1/Google.Cloud.Video.Stitcher.V1.GeneratedSnippets/VideoStitcherServiceClient.CreateCdnKeyAsyncSnippet.g.cs**

Would you mind double-checking with the devrel team if your region tag is fine as-is, or if the latter part should capitalized in upperCamelCase?

I also plan to enable my autogenerated snippets for all repos at some point; not sure why videostitcher was excluded (if the Video Stitcher team would rather keep these handwritten samples and not get autogenerated ones, that's fine) in the first place.

irataxy:
@alicejli The region tags for these handwritten samples conform to the canonical DLP example per go/cloud-java-samples. These region tags are the same for all handwritten samples (includes multiple programming languages) for this library.

I think it is fine to include both handwritten samples and autogenerated ones in the repo.

alicejli (approved):
Understood, thank you for clarifying. Thank you for your work on this!

PR #7339, opened 2022-10-12
test: Temporarily ignore flaky Spanner test

kweinmeister:
#6831 has been open since Feb 26. It is a major source of build flakiness. It is also blocking important updates such as #6966, which updates the core library BOM across 46 samples.

This change is intended to restore build health while #6831 is open. That fix needs to include removing the `@Ignore` annotation added here.


Sita04 (approved):
LGTM. @kurtisvg PTAL.

PR #7353, opened 2022-10-20
test: address "functions in use" build failures

### LINKED ISSUE #7351: Flaky GCF tests due to functions still running
The [build](https://source.cloud.google.com/results/invocations/486c6b48-2667-4910-b3a2-9dcb4354f111/targets/cloud-devrel%2Fjava%2Fjava-docs-samples%2Fjava11%2Fperiodic/log) occasionally has these errors:

```
ERROR: (gcloud.functions.deploy) ResponseError: status=[400], code=[Ok], message=[An operation on function pubsub-*** in region us-central1 in project *** is already in progress. Please try again later.]
```

kweinmeister:
Fixes #7351

### DIFF
--- a/.kokoro/tests/build_cloud_functions.sh
+++ b/.kokoro/tests/build_cloud_functions.sh
@@ -33,7 +33,7 @@ requireEnv "FUNCTIONS_BUCKET"
 # to run them concurrently.
 export SAMPLE_VERSION="${KOKORO_GIT_COMMIT:-latest}"
 # Builds not triggered by a PR will fall back to the commit hash then "latest".
-SUFFIX=${KOKORO_GITHUB_PULL_REQUEST_NUMBER:-${SAMPLE_VERSION:0:12}}-$(date +%s)
+SUFFIX=${KOKORO_GITHUB_PULL_REQUEST_NUMBER:-${SAMPLE_VERSION:0:12}}-$(date +%s%N)
 
 export FUNCTIONS_HTTP_FN_NAME="http-${SUFFIX}"
 export FUNCTIONS_PUBSUB_FN_NAME="pubsub-${SUFFIX}"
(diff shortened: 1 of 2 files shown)


.kokoro/tests/build_cloud_functions.sh
```diff
@@ -56,28 +56,33 @@ fi
...
-  echo "Deploying function HelloGcs to: ${FUNCTIONS_GCS_FN_NAME}"
-  gcloud functions deploy $FUNCTIONS_GCS_FN_NAME \
-    --region $FUNCTIONS_REGION \
-    --runtime $FUNCTIONS_JAVA_RUNTIME \
-    --entry-point "functions.HelloGcs" \
-    --trigger-bucket $FUNCTIONS_BUCKET
-fi
+for i in {1..5}; do # Retry up to 5 times
+  if [[ "$file" == *"hello-http"* ]]; then
+    echo "Deploying function HelloHttp to: ${FUNCTIONS_HTTP_FN_NAME}"
```
kurtisvg, line 61: Look at the issue, it looks like the error `An operation on function pubsub-37ee50795f85 in region us-central1 in project java-docs-samples-testing is already in progress. Please try again later.` happens when you try to deploy a new function when the existing on already has an ongoing operation. 

I'm concerned this isn't the right fix because the root cause is a shared function name. It looks like the function name is derived from:
```
SUFFIX=${KOKORO_GITHUB_PULL_REQUEST_NUMBER:-${SAMPLE_VERSION:0:12}}
```

Which is not sufficiently different between parallel builds using the same PR/commit. Can we update the function name to use a uuid instead? Otherwise, it's likely that a retry here will just re-deploy the function and create a race condition between different builds.

kurtisvg (changes requested):

kurtisvg (approved):

PR #7509, opened 2022-12-15
docs(webrisk-samples): add new samples (search, threat diff) and refa…

Sita04:
…ctor existing

1. Adding two new samples:
    Search Hashes
    Compute Threat List diff

2. Refactor existing samples and updated tests to match Java standard guide.


webrisk/src/test/java/webrisk/SnippetsIT.java
```diff
@@ -0,0 +1,138 @@
+/*
+ * Copyright 2020 Google LLC
```
lesv, line 2: Is this new code?  Then it should be 2022

webrisk/src/main/java/webrisk/SubmitUri.java
```diff
@@ -0,0 +1,72 @@
+/*
+ * Copyright 2020 Google LLC
```
lesv, line 2: This appears to be new code, so it should be 2022

webrisk/src/main/java/webrisk/SearchUri.java
```diff
@@ -0,0 +1,68 @@
+/*
+ * Copyright 2020 Google LLC
```
lesv, line 2: 2022?

webrisk/src/test/java/webrisk/SnippetsIT.java
```diff
@@ -0,0 +1,138 @@
...
+    assertWithMessage(String.format("Missing environment variable '%s' ", envVarName))
+        .that(System.getenv(envVarName))
+        .isNotEmpty();
+  }
+
+  @BeforeClass
+  public static void setUp() throws IOException {
+    final PrintStream out = System.out;
+    ByteArrayOutputStream stdOut = new ByteArrayOutputStream();
+    System.setOut(new PrintStream(stdOut));
```
lesv, line 61: This seems wrong here.

webrisk/src/test/java/webrisk/SnippetsIT.java
```diff
@@ -0,0 +1,138 @@
...
+  }
+
+  @AfterClass
+  public static void cleanup() throws IOException {
+    final PrintStream out = System.out;
+    ByteArrayOutputStream stdOut = new ByteArrayOutputStream();
+    System.setOut(new PrintStream(stdOut));
+
+    stdOut.close();
+    System.setOut(out);
```
lesv, line 77: Seems a bit pointless.

webrisk/src/test/java/webrisk/SnippetsIT.java
```diff
@@ -0,0 +1,138 @@
...
+  @Before
+  public void beforeEach() {
+    stdOut = new ByteArrayOutputStream();
+    System.setOut(new PrintStream(stdOut));
+  }
+
+  @After
+  public void afterEach() {
+    stdOut = null;
+    System.setOut(null);
```
lesv, line 89: Should always be reset to `out` -- Never to `null`.

lesv (changes requested):
PTAL

webrisk/src/test/java/webrisk/SnippetsIT.java
Sita04: Not new code. I just renamed the file and made some updates.

webrisk/src/main/java/webrisk/SearchUri.java
Sita04: Ditto. Not new code.

webrisk/src/test/java/webrisk/SnippetsIT.java
Sita04: I removed this code as it was redundant.
But I don't follow your comment here.

Sita04:
@lesv PTAL at the changes. Thanks.

webrisk/src/test/java/webrisk/SnippetsIT.java
```diff
@@ -0,0 +1,122 @@
...
+
+  @Before
+  public void beforeEach() {
+    stdOut = new ByteArrayOutputStream();
+    System.setOut(new PrintStream(stdOut));
+  }
+
+  @After
+  public void afterEach() {
+    stdOut = null;
```
lesv, line 71: This ideally should be restored `out`, not `null`.

lesv: setting to `null` makes debugging harder.

lesv: This may also mess with later test files.

lesv (changes requested):
It's bad to set `stdOut` to `null`.  You should always restore to a saved version of `System.out`

webrisk/src/test/java/webrisk/SnippetsIT.java
Sita04: Done. I've modified the code. PTAL, thanks :)

webrisk/src/test/java/webrisk/SnippetsIT.java
```diff
@@ -0,0 +1,126 @@
...
+import org.junit.BeforeClass;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.junit.runners.JUnit4;
+
+@RunWith(JUnit4.class)
+public class SnippetsIT {
+
+  private static final String PROJECT_ID = System.getenv("GOOGLE_CLOUD_PROJECT");
+  private static ByteArrayOutputStream originalOut = new ByteArrayOutputStream();
```
lesv, line 48: You probably want `System.out` here.

lesv (changes requested):

lesv (changes requested):
PTAL

lesv (approved):

PR #7580, opened 2023-01-17
docs(recaptcha-samples): update account defender samples

Sita04:
Update Create account defender assessment sample to use HMAC SHA 256.
Update SearchRelatedAccountGroupMemberships to reflect API changes. (setParent to setProject)

### DIFF
--- a/recaptcha_enterprise/snippets/src/main/java/recaptcha/account_defender/AccountDefenderAssessment.java
+++ b/recaptcha_enterprise/snippets/src/main/java/recaptcha/account_defender/AccountDefenderAssessment.java
@@ -60,14 +60,13 @@ public class AccountDefenderAssessment {
     String userIdentifier = "default" + UUID.randomUUID().toString().split("-")[0];
 
     // Change this to a secret not shared with Google.
-    String HMAC_KEY = "SOME_INTERNAL_UNSHARED_KEY";
+    final String HMAC_KEY = "SOME_INTERNAL_UNSHARED_KEY";
 
     // Get instance of Mac object implementing HmacSHA256, and initialize it with the above
     // secret key.
     Mac mac = Mac.getInstance("HmacSHA256");
-    SecretKeySpec secretKeySpec = new SecretKeySpec(HMAC_KEY.getBytes(StandardCharsets.UTF_8),
-        "HmacSHA256");
-    mac.init(secretKeySpec);
+    mac.init(new SecretKeySpec(HMAC_KEY.getBytes(StandardCharsets.UTF_8),
+        "HmacSHA256"));
     byte[] hashBytes = mac.doFinal(userIdentifier.getBytes(StandardCharsets.UTF_8));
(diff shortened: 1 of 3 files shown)


recaptcha_enterprise/snippets/src/main/java/recaptcha/account_defender/AccountDefenderAssessment.java
```diff
@@ -56,9 +59,16 @@ public static void main(String[] args) throws IOException, NoSuchAlgorithmExcept
     // Unique ID of the customer, such as email, customer ID, etc.
     String userIdentifier = "default" + UUID.randomUUID().toString().split("-")[0];
 
-    // Hash the unique customer ID using HMAC SHA-256.
-    MessageDigest digest = MessageDigest.getInstance("SHA-256");
-    byte[] hashBytes = digest.digest(userIdentifier.getBytes(StandardCharsets.UTF_8));
+    // Change this to a secret not shared with Google.
+    String HMAC_KEY = "SOME_INTERNAL_UNSHARED_KEY";
```
lesv, line 63: Might this be better near the top so that users can spot it?  Or is it better where it is so it will be in the snippet?

lesv: `static final` ?

recaptcha_enterprise/snippets/src/main/java/recaptcha/account_defender/AccountDefenderAssessment.java
```diff
@@ -56,9 +59,16 @@ public static void main(String[] args) throws IOException, NoSuchAlgorithmExcept
...
+
+    // Get instance of Mac object implementing HmacSHA256, and initialize it with the above
+    // secret key.
+    Mac mac = Mac.getInstance("HmacSHA256");
+    SecretKeySpec secretKeySpec = new SecretKeySpec(HMAC_KEY.getBytes(StandardCharsets.UTF_8),
+        "HmacSHA256");
```
lesv, line 69: Got to wonder if you shouldn't define a constant of this, probably not, but, I wonder.

lesv (approved):
Consider the questions, otherwise LGTM

recaptcha_enterprise/snippets/src/main/java/recaptcha/account_defender/AccountDefenderAssessment.java
Sita04: Leaning towards leaving it there at the snippet as it will be cohesive.
Making it final.

recaptcha_enterprise/snippets/src/main/java/recaptcha/account_defender/AccountDefenderAssessment.java
Sita04: Thanks for the catch. Better to inline with the mac.init so as to not risk exposing during debugging or logging.

PR #7581, opened 2023-01-17
feat(speech): adds adaptation samples for V2

telpirion:
This PR (2 of 3) adds the Speech V2 samples for Java.

Canonical samples are [here](https://github.com/googleapis/python-speech/pull/460).


anguillanneuf (changes requested):
The samples look to be missing region tags.

Tests need to be suffixed with IT.java.

If you want to reduce boilerplate code, this is how Pub/Sub does it, but it's up to you -> https://github.com/googleapis/java-pubsub/blob/main/samples/snippets/src/test/java/pubsub/PublisherIT.java

anguillanneuf (approved):

PR #7633, opened 2023-01-24
Change default port from 8000 to 8080.

Cory-Kramer:
This is to be consistent with the Python and Node.js reCAPTCHA demos.


lesv (dismissed):

Cory-Kramer:
I am indeed a Googler :)
I think I must have both my corp and personal emails associated with my account

lesv (approved):

PR #7654, opened 2023-02-03
fix(healthcare-api): clean up de-identified dataset after tests

noerog:
Fixes de-identified datasets created during tests not being deleted.

### DIFF
--- a/healthcare/v1/src/test/java/snippets/healthcare/DatasetTests.java
+++ b/healthcare/v1/src/test/java/snippets/healthcare/DatasetTests.java
@@ -67,10 +67,17 @@ public class DatasetTests {
 
   @AfterClass
   public static void deleteTempItems() throws IOException {
+    // Delete the destination dataset created during de-identification.
     try {
       DatasetDelete.datasetDelete(destinationDatasetName);
     } catch (GoogleJsonResponseException ex) {
-      // Destination dataset already deleted, continue.
+      if (ex.getStatusCode() == 404) {
+        System.out.println("Destination dataset already deleted, continue.");
+      } else {
+        System.out.println(
+            "Error during deleteTempItems while deleting de-identified destination dataset: \n"
+                + ex.toString());
+      }
     }
   }
 
@@ -94,7 +101,11 @@ public class DatasetTests {
     try {
       DatasetDelete.datasetDelete(datasetName);
     } catch (GoogleJsonResponseException ex) {
-      // Dataset already deleted, continue.
+      if (ex.getStatusCode() == 404) {
+        System.out.println("Dataset already deleted, continue.");
+      } else {
+        System.out.println("Error during tearDown while deleting dataset: \n" + ex.toString());
+      }
     }
     bout.reset();
   }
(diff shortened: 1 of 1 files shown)


healthcare/v1/src/test/java/snippets/healthcare/DatasetTests.java
```diff
@@ -63,6 +65,15 @@ public static void checkRequirements() {
     requireEnvVar("GOOGLE_CLOUD_PROJECT");
   }
 
+  @AfterClass
+  public static void deleteTempItems() throws IOException {
+    try {
+      DatasetDelete.datasetDelete(destinationDatasetName);
+    } catch (GoogleJsonResponseException ex) {
+      // Destination dataset already deleted, continue.
```
Shabirmean, line 73: Is this the only case when this exception is thrown? Are there other times this can happen? `GoogleJsonResponseException` seems too generic. Should we check for more specific details to ensure what was the reason for the execption?

noerog: I think we just need to check for non-2xx responses, which GoogleJsonResponseException does. Any non-2xx response will cause DatasetTests.deleteTempItems to fail.

Shabirmean: Thank you @noerog . But I am trying to understand if this exception can be thrown for cases where we should not just ignore the exception. 

I am not convinced that the `GoogleJsonResponseException` is thrown only when the said dataset has already been deleted. Correct me if I am wrong. Non-2xx can mean different reasons for the failure. I feel that if someone else updates this test and for some reason a failure is introduced for a reason other than the resource being already deleted, this portion of the test will silently fail always.

noerog: You're right, it's not only thrown when the dataset has already been deleted, it's thrown whenever there's a non-2xx error code. Do you want to individually check for every possible error that can be returned? If it fails because of something like a quota issue, it'll still fail the test and return an error message stating why it failed. Would changing the comment help?

noerog:
Hi @Shabirmean, sorry for the delay. PTAL and let me know if I've addressed your comments. Thanks.

healthcare/v1/src/test/java/snippets/healthcare/DatasetTests.java
```diff
@@ -63,6 +65,22 @@ public static void checkRequirements() {
...
+  @AfterClass
+  public static void deleteTempItems() throws IOException {
+    // Delete the destination dataset created during de-identification.
+    try {
+      DatasetDelete.datasetDelete(destinationDatasetName);
+    } catch (GoogleJsonResponseException ex) {
+      if (ex.getStatusCode() == 404) {
+        System.out.println("Destination dataset already deleted, continue.");
+      } else {
+        System.out.println(
```
Shabirmean, line 77: If it was some other error,  we must surface that error. So I think we should throw the exception in the else case. My suggestion is to have something as below to have the tests clean and simple.

```java
} catch (GoogleJsonResponseException ex) {
     // if 404, then the dataset has already been deleted
      if (ex.getStatusCode() != 404) {
           throw ex;
      }
}
```

healthcare/v1/src/test/java/snippets/healthcare/DatasetTests.java
```diff
@@ -84,16 +101,20 @@ public void tearDown() throws IOException {
     try {
       DatasetDelete.datasetDelete(datasetName);
     } catch (GoogleJsonResponseException ex) {
-      // Dataset already deleted, continue.
+      if (ex.getStatusCode() == 404) {
```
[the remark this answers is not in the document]
noerog: Done.

PR #7682, opened 2023-02-10
test: temporarily ignore failing featurestore tests

kweinmeister:
Ignores failing FeatureStore tests which are blocking other PRs.

The tests should be enabled with #7642 and #7651.


averikitsch:
Do we have an ETA on the fix? In the past ignoring tests has not been a good solution.

averikitsch (approved):

PR #7691, opened 2023-02-15
docs(samples): Add Dataflow TextIO snippet

VeronicaWasson:
Adds a snippet for a Dataflow doc in progress

Also adds a snippets dataflow/snippets subdirectory with POM file

### DIFF
--- a/dataflow/snippets/pom.xml
+++ b/dataflow/snippets/pom.xml
@@ -1,0 +1,131 @@
+<!--
+ Copyright 2023 Google LLC.
+
+ Licensed under the Apache License, Version 2.0 (the "License");
+ you may not use this file except in compliance with the License.
+ You may obtain a copy of the License at
+
+       http://www.apache.org/licenses/LICENSE-2.0
+
+ Unless required by applicable law or agreed to in writing, software
+ distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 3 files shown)


dataflow/snippets/pom.xml
```diff
@@ -0,0 +1,131 @@
...
+    <dependency>
+      <groupId>org.apache.beam</groupId>
+      <artifactId>beam-sdks-java-core</artifactId>
+      <version>${apache_beam.version}</version>
+    </dependency>
+
+    <dependency>
+      <groupId>org.apache.beam</groupId>
+      <artifactId>beam-sdks-java-io-google-cloud-platform</artifactId>
+      <version>${apache_beam.version}</version>
```
lesv, line 87: You might wish to verify that BEAM isn't in the Libraries-BOM.  If it is, you don't need the `<version>` line.

lesv (approved):

dataflow/snippets/pom.xml
VeronicaWasson: thanks! If I remove the Beam version, I get a `dependencies.dependency.version` missing error from mvn.

lesv:
SGTM - please visit the internal site `github/` and register.  Join Googlers, GoogleCloudPlatform, and GoogleAPIs

PR #7697, opened 2023-02-16
docs(ca-samples): delete stale ca pool in test project

Sita04:
Adding util to delete stale resource in JDST project.


rsamborski (approved):
We need a similar fix for python as we hit quota there as well.

Sita04:
> We need a similar fix for python as we hit quota there as well.

Yes, working on it. Will send it for review once done. Thanks

averikitsch (approved):

PR #7707, opened 2023-02-22
docs(compute-samples): adding machine type update and regional disk samples

Sita04:
Adding samples needed for [this page](https://cloud.google.com/compute/docs/disks/regional-persistent-disk).

### DIFF
--- a/compute/cloud-client/src/main/java/compute/ChangeInstanceMachineType.java
+++ b/compute/cloud-client/src/main/java/compute/ChangeInstanceMachineType.java
@@ -1,0 +1,84 @@
+/*
+ * Copyright 2023 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package compute;
+
+// [START compute_change_machine_type]
+
+import com.google.cloud.compute.v1.Instance;
+import com.google.cloud.compute.v1.Instance.Status;
+import com.google.cloud.compute.v1.InstancesClient;
+import com.google.cloud.compute.v1.InstancesSetMachineTypeRequest;
+import com.google.cloud.compute.v1.Operation;
+import java.io.IOException;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
+
+public class ChangeInstanceMachineType {
+
+  public static void main(String[] args)
+      throws IOException, ExecutionException, InterruptedException, TimeoutException {
+    // TODO(developer): Replace these variables before running the sample.
+    // Project ID or project number of the Cloud project you want to use.
+    String projectId = "your-project-id";
+    // Name of the zone your instance belongs to.
+    String zone = "zone-name";
+    // Name of the VM you want to modify.
+    String instanceName = "instance-name";
+    // The new machine type you want to use for the VM.
+    // For example: "e2-standard-8", "e2-custom-4-2048" or "m1-ultramem-40"
+    // More about machine types: https://cloud.google.com/compute/docs/machine-resource
+    String newMachineType = "e2-standard-8";
+    changeMachineType(projectId, zone, instanceName, newMachineType);
+  }
+
+  // Changes the machine type of VM.
+  // The VM needs to be in the 'TERMINATED' state for this operation to be successful.
+  public static void changeMachineType(String projectId, String zone, String instanceName,
+      String newMachineType)
+      throws IOException, ExecutionException, InterruptedException, TimeoutException {
+    // Initialize client that will be used to send requests. This client only needs to be created
+    // once, and can be reused for multiple requests. After completing all of your requests, call
+    // the `instancesClient.close()` method on the client to safely
+    // clean up any remaining background resources.
+    try (InstancesClient instancesClient = InstancesClient.create()) {
+
+      Instance instance = instancesClient.get(projectId, zone, instanceName);
+      if (!instance.getStatus().equals(Status.TERMINATED.name())) {
+        throw new Error(String.format(
+            "Only machines in TERMINATED state can have their machine type changed. %s is in %s state.",
+            instance.getName(), instance.getStatus()));
+      }
+
+      InstancesSetMachineTypeRequest machineTypeRequest = InstancesSetMachineTypeRequest.newBuilder()
+          .setMachineType(String.format("projects/%s/zones/%s/machineTypes/%s",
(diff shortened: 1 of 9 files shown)


compute/cloud-client/src/main/java/compute/disks/AttachDisk.java
```diff
@@ -0,0 +1,97 @@
...
+    // Project ID or project number of the Cloud project you want to use.
+    String projectId = "your-project-id";
+
+    // Name of the zone in which the instance you want to use resides.
+    String zone = "zone-name";
+
+    // Name of the compute instance you want to attach a disk to.
+    String instanceName = "instance-name";
+
+    // Full or partial URL to a persistent disk that you want to attach. This can be either
```
savijatv, line 45: Full or partial URL for a persistent disk...
'for' or 'of' a persistent disk sounds better. WDYT?

compute/cloud-client/src/main/java/compute/disks/AttachDisk.java
```diff
@@ -0,0 +1,97 @@
...
+    // regional or zonal disk.
```
savijatv, line 46: either be a regional or zonal disk.

compute/cloud-client/src/main/java/compute/disks/AttachDisk.java
```diff
@@ -0,0 +1,97 @@
...
+    // Expected formats:
```
savijatv, line 47: Valid formats?

compute/cloud-client/src/main/java/compute/disks/ResizeDisk.java
```diff
@@ -0,0 +1,91 @@
...
+
+public class ResizeDisk {
+
+  public static void main(String[] args)
+      throws IOException, ExecutionException, InterruptedException, TimeoutException {
+    // TODO(developer): Replace these variables before running the sample.
+    // Project ID or project number of the Cloud project you want to use.
+    String projectId = "your-project-id";
+
+    // A link to the disk you want to resize.
```
savijatv, line 40: Full or partial URL for the disk that you want to resize.

compute/cloud-client/src/main/java/compute/disks/ResizeRegionalDisk.java
```diff
@@ -0,0 +1,90 @@
...
+
+public class ResizeRegionalDisk {
+
+  public static void main(String[] args)
+      throws IOException, ExecutionException, InterruptedException, TimeoutException {
+    // TODO(developer): Replace these variables before running the sample.
+    // Project ID or project number of the Cloud project you want to use.
+    String projectId = "your-project-id";
+
+    // A link to the disk you want to resize.
```
savijatv, line 40: Full or partial URL for the disk that you...

compute/cloud-client/src/main/java/compute/disks/ResizeRegionalDisk.java
```diff
@@ -0,0 +1,90 @@
...
+    // This value uses the following format:
+    //     * projects/{project_name}/regions/{region}/disks/{disk_name}
```
savijatv, line 42: Just checking if this is the only format. Does it not accept the full URL?

compute/cloud-client/src/main/java/compute/disks/AttachDisk.java
Sita04: Modifying it to "of". Thank you.

compute/cloud-client/src/main/java/compute/disks/ResizeRegionalDisk.java
Sita04: The input can be in any format as we have the pattern matcher to extract the relevant portion. The format listed here is irrelevant to the API format and hence I prefer to not mention it here.

compute/cloud-client/src/main/java/compute/ChangeInstanceMachineType.java
```diff
@@ -0,0 +1,85 @@
...
+
+      InstancesSetMachineTypeRequest machineTypeRequest =
+          InstancesSetMachineTypeRequest.newBuilder()
+              .setMachineType(String.format("projects/%s/zones/%s/machineTypes/%s",
+                  projectId, zone, newMachineType))
+              .build();
+
+      Operation response = instancesClient
+          .setMachineTypeAsync(projectId, zone, instanceName, machineTypeRequest)
+          .get(3, TimeUnit.MINUTES);
```
averikitsch, line 75: Any reason to set 3 minutes for the timeout?

compute/cloud-client/src/main/java/compute/ChangeInstanceMachineType.java
```diff
@@ -0,0 +1,85 @@
...
+
+      if (response.hasError()) {
+        System.out.println("Machine type update failed! " + response);
```
averikitsch, line 78: What types of errors do we expect that wouldn't be thrown during the operation? Can we show users how to handle them?

compute/cloud-client/src/main/java/compute/disks/ResizeDisk.java
```diff
@@ -0,0 +1,90 @@
...
+      while (matcher.find()) {
+        match = matcher.group().split("/");
+      }
+
+      ResizeDiskRequest resizeDiskRequest = ResizeDiskRequest.newBuilder()
+          .setZone(match[3])
+          .setDisksResizeRequestResource(DisksResizeRequest.newBuilder()
+              .setSizeGb(newSizeGb)
+              .build())
+          .setDisk(match[5])
```
averikitsch, line 75: This doesn't take the full disk resource path? This seems unnecessarily hard. Is there another way to achieve this? If not You should add more comments to why you are using the matcher.

averikitsch (changes requested):

compute/cloud-client/src/main/java/compute/ChangeInstanceMachineType.java
Sita04: ditto, Followed the same pattern from the other compute samples. 
Python Reference: https://github.com/GoogleCloudPlatform/python-docs-samples/pull/8992/files#diff-9dcee92af6eb54e1cd2c96bbdc292fc4857452f209215aa0204e461c7ee49d9bR62-R69

But AFAIK, I haven't encountered any errors that were not raised as exceptions.

compute/cloud-client/src/main/java/compute/ChangeInstanceMachineType.java
Sita04: Followed the same pattern from the other compute samples. No specific reason.

compute/cloud-client/src/main/java/compute/disks/ResizeDisk.java
Sita04: Yes, I believe this is due to multiple formats being accepted. See: https://github.com/GoogleCloudPlatform/python-docs-samples/pull/8992/files#diff-00f22f3a9fa3fb85b7b4340eaeb0aefece4b60556f7efd7e7bf58d684b496f2bR88-R91

I'll double check this and update the sample. Thanks for calling this out.

Sita04: Updated the code to simplify. This will result in code discrepancy between the Java and Python versions. But using matcher is an over-kill for this situation and hence added additional params.

averikitsch (approved):

PR #7716, opened 2023-03-01
fix: add MultipleAttemptsRule for Batch template creation

Sita04:
Fixes #7696

### DIFF
--- a/batch/snippets/src/test/java/BatchTemplateIT.java
+++ b/batch/snippets/src/test/java/BatchTemplateIT.java
@@ -61,7 +61,7 @@ public class BatchTemplateIT {
   private static InstanceTemplate INSTANCE_TEMPLATE;
 
   @Rule
-  private final MultipleAttemptsRule multipleAttemptsRule = new MultipleAttemptsRule(
+  public final MultipleAttemptsRule multipleAttemptsRule = new MultipleAttemptsRule(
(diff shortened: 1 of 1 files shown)


batch/snippets/src/test/java/BatchTemplateIT.java
```diff
@@ -52,10 +54,16 @@ public class BatchTemplateIT {
 
   private static final String PROJECT_ID = System.getenv("GOOGLE_CLOUD_PROJECT");
   private static final String REGION = "europe-north1";
+  private static final int MAX_ATTEMPT_COUNT = 3;
+  private static final int INITIAL_BACKOFF_MILLIS = 120000; // 2 minutes
```
rsamborski, line 58: Why do we choose 2 minutes? Would it still work with a lower initial backoff, for example 30 or 60s?

rsamborski (approved):
Approved with minor suggestion.

batch/snippets/src/test/java/BatchTemplateIT.java
realjordanna: it's timing out https://source.cloud.google.com/results/invocations/37167479-1507-417f-88b2-f8e458c387b3/targets/github%2Fjava-docs-samples%2Fbatch%2Fsnippets%2Ftarget%2Ffailsafe-reports/tests

PR #7723, opened 2023-03-03
chore: recaptcha - update snippet lint errors and restructure pom

Sita04:
Updates lint errors in snippets.
Moved pom location to snippets/


rsamborski (approved):

recaptcha_enterprise/snippets/src/test/java/app/SnippetsIT.java
```diff
@@ -63,6 +58,12 @@
 import org.springframework.web.util.UriComponentsBuilder;
 import recaptcha.AnnotateAssessment;
 import recaptcha.GetMetrics;
+import recaptcha.accountdefender.AccountDefenderAssessment;
```
averikitsch, line 61: You might not have to do this if you use the path `test/java/recaptcha` and package `recaptcha`.

averikitsch (commented):
There is still a pom.xml under recaptcha_enterprise that isn't needed. Make sure this doesn't break the docs, see snippetbot issue.

averikitsch (changes requested):
Please confirm that the tests are actually running.

PR #7725, opened 2023-03-06
chore: Adding self-service DLP team to labels and blunderbuss

rsamborski:
Adding self-service DLP team to labels and blunderbuss

### DIFF
--- a/.github/auto-label.yaml
+++ b/.github/auto-label.yaml
@@ -69,3 +69,4 @@ path:
     vision: "vision"
     webrisk: "webrisk"
     workflows: "workflows"
+    dlp: "dlp"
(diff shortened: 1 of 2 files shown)


Sita04 (approved):
Can we add a comment next to DLP marking it as self-service?

lesv (approved):

rsamborski:
@Sita04 There is a comment about self-service in CODEOWNERS file and I don't see other comments in these yaml files, so not sure how they are treated. As the PR was merged already - let me know if you'd like to proceed with another one to add comments.

Sita04:
No, thanks. As long as we have it captured in CODEOWNERS, it's good :)

PR #7816, opened 2023-03-29
fix: Batch issue fix

Sita04:
Unblocks PR: https://github.com/GoogleCloudPlatform/java-docs-samples/pull/7752

Fixes timeout issue with "MultipleAttemptsRule".
Fixes Job exists issue by moving the Job creation to setup, so that the job is created only once when the MultipleAttemptsRule runs.

### DIFF
--- a/batch/snippets/src/test/java/BatchBucketIT.java
+++ b/batch/snippets/src/test/java/BatchBucketIT.java
@@ -135,7 +135,7 @@ public class BatchBucketIT {
     String fileNameTemplate = "output_task_%s.txt";
     String fileContentTemplate;
 
-    Storage storage = StorageOptions.getDefaultInstance().getService();
+    Storage storage = StorageOptions.newBuilder().setProjectId(PROJECT_ID).build().getService();
     Bucket bucket = storage.get(BUCKET_NAME);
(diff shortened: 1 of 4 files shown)


batch/snippets/src/test/java/BatchBucketIT.java
```diff
@@ -131,11 +131,12 @@ public void testBucketJob()
     testBucketContent();
   }
 
+  @Test
   public void testBucketContent() {
     String fileNameTemplate = "output_task_%s.txt";
     String fileContentTemplate;
 
-    Storage storage = StorageOptions.getDefaultInstance().getService();
+    Storage storage = StorageOptions.newBuilder().setProjectId(PROJECT_ID).build().getService();
```
averikitsch, line 139: Is there a reason why project ID is not recognized automatically?

averikitsch (approved):

batch/snippets/src/test/java/BatchBucketIT.java
Sita04: IIUC, this is the equivalent of passing projectId in a client request, right?
Please lmk if the assumption is wrong and I can verify it.

PR #7817, opened 2023-03-29
docs(samples): Add Dataflow BigQueryIO read snippets

VeronicaWasson:
Adds BigQuery I/O snippets for a Dataflow doc in progress

### DIFF
--- a/dataflow/snippets/src/main/java/com/example/dataflow/BigQueryReadAvro.java
+++ b/dataflow/snippets/src/main/java/com/example/dataflow/BigQueryReadAvro.java
@@ -53,9 +53,7 @@ public class BigQueryReadAvro {
 
   public static void main(String[] args) {
     // Parse the pipeline options passed into the application. Example:
-    //   --projectId=$PROJECT_ID
-    //   --datasetName=$DATASET_NAME
-    //   --tableName=$TABLE_NAME
+    //   --projectId=$PROJECT_ID --datasetName=$DATASET_NAME --tableName=$TABLE_NAME
     // For more information, see https://beam.apache.org/documentation/programming-guide/#configuring-pipeline-options
     PipelineOptionsFactory.register(BigQueryReadOptions.class);
     BigQueryReadOptions options = PipelineOptionsFactory.fromArgs(args)
--- a/dataflow/snippets/src/main/java/com/example/dataflow/BigQueryReadAvro.java
+++ b/dataflow/snippets/src/main/java/com/example/dataflow/BigQueryReadAvro.java
@@ -17,6 +17,7 @@
 package com.example.dataflow;
 
 // [START dataflow_bigquery_read_avro]
+import org.apache.avro.generic.GenericRecord;
 import org.apache.avro.util.Utf8;
 import org.apache.beam.sdk.Pipeline;
 import org.apache.beam.sdk.coders.AvroCoder;
@@ -41,8 +42,8 @@ public class BigQueryReadAvro {
     public static class FromSchemaAndRecord
             implements SerializableFunction<SchemaAndRecord, MyData> {
       @Override public MyData apply(SchemaAndRecord elem) {
-        var data = new MyData();
-        var record = elem.getRecord();
+        MyData data = new MyData();
+        GenericRecord record = elem.getRecord();
         data.name = ((Utf8) record.get("user_name")).toString();
         data.age = (Long) record.get("age");
         return data;
@@ -51,12 +52,13 @@ public class BigQueryReadAvro {
   }
 
   public static void main(String[] args) {
+    // Parse the pipeline options passed into the application.
     PipelineOptionsFactory.register(BigQueryReadOptions.class);
-    var options = PipelineOptionsFactory.fromArgs(args)
+    BigQueryReadOptions options = PipelineOptionsFactory.fromArgs(args)
         .withValidation()
         .as(BigQueryReadOptions.class);
 
-    var pipeline = Pipeline.create(options);
+    Pipeline pipeline = Pipeline.create(options);
     pipeline
         // Read table data into Avro records, using an application-defined parsing function.
         .apply(BigQueryIO.read(new MyData.FromSchemaAndRecord())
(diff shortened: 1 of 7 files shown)


VeronicaWasson:
@reuvenlax Could you take a look at these?

dataflow/snippets/src/main/java/com/example/dataflow/BigQueryReadAvro.java
```diff
@@ -0,0 +1,78 @@
...
+
+    // Function to convert Avro records to MyData instances.
+    public static class FromSchemaAndRecord
+            implements SerializableFunction<SchemaAndRecord, MyData> {
+      @Override public MyData apply(SchemaAndRecord elem) {
+        var data = new MyData();
+        var record = elem.getRecord();
+        data.name = ((Utf8) record.get("user_name")).toString();
+        data.age = (Long) record.get("age");
+        return data;
```
Shabirmean, line 48: @averikitsch  Do we have a precedence for allowing the use of `var` in samples? 

Since they are samples, I think we used the full type so readers can easily understand the returns and what each line is doing.

dataflow/snippets/src/main/java/com/example/dataflow/BigQueryReadOptions.java
```diff
@@ -0,0 +1,37 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.example.dataflow;
+
+import org.apache.beam.sdk.options.Description;
+import org.apache.beam.sdk.options.PipelineOptions;
+
+public interface BigQueryReadOptions extends PipelineOptions {
```
Shabirmean, line 22: Can we add a java doc comment for this interface?

dataflow/snippets/src/main/java/com/example/dataflow/BigQueryReadWithProjectionAndFiltering.java
```diff
@@ -0,0 +1,61 @@
...
+import org.apache.beam.sdk.io.gcp.bigquery.BigQueryIO;
+import org.apache.beam.sdk.io.gcp.bigquery.BigQueryIO.TypedRead;
+import org.apache.beam.sdk.options.PipelineOptionsFactory;
+import org.apache.beam.sdk.transforms.MapElements;
+import org.apache.beam.sdk.values.TypeDescriptor;
+
+public class BigQueryReadWithProjectionAndFiltering {
+  public static void main(String[] args) {
+    PipelineOptionsFactory.register(BigQueryReadOptions.class);
+    var options = PipelineOptionsFactory.fromArgs(args)
```
Shabirmean, line 32: What are some examples for the `args` here? Can we add some comments to explain that?

dataflow/snippets/src/main/java/com/example/dataflow/BigQueryReadAvro.java
VeronicaWasson: Updated to replace 'var' throughout

dataflow/snippets/src/main/java/com/example/dataflow/BigQueryReadOptions.java
VeronicaWasson: done

dataflow/snippets/src/main/java/com/example/dataflow/BigQueryReadWithProjectionAndFiltering.java
VeronicaWasson: I added a comment. It's a bit vague because pipeline options are open-ended. However this is a basic Apache Beam pattern so I think it should be OK. (Dataflow users will see this pattern pretty much every where in Beam code.)

dataflow/snippets/src/main/java/com/example/dataflow/BigQueryReadAvro.java
```diff
@@ -0,0 +1,78 @@
...
+      }
+    }
+  }
+
+  public static void main(String[] args) {
+    PipelineOptionsFactory.register(BigQueryReadOptions.class);
+    var options = PipelineOptionsFactory.fromArgs(args)
```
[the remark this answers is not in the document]
VeronicaWasson: Added - see my other comment.

VeronicaWasson:
@Shabirmean  thanks for reviewing!

dataflow/snippets/src/main/java/com/example/dataflow/BigQueryReadWithProjectionAndFiltering.java
Shabirmean: I think we should still provide some context or starter guidance for developers. Atleast point to a doc which a Beam example. Or maybe an example as you have in your test:

```
new String[] {
            "--runner=DirectRunner",
            "--projectId=" + projectId,
            "--datasetName=" + datasetName,
            "--tableName=" + tableName
        });
```
WDYT?

Shabirmean (commented):
Thank you for addressing the comments. I left one more followup!

dataflow/snippets/src/main/java/com/example/dataflow/BigQueryReadWithProjectionAndFiltering.java
VeronicaWasson: I like that suggestion, thanks.

Usually the arguments get passed indirectly -- you wouldn't directly invoke Main like in the test. But I added comments showing what they might look like, plus a link to a Beam doc.

Shabirmean: Thank you very much for adding more details and I agree, usually it will be passed into the `java ....` call. What you have now looks good. 

One last suggestion before we merge it 🙏🏼 : can you flatten our the arguments like how it will be passed into an actual call for the java program execution:
```
--projectId=$PROJECT_ID --datasetName=$DATASET_NAME --tableName=$TABLE_NAME
```

VeronicaWasson: done!

Shabirmean (approved):
LGTM! Thank you for addressing all the comments!

PR #7827, opened 2023-04-03
docs(samples): Created Initial Sample for Generative AI app builder - Search

holtskinner:
- The Updated Client Library has been published.
- The Test Project has the API/Search Engine configured.
- Snippet Prefix exposed in tool.


genappbuilder/src/main/java/genappbuilder/v1beta/Search.java
```diff
@@ -0,0 +1,61 @@
...
+      SearchRequest request =
+          SearchRequest.newBuilder()
+              .setServingConfig(
+                      ServingConfigName.ofProjectLocationDataStoreServingConfigName(
+                          projectId, location, searchEngineId, servingConfigId)
+                      .toString())
+              .setQuery(searchQuery)
+              .setPageSize(10)
+              .build();
+      SearchResponse response = searchClient.search(request).getPage().getResponse();
```
Shabirmean, line 54: This must be `com.google.api.grpc.search(...`

genappbuilder/src/main/java/genappbuilder/v1beta/Search.java
```diff
@@ -0,0 +1,61 @@
...
+      for (SearchResponse.SearchResult element : response.getResultList()) {
```
Shabirmean, line 55: This must be `getResultsList()`

genappbuilder/pom.xml
```diff
@@ -0,0 +1,59 @@
...
+        <type>pom</type>
+        <scope>import</scope>
+      </dependency>
+    </dependencies>
+  </dependencyManagement>
+
+  <dependencies>
+    <dependency>
+      <groupId>com.google.cloud</groupId>
+      <artifactId>google-cloud-discoveryengine</artifactId>
```
Shabirmean, line 43: The latest libraries bom only imports: `0.11.0` of the discoveryengine client. So I suggest you add the version explicitly here and add a comment to remove it when the `libraries-bom` has updated the version.

`<version>0.12.0</version>`

genappbuilder/src/main/java/genappbuilder/v1beta/Search.java
```diff
@@ -0,0 +1,61 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package genappbuilder.v1beta;
+
+// [START genappbuilder_search]
+
+import com.google.cloud.discoveryengine.v1beta.SearchRequest;
+import com.google.cloud.discoveryengine.v1beta.SearchResponse;
```
Shabirmean, line 22: The `SearchRequest` and `SearchRequest` classes are defined in the `proto-google-cloud-discoveryengine-v1beta` package. Thus, you have to add that in the pom.xml after the `gproto-google-cloud-discoveryengine-v1beta`

```
<dependency>
      <groupId>com.google.api.grpc</groupId>
      <artifactId>proto-google-cloud-discoveryengine-v1beta</artifactId>
      <version>0.12.0</version>
    </dependency>
```

genappbuilder/pom.xml
```diff
@@ -0,0 +1,59 @@
+<?xml version='1.0' encoding='UTF-8'?>
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+  <groupId>com.example.genappbuilder</groupId>
+  <artifactId>genappbuilder-snippets</artifactId>
+  <packaging>jar</packaging>
+  <name>Generative AI App Builder Snippets</name>
+  <url>https://github.com/GoogleCloudPlatform/java-docs-samples/tree/main/genappbuilder</url>
```
Shabirmean, line 8: Update here as well to be `discoveryengine`

genappbuilder/pom.xml
```diff
@@ -0,0 +1,59 @@
+<?xml version='1.0' encoding='UTF-8'?>
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+  <groupId>com.example.genappbuilder</groupId>
```
Shabirmean, line 4: The id should use the API/product name. So I suggest you use `com.example. discoveryengine `

genappbuilder/pom.xml
```diff
@@ -0,0 +1,59 @@
+<?xml version='1.0' encoding='UTF-8'?>
```
Shabirmean, line 1: The folder name must resemble the product or the API. So please use `discoveryengine` instead of genappbuilder`

genappbuilder/src/main/java/genappbuilder/v1beta/Search.java
```diff
@@ -0,0 +1,61 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package genappbuilder.v1beta;
+
+// [START genappbuilder_search]
```
Shabirmean, line 19: Region tag prefix must match the product/API name. Please use `discoveryengine_search`
Update the END tag as well 
Also update all place where the region tag is used

Shabirmean (commented):
Thank you for contributing this sample. Please TAL at the comments!

Shabirmean (changes requested):
Needs making sure the CI runs

genappbuilder/src/main/java/genappbuilder/v1beta/Search.java
holtskinner: This is a special case... Generative AI App Builder is using the Discovery Engine API and it has a distinct region tag. https://devrel.corp.google.com/admin/products/649

genappbuilder/src/main/java/genappbuilder/v1beta/Search.java
holtskinner: That's really strange, why does this client library need it, but others don't?

genappbuilder/src/main/java/genappbuilder/v1beta/Search.java
holtskinner: Not sure I understand. Why would this not use the search client?

genappbuilder/pom.xml
holtskinner: See previous comment

genappbuilder/pom.xml
holtskinner: Previous comment about product naming

genappbuilder/pom.xml
holtskinner: previous comment about naming

genappbuilder/pom.xml
holtskinner: That's interesting. It also didn't work when I added the version explicitly.
Does libraries-bom have access to the 0.12.0 version? It was just published recently

genappbuilder/src/main/java/genappbuilder/v1beta/Search.java
holtskinner: Good catch, I missed this

genappbuilder/src/main/java/genappbuilder/v1beta/Search.java
Shabirmean: Sorry. That is what I was pointing to. I might have copy pasted something else.

genappbuilder/pom.xml
Shabirmean: libraries-bom will automatically include it in the next release. But we are not sure when the cloud sdk team will publish the next release. SO until then we can explicitly force to use `0.12.0` by adding the version tag

genappbuilder/src/main/java/genappbuilder/v1beta/Search.java
Shabirmean: It's because of the use of the `SearchRequest` and `SearchResponse` classes which are part of this dependency

genappbuilder/pom.xml
Shabirmean: Please leave the comments `unresolved` for the original commentor to see your updates - the commentor shall resolve it once they find it addressed 🙌🏼 

Wrt to the naming - I still think the name here should match the client API. We can have a subdirectory inside it called `genappbuilder`. Also the region tag prefix can match `genappbuilder` since we have it added in devrel services. But the sample directory and the package names must be the client API:

Toplevel directory: `discoveryengine`
Subdirectory - `genappbuilder`
Pom xml artifact details:

```
<groupId>com.example.discoveryengine</groupId>
 <artifactId>genappbuilder-snippets</artifactId>
```

Shabirmean: CC: @averikitsch

genappbuilder/src/main/java/genappbuilder/v1beta/Search.java
holtskinner: That is bizarre, but it seems to have fixed the issue!

Shabirmean (changes requested):
This looks great. Thanks for contributing the sample and also fixing the import issues. I have left a few more comments to polish up the samples to be consistent with the other samples in this repository.

Also one additional comment:
- Thanks for moving the sample to a new folder called `discoveryengine`
- I am assuming there will be more samples around discovery engine, which are not `genappbuilder`
- Thus, can we add the contents of this sample inside a new subdirectory called `genappbuilder` 
- So the path would be `discoveryengine/genappbuilder`

Also we need to add CODEOWNERS for this sample:
- Can you update the following file with the right Product/DPE team that must own this sample? [CODEOWNERS file](https://github.com/GoogleCloudPlatform/java-docs-samples/blob/main/.github/CODEOWNERS)

Shabirmean (approved):

Shabirmean:
@balajismaniam @anguillanneuf Can someone from the @GoogleCloudPlatform/dee-data-ai team TAL at this before I can merge :)

discoveryengine/genappbuilder/src/main/java/genappbuilder/v1beta/Search.java
```diff
@@ -0,0 +1,71 @@
...
+    // Initialize client that will be used to send requests. This client only needs to be created
+    // once, and can be reused for multiple requests. After completing all of your requests, call
+    // the `searchServiceClient.close()` method on the client to safely
+    // clean up any remaining background resources.
+    try (SearchServiceClient searchServiceClient = SearchServiceClient.create()) {
+      SearchRequest request =
+          SearchRequest.newBuilder()
+              .setServingConfig(
+                      ServingConfigName.ofProjectLocationDataStoreServingConfigName(
+                          projectId, location, searchEngineId, servingConfigId)
```
chenlei1216, line 59: Let's add collectionId as well?

holtskinner: Done

chenlei1216 (approved):

PR #7842, opened 2023-04-05
Enable live reload for Java demosite

faube:
- Use Spring Boot devtools to restart on backend changes
- Use the static-locations config to load the latest version of frontend changes

https://docs.spring.io/spring-boot/docs/1.5.16.RELEASE/reference/html/using-boot-devtools.html#using-boot-devtools-restart
https://www.baeldung.com/spring-mvc-static-resources


serargz (approved):
LGTM, although live reload won't work if the service is started with docker, right? if we want that we probably need to mount the directory instead of copying the contents.

faube:
It does work with docker.

PR #7849, opened 2023-04-07
[DLP] Implemented Reidentify With Deterministic Encryption

bhavika-dhanwani:
Implemented [Re-identify content encrypted by deterministic encryption](https://cloud.google.com/dlp/docs/samples/dlp-reidentify-deterministic)


dlp/snippets/src/main/java/dlp/snippets/ReidentifyWithDeterministicEncryption.java
```diff
@@ -0,0 +1,137 @@
...
+
+public class ReidentifyWithDeterministicEncryption {
+
+  public static void main(String[] args) throws Exception {
+    // TODO(developer): Replace these variables before running the sample.
+
+    // The Google Cloud project id to use as a parent resource.
+    String projectId = "your-project-id";
+    // The string to re-identify
+    String textToReIdentify = "My SSN is SSN_TOKEN(36):Abmg1/Wql5D6stA1s467OQsGyegFrV0Qe/E=";
```
micheldavid, line 48: this encrypted text only works with the wrappedkey you used to generate it... it will fail if someone uses a different wrapped key... can we add a DeidentifyWithDeterministicEncryption class, and use that here to generate this text?

micheldavid (changes requested):

dlp/snippets/src/main/java/dlp/snippets/ReidentifyWithDeterministicEncryption.java
micheldavid: I see the PR for deid with deterministic encryption here: https://github.com/GoogleCloudPlatform/java-docs-samples/pull/7848

bhavika-dhanwani: Thank you for reviewing the sample. According to me, this is a re-identification code sample, and for the creation of encrypted text, the user can refer to a de-identification code sample. Additionally, we have not used the de-identification class for the other reidentification code provided [here](https://cloud.google.com/dlp/docs/samples). With that said, please let me know how should I proceed with this and any other re-identification examples.

averikitsch (approved):

dlp/snippets/src/main/java/dlp/snippets/ReidentifyWithDeterministicEncryption.java
micheldavid: I see - python embeds the token here: https://github.com/GoogleCloudPlatform/python-docs-samples/blob/636039e54c227629988838155ee8c9345e66bd13/dlp/snippets/deid_test.py#L236
It also embeds the wrapped and unwrapped, which won't work if someone uses a different google project (e.g. an external contributor).

I also see you have used the deid code to generate the input for reid. I think that is safer and more portable.

dlp/snippets/src/test/java/dlp/snippets/DeIdentificationTests.java
```diff
@@ -555,6 +555,17 @@ public void testDeIdentifyWithDeterministicEncryption() throws IOException {
...
 
+  @Test
+  public void testReIdentifyWithDeterministicEncryption() throws IOException {
+    String textToReIdentify =
+        DeIdenitfyWithDeterministicEncryption.deIdentifyWithDeterministicEncryption(
+            PROJECT_ID, "My SSN is 372819127", WRAPPED_KEY, KMS_KEY_NAME);
+    ReidentifyWithDeterministicEncryption.reIdentifyWithDeterminsiticEncryption(
+        PROJECT_ID, textToReIdentify, WRAPPED_KEY, KMS_KEY_NAME);
+    String output = bout.toString();
+    assertThat(output).contains("Text after re-identification: ");
```
micheldavid, line 566: it should contain the entire input that was provided in the deid method: "My SSN is 372819127"

micheldavid (approved):

PR #7863, opened 2023-04-09
chore: remove versions of managed dependencies

kweinmeister:
Some samples are still explicitly setting dependency versions, even though they are managed by the [libraries-bom](https://github.com/googleapis/java-cloud-bom).

This creates unnecessary dependency updates that would otherwise be managed through one update to the `libraries-bom`. It can also lead to potential conflicts between dependency versions that don't work well together.

This PR removes `<version/>` from dependencies that are already centrally managed.

Several samples were excluded due to failures, and will be revisited later:
* appengine-java8/springboot-helloworld
* automl
* compute/cloud-client
* jobs/v4
* privateca
* pubsub/spring
* video

### DIFF
--- a/appengine-java11/guestbook-cloud-firestore/pom.xml
+++ b/appengine-java11/guestbook-cloud-firestore/pom.xml
@@ -72,8 +72,7 @@
     <dependency>
       <groupId>com.google.guava</groupId>
       <artifactId>guava</artifactId>
-      <version>31.1-jre</version>
-    </dependency>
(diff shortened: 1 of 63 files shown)


ngorchakovaInG (approved):
approved for datacatalog pom.xml changes

irataxy (approved):
approved for media/transcoder pom.xml changes

Shabirmean (approved):
Thank you very much Karl!

PR #7891, opened 2023-04-14
feat: add experimental support for reverse scans public preview

igorbernstein2:


mutianf (approved):

averikitsch:
@igorbernstein2 do you have an ETA on the completion of this PR?

mutianf (approved):

bigtable/hbase/snippets/src/main/java/com/example/bigtable/Reads.java
```diff
@@ -237,6 +236,38 @@ public static void readPrefix(String projectId, String instanceId, String tableI
...
+          new Scan()
+              .setReversed(true)
+              .setLimit(2)
+              .withStartRow(Bytes.toBytes("phone#4c410523#20190505"));
+      ResultScanner rows = table.getScanner(revScan);
+
+      for (Result row : rows) {
+        printRow(row);
+      }
+    } catch (IOException e) {
```
billyjacobson, line 264: can you get rid of the try catch and just let the method throw the exception 

https://github.com/GoogleCloudPlatform/java-docs-samples/blob/main/SAMPLE_FORMAT.md#exception-handling

billyjacobson (approved):
looks good

bigtable/hbase/snippets/src/main/java/com/example/bigtable/Reads.java
igorbernstein2: I think that should come separately, to keep the structure consistent with the rest of the file.

billyjacobson: – will refactor in a larger PR

bigtable/hbase/snippets/pom.xml
```diff
@@ -62,7 +62,7 @@
     <dependency>
       <groupId>com.google.cloud.bigtable</groupId>
       <artifactId>bigtable-hbase-1.x</artifactId>
-      <version>2.9.0</version>
+      <version>2.10.0</version>
```
Sita04, line 65: Do you need to manually set this version? Can it be pulled from the bom?

bigtable/hbase/snippets/src/main/java/com/example/bigtable/Reads.java
```diff
@@ -237,6 +236,38 @@ public static void readPrefix(String projectId, String instanceId, String tableI
...
+  // [START bigtable_reverse_scan_hbase]
+  public static void readRowsReversed() {
+    // TODO(developer): Replace these variables before running the sample.
+    String projectId = "my-project-id";
+    String instanceId = "my-instance-id";
+    String tableId = "mobile-time-series";
+    readRowsReversed(projectId, instanceId, tableId);
+  }
+
+  public static void readRowsReversed(String projectId, String instanceId, String tableId) {
```
Sita04, line 248: Kindly add a method comment here.

bigtable/hbase/snippets/src/main/java/com/example/bigtable/Reads.java
```diff
@@ -237,6 +236,38 @@ public static void readPrefix(String projectId, String instanceId, String tableI
...
+    // Initialize client that will be used to send requests. This client only needs to be created
+    // once, and can be reused for multiple requests. After completing all of your requests, call
+    // the "close" method on the client to safely clean up any remaining background resources.
```
Sita04, line 251:
```suggestion
    // Initialize client that will be used to send requests. This client only needs to be created
    // once, and can be reused for multiple requests.
```

bigtable/hbase/snippets/src/main/java/com/example/bigtable/Reads.java
Sita04: +1 to that. If we are not demonstrating what to do with the exception, it should be thrown.

Sita04 (changes requested):

bigtable/hbase/snippets/pom.xml
igorbernstein2: These version shouldnt be here to begin with. I will remove them in a follow up PR, but I didnt want to expand scope here

Sita04 (approved):

PR #7894, opened 2023-04-14
chore: add managed dependencies

kweinmeister:
Added `libraries-bom` support to samples missing it in `pom.xml`, and where at least 1 dependency would be used from the BOM.

PR #7906, opened 2023-04-18
ci: remove managed dependencies

kweinmeister:
This PR takes a second pass through dependencies that were missed in #7894.


kweinmeister:
FYI @kappratiksha re: `flexible/java-11` use of managed dependencies from`spring-boot-dependencies`, and overall dependency clean-up.

PR #7507, opened 2022-12-15
docs(recaptcha-sample): init add mfa sample

Sita04:
Adding sample and test to demonstrate reCAPTCHA Enterprise Multi Factor Authentication workflow.

PR #7917, opened 2023-04-20
fix: recaptcha password leak sample update

Sita04:
## Description

Based on the comments from Product team in the Typescript PR (https://github.com/GoogleCloudPlatform/nodejs-docs-samples/pull/3059#discussion_r1157666942), site key and token is not necessary for Password leak sample. In fact, it counts as an additional assessment call.

Removing the unnecessary params from the samples and tests.


Sita04:
Tests are not running currently. Should be fixed by this PR: https://github.com/GoogleCloudPlatform/java-docs-samples/pull/7507
Until then, marking the PR as draft.

serargz (approved):

anguillanneuf (approved):

PR #7922, opened 2023-04-21
[DLP] Implemented Inspect Gcs File and Send findings to SCC

bhavika-dhanwani:
Implemented [Inspect Gcs File and Send findings to SCC](https://cloud.google.com/dlp/docs/sending-results-to-scc.md?g=0&l=416#cloud-storage-data)

## Checklist

### DIFF
--- a/dlp/snippets/src/main/java/dlp/snippets/InspectGcsFileSendToScc.java
+++ b/dlp/snippets/src/main/java/dlp/snippets/InspectGcsFileSendToScc.java
@@ -113,4 +113,4 @@ public class InspectGcsFileSendToScc {
   }
 }
 
-// [START dlp_inspect_gcs_send_to_scc]
+// [END dlp_inspect_gcs_send_to_scc]
--- a/dlp/snippets/src/main/java/dlp/snippets/InspectGcsFileSendToScc.java
+++ b/dlp/snippets/src/main/java/dlp/snippets/InspectGcsFileSendToScc.java
@@ -1,0 +1,116 @@
+/*
+ * Copyright 2023 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package dlp.snippets;
+
+// [START dlp_inspect_gcs_send_to_scc]
+
+import com.google.cloud.dlp.v2.DlpServiceClient;
+import com.google.privacy.dlp.v2.Action;
+import com.google.privacy.dlp.v2.CloudStorageOptions;
+import com.google.privacy.dlp.v2.CreateDlpJobRequest;
+import com.google.privacy.dlp.v2.DlpJob;
+import com.google.privacy.dlp.v2.InfoType;
+import com.google.privacy.dlp.v2.InspectConfig;
+import com.google.privacy.dlp.v2.InspectJobConfig;
+import com.google.privacy.dlp.v2.Likelihood;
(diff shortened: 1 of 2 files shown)


dlp/snippets/src/main/java/dlp/snippets/InspectGcsFileSendToScc.java
```diff
@@ -0,0 +1,116 @@
...
+  public static void main(String[] args) throws Exception {
+    // TODO(developer): Replace these variables before running the sample.
+    // The Google Cloud project id to use as a parent resource.
+    String projectId = "your-project-id";
+    // The name of the file in the bucket
+    String gcsPath = "gs://" + "your-bucket-name" + "path/to/file.txt";
+    createJobSendToScc(projectId, gcsPath);
+  }
+
+  // Creates a DLP Job
```
Sita04, line 48: Kindly provide the full description.
"Creates a DLP Job to ..."

dlp/snippets/src/main/java/dlp/snippets/InspectGcsFileSendToScc.java
```diff
@@ -0,0 +1,116 @@
...
+          InspectJobConfig.newBuilder()
+              .setInspectConfig(inspectConfig)
+              .setStorageConfig(storageConfig)
+              .addActions(action)
+              .build();
+
+      // Construct the job creation request to be sent by the client.
+      CreateDlpJobRequest createDlpJobRequest =
+          CreateDlpJobRequest.newBuilder()
+              .setParent(LocationName.of(projectId, "global").toString())
```
Sita04, line 105: nit: Consider moving "global" to a variable in main method.

Sita04 (changes requested):
Thanks for the sample.
Kindly make sure to propagate few of these comments (whichever applicable) in all incoming PRs.

dlp/snippets/src/main/java/dlp/snippets/InspectGcsFileSendToScc.java
bhavika-dhanwani: Generally, we are not declaring "global" in main() for all the code samples.
So if I make changes in current PR, we need to incorporate the changes for all the code sample thus I'm not making the requested changes.

dlp/snippets/src/test/java/dlp/snippets/JobsTests.java
```diff
@@ -125,4 +134,19 @@ public void testDeleteJobs() throws Exception {
...
+
+  @Test
+  public void testCreateJobsSendScc() throws Exception {
+    // Call createJobs to create a Dlp job from project id and gcs path and send data to SCC.
+    InspectGcsFileSendToScc.createJobSendToScc(PROJECT_ID, GCS_PATH);
+    String output = bout.toString();
+    assertThat(output).contains("Job created successfully:");
+
+    // Delete the created Dlp Job
+    String dlpJobName = output.split("Job created successfully: ")[1].split("\n")[0];
```
jakubrauch, line 146: Please clarify why are we plucking out [1] and [0] here.

bhavika-dhanwani: I have modified the extraction here, we don't need [0] here.
So now .split("Job created successfully: ") splits the output string into an array of substrings using the string "Job created successfully: " as the delimiter. The result is an array with at least two elements. [1] accesses the second element (index 1) of the resulting array.

Sita04 (approved):

Sita04:
PTAL at the failing tests.

dlp/snippets/src/test/java/dlp/snippets/JobsTests.java
bhavika-dhanwani: @jakubrauch Since I have used `System.out.println` [here](https://github.com/GoogleCloudPlatform/java-docs-samples/pull/7922/files#diff-59a0bdcd72ddfe723b474c967233c2b82cdda57a211e7a01b1fb6f90d893c29eR112) , `.split("Job created successfully: ") ` splits the output string using the delimiter "Job created successfully: ", which returns an array of strings with at least two elements. The second element of the resulting array contains the job name along with the newline character.

The second split() method is used to remove the newline character from the job name. It splits the second element of the resulting array using the newline character (\n) as the delimiter, and accesses the first element of the resulting array, which should be the job name itself.

PR #7955, opened 2023-04-27
Fix live reload of HTML files.

faube:
We also need to restore the spring-boot-devtools dependency removed by 
https://github.com/GoogleCloudPlatform/java-docs-samples/pull/7927/files#diff-8244e1351f473dc033a31c273f3c5999a07b1cdb678725da1338908330a30393


msampathkumar (approved):
LGMT
No Errors & all the tests are passing.

Kindly wait for the reviewers to do LGMT.

PR #7979, opened 2023-05-02
fix: update demosite to reflect error messages in client 

Sita04:
1. Update returned `response` from backend to include `reason` for "bad" actions.
2. Update client messaging to display the bad action messages.
3. Update client response to fetch 'Invalid token' value based on the reason returned.


Sita04:
@faube Removed the thymeleaf setting since the templates are in the default location. Was there any other intrinsic behavior this property achieved?

recaptcha_enterprise/demosite/src/main/java/app/MainController.java
```diff
@@ -140,7 +185,8 @@ public static ModelAndView signup() {
     String recaptchaAction = PROPERTIES.getProperty("recaptcha_action.signup");
     HashMap<String, HashMap<String, String>> data = new HashMap<>();
     Assessment assessmentResponse;
-    String verdict;
+    String reason = "";
```
averikitsch, line 188: Nit: a lot of this looks copy and paste. Would it be best to refactor this logic into a single function?

averikitsch (approved):

recaptcha_enterprise/demosite/src/main/java/app/MainController.java
Sita04: Thanks Averi. Agreed.
Initially it made sense to have it in individual methods because it was not a lot of code to encapsulate. But it totally makes sense now to have a separate method for the common parts.

recaptcha_enterprise/demosite/src/main/java/app/MainController.java
```diff
@@ -358,4 +366,33 @@ public static ModelAndView comment() {
...
+      label = Label.BAD.getLabel();
+    }
+
+    // Check if the returned recaptcha action matches the expected.
+    if (!assessmentResponse.getTokenProperties().getAction().equals(recaptchaAction)) {
+      reason = Error.ACTION_MISMATCH.getErrorMessage();
+      label = Label.BAD.getLabel();
+    }
+
+    // Check if the returned score is greater less than or equal to the threshold set.
```
faube: "greater less than or equal to" -> "greater than"

Sita04: Updated it in previous commit. Thanks

recaptcha_enterprise/demosite/src/main/resources/application.properties
```diff
@@ -1,2 +1 @@
-spring.thymeleaf.prefix=file:src/main/resources/templates/
```
faube, line 1: We need this line!!!

faube (approved):

recaptcha_enterprise/demosite/src/main/resources/application.properties
Sita04: Local testing threw errors (guessing it's because of some local config). But console testing works. Hence restored.

PR #8005, opened 2023-05-06
fix: address generic type warnings

kweinmeister:


appengine-java8/datastore/src/test/java/com/example/appengine/MetadataKindsTest.java
```diff
@@ -75,7 +75,7 @@ void printLowercaseKinds(DatastoreService ds, PrintWriter writer) {
     // Start with unrestricted kind query
     Query q = new Query(Entities.KIND_METADATA_KIND);
 
-    List<Filter> subFils = new ArrayList();
+    List<Filter> subFils = new ArrayList<Filter>();
```
sriram-mahavadi, line 78: nitpick. want to use the diamond operator instead?  (applies to multiple occurrences below).

`List<Filter> subFils = new ArrayList<>();`

appengine-java8/datastore/src/test/java/com/example/appengine/MetadataPropertiesTest.java
```diff
@@ -199,6 +199,7 @@ public void propertiesOfKind_returnsProperties() throws Exception {
     assertThat(properties).containsExactly("beta", "charlie", "delta");
   }
 
+  @SuppressWarnings("unchecked")
```
sriram-mahavadi, line 202: Any particular reason as to why we are suppressing the check here? (applies to multiple occurrences below).

appengine-java8/datastore/src/test/java/com/example/appengine/MetadataKindsTest.java
kweinmeister: Sure, can use that wherever possible.

appengine-java8/datastore/src/test/java/com/example/appengine/MetadataPropertiesTest.java
kweinmeister: There is an existing warning `Type safety: Unchecked cast from Object to Collection<String>`

here: `return (Collection<String>) propInfo.getProperty("property_representation");`

as `propInfo.getProperty()` returns type `Object`. Is there a change you suggest to address this? I can also remove these changes, leaving the warnings.

functions/v2/ocr/ocr-translate-text/src/main/java/functions/OcrTranslateApiMessage.java
```diff
@@ -59,7 +59,8 @@ public String getLang() {
     return lang;
   }
 
-  public static OcrTranslateApiMessage fromPubsubData(byte[] data) {
+  @SuppressWarnings("unchecked")
```
averikitsch, line 62: Extra spaces before many of the the "@SuppressWarnings" annotations

averikitsch (changes requested):

functions/v2/ocr/ocr-translate-text/src/main/java/functions/OcrTranslateApiMessage.java
kweinmeister: Good catch!

Shabirmean (approved):

averikitsch (approved):

PR #8006, opened 2023-05-06
chore: update pubsublite/streaming-analytics

kweinmeister:
Updates to `pubsublite/streaming-analytics`:

* **Dependencies**: Upgraded to latest dependencies in both Maven and Gradle build files. Removed unused dependencies. Used Libraries BOM to avoid setting a version manually for supported dependencies.
  * Note: Beam GCP dependencies were also set as a managed dependency, with precedence over Google Cloud Libraries BOM. This resolved dependency conflicts between the Libraries BOM and Beam libraries that led to missing class and method exceptions.

* **Code**: resolved several warnings in the test and sample classes.

* **README**: Resolved linting/formatting warnings and observed 80 character line-length.

* **Testing**: Manually tested both Maven and Gradle builds.


anguillanneuf (approved):
Thanks for your diligence @kweinmeister! 😃

PR #8083, opened 2023-05-16
fixit: add 7 iam samples

anguillanneuf:
add 7 missing samples for https://cloud.google.com/iam/docs/creating-custom-roles

running `mvn com.coveo:fmt-maven-plugin:format` touched files in the `iam/` directory without me intending to change them

### DIFF
--- a/iam/snippets/src/main/java/CreateRole.java
+++ b/iam/snippets/src/main/java/CreateRole.java
@@ -34,8 +34,11 @@ public class CreateRole {
     String description = "a description of the role";
     Iterable<String> includedPermissions =
         Arrays.asList("roles/iam.roleViewer", "roles/logging.viewer");
+    // See launch stage enums at
+    // https://cloud.google.com/iam/docs/reference/rpc/google.iam.admin.v1#rolelaunchstage
+    String launchStage = "BETA";
 
-    createRole(projectId, title, description, includedPermissions, roleId);
+    createRole(projectId, title, description, includedPermissions, roleId, launchStage);
   }
 
   public static void createRole(
@@ -43,15 +46,16 @@ public class CreateRole {
       String title,
       String description,
       Iterable<String> includedPermissions,
-      String roleId) {
+      String roleId,
+      String launchStage) {
     Role.Builder roleBuilder =
         Role.newBuilder()
             .setTitle(title)
             .setDescription(description)
             .addAllIncludedPermissions(includedPermissions)
-            // More available enums at
-            // https://cloud.google.com/iam/docs/reference/rpc/google.iam.admin.v1#rolelaunchstage
-            .setStage(RoleLaunchStage.BETA);
+            // Generally we don't recommend accessing RoleLaunchStage enums via strings.
+            // Instead, retrieve them using the dot notation (e.g. RoleLaunchStage.BETA).
+            .setStage(RoleLaunchStage.valueOf(launchStage));
 
     CreateRoleRequest createRoleRequest =
         CreateRoleRequest.newBuilder()
--- a/iam/snippets/src/main/java/CreateRole.java
+++ b/iam/snippets/src/main/java/CreateRole.java
@@ -14,6 +14,8 @@
  * limitations under the License.
  */
 
+// [START iam_create_role]
+
 import com.google.api.gax.rpc.AlreadyExistsException;
 import com.google.cloud.iam.admin.v1.IAMClient;
 import com.google.iam.admin.v1.CreateRoleRequest;
@@ -22,15 +24,15 @@ import com.google.iam.admin.v1.Role.RoleLaunchStage;
 import java.io.IOException;
 import java.util.Arrays;
 
-// [START iam_create_role]
 public class CreateRole {
   public static void main(String[] args) {
     // TODO(developer): Replace the variables before running the sample.
     String projectId = "your-project-id";
-    String title = "Role Title";
-    String description = "A description";
-    Iterable<String> includedPermissions = Arrays.asList("roles/iam.roleViewer");
-    String roleId = "your-role-id";
+    String roleId = "a unique identifier (e.g. testViewer)";
+    String title = "a title for your role (e.g. IAM Role Viewer)";
+    String description = "a description of the role";
+    Iterable<String> includedPermissions =
+        Arrays.asList("roles/iam.roleViewer", "roles/logging.viewer");
 
     createRole(projectId, title, description, includedPermissions, roleId);
   }
@@ -55,6 +57,8 @@ public class CreateRole {
             .setRole(roleBuilder)
             .build();
 
+    // Initialize client for sending requests. This client only needs to be created
+    // once, and can be reused for multiple requests.
     try (IAMClient iamClient = IAMClient.create()) {
       Role result = iamClient.createRole(createRoleRequest);
       System.out.println("Created role: " + result.getName());
--- a/iam/snippets/src/main/java/CreateRole.java
+++ b/iam/snippets/src/main/java/CreateRole.java
@@ -33,11 +33,8 @@ public class CreateRole {
     String description = "a description of the role";
     Iterable<String> includedPermissions =
         Arrays.asList("roles/iam.roleViewer", "roles/logging.viewer");
-    // See launch stage enums at
-    // https://cloud.google.com/iam/docs/reference/rpc/google.iam.admin.v1#rolelaunchstage
-    String launchStage = "BETA";
 
-    createRole(projectId, title, description, includedPermissions, roleId, launchStage);
+    createRole(projectId, title, description, includedPermissions, roleId);
   }
 
   public static void createRole(
@@ -45,17 +42,16 @@ public class CreateRole {
       String title,
       String description,
       Iterable<String> includedPermissions,
-      String roleId,
-      String launchStage)
+      String roleId)
       throws IOException {
     Role.Builder roleBuilder =
         Role.newBuilder()
             .setTitle(title)
             .setDescription(description)
             .addAllIncludedPermissions(includedPermissions)
-            // Generally we don't recommend accessing RoleLaunchStage enums via strings.
-            // Instead, retrieve them using the dot notation (e.g. RoleLaunchStage.BETA).
-            .setStage(RoleLaunchStage.valueOf(launchStage));
+            // See launch stage enums at
+            // https://cloud.google.com/iam/docs/reference/rpc/google.iam.admin.v1#rolelaunchstage
+            .setStage(RoleLaunchStage.BETA);
     CreateRoleRequest createRoleRequest =
         CreateRoleRequest.newBuilder()
             .setParent("projects/" + projectId)
--- a/iam/snippets/src/main/java/CreateRole.java
+++ b/iam/snippets/src/main/java/CreateRole.java
@@ -16,7 +16,6 @@
 
 // [START iam_create_role]
 
-import com.google.api.gax.rpc.AlreadyExistsException;
 import com.google.cloud.iam.admin.v1.IAMClient;
 import com.google.iam.admin.v1.CreateRoleRequest;
 import com.google.iam.admin.v1.Role;
@@ -26,7 +25,7 @@ import java.util.Arrays;
 
 /** Create role. */
 public class CreateRole {
-  public static void main(String[] args) {
+  public static void main(String[] args) throws IOException {
     // TODO(developer): Replace the variables before running the sample.
     String projectId = "your-project-id";
     String roleId = "a unique identifier (e.g. testViewer)";
@@ -47,7 +46,8 @@ public class CreateRole {
       String description,
       Iterable<String> includedPermissions,
       String roleId,
-      String launchStage) {
+      String launchStage)
+      throws IOException {
     Role.Builder roleBuilder =
         Role.newBuilder()
             .setTitle(title)
@@ -56,7 +56,6 @@ public class CreateRole {
             // Generally we don't recommend accessing RoleLaunchStage enums via strings.
             // Instead, retrieve them using the dot notation (e.g. RoleLaunchStage.BETA).
             .setStage(RoleLaunchStage.valueOf(launchStage));
-
     CreateRoleRequest createRoleRequest =
         CreateRoleRequest.newBuilder()
             .setParent("projects/" + projectId)
@@ -69,8 +68,6 @@ public class CreateRole {
     try (IAMClient iamClient = IAMClient.create()) {
       Role result = iamClient.createRole(createRoleRequest);
       System.out.println("Created role: " + result.getName());
-    } catch (AlreadyExistsException | IOException e) {
-      throw new RuntimeException(e);
     }
   }
 }
--- a/iam/snippets/src/main/java/CreateRole.java
+++ b/iam/snippets/src/main/java/CreateRole.java
@@ -1,0 +1,66 @@
+/*
+ * Copyright 2023 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import com.google.api.gax.rpc.AlreadyExistsException;
+import com.google.cloud.iam.admin.v1.IAMClient;
+import com.google.iam.admin.v1.CreateRoleRequest;
+import com.google.iam.admin.v1.Role;
+import com.google.iam.admin.v1.Role.RoleLaunchStage;
+import java.io.IOException;
+import java.util.Arrays;
+
+// [START iam_create_role]
+public class CreateRole {
+  public static void main(String[] args) {
+    // TODO(developer): Replace the variables before running the sample.
+    String projectId = "your-project-id";
+    String title = "Role Title";
+    String description = "A description";
+    Iterable<String> includedPermissions = Arrays.asList("roles/iam.roleViewer");
+    String roleId = "your-role-id";
+
+    createRole(projectId, title, description, includedPermissions, roleId);
+  }
+
+  public static void createRole(
+      String projectId,
+      String title,
+      String description,
+      Iterable<String> includedPermissions,
+      String roleId) {
+    Role.Builder roleBuilder =
+        Role.newBuilder()
+            .setTitle(title)
+            .setDescription(description)
+            .addAllIncludedPermissions(includedPermissions)
+            .setStage(RoleLaunchStage.BETA);
+
+    CreateRoleRequest createRoleRequest =
+        CreateRoleRequest.newBuilder()
+            .setParent("projects/" + projectId)
+            .setRoleId(roleId)
+            .setRole(roleBuilder)
+            .build();
+
+    try (IAMClient iamClient = IAMClient.create()) {
+      Role result = iamClient.createRole(createRoleRequest);
+      System.out.println("Created role: " + result.getName());
+    } catch (AlreadyExistsException | IOException e) {
+      throw new RuntimeException(e);
+    }
+  }
+}
+// [END iam_create_role]
--- a/iam/snippets/src/main/java/CreateRole.java
+++ b/iam/snippets/src/main/java/CreateRole.java
@@ -24,6 +24,7 @@ import com.google.iam.admin.v1.Role.RoleLaunchStage;
 import java.io.IOException;
 import java.util.Arrays;
 
+/** Create role. */
 public class CreateRole {
   public static void main(String[] args) {
     // TODO(developer): Replace the variables before running the sample.
@@ -48,6 +49,8 @@ public class CreateRole {
             .setTitle(title)
             .setDescription(description)
             .addAllIncludedPermissions(includedPermissions)
+            // More available enums at
+            // https://cloud.google.com/iam/docs/reference/rpc/google.iam.admin.v1#rolelaunchstage
             .setStage(RoleLaunchStage.BETA);
 
     CreateRoleRequest createRoleRequest =
--- a/iam/snippets/src/main/java/DeleteRole.java
+++ b/iam/snippets/src/main/java/DeleteRole.java
@@ -21,6 +21,7 @@ import com.google.cloud.iam.admin.v1.IAMClient;
 import com.google.iam.admin.v1.DeleteRoleRequest;
 import java.io.IOException;
 
+/** Delete role. */
 public class DeleteRole {
 
   public static void main(String[] args) {
--- a/iam/snippets/src/main/java/DeleteRole.java
+++ b/iam/snippets/src/main/java/DeleteRole.java
@@ -16,7 +16,6 @@
 
 // [START iam_delete_role]
 
-import com.google.api.gax.rpc.NotFoundException;
 import com.google.cloud.iam.admin.v1.IAMClient;
 import com.google.iam.admin.v1.DeleteRoleRequest;
 import java.io.IOException;
@@ -24,7 +23,7 @@ import java.io.IOException;
 /** Delete role. */
 public class DeleteRole {
 
-  public static void main(String[] args) {
+  public static void main(String[] args) throws IOException {
     // TODO(developer): Replace the variables before running the sample.
     // Role ID must point to an existing role.
     String projectId = "your-project-id";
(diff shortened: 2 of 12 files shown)


anguillanneuf:
Faile due to `com.google.api.gax.rpc.PermissionDeniedException: io.grpc.StatusRuntimeException: PERMISSION_DENIED: You don't have permission to create a role in projects/java-docs-samples-testing.
Caused by: io.grpc.StatusRuntimeException: PERMISSION_DENIED: You don't have permission to create a role in projects/java-docs-samples-testing.`

hegemonic:
@melaniedejong Can you handle this review?

iam/snippets/src/main/java/CreateRole.java
```diff
@@ -0,0 +1,66 @@
...
+
+// [START iam_create_role]
+public class CreateRole {
+  public static void main(String[] args) {
+    // TODO(developer): Replace the variables before running the sample.
+    String projectId = "your-project-id";
+    String title = "Role Title";
+    String description = "A description";
+    Iterable<String> includedPermissions = Arrays.asList("roles/iam.roleViewer");
+    String roleId = "your-role-id";
```
Shabirmean, line 33: (nit) can you move `roleId` above `includedPermissions` so that it is amongst the other `String` variables?

Shabirmean: Can we add some comments to describe some of these variables? What are accepted values for the `roleid` string? Can it be anything? Or should be something specific?

Shabirmean: Can we add more than one role to the includedPermissions list and also add a link to where one can find all the permissions?

https://cloud.google.com/iam/docs/understanding-roles

Shabirmean, line 33:
```suggestion
    String projectId = "your-project-id";
    String title = "a title for your role (e.g. Video Stream Viewer)";
    String description = "detailed description of what the role does";
    String roleId = "a unique identifier (e.g. ?????)";
    Iterable<String> includedPermissions = Arrays.asList("roles/iam.roleViewer");
    
```

iam/snippets/src/main/java/CreateRole.java
```diff
@@ -0,0 +1,66 @@
...
+
+    createRole(projectId, title, description, includedPermissions, roleId);
+  }
+
+  public static void createRole(
+      String projectId,
+      String title,
+      String description,
+      Iterable<String> includedPermissions,
+      String roleId) {
```
Shabirmean, line 43: Is one argument per line fromt he style guide? Or did the formatter break it like this? If not can be put the arguments in the same line and add line breaks only if the maximum character limit per line is exceeded?

iam/snippets/src/main/java/CreateRole.java
```diff
@@ -0,0 +1,66 @@
...
+            .setRoleId(roleId)
+            .setRole(roleBuilder)
+            .build();
+
+    try (IAMClient iamClient = IAMClient.create()) {
+      Role result = iamClient.createRole(createRoleRequest);
+      System.out.println("Created role: " + result.getName());
+    } catch (AlreadyExistsException | IOException e) {
+      throw new RuntimeException(e);
+    }
```
Shabirmean, line 63: Did you follow this `catch` guidelines from somewhere else? The other samples just add the throws clause to the method signature and does not catch the exception. So we should follow the same. When a user adopts the sample to their own environment they would handle it in the best way suitable to their environment

iam/snippets/src/main/java/CreateRole.java
```diff
@@ -0,0 +1,66 @@
...
+            .setStage(RoleLaunchStage.BETA);
+
+    CreateRoleRequest createRoleRequest =
+        CreateRoleRequest.newBuilder()
+            .setParent("projects/" + projectId)
+            .setRoleId(roleId)
+            .setRole(roleBuilder)
+            .build();
+
+    try (IAMClient iamClient = IAMClient.create()) {
```
Shabirmean, line 58: Add generic line comment to the `try` block

```suggestion
   // Initialize client that will be used to send requests. This client only needs to be created
    // once, and can be reused for multiple requests. After completing all of your requests, call
    // the "close" method on the client to safely clean up any remaining background resources.
    try (IAMClient iamClient = IAMClient.create()) {}
```

iam/snippets/src/main/java/CreateRole.java
```diff
@@ -0,0 +1,66 @@
...
+
+import com.google.api.gax.rpc.AlreadyExistsException;
+import com.google.cloud.iam.admin.v1.IAMClient;
+import com.google.iam.admin.v1.CreateRoleRequest;
+import com.google.iam.admin.v1.Role;
+import com.google.iam.admin.v1.Role.RoleLaunchStage;
+import java.io.IOException;
+import java.util.Arrays;
+
+// [START iam_create_role]
```
Shabirmean, line 25: Region tag must include the imports. Only exclude the package definition and the license header.

Shabirmean: This applies to all the samples

iam/snippets/src/main/java/DeleteRole.java
```diff
@@ -0,0 +1,48 @@
...
+    deleteRole(projectId, roleId);
+  }
+
+  public static void deleteRole(String projectId, String roleId) {
+    DeleteRoleRequest deleteRoleRequest =
+        DeleteRoleRequest.newBuilder()
+            .setName("projects/" + projectId + "/roles/" + roleId)
+            .build();
+
+    try (IAMClient iamClient = IAMClient.create()) {
```
Shabirmean, line 40: See comment about try catch block in the `CreateRole` sample

iam/snippets/src/main/java/EditRole.java
```diff
@@ -0,0 +1,59 @@
...
+    String projectId = "your-project-id";
+    String roleId = "your-role-id";
+
+    editRole(projectId, roleId);
+  }
+
+  public static void editRole(String projectId, String roleId) {
+
+    String roleName = "projects/" + projectId + "/roles/" + roleId;
+
```
Shabirmean, line 39: Remove additional spaces here

iam/snippets/src/main/java/DeleteRole.java
```diff
@@ -0,0 +1,48 @@
...
+    String projectId = "your-project-id";
+    String roleId = "your-role-id";
+
+    deleteRole(projectId, roleId);
+  }
+
+  public static void deleteRole(String projectId, String roleId) {
+    DeleteRoleRequest deleteRoleRequest =
+        DeleteRoleRequest.newBuilder()
+            .setName("projects/" + projectId + "/roles/" + roleId)
```
Shabirmean, line 37: Here you have the roleName inline the `setName` method, but in the next sample you have a variable called `roleName` defined and you are using that variable. Can we have a consistent way of doing it in all the samples?

Shabirmean: In GCP which is the `name` and which is the `id`?

I always thought that the single string was the name and the one with `/projects/<PROJCT>/region/<REGION>/roles/<name>` was the **id**. Can we verify this properly and name variables accordingly?

Shabirmean: Applies to all the files

iam/snippets/src/main/java/EditRole.java
```diff
@@ -0,0 +1,59 @@
...
+    Role.Builder roleBuilder = Role.newBuilder().setName(roleName).setStage(RoleLaunchStage.GA);
```
Shabirmean, line 40: Can we format the builder to be in multiple lines like with `CreateRole` sample?

iam/snippets/src/main/java/EditRole.java
```diff
@@ -0,0 +1,59 @@
...
+
+    FieldMask fieldMask = FieldMask.newBuilder().addPaths("stage").build();
```
Shabirmean, line 42: Can we break builder patterns in multiple lines?

iam/snippets/src/main/java/EditRole.java
```diff
@@ -0,0 +1,59 @@
...
+
+    UpdateRoleRequest updateRoleRequest =
+        UpdateRoleRequest.newBuilder()
+            .setName(roleName)
+            .setRole(roleBuilder)
+            .setUpdateMask(fieldMask)
+            .build();
+
+    try (IAMClient iamClient = IAMClient.create()) {
```
Shabirmean, line 51: See comment about try catch block in the `CreateRole` sample

iam/snippets/src/main/java/GetRole.java
```diff
@@ -0,0 +1,44 @@
...
+import com.google.iam.admin.v1.GetRoleRequest;
+import com.google.iam.admin.v1.Role;
+import java.io.IOException;
+
+// [START iam_get_role]
+public class GetRole {
+
+  public static void main(String[] args) {
+    // TODO(developer): Replace the variable before running the sample.
+    String roleId = "roles/iam.roleViewer";
```
Shabirmean, line 28: Is this a correct example of a role id? Should it be in the format of `projects/PROJ/roles/<ROELNAME>`?

iam/snippets/src/main/java/GetRole.java
```diff
@@ -0,0 +1,44 @@
...
+
+    getRole(roleId);
+  }
+
+  public static void getRole(String roleId) {
+    GetRoleRequest getRoleRequest = GetRoleRequest.newBuilder().setName(roleId).build();
+
+    try (IAMClient iamClient = IAMClient.create()) {
```
Shabirmean, line 36: See comment about try catch block in the `CreateRole` sample

iam/snippets/src/main/java/ListRoles.java
```diff
@@ -0,0 +1,45 @@
...
+  public static void main(String[] args) {
+    // TODO(developer): Replace the variable before running the sample.
+    String projectId = "your-project-id";
+
+    listRoles(projectId);
+  }
+
+  public static void listRoles(String projectId) {
+    ListRolesRequest listRolesRequest =
+        ListRolesRequest.newBuilder().setParent("projects/" + projectId).build();
```
Shabirmean, line 35: Break builder pattern into multiple lines

iam/snippets/src/main/java/ListRoles.java
```diff
@@ -0,0 +1,45 @@
...
+
+    try (IAMClient iamClient = IAMClient.create()) {
+      ListRolesPagedResponse listRolesResponse = iamClient.listRoles(listRolesRequest);
+      listRolesResponse.iterateAll().forEach(role -> System.out.println(role));
```
Shabirmean, line 39:
```suggestion
      listRolesResponse.iterateAll()
          .forEach(role -> System.out.println(role));
```

iam/snippets/src/main/java/ListRoles.java
```diff
@@ -0,0 +1,45 @@
...
+    String projectId = "your-project-id";
+
+    listRoles(projectId);
+  }
+
+  public static void listRoles(String projectId) {
+    ListRolesRequest listRolesRequest =
+        ListRolesRequest.newBuilder().setParent("projects/" + projectId).build();
+
+    try (IAMClient iamClient = IAMClient.create()) {
```
Shabirmean, line 37: See comment about try catch block in the `CreateRole` sample

iam/snippets/src/main/java/QueryTestablePermissions.java
```diff
@@ -0,0 +1,49 @@
...
+    queryTestablePermissions(projectId);
+  }
+
+  public static void queryTestablePermissions(String projectId) {
+    QueryTestablePermissionsRequest queryTestablePermissionsRequest =
+        QueryTestablePermissionsRequest.newBuilder()
+            .setFullResourceName("//cloudresourcemanager.googleapis.com/projects/" + projectId)
+            .build();
+
+    try (IAMClient iamClient = IAMClient.create()) {
```
Shabirmean, line 38: See comment about try catch block in the `CreateRole` sample

iam/snippets/src/main/java/QueryTestablePermissions.java
```diff
@@ -0,0 +1,49 @@
...
+    // TODO(developer): Replace the variable before running the sample.
+    String projectId = "your-project-id";
+
+    queryTestablePermissions(projectId);
+  }
+
+  public static void queryTestablePermissions(String projectId) {
+    QueryTestablePermissionsRequest queryTestablePermissionsRequest =
+        QueryTestablePermissionsRequest.newBuilder()
+            .setFullResourceName("//cloudresourcemanager.googleapis.com/projects/" + projectId)
```
Shabirmean, line 35: Should the resource name be a variable passed into the method? 
Can we move it to the main method and ask the developer to update that variable based on their needs and give them an example?

iam/snippets/src/main/java/UndeleteRole.java
```diff
@@ -0,0 +1,49 @@
...
+    String roleId = "your-role-id";
+
+    undeleteRole(projectId, roleId);
+  }
+
+  public static void undeleteRole(String projectId, String roleId) {
+
+    UndeleteRoleRequest undeleteRoleRequest =
+        UndeleteRoleRequest.newBuilder()
+            .setName("projects/" + projectId + "/roles/" + roleId)
```
Shabirmean, line 38: Same comment as earlier about being consistent with the way we define this roleName variable

iam/snippets/src/main/java/UndeleteRole.java
```diff
@@ -0,0 +1,49 @@
...
+            .build();
+
+    try (IAMClient iamClient = IAMClient.create()) {
```
Shabirmean, line 41: See comment about try catch block in the `CreateRole` sample

iam/snippets/src/test/java/RoleIT.java
```diff
@@ -0,0 +1,129 @@
...
+  public void tearDown() {
+    System.setOut(null);
+  }
+
+  @AfterClass
+  public static void cleanUp() {
+    try (IAMClient iamClient = IAMClient.create()) {
+      iamClient.deleteRole(DeleteRoleRequest.newBuilder().setName(roleName).build());
+    } catch (NotFoundException e) {
+      System.out.println("Role no longer exists already.");
```
Shabirmean, line 82: Should this be:
- `Role no longer exists.`
OR
- `Role deleted already.`

iam/snippets/src/test/java/RoleIT.java
```diff
@@ -0,0 +1,129 @@
...
+    } catch (IOException e) {
+      throw new RuntimeException(e);
+    }
+  }
+
+  @Test
+  public void testRole() {
```
Shabirmean, line 89: Can we break all of this into their own individual test cases instead of all being in one?

Shabirmean (changes requested):
Thank you for adding all this samples 🙏🏼 

I have left some comments, mostly related to maintaining code consistency compared to other samples and improving readability.  

Also can you please add a simple **comment** above each `Class` or the role specific method about what it does/intended to? In most cases the name itself makes it self evident but it's still important and in some cases it is necessary like with `undelete` - a developer can think what does and undelete do.

iam/snippets/src/test/java/RoleIT.java
anguillanneuf: why do we do that? `@Before`, `@After`, `@BeforeClass`, `@AfterClass` let us achieve test separation. i've also found lots of precedences.

- https://github.com/GoogleCloudPlatform/java-docs-samples/blob/main/iam/snippets/src/test/java/DenyIT.java
- https://github.com/googleapis/java-pubsub/tree/main/samples/snippets/src/test/java/pubsub

iam/snippets/src/main/java/CreateRole.java
anguillanneuf: re: the second comment, these samples are to be added to https://cloud.google.com/iam/docs/creating-custom-roles, and the docs write-up already describes what these values are.

iam/snippets/src/test/java/RoleIT.java
Shabirmean: Im leaning towards having it broken in light of making it easy for future contributors not having to worry about samples they didn't touch. 

@averikitsch and @lesv to chime in what would be the best direction to take here.

iam/snippets/src/main/java/CreateRole.java
anguillanneuf: to answer your question, it's the formatter. lint is passed :)

iam/snippets/src/main/java/DeleteRole.java
anguillanneuf: The fully qualified name refers to the long string. Just like when we say project ID, we refer to the short string. See a pub/sub example: https://cloud.google.com/pubsub/docs/reference/rpc/google.pubsub.v1#topic

iam/snippets/src/main/java/EditRole.java
anguillanneuf: lint is passing, ignore.

iam/snippets/src/main/java/GetRole.java
anguillanneuf: ditto.

iam/snippets/src/main/java/ListRoles.java
anguillanneuf: I'll let the formatter dictate this one.

iam/snippets/src/main/java/EditRole.java
anguillanneuf: Letting formatter dictate. Linter didn't complain.

iam/snippets/src/main/java/EditRole.java
anguillanneuf: Ditto.

iam/snippets/src/main/java/EditRole.java
anguillanneuf: Ditto.

iam/snippets/src/main/java/ListRoles.java
anguillanneuf: Ditto.

iam/snippets/src/main/java/QueryTestablePermissions.java
anguillanneuf: Ditto.

iam/snippets/src/main/java/DeleteRole.java
anguillanneuf: Ditto.

iam/snippets/src/main/java/GetRole.java
anguillanneuf: Ditto.

iam/snippets/src/main/java/UndeleteRole.java
anguillanneuf: Ditto.

iam/snippets/src/main/java/EditRole.java
Shabirmean: Maybe the linter is not configured to capture it, but in the interest of having a clean code can we remove these unnecessary lines?

iam/snippets/src/main/java/CreateRole.java
Shabirmean: the comment was not about this sample being un-usable as is. It was only pointing to how the other samples have handled it. So seeking consistency with other samples. Also I don't see the need for catching specific Exceptions and throwing a generic RuntimeException - if we are only going to throw a new exception, we might as well throw the original exception as is.

iam/snippets/src/main/java/CreateRole.java
Shabirmean: Makes sense

iam/snippets/src/main/java/CreateRole.java
averikitsch: We should use try with resources, but we don't need to catch these exceptions unless we show the user how to recover from them. Allow the createRole method to throw these exceptions

iam/snippets/src/main/java/EditRole.java
averikitsch: Removing the lines is more idiomatic

iam/snippets/src/main/java/GetRole.java
averikitsch: We may want Sita to weigh in on consistency. I think we should be careful about role name and role id. This string may just signal a "role" (https://cloud.google.com/iam/docs/reference/rest/v1/roles/get)

iam/snippets/src/main/java/QueryTestablePermissions.java
averikitsch: It would be nice to know where this value comes from and if it is changeable.

iam/snippets/src/test/java/RoleIT.java
averikitsch: This repo does have technical debt so precedence should not outweigh a best practice. It is best practice to keep tests more isolated; however, with these samples being highly scoped, I don't think there's a high risk of keeping the test this way. Following best practices here may double the size of this file and make maintenance harder.

iam/snippets/src/main/java/CreateRole.java
anguillanneuf: @averikitsch fyi the existing samples aren't using a package. what should this package be?

https://github.com/GoogleCloudPlatform/java-docs-samples/tree/main/iam/snippets/src/main/java

iam/snippets/src/main/java/CreateRole.java
```diff
@@ -0,0 +1,70 @@
...
+      String title,
+      String description,
+      Iterable<String> includedPermissions,
+      String roleId) {
+    Role.Builder roleBuilder =
+        Role.newBuilder()
+            .setTitle(title)
+            .setDescription(description)
+            .addAllIncludedPermissions(includedPermissions)
+            .setStage(RoleLaunchStage.BETA);
```
melaniedejong, line 51: Can we make the launch stage an option for the user to set as well?

iam/snippets/src/main/java/EditRole.java
```diff
@@ -0,0 +1,62 @@
...
+    String roleId = "a unique identifier (e.g. testViewer)";
+
+    editRole(projectId, roleId);
+  }
+
+  public static void editRole(String projectId, String roleId) {
+
+    String roleName = "projects/" + projectId + "/roles/" + roleId;
+
+    Role.Builder roleBuilder = Role.newBuilder().setName(roleName).setStage(RoleLaunchStage.GA);
```
melaniedejong, line 41: Instead of editing the launch stage, can we let the user input new information to edit the role? For examples, see other language samples on https://cloud.google.com/iam/docs/creating-custom-roles#edit-role

iam/snippets/src/main/java/QueryTestablePermissions.java
```diff
@@ -0,0 +1,53 @@
...
+import com.google.api.gax.rpc.NotFoundException;
+import com.google.cloud.iam.admin.v1.IAMClient;
+import com.google.cloud.iam.admin.v1.IAMClient.QueryTestablePermissionsPagedResponse;
+import com.google.iam.admin.v1.QueryTestablePermissionsRequest;
+import java.io.IOException;
+
+public class QueryTestablePermissions {
+  public static void main(String[] args) {
+    // TODO(developer): Replace the variable before running the sample.
+    String projectId = "your-project-id";
```
melaniedejong, line 28: Instead of assuming they want to list grantable roles for a project, can we ask them to input the full resource ID, like we do in the other samples (https://cloud.google.com/iam/docs/creating-custom-roles#viewing-resource-permissions)?

iam/snippets/src/main/java/UndeleteRole.java
```diff
@@ -0,0 +1,51 @@
...
+import com.google.cloud.iam.admin.v1.IAMClient;
+import com.google.iam.admin.v1.Role;
+import com.google.iam.admin.v1.UndeleteRoleRequest;
+import java.io.IOException;
+
+public class UndeleteRole {
+
+  public static void main(String[] args) {
+    // TODO(developer): Replace the variables before running the sample.
+    // Role ID must point to a deleted role within 7 days.
```
melaniedejong, line 28: Instead of saying "a deleted role within 7 days", can we say something like, "a role that was deleted in the past 7 days"?

Shabirmean (approved):
LGTM! 🚀 

Thank you very much for addressing all the comments! LGTM - except the two unresolved comments with @averikitsch's comments. 

Please consider removing `catch` blocks  and throw the exceptions to make it consistent with other samples and to get rid of a redundant wrap-around RuntimeException. Also consider removing unnecessary blank lines!

Thanks alot 🙌🏼

iam/snippets/src/main/java/QueryTestablePermissions.java
```diff
@@ -0,0 +1,50 @@
...
+import com.google.iam.admin.v1.QueryTestablePermissionsRequest;
+import java.io.IOException;
+
+/** View available permissions in a project. */
+public class QueryTestablePermissions {
+  public static void main(String[] args) throws IOException {
+    // TODO(developer): Replace the variable before running the sample.
+    // Full resource names can take one of the following forms:
+    // cloudresourcemanager.googleapis.com/projects/PROJECT_ID
+    // To view available permissions in an organization, use:
```
melaniedejong, line 30: Remove (since the preceding sentence says that the resource names "can take one of the following forms")

melaniedejong (approved):

PR #8103, opened 2023-05-17
fixit: add sample for cloudcdn sign url prefix

zhumin8:
add sample for cloudcdn_sign_url_prefix as part of fixit.
Doc page: https://cloud.google.com/cdn/docs/using-signed-urls#programmatically_create_signed_urls_with_a_url_prefix

### DIFF
--- a/cdn/signed-urls/src/main/java/com/google/cdn/SignedUrlWithPrefix.java
+++ b/cdn/signed-urls/src/main/java/com/google/cdn/SignedUrlWithPrefix.java
@@ -35,20 +35,20 @@ public class SignedUrlWithPrefix {
   public static void main(String[] args) throws Exception {
     // TODO(developer): Replace these variables before running the sample.
 
-    // The name of the signing key added to the back end bucket or service
+    // The name of the signing key must match a key added to the back end bucket or service.
     String keyName = "YOUR-KEY-NAME";
-    // Path to the URL signing key uploaded to the backend service/bucket
+    // Path to the URL signing key uploaded to the backend service/bucket.
     String keyPath = "/path/to/key";
-    // The date that the signed URL expires
+    // The date that the signed URL expires.
     long expirationTime = ZonedDateTime.now().plusDays(1).toEpochSecond();
     // URL of request
     String requestUrl = "https://media.example.com/videos/id/main.m3u8?userID=abc123&starting_profile=1";
-    // URL prefix to sign as a string. urlPrefix must start with either http:// or
-    // https:// and should not include query parameters
+    // URL prefix to sign as a string. URL prefix must start with either "http://" or "https://"
+    // and must not include query parameters.
     String urlPrefix = "https://media.example.com/videos/";
 
-    // Read the key as a base64 url-safe encoded string, then convert to byte array
-    // Key used in signing should be in raw form (not base64url-encoded).
+    // Read the key as a base64 url-safe encoded string, then convert to byte array.
+    // Key used in signing must be in raw form (not base64url-encoded).
     String base64String = new String(Files.readAllBytes(Paths.get(keyPath)),
         StandardCharsets.UTF_8);
     byte[] keyBytes = Base64.getUrlDecoder().decode(base64String);
@@ -66,7 +66,7 @@ public class SignedUrlWithPrefix {
       long expirationTime)
       throws InvalidKeyException, NoSuchAlgorithmException {
 
-    // Validate input URL prefix
+    // Validate input URL prefix.
     try {
       URL validatedUrlPrefix = new URL(urlPrefix);
       if (!validatedUrlPrefix.getProtocol().startsWith("http")) {
@@ -90,7 +90,7 @@ public class SignedUrlWithPrefix {
     return requestUrl + "&" + urlToSign + "&Signature=" + encoded;
   }
 
-  // Creates signature for input url with private key
+  // Creates signature for input url with private key.
   private static String getSignatureForUrl(byte[] privateKey, String input)
       throws InvalidKeyException, NoSuchAlgorithmException {
 


cdn/signed-urls/src/main/java/com/google/cdn/SignedUrls.java
```diff
@@ -75,6 +76,61 @@ public static String getSignature(byte[] privateKey, String input)
   }
   // [END cloudcdn_sign_url]
 
+  // [START cloudcdn_sign_url_prefix]
```
Sita04, line 79: One sample per file. Kindly move this sample to a new file.

cdn/signed-urls/src/main/java/com/google/cdn/SignedUrls.java
```diff
@@ -75,6 +76,61 @@ public static String getSignature(byte[] privateKey, String input)
...
+  /**
+   * Creates a signed URL with a URL prefix for a Cloud CDN endpoint with the given key. Prefixes
+   * allow access to any URL with the same prefix, and can be useful for granting access broader
+   * content without signing multiple URLs.
+   *
+   * @param requestUrl URL of request
```
Sita04, line 85: Applies to all parameters below.
Consider moving these comments to the main method in the new file and refrain from javadoc style comments.

cdn/signed-urls/src/main/java/com/google/cdn/SignedUrls.java
```diff
@@ -75,6 +76,61 @@ public static String getSignature(byte[] privateKey, String input)
...
+      String urlPrefix,
+      byte[] key,
+      String keyName,
+      Date expirationTime)
+      throws InvalidKeyException, NoSuchAlgorithmException, IllegalArgumentException {
+
+    if (urlPrefix.contains("?") || urlPrefix.contains("#")) {
+      throw new IllegalArgumentException("urlPrefix must not include query params: " + urlPrefix);
+    }
+    if (!urlPrefix.startsWith("http://") && !urlPrefix.startsWith("https://")) {
```
Sita04, line 106:
```suggestion
    if (!urlPrefix.startsWith("http://")) {
```

cdn/signed-urls/src/main/java/com/google/cdn/SignedUrls.java
```diff
@@ -75,6 +76,61 @@ public static String getSignature(byte[] privateKey, String input)
...
+   * @return a properly formatted signed URL with URL prefix
+   * @throws InvalidKeyException when there is an error generating the signature for the input key
+   * @throws NoSuchAlgorithmException when HmacSHA1 algorithm is not available in the environment
+   * @throws IllegalArgumentException when urlPrefix string is malformed.
+   */
+  static String signUrlWithPrefix(String requestUrl,
+      String urlPrefix,
+      byte[] key,
+      String keyName,
+      Date expirationTime)
```
Sita04, line 100: Kindly move the parameters to a single line and adding line breaks only if they reach the character limit.

cdn/signed-urls/src/main/java/com/google/cdn/SignedUrls.java
Sita04: Example: https://github.com/GoogleCloudPlatform/java-docs-samples/blob/main/dlp/snippets/src/main/java/dlp/snippets/DeIdenitfyWithDeterministicEncryption.java#L42-L58

Sita04 (changes requested):

cdn/signed-urls/src/main/java/com/google/cdn/SignedUrls.java
```diff
@@ -75,6 +76,62 @@ public static String getSignature(byte[] privateKey, String input)
...
+      throws InvalidKeyException, NoSuchAlgorithmException {
+
+    final String algorithm = "HmacSHA1";
+    final int offset = 0;
+    Key key = new SecretKeySpec(privateKey, offset, privateKey.length, algorithm);
+    Mac mac = Mac.getInstance(algorithm);
+    mac.init(key);
+    return Base64.getUrlEncoder()
+        .encodeToString(mac.doFinal(input.getBytes(StandardCharsets.UTF_8)));
+  }
```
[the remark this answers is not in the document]
Sita04: You can add multiple region tags for a method. But the guideline is to have one sample per file. So, you would need to copy over the method to the new file.

Sita04:
Thanks for the contribution! 

Kindly also address the Inclusive language warnings.

cdn/signed-urls/src/main/java/com/google/cdn/SignedUrls.java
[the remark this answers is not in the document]
zhumin8: Got it. That's done in the recent update.

cdn/signed-urls/src/main/java/com/google/cdn/SignedUrls.java
Sita04: Since we only need to check if the urlPrefix does not start with "http://", there is no need to explicitly check for "https://" as well.
The second condition will be captured as part of the first condition.

cdn/signed-urls/src/main/java/com/google/cdn/SignedUrlWithPrefix.java
```diff
@@ -0,0 +1,107 @@
...
+    String keyPath = "/path/to/key";
+    // The date that the signed URL expires
+    Date expirationTime = getTomorrow();
+    // URL of request
+    String requestUrl = "https://media.example.com/videos/id/main.m3u8?userID=abc123&starting_profile=1";
+    // URL prefix to sign as a string. urlPrefix must start with either http:// or
+    // https:// and should not include query parameters
+    String urlPrefix = "https://media.example.com/videos/";
+
+    //read the key as a base 64 url-safe encoded string, then convert to byte array
```
Sita04, line 51:
```suggestion
    // Read the key as a base64 url-safe encoded string, then convert to byte array
```

cdn/signed-urls/src/main/java/com/google/cdn/SignedUrlWithPrefix.java
```diff
@@ -0,0 +1,107 @@
...
+
+    final String algorithm = "HmacSHA1";
+    final int offset = 0;
+    Key key = new SecretKeySpec(privateKey, offset, privateKey.length, algorithm);
+    Mac mac = Mac.getInstance(algorithm);
+    mac.init(key);
+    return Base64.getUrlEncoder()
+        .encodeToString(mac.doFinal(input.getBytes(StandardCharsets.UTF_8)));
+  }
+  // [END cloudcdn_sign_url_prefix]
```
Sita04, line 106: Move to line 108

cdn/signed-urls/src/test/java/com/google/cdn/SignedUrlsTest.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright 2018 Google Inc.
+ * Copyright 2018 Google LLC
```
Sita04, line 2: Kindly revert this change. Headers for existing code shouldn't be modified.

cdn/signed-urls/src/test/java/com/google/cdn/SignedUrlWithPrefixTest.java
```diff
@@ -0,0 +1,57 @@
...
+
+import static com.google.cdn.SignedUrlWithPrefix.signUrlWithPrefix;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertThrows;
+
+import java.util.Base64;
+import java.util.Date;
+import org.junit.Test;
+
+public class SignedUrlWithPrefixTest {
```
Sita04, line 27:
```suggestion
@RunWith(JUnit4.class)
public class SignedUrlWithPrefixTest {
```

Sita04 (changes requested):
Thanks for the changes. Few minor nits.

cdn/signed-urls/src/test/java/com/google/cdn/SignedUrlWithPrefixTest.java
```diff
@@ -0,0 +1,60 @@
...
+    assertThrows(IllegalArgumentException.class,
+        () -> {
+          signUrlWithPrefix(REQUEST_URL, INVALID_URL_PREFIX_1, KEY_BYTES, KEY_NAME, EXPIRATION);
+        });
+    assertThrows(IllegalArgumentException.class,
+        () -> {
+          signUrlWithPrefix(REQUEST_URL, INVALID_URL_PREFIX_2, KEY_BYTES, KEY_NAME, EXPIRATION);
+        });
+  }
+}
```
Sita04, line 60: Kindly add a new line at the end of the file.

Sita04 (approved):
LGTM. Thanks.
Kindly fix the 2 open comments before merging.

cdn/signed-urls/src/main/java/com/google/cdn/SignedUrls.java
zhumin8: > The second condition will be captured as part of the first condition.

I disagree, both condition are needed. With only the first condition,  `https://` will be falsely identified as invalid. Simply try`assertFalse("https://".startsWith("http://"));`
Unless you suggest to capture invalid cases with _"not starting with `http`"_, which still is not the same as saying "only starting with `http://` and `https://` are valid"

Sita04: Apologies for the confusion. The suggestion was to make it`(!urlPrefix.startsWith("http")` which would capture both the conditions.
Unless you have to explicitly put both conditions in there, might be okay to change it.

Sita04: Actually on a second thought, I think it's better to demonstrate 2 conditions instead of one. That way the customer knows the URL can start with either "http" or "https".
By simplifying the condition, we'll be only checking if it starts with "http".

Hence, resolving this comment. Thank you.

msampathkumar:
Nice work, @zhumin8 @Sita04. LGTM but let me include a TW, to look for any improvements that can be made to the code comments.

@upeeth - PTAL at this request.

upeeth:
Please rephrase as follows the comments used in the sample published at: https://cloud.google.com/cdn/docs/using-signed-urls#go_1 .

// - urlPrefix must start with "https://" and must not include query parameters.
// - key must be in raw form (not base64-encoded) and up to 16 bytes.
// - keyName must match a key added to the backend service or bucket.

zhumin8:
@upeeth 
Regarding the comment style and wording:
- Updated to include key requirement in comment https://github.com/GoogleCloudPlatform/java-docs-samples/pull/8103/commits/2cf9d00222f4802b31dbc73e6eb230e11a33047a
- I don't think Go sample quoted in your comment is a good candidate to follow on wording. The `urlPrefix must start with "https://" ` conflicts with documentation page highlighted [here](https://cloud.google.com/cdn/docs/using-signed-urls#url-prefix:~:text=A%20URLPrefix%20encodes%20a%20scheme%20(either%20http%3A//%20or%20https%3A//)%2C) 
- The current comment style is updated to follow guide and example from @Sita04 (see comment [here](https://github.com/GoogleCloudPlatform/java-docs-samples/pull/8103#discussion_r1196825659)) and in the spirit of avoiding javadoc comments as stated in [SAMPLE_FORMAT.md](https://github.com/GoogleCloudPlatform/java-docs-samples/pull/8103#discussion_r1196825659). So the comments that corresponds to the Go sample given in your comment, is within main method [here](https://github.com/GoogleCloudPlatform/java-docs-samples/blob/2cf9d00222f4802b31dbc73e6eb230e11a33047a/cdn/signed-urls/src/main/java/com/google/cdn/SignedUrlWithPrefix.java#L38-L49). Let me know if this looks okay or the style is conflicting with any wider guidelines we are trying to follow?

upeeth:
Please compare my suggestions with the comments. The changes suggested are purely editorial and not technical. Use "must" instead of "should" etc.

Shabirmean:
> Please compare my suggestions with the comments. The changes suggested are purely editorial and not technical. Use "must" instead of "should" etc.

@zhumin8 - I think we can go ahead with the PR if you can fix the `should` -> `must` suggestion!. Let me know once it's done.

zhumin8:
@upeeth Ah, thanks for the clarification. Updated in https://github.com/GoogleCloudPlatform/java-docs-samples/pull/8103/commits/b82a1f74cd3481a1cefb20e82234cb9f800d1e51.
@Shabirmean Done. PTAL

Shabirmean (approved):

PR #8130, opened 2023-05-18
chore: group SLF4J dependency updates

kweinmeister:
This will group 3+ SLF4J dependency updates that have a matching version number.


gericdong (approved):
LGTM

PR #8160, opened 2023-05-22
Remove gaeinfo project. We no longer support flex compat applications. 

kappratiksha:
Delete gaeinfo projects from flex samples as we no longer support them.


jinglundong (approved):
Thanks for cleaning this up! Flex compat was deprecated and turned down in late 2019, early 2020.

PR #8165, opened 2023-05-23
cleanup: add e2e test for appengine flex (Not ready)

kappratiksha:
Draft PR to trigger kokoro pipeline


.kokoro/tests/run_test_java.sh
```diff
@@ -60,6 +60,23 @@ if [[ "$file" == *"functions/helloworld/"* ]]; then
...
+    source "$SCRIPT_DIR"/build_appengine_flex.sh
+    EXIT=$?
+
+    if [[ $EXIT -ne 0 ]]; then
+        RTN=1
+        echo -e "\n Appengine Flex build/deploy failed: gcloud returned a non-zero exit code. \n"
+    else
+        echo -e "\n Appengine Flex build/deploy completed.\n"
+
+        # Wait for functions to warm up (and start detecting events)
```
jinglundong, line 74: functions -> app?

.kokoro/tests/run_test_java.sh
```diff
@@ -60,6 +60,23 @@ if [[ "$file" == *"functions/helloworld/"* ]]; then
...
+        sleep 1m
+    fi
+fi
+
+
 # Use maven to execute the tests for the project.
```
jinglundong, line 80: What does it test or verify exactly?

jinglundong (approved):
LGTM over all. I think it's quite reasonable to follow the pattern being set in this repo and only cover the fundamental use cases. 

I think you also need to rebase this PR to avoid some of those deleted files.

.kokoro/tests/run_test_java.sh
```diff
@@ -60,6 +60,23 @@ if [[ "$file" == *"functions/helloworld/"* ]]; then
     fi
 fi
 
+# Build and deploy Appengine flex samples
+if [[ "$file" == *"flexible/java-11/"* ]]; then
```
averikitsch, line 64: Can we deploy everything under `flexible/`?

.kokoro/tests/run_test_java.sh
```diff
@@ -60,6 +60,23 @@ if [[ "$file" == *"functions/helloworld/"* ]]; then
...
+    EXIT=$?
+
+    if [[ $EXIT -ne 0 ]]; then
+        RTN=1
+        echo -e "\n Appengine Flex build/deploy failed: gcloud returned a non-zero exit code. \n"
+    else
+        echo -e "\n Appengine Flex build/deploy completed.\n"
+
+        # Wait for functions to warm up (and start detecting events)
+        sleep 1m
```
averikitsch, line 75: I'm not sure we need this wait. Can we try it without?

.kokoro/tests/run_test_java.sh
averikitsch: This is not my preferred approach, since we are not testing requests against the service deployed. I asked for the service to be deployed as a part of the Java tests in order to make this pattern feasible. But at minimum we will know samples do not break during build/deployment.

.kokoro/tests/build_appengine_flex.sh
```diff
@@ -0,0 +1,42 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -eo pipefail
+
+file="$(pwd)"
+APP_REGION="us-central1"
+export SAMPLE_VERSION="${KOKORO_GIT_COMMIT:-latest}"
+# Builds not triggered by a PR will fall back to the commit hash then "latest".
+SUFFIX=${KOKORO_GITHUB_PULL_REQUEST_NUMBER:-${SAMPLE_VERSION:0:12}}-$(date +%s%N)
```
averikitsch, line 23: I would be great to use this SUFFIX or SERVICE_NAME to deploy a nondefault version of the app and use that to make requests against

.kokoro/tests/build_appengine_flex.sh
```diff
@@ -0,0 +1,42 @@
...
+export SERVICE_NAME="${SAMPLE_NAME}-${SUFFIX}"
+
+set -x
+
+if [[ "$file" == *"java-11/hello"* ]]; then
```
averikitsch, line 28: Please deploy all apps

averikitsch (changes requested):

.kokoro/tests/run_test_java.sh
kweinmeister: +1 (and to the similar comment [here](https://github.com/GoogleCloudPlatform/java-docs-samples/pull/8165#discussion_r1203155868)).

If there's a reason we have to limit the scope, these filters should be a constant at the top of each file. Also a comment about why we selected these specific folders to be tested should be included.

bigtable/beam/change-streams/pom.xml
```diff
@@ -102,4 +102,4 @@
       <artifactId>gax</artifactId>
     </dependency>
   </dependencies>
-</project>
+</project>
```
kweinmeister, line 105: Nit: consider reverting this change (GitHub flags missing new lines at the end of files). Also present in the `kotlin-ktor` `pom.xml` change. [Background info](https://stackoverflow.com/questions/729692/why-should-text-files-end-with-a-newline)

PR #8170, opened 2023-05-23
fixit: create new Java sample for aiplatform_create_custom_job_sample

gericdong:
## Description

Fixes #281693980

Added missing Java sample for aiplatform_create_custom_job_sample

## Checklist

### DIFF
--- a/aiplatform/src/main/java/aiplatform/CreateCustomJobSample.java
+++ b/aiplatform/src/main/java/aiplatform/CreateCustomJobSample.java
@@ -1,0 +1,86 @@
+/*
+ * Copyright 2023 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package aiplatform;
+
+// [START aiplatform_create_custom_job_sample]
+import com.google.cloud.aiplatform.v1.AcceleratorType;
+import com.google.cloud.aiplatform.v1.ContainerSpec;
+import com.google.cloud.aiplatform.v1.CustomJob;
+import com.google.cloud.aiplatform.v1.CustomJobSpec;
+import com.google.cloud.aiplatform.v1.JobServiceClient;
+import com.google.cloud.aiplatform.v1.JobServiceSettings;
+import com.google.cloud.aiplatform.v1.LocationName;
+import com.google.cloud.aiplatform.v1.MachineSpec;
+import com.google.cloud.aiplatform.v1.WorkerPoolSpec;
+import java.io.IOException;
+
+public class CreateCustomJobSample {
+
+  public static void main(String[] args) throws IOException {
+    // TODO(developer): Replace these variables before running the sample.
+    String project = "PROJECT";
+    String displayName = "DISPLAY_NAME";
+    String containerImageUri = "CONTAINER_IMAGE_URI";
+    createCustomJobSample(project, displayName, containerImageUri);
+  }
+
+  static void createCustomJobSample(String project, String displayName, String containerImageUri)
+      throws IOException {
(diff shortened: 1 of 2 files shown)


aiplatform/src/main/java/aiplatform/CreateCustomJobSample.java
```diff
@@ -0,0 +1,86 @@
...
+      throws IOException {
+    JobServiceSettings settings =
+        JobServiceSettings.newBuilder()
+            .setEndpoint("us-central1-aiplatform.googleapis.com:443")
+            .build();
+    String location = "us-central1";
+
+    // Initialize client that will be used to send requests. This client only needs to be created
+    // once, and can be reused for multiple requests. After completing all of your requests, call
+    // the "close" method on the client to safely clean up any remaining background resources.
```
Shabirmean, line 51: Since we are using Try with resources here; we do not need the statement about `close`. Once the try block is finished the close will be called automatically.  So we can remove the last line of comment here

aiplatform/src/main/java/aiplatform/CreateCustomJobSample.java
```diff
@@ -0,0 +1,86 @@
...
+import com.google.cloud.aiplatform.v1.WorkerPoolSpec;
+import java.io.IOException;
+
+public class CreateCustomJobSample {
+
+  public static void main(String[] args) throws IOException {
+    // TODO(developer): Replace these variables before running the sample.
+    String project = "PROJECT";
+    String displayName = "DISPLAY_NAME";
+    String containerImageUri = "CONTAINER_IMAGE_URI";
```
Shabirmean, line 37: Cn we add a little more description for what this container image is? Also if possible provide a link to an example so the developer can quickly get going with testing this sample instead of having to generate their own image?

aiplatform/src/main/java/aiplatform/CreateCustomJobSample.java
```diff
@@ -0,0 +1,86 @@
...
+import com.google.cloud.aiplatform.v1.CustomJob;
+import com.google.cloud.aiplatform.v1.CustomJobSpec;
+import com.google.cloud.aiplatform.v1.JobServiceClient;
+import com.google.cloud.aiplatform.v1.JobServiceSettings;
+import com.google.cloud.aiplatform.v1.LocationName;
+import com.google.cloud.aiplatform.v1.MachineSpec;
+import com.google.cloud.aiplatform.v1.WorkerPoolSpec;
+import java.io.IOException;
+
+public class CreateCustomJobSample {
```
Shabirmean, line 31: Can we add a simple one/two line comment to explain what this sample is for?

(something like)
```java
Run your custom machine learning training job in VertexAI

```

aiplatform/src/test/java/aiplatform/CreateCustomJobSampleTest.java
```diff
@@ -0,0 +1,108 @@
...
+import org.junit.Before;
+import org.junit.BeforeClass;
+import org.junit.Rule;
+import org.junit.Test;
+
+public class CreateCustomJobSampleTest {
+
+  @Rule public final MultipleAttemptsRule multipleAttemptsRule = new MultipleAttemptsRule(5);
+
+  private static final String PROJECT = System.getenv("UCAIP_PROJECT_ID");
```
Shabirmean, line 43: (nit) since you are using `UCAIP_PROJECT_ID` in more than one place, assign it to a variable and re-use that variable

Shabirmean (approved):
LGTM!
Please respond to comments before we can merge 🙌🏼

aiplatform/src/main/java/aiplatform/CreateCustomJobSample.java
gericdong: Changed

aiplatform/src/main/java/aiplatform/CreateCustomJobSample.java
gericdong: Comment added

aiplatform/src/main/java/aiplatform/CreateCustomJobSample.java
gericdong: Comment added

aiplatform/src/test/java/aiplatform/CreateCustomJobSampleTest.java
gericdong: It makes sense but all of our tests (70+) are using the same pattern so for consistency, I am hesitant to make change for this test. I am thinking if it may be better to keep it as it is for now and change in a batch in the future if it's necessary.

PR #8191, opened 2023-05-25
Use 'else if' to avoid re-classifying an interaction.

faube:
The previous code showed the SCORE_LESS_THAN_THRESHOLD message for invalid tokens.

This matches Javascript and Python samples


Shabirmean:
@Sita04 - PTAL and merge

PR #8194, opened 2023-05-25
fix: update to remove region tag conflict

iennae:
related: b/281695703

identified that there are 2 samples in this repository with the same region tag 
https://github.com/GoogleCloudPlatform/java-docs-samples/blob/4eccabd9f0dc1f783edf5caa1841fc3066650f16/dialogflow/basic-webhook/src/main/java/com/example/dialogflow/cx/BasicWebhook.java#L21-L75
https://github.com/GoogleCloudPlatform/java-docs-samples/blob/234aa56b1398a24584234a619e1ac51fe7f2d29f/dialogflow/snippets/src/main/dialogflow/Example.java#L19-L72

Adding the _es_ for the one that works with diagflow es

### DIFF
--- a/dialogflow/snippets/src/main/dialogflow/Example.java
+++ b/dialogflow/snippets/src/main/dialogflow/Example.java
@@ -17,6 +17,7 @@
 package dialogflow;
 
 // [START dialogflow_es_webhook]
+// [START dialogflow_webhook]
 
 // TODO: add GSON dependency to Pom file
 // (https://mvnrepository.com/artifact/com.google.code.gson/gson/2.8.5)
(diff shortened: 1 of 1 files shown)


averikitsch (approved):

dialogflow/snippets/src/main/dialogflow/Example.java
```diff
@@ -16,6 +16,7 @@
 
 package dialogflow;
 
+// [START dialogflow_es_webhook]
```
Shabirmean, line 19: Should we move this below the other (`dialogflow_webhook`) region tag so that that is not visible with only the `START` and not the `END` when included in a sample?

averikitsch (approved):

PR #8203, opened 2023-05-30
chore: delete duplicate jobs/v4 samples

### LINKED ISSUE #7872: Duplicate samples: jobs/v4 and talent/snippets
These samples appear to be duplicated. We should delete one of these directories.

In a quick search, it looks like jobs/v4 is used in the docs, so talent/snippets can be deleted. This should be verified.

kweinmeister:
Deletes `/jobs/v4` samples, which are duplicated in `/talent`. Talent Solution is the current name of the product, and doc references for API v4 current point to `/talent`.

In the future, we can consider consolidating `/jobs/v3` samples in `/talent`. That will require updating the directory structure to handle both v3 and v4, and updating docs references for `/jobs/v3` samples.

Fixes #7872


kweinmeister:
Closing this PR to instead delete `/talent`, to keep v3 and v4 samples grouped together.

PR #8267, opened 2023-06-16
Feature/fetch acl sample

chasinandrew:
## Description

Fixes #<ISSUE-NUMBER>

## Checklist


chasinandrew:
Looks like the project connected to the unit test has universal access, so no ACL will be fetched. Can we change the project that runs in the CI build to one that has ACL?

chasinandrew:
@averikitsch @telpirion @holtskinner would any of you be able to help review this? Also is it possible to change the project ID in the CI pipeline? The one currently there does not have any ACL so the test is failing.

content-warehouse/src/main/java/contentwarehouse/v1/FetchAcl.java
```diff
@@ -0,0 +1,95 @@
...
+  private static String getProjectNumber(String projectId) throws IOException { 
+    try (ProjectsClient projectsClient = ProjectsClient.create()) { 
+      ProjectName projectName = ProjectName.of(projectId); 
+      Project project = projectsClient.getProject(projectName);
+      String projectNumber = project.getName(); // Format returned is projects/xxxxxx
+      return projectNumber.substring(projectNumber.lastIndexOf("/") + 1);
+    } 
+  }
+}
+// [END contentwarehouse_fetch_acl]
```
Sita04, line 95: Please add a new line.

content-warehouse/src/main/java/contentwarehouse/v1/FetchAcl.java
```diff
@@ -0,0 +1,95 @@
...
+import com.google.cloud.contentwarehouse.v1.RequestMetadata;
+import com.google.cloud.contentwarehouse.v1.UserInfo;
+import com.google.cloud.resourcemanager.v3.Project;
+import com.google.cloud.resourcemanager.v3.ProjectName;
+import com.google.cloud.resourcemanager.v3.ProjectsClient;
+import java.io.IOException;
+
+
+public class FetchAcl {
+  public static void fetchAcl() throws IOException {
```
Sita04, line 34:
```suggestion
  public static void main(String[] args) throws IOException {
```
The format is to use a main method and then call the snippet.

content-warehouse/src/main/java/contentwarehouse/v1/FetchAcl.java
```diff
@@ -0,0 +1,95 @@
...
+    // TODO(developer): Replace these variables before running the sample.
+    String projectId = "your-project-id";
+    String location = "your-location"; // Format is "us" or "eu".
+    String userId = "your-user-id"; // Format is user:<user-id>
+    String documentId = "your-documentid";
```
Sita04, line 39: Can you kindly provide more details about these variables and possibly a link to docs?

content-warehouse/src/main/java/contentwarehouse/v1/FetchAcl.java
```diff
@@ -0,0 +1,95 @@
...
+    fetchAcl(projectId, location, userId, documentId);
+  }
+
+  public static void fetchAcl(String projectId, String location, String userId, String documentId)
```
Sita04, line 43: Please add a method comment in line 42, explaining the goal of this method.

content-warehouse/src/main/java/contentwarehouse/v1/FetchAcl.java
```diff
@@ -0,0 +1,95 @@
...
+
+  public static void fetchAcl(String projectId, String location, String userId, String documentId) 
+      throws IOException {
+    String projectNumber = getProjectNumber(projectId); 
+    String endpoint = String.format("%s-contentwarehouse.googleapis.com:443", location);
+
+    DocumentServiceSettings documentServiceSettings = 
+        DocumentServiceSettings.newBuilder().setEndpoint(endpoint).build();
+        
+    try (DocumentServiceClient documentServiceClient =
```
Sita04, line 51: This comment should be added before the client initialization.
```suggestion
    // Initialize client that will be used to send requests. This client only needs to be
    // created once, and can be reused for multiple requests.
    try (DocumentServiceClient documentServiceClient = 
```

content-warehouse/src/main/java/contentwarehouse/v1/FetchAcl.java
```diff
@@ -0,0 +1,95 @@
...
+                .setResource(projectNumber)
+                .setProjectOwner(true).build();
+        fetchAclResponse = documentServiceClient.fetchAcl(fetchAclRequest);
+      }
+      System.out.println(fetchAclResponse);
+    }
+  }
+
+  private static String getProjectNumber(String projectId) throws IOException { 
+    try (ProjectsClient projectsClient = ProjectsClient.create()) {
```
Sita04, line 87:
```suggestion
    // Initialize client that will be used to send requests. This client only needs to be
    // created once, and can be reused for multiple requests.
    try (ProjectsClient projectsClient = ProjectsClient.create()) { 
```

content-warehouse/src/main/java/contentwarehouse/v1/FetchAcl.java
```diff
@@ -0,0 +1,95 @@
...
+                .setResource(documentName.toString()).build(); 
+        fetchAclResponse = documentServiceClient.fetchAcl(fetchAclRequest);
+      } else {
+        FetchAclRequest fetchAclRequest = 
+            FetchAclRequest.newBuilder()
+                .setRequestMetadata(requestMetadata)
+                .setResource(projectNumber)
+                .setProjectOwner(true).build();
+        fetchAclResponse = documentServiceClient.fetchAcl(fetchAclRequest);
+      }
```
Sita04, line 81: Instead of sending FetchAclRequest in the if else condition, wdyt about using the condition to create a new resource string and then outside the block, create the request?

Something like, 

```
String resource = "";

if (condition) {
 resource = documentName.toString();
} else {
 resource = projectNumber
}

.setResource(resource);
```

content-warehouse/src/test/java/contentwarehouse/v1/FetchAclTest.java
```diff
@@ -0,0 +1,80 @@
...
+import java.io.IOException;
+import java.io.PrintStream;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.TimeoutException;
+import org.junit.After;
+import org.junit.Before;
+import org.junit.Rule;
+import org.junit.Test;
+ 
+public class FetchAclTest {
```
Sita04, line 33:
```suggestion
@RunWith(JUnit4.class)
public class FetchAclTest {
```

Sita04 (changes requested):
Thanks for the sample!
PTAL at the comments and also the test failures.

content-warehouse/src/main/java/contentwarehouse/v1/FetchAcl.java
chasinandrew: I'm not sure that this would work due to .setProjectOwner(true), it is only applicable when the ACL request is for a project. It would throw an error if it's set on the document scope.

content-warehouse/src/main/java/contentwarehouse/v1/FetchAcl.java
chasinandrew: Done, please let me know if my description is sufficient

chasinandrew:
Pull request below will unblock this PR once merged. 
https://github.com/GoogleCloudPlatform/java-docs-samples/pull/8461

chasinandrew:
Closing this PR as this sample is deprioritized

PR #8363, opened 2023-06-27
feat: Proto Columns samples for Spanner JDBC

harshachinta:
Adds samples for the new Proto Columns (Proto Messages and Proto Enum) data type for Spanner JDBC.

Do not merge yet: Depends on changes in the Spanner JDBC driver that have not yet been included in the libraries bom.


Sita04:
@harshachinta Pinging to see if there's an update in this PR?

PR #8380, opened 2023-06-30
fix: cleanup testMonitorCertificateAuthority() test

### LINKED ISSUE #8320: Monitoring policies need clean up
* `privateca/snippets/src/main/java/privateca/MonitorCertificateAuthority.java` tested in `privateca/snippets/src/test/java/privateca/SnippetsIT.java` creates a monitor policy: "policy-name". 
* Many tests in `monitoring/v3/src/test/java/com/example/monitoring/*IT.java` create policies: "alert_policy_name_*" and"test-policy*"

Expectations:
* Tests clean up these resources
* Test project is cleanup manually of these test resources

This is blocking current tests.

minherz:
### Description

Performs cleanup for the `testMonitorCertificateAuthority()` by deleting an alert policy that is created by `createCaMonitoringPolicy()`.

Fixes #8320

### DIFF
--- a/privateca/snippets/src/test/java/privateca/SnippetsIT.java
+++ b/privateca/snippets/src/test/java/privateca/SnippetsIT.java
@@ -19,6 +19,7 @@ package privateca;
 import static com.google.common.truth.Truth.assertThat;
 import static com.google.common.truth.Truth.assertWithMessage;
 
+import com.google.cloud.monitoring.v3.AlertPolicyServiceClient;
 import com.google.cloud.security.privateca.v1.CaPool.IssuancePolicy;
 import com.google.cloud.security.privateca.v1.CaPoolName;
 import com.google.cloud.security.privateca.v1.Certificate;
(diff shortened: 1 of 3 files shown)


privateca/snippets/src/test/java/privateca/SnippetsIT.java
```diff
@@ -346,6 +347,15 @@ public void testUpdateCertificateAuthority()
   public void testMonitorCertificateAuthority() throws IOException, InterruptedException {
     privateca.MonitorCertificateAuthority.createCaMonitoringPolicy(PROJECT_ID);
     assertThat(stdOut.toString()).contains("Monitoring policy successfully created !");
+
+    // cleanup created policy
```
kweinmeister, line 351: Would this fit best in [`cleanUp()`](https://github.com/GoogleCloudPlatform/java-docs-samples/blob/34e41ee46a1647c19ee7556d0a6e6fa4fb74c571/privateca/snippets/src/test/java/privateca/SnippetsIT.java#L185)?

minherz: It cannot because it is contextual for this particular test

kweinmeister: I see in other tests that we create resources in `setUp()` and then maintain a reference to these resources in an instance variable to be deleted later. See the use of `queue` [here](https://github.com/GoogleCloudPlatform/java-docs-samples/blob/e97ca11ce87ffd28f3336c8fb675afb423baa6ff/tasks/snippets/src/test/java/com/example/task/ListQueuesIT.java#L44) or `CA_poolid` [here](https://github.com/GoogleCloudPlatform/java-docs-samples/blob/e97ca11ce87ffd28f3336c8fb675afb423baa6ff/privateca/snippets/src/test/java/privateca/SnippetsIT.java#L66).

I wonder if this test should actually be pulled into the setup, with the new cleanup code moved into teardown. It works as-is, this is just something to consider.

minherz: I think the use case is different. [SnippetIT][1] has many tests with only one that makes use of the specific example. There is no need to create such complex code to run it only once IMHO.

We can always refactor the solution when we will have more places that need to cleanup specific resources.

[1]: https://github.com/GoogleCloudPlatform/java-docs-samples/pull/8380/files#diff-995861e62924ab2481c923f27a70073a4d0284e26de93b111e7c966268b011dd

kweinmeister (approved):

PR #8385, opened 2023-07-01
fix: update batch region to address quota alerts

kweinmeister:
Change region to `us-central1` to address CPU quota alerts.

Also, addressed code warnings that added `com.example.batch` package and disposed of resources.


rsamborski (approved):

rsamborski:
@Sita04 LGTM from my side, but please take another look before merging.

Sita04 (approved):

PR #8438, opened 2023-07-17
fix: remove deprecated method from bigtable change-streams sample

kweinmeister:
The `withHeartbeatDuration()` method was [removed](https://github.com/apache/beam/pull/27249) in the 2.49.0 Beam release, causing this sample to no longer compile with 2.49.0. The sample will now use the default value of 5 seconds.


billyjacobson (approved):

bourgeoisor (approved):
LGTM

PR #8480, opened 2023-07-28
docs(samples): Add error handling for BigQueryIO streaming example

VeronicaWasson:
Adds error handling to show the [dead-letter pattern](https://beam.apache.org/documentation/patterns/bigqueryio/#bigqueryio-deadletter-pattern).

Also removes the call to withNumStorageWriteApiStreams, because this setting now [defaults to autosharding](https://beam.apache.org/documentation/io/built-in/google-bigquery/#exactly-once-semantics), so there is no need to explicitly set a value.


lesv (approved):
I like to see these kinds of samples - LGTM

PR #8494, opened 2023-08-03
feat: add assets (for slates) and pools samples for Live Stream

irataxy:
b:264616218

Please **merge** this PR for me once it is approved

### DIFF
--- a/media/livestream/src/main/java/com/example/livestream/CreateAsset.java
+++ b/media/livestream/src/main/java/com/example/livestream/CreateAsset.java
@@ -45,24 +45,24 @@ public class CreateAsset {
     // Initialize client that will be used to send requests. This client only needs to be created
     // once, and can be reused for multiple requests. After completing all of your requests, call
     // the "close" method on the client to safely clean up any remaining background resources.
-    try (LivestreamServiceClient livestreamServiceClient = LivestreamServiceClient.create()) {
-      var createAssetRequest =
-          CreateAssetRequest.newBuilder()
-              .setParent(LocationName.of(projectId, location).toString())
-              .setAssetId(assetId)
-              .setAsset(
-                  Asset.newBuilder()
-                      .setVideo(
-                          VideoAsset.newBuilder()
-                              .setUri(assetUri)
-                              .build())
-                      .build())
-              .build();
-      // First API call in a project can take up to 15 minutes.
-      Asset result =
-          livestreamServiceClient.createAssetAsync(createAssetRequest).get(15, TimeUnit.MINUTES);
-      System.out.println("Asset: " + result.getName());
-    }
+    LivestreamServiceClient livestreamServiceClient = LivestreamServiceClient.create();
+    var createAssetRequest =
+        CreateAssetRequest.newBuilder()
+            .setParent(LocationName.of(projectId, location).toString())
+            .setAssetId(assetId)
+            .setAsset(
+                Asset.newBuilder()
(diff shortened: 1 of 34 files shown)


media/livestream/src/main/java/com/example/livestream/CreateAsset.java
```diff
@@ -0,0 +1,68 @@
...
+
+    createAsset(projectId, location, assetId, assetUri);
+  }
+
+  public static void createAsset(String projectId, String location, String assetId, String assetUri)
+      throws InterruptedException, ExecutionException, TimeoutException, IOException {
+    // Initialize client that will be used to send requests. This client only needs to be created
+    // once, and can be reused for multiple requests. After completing all of your requests, call
+    // the "close" method on the client to safely clean up any remaining background resources.
+    try (LivestreamServiceClient livestreamServiceClient = LivestreamServiceClient.create()) {
```
lesv, line 48: We tend to recommend try with resources (which your using) for short operations, but here, where it's a long running operation, why are you doing it?  Especially when you talk about calling close at a later time.

media/livestream/src/main/java/com/example/livestream/DeleteAsset.java
```diff
@@ -0,0 +1,56 @@
...
+
+    deleteAsset(projectId, location, assetId);
+  }
+
+  public static void deleteAsset(String projectId, String location, String assetId)
+      throws InterruptedException, ExecutionException, TimeoutException, IOException {
+    // Initialize client that will be used to send requests. This client only needs to be created
+    // once, and can be reused for multiple requests. After completing all of your requests, call
+    // the "close" method on the client to safely clean up any remaining background resources.
+    try (LivestreamServiceClient livestreamServiceClient = LivestreamServiceClient.create()) {
```
lesv, line 45: Again, your calling try w/ resources which will automatically close the client.  You might wish to be a bit clearer with your comments.

lesv (changes requested):
Generally LGTM, but for a long running API like this, I'm not sure using try w/ resources (which will automatically close the client) is the right thing to be demonstrating especially when we also talk about calling close() at some point.  This is throughout the code, not just where I commented.  I would argue this might be a time to break with our standard ways of doing things.  That said, these comments are advisory only, consider them, if you really want to go this way, then at least consider changing the comments, but I won't object.

media/livestream/src/main/java/com/example/livestream/CreateAsset.java
irataxy: Removed try block for LRO API calls.

media/livestream/src/main/java/com/example/livestream/DeleteAsset.java
irataxy: Removed try and added explicit close() calls for LROs.

lesv (approved):
@irataxy LGTM - thanks for making those changes.  How long do the tests take to run?
I'm not going to merge, but ask @averikitsch or @kweinmeister to do so once they see your answer as they may need to update the timeouts for nightly | period testing, as well as for updates that touch a lot of files.

irataxy:
It runs locally in under 10 minutes. The only caveat here is that it will take 2-3x that time for the first run in a new project/location. This shouldn't be a concern for the existing test projects that have already run the tests.

lesv:
Ok - then I'll merge it.  Thanks!

PR #8648, opened 2023-09-15
fix: refactor error reporting code sample

minherz:
align code sample for error reporting with code in other languages (b/275730159):
- demonstrate the use of Error Reporting for reporting exceptions
- use the "fixed" error message
- delegate error context processing to Error Reporting backend

### DIFF
--- a/errorreporting/src/main/java/com/example/errorreporting/QuickStart.java
+++ b/errorreporting/src/main/java/com/example/errorreporting/QuickStart.java
@@ -27,16 +27,10 @@ import java.io.PrintWriter;
 import java.io.StringWriter;
 
 /**
- * Snippet demonstrates using the Error Reporting API to report a custom error
- * event.
- *
+ * Snippet demonstrates using the Error Reporting API to report an exception.
  * <p>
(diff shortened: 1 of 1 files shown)


errorreporting/src/main/java/com/example/errorreporting/QuickStart.java
```diff
@@ -18,52 +18,69 @@
...
- * Snippet demonstrates using the Error Reporting API to report a custom error event.
+ * Snippet demonstrates using the Error Reporting API to report a custom error
+ * event.
  *
- * <p>This library is not required on App Engine, errors written to stderr are automatically written
- * to Error Reporting. It is also not required if you are writing logs to Cloud Logging. Errors
- * written to Cloud Logging that contain an exception or stack trace are automatically written out
+ * <p>
+ * This library is not required on App Engine, errors written to stderr are
+ * automatically written
```
kweinmeister, line 35: Nit: we addressed the long lines, but I think we created extra short lines in the process. Is there a way to make them more balanced?

kweinmeister (approved):

PR #8747, opened 2023-10-11
feat: add sample and test for getting an access token from an imperso…

irataxy:
…nated SA

## Description

Fixes b:302748677

### DIFF
--- a/auth/src/test/java/com/google/cloud/auth/samples/AccessTokenFromImpersonatedCredentialsIT.java
+++ b/auth/src/test/java/com/google/cloud/auth/samples/AccessTokenFromImpersonatedCredentialsIT.java
@@ -31,7 +31,7 @@ import org.junit.runners.JUnit4;
 //CHECKSTYLE OFF: AbbreviationAsWordInName
 public class AccessTokenFromImpersonatedCredentialsIT {
   //CHECKSTYLE ON: AbbreviationAsWordInName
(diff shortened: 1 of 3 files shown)


auth/src/test/java/com/google/cloud/auth/samples/AccessTokenFromImpersonatedCredentialsIT.java
```diff
@@ -0,0 +1,57 @@
...
+
+  @Before
+  public void setUp() {
+    bout = new ByteArrayOutputStream();
+    out = new PrintStream(bout);
+    System.setOut(out);
+    credentials = System.getenv("GOOGLE_APPLICATION_CREDENTIALS");
+    assertNotNull(credentials);
+  }
+
```
Sita04, line 49: It's always good to include an @After method to cleanup sysout (and any resources if created)

Sita04 (changes requested):
Thanks for the sample. Please add a cleanup method as suggested in the comment.

auth/src/test/java/com/google/cloud/auth/samples/AccessTokenFromImpersonatedCredentialsIT.java
irataxy: Done

Sita04 (approved):

PR #8844, opened 2023-11-06
fix: update references to deprecated `hashed_account_id`

serargz:
The new recommended way to provide the user identifier is through the `userInfo.accountId` field in the assessment's `event`.

Change-Id: Icd7bdecb8a783e9a148db2f2b50d1df3d222347c

### DIFF
--- a/recaptcha_enterprise/snippets/src/main/java/recaptcha/account_defender/AccountDefenderAssessment.java
+++ b/recaptcha_enterprise/snippets/src/main/java/recaptcha/account_defender/AccountDefenderAssessment.java
@@ -27,6 +27,8 @@ import com.google.recaptchaenterprise.v1.Event;
 import com.google.recaptchaenterprise.v1.ProjectName;
 import com.google.recaptchaenterprise.v1.RiskAnalysis.ClassificationReason;
 import com.google.recaptchaenterprise.v1.TokenProperties;
+import com.google.recaptchaenterprise.v1.UserId;
+import com.google.recaptchaenterprise.v1.UserInfo;
 import java.io.IOException;
 import java.nio.charset.StandardCharsets;
 import java.security.InvalidKeyException;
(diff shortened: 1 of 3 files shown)


recaptcha_enterprise/snippets/src/main/java/recaptcha/account_defender/AccountDefenderAssessment.java
```diff
@@ -84,19 +79,26 @@ public static void accountDefenderAssessment(
...
-      ByteString hashedAccountId)
+      String accountId,
+      String phoneNumber,
+      String emailAddress)
       throws IOException {
     try (RecaptchaEnterpriseServiceClient client = RecaptchaEnterpriseServiceClient.create()) {
 
       // Set the properties of the event to be tracked.
-      Event event =
+      Event.Builder eventBuilder =
```
Sita04, line 89: Can we combine two separate property settings into one instance like:

Event event = Event.newBuilder()
                               .setSiteKey()
                               .setToken()
                               .setAcountId()
.....

recaptcha_enterprise/snippets/src/main/java/recaptcha/account_defender/AnnotateAccountDefenderAssessment.java
```diff
@@ -38,29 +38,29 @@ public static void main(String[] args) throws IOException, NoSuchAlgorithmExcept
...
   public static void annotateAssessment(
-      String projectID, String assessmentId, ByteString hashedAccountId) throws IOException {
+      String projectID, String assessmentId, String accountId) throws IOException {
 
     try (RecaptchaEnterpriseServiceClient client = RecaptchaEnterpriseServiceClient.create()) {
       // Build the annotation request.
       // For more info on when/how to annotate, see:
       // https://cloud.google.com/recaptcha-enterprise/docs/annotate-assessment#when_to_annotate
-      AnnotateAssessmentRequest annotateAssessmentRequest =
+      AnnotateAssessmentRequest.Builder annotateAssessmentRequest =
```
Sita04, line 58: .Builder can be removed here and the object can be directly constructed.

recaptcha_enterprise/snippets/src/main/java/recaptcha/account_defender/AccountDefenderAssessment.java
Sita04, line 1: Kindly update the corresponding test cases as well.

Sita04 (changes requested):

recaptcha_enterprise/snippets/src/main/java/recaptcha/account_defender/AccountDefenderAssessment.java
serargz: we can't because of the

PR #8863, opened 2023-11-09
feat: Add Makefile for constistent sample interface

muncus:
This is a work in progress.

The goal is to provide a consistent way for contributors to built, test, and
lint their samples, regardless of the language repo they are working in.
(Similar PRs are open in other language repos)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,34 @@
+# Makefile for running typical developer workflow actions.
+# To run actions in a subdirectory of the repo:
+#   make lint build dir=translate/snippets
+
+INTERFACE_ACTIONS="build test lint"
+
+.ONESHELL: #ease subdirectory work by using the same subshell for all commands
+.-PHONY: *
+
+# Default to current dir if not specified.
+dir ?= $(shell pwd)
+
+GOOGLE_CLOUD_PROJECT="${GOOGLE_SAMPLE_PROJECT}"
+
+build:
+	cd ${dir}
+	mvn compile
+
+test: check-env build
+	cd ${dir}
+	mvn verify
+
+lint:
+	cd ${dir}
+	mvn -P lint checkstyle:check
+
+check-env:
+ifndef GOOGLE_SAMPLE_PROJECT
+	$(error GOOGLE_SAMPLE_PROJECT environment variable is required to perform this action)
+endif
+
+list-actions:
+	@ echo ${INTERFACE_ACTIONS}
+
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,8 @@
 # Makefile for running typical developer workflow actions.
 # To run actions in a subdirectory of the repo:
 #   make lint build dir=translate/snippets
+# Note: testing requires Application Default Credentials.
+# For details about ADC, see https://cloud.google.com/docs/authentication/application-default-credentials
 
 INTERFACE_ACTIONS="build test lint"
 
--- a/Makefile
+++ b/Makefile
@@ -10,7 +10,7 @@ INTERFACE_ACTIONS="build test lint"
 # Default to current dir if not specified.
 dir ?= $(shell pwd)
 
-GOOGLE_CLOUD_PROJECT="${GOOGLE_SAMPLE_PROJECT}"
+export GOOGLE_CLOUD_PROJECT = ${GOOGLE_SAMPLES_PROJECT}
 
 build:
 	cd ${dir}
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -0,0 +1,39 @@
...
+	mvn compile
+
+test: check-env build
+	cd ${dir}
+	mvn --quiet --batch-mode --fail-at-end clean verify \
+    -Dfile.encoding="UTF-8" \
+    -Dorg.slf4j.simpleLogger.log.org.apache.maven.cli.transfer.Slf4jMavenTransferListener=warn \
+    -Dmaven.test.redirectTestOutputToFile=true \
+    -Dbigtable.projectID="${GOOGLE_CLOUD_PROJECT}" \
+    -Dbigtable.instanceID=instance
```
Sita04, line 26: Both these lines can be removed for local testing.

Makefile
```diff
@@ -0,0 +1,39 @@
...
+
+lint:
+	cd ${dir}
+	mvn -P lint checkstyle:check
+
+check-env:
```
Sita04, line 32: Suggesting to add a check for "mvn". If not, provide a link to the installation page?

Makefile
```diff
@@ -0,0 +1,39 @@
...
+ifndef GOOGLE_SAMPLES_PROJECT
```
Sita04, line 33: Currently, the tests in the repo requires another env variable "GOOGLE_APPLICATION_CREDENTIALS" pointing to the Service account JSON key for testing.

Makefile
```diff
@@ -0,0 +1,39 @@
...
+
+.ONESHELL: #ease subdirectory work by using the same subshell for all commands
+.-PHONY: *
+
+# Default to current dir if not specified.
+dir ?= $(shell pwd)
+
+export GOOGLE_CLOUD_PROJECT = ${GOOGLE_SAMPLES_PROJECT}
+
+build:
```
Sita04, line 15: The commands after "cd" is executed at the root of the project and not in the directory specified. Applies to all actions.

Makefile
Sita04, line 33: Can we use "GOOGLE_CLOUD_PROJECT" instead of this var?

Sita04 (changes requested):
Thanks Marc!

Makefile
muncus: That's normally the case with a Makefile, but the [`.ONESHELL` option](https://www.gnu.org/software/make/manual/html_node/One-Shell.html) above ensures all commands are run in the same shell, so the `cd` does take effect for all commands.

Makefile
muncus: this command comes from the one currently used in the CI testing.
I'd like to eventually move CI to use the Makefile, so i'd prefer to leave them to avoid issues later. Unless there's a reason these options might interfere with local testing?

Makefile
muncus: I discussed this with @minherz yesterday, and we came to the agreement that we should check both $GOOGLE_SAMPLES_PROJECT and $GOOGLE_CLOUD_PROJECT in that order.
I had this change made but not yet pushed to the branch 😄

muncus: This may be true for using a Service Account, but they can also be run without this variable, using Application Default Credentials.
I looked for a way to verify ADCs are present, but could not find a way that would catch both cases 😞 
I think we'll have to rely on the [existing setup docs](https://github.com/GoogleCloudPlatform/java-docs-samples/tree/main#set-up) for this aspect.

Makefile
muncus: This is a nice enhancement. Is there a way to do this that will work also work on Windows?
(I'm not very familiar with the Windows dev process - do you know anyone that could help us verify that these work on windows?)

Makefile
Sita04: Yeah, initially I wasn't able to get the makefile run with the ONESHELL option. But turns out its my corrupted IDE. I verified it works with Cloud Shell. Thanks for clarifying.

Makefile
Sita04: There's no impact to the samples except for testing bigtable. The suggestion was to keep the Makefile tightly coupled to local testing if that's what you are targeting now.

I suggest the two options:
1. We can either remove these and add back when the Makefile is being made CI compatible
2. Or, init both these params to empty values to avoid any resource mixups.

Makefile
Sita04: If that's the case, can we add a comment saying the user needs to be authenticated before running the Makefile?

Makefile
Sita04: Parking this for addition in the future PRs.

Sita04 (approved):
Couple minor nits. LGTM.

Makefile
muncus: Done. I've added a note about this to the top of the file, near the other usage information. I'll also be sure to include it in the docs, which are the next step.

Makefile
muncus: it looks like most of the bigtable tests use [`BIGTABLE_TESTING_INSTANCE`](https://github.com/GoogleCloudPlatform/java-docs-samples/blob/d3170e9b64520e667475480494c9353b7a28fa60/bigtable/hbase/snippets/src/test/java/com/example/bigtable/ConfigureConnectionPoolTest.java#L46).

these properties appear to only be used in [a single test](https://github.com/GoogleCloudPlatform/java-docs-samples/blob/d3170e9b64520e667475480494c9353b7a28fa60/bigtable/hbase/snippets/src/test/java/com/example/bigtable/BigtableConnectTest.java#L34).

I'm inclined to leave these in place for now, to avoid breaking even the single test that depends on them. We could also file an issue to update the single test to use the environment variable instead, which would let us safely remove these properties from both this invocation, and the CI script. WDYT?

PR #8866, opened 2023-11-10
docs(samples): Add Dataflow snippet for Pub/Sub write

VeronicaWasson:
## Description

Add a sample for writing from Dataflow to Pub/Sub.  Doc bug: b/309130866

This code is adapted from the [Dataflow cookbook](https://github.com/GoogleCloudPlatform/dataflow-cookbook/blob/main/Java/src/main/java/pubsub/WriteWithAttributesPubSub.java)

## Checklist


dataflow/snippets/src/test/java/com/example/dataflow/PubSubWriteIT.java
```diff
@@ -0,0 +1,113 @@
...
+      ProjectSubscriptionName subscriptionName =
+          ProjectSubscriptionName.of(projectId, subscriptionId);
+
+      subscriber = Subscriber.newBuilder(subscriptionName, receiver).build();
+      subscriber.startAsync().awaitRunning();
+      subscriber.awaitTerminated(30, TimeUnit.SECONDS);
+    } catch (TimeoutException timeoutException) {
+      subscriber.stopAsync();
+    }
+    assertEquals(4, count.get());
```
minherz, line 111: AFAIK subscription can get more than one copy of the message at some events. I suggest to validate that the each of the sent messages is received.

dataflow/snippets/src/main/java/com/example/dataflow/PubSubWriteWithAttributes.java
```diff
@@ -0,0 +1,66 @@
...
+    String getTopic();
+
+    void setTopic(String value);
+  }
+
+  // Write messages to a Pub/Sub topic.
+  public static void main(String[] args) {
+    final List<String> messages = Arrays.asList("message1", "message2", "message3", "message4");
+
+    var options = PipelineOptionsFactory.fromArgs(args).withValidation().as(Options.class);
```
minherz, line 46: Please modify the sample to follow the [minimal arguments](https://googlecloudplatform.github.io/samples-style-guide/#minimal-arguments) guidelines. In its current form, the sample requires additional explanations to figure out the parameters. I recommend to host the code samples in a separate function with explicit parameters and to use `main()` as a wrapper to call this function if required (although it is not necessary). Arguments to the public method make the parameters more verbose and easy to understanding from the code.

dataflow/snippets/src/main/java/com/example/dataflow/PubSubWriteWithAttributes.java
```diff
@@ -0,0 +1,66 @@
...
+        // Create some data to write to Pub/Sub.
+        .apply(Create.of(messages))
+        // Convert the data to Pub/Sub messages.
+        .apply(MapElements
+            .into(TypeDescriptor.of(PubsubMessage.class))
+            .via((payload -> {
+              // Create attributes for each message.
+              HashMap<String, String> attributes = new HashMap<String, String>();
+              attributes.put("key1", payload + " value");
+              attributes.put("key2", payload + " value2");
```
minherz, line 58: nit: given the message payloads are hardcoded, would it be more readable to define both the attributes and the payload together? Something like

```java
static class ExampleMessage {
  public String payload;
  public HashMap<String, String> attributes;
  public ExampleMessage(String payload, HashMap<String, String>attributes) {
    this.payload = payload;
    this.attributes = attributes;
  }
}
//...
// then after public static void main(String[] args) {
final List<ExampleMessage> messages = Arrays.asList(
  new ExampleMessage("message1", new HashMap<String, String>() {{
                  put("key1", "value 1/1");
                  put("key2", "value 1/2");
              }},
  new ExampleMessage("message2", new HashMap<String, String>() {{
                  put("key1", "value 2/1");
                  put("key2", "value 2/2");
              }},
};
```
it is assuming that you build, as expected, with Java 8.

dataflow/snippets/src/test/java/com/example/dataflow/PubSubWriteIT.java
```diff
@@ -0,0 +1,113 @@
...
+import java.util.concurrent.atomic.AtomicInteger;
+import org.junit.After;
+import org.junit.Before;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.junit.runners.JUnit4;
+
+@RunWith(JUnit4.class)
+public class PubSubWriteIT {
+  private static final String projectId = System.getenv("GOOGLE_CLOUD_PROJECT");
```
minherz, line 47: Mind that the project might not be defined via this environment variable. Review other code samples to see how the project ID is defined

minherz (changes requested):
The project is designed to host code samples that are used in the public documentation. Can you please share in the PR description the bug number for updating the public documentation with the code samples?

dataflow/snippets/src/test/java/com/example/dataflow/PubSubWriteIT.java
VeronicaWasson: Thanks -- yes, that's correct. By default, it's possible to receive a message more than once. I updated the test to collect the messages, by unique message ID.

dataflow/snippets/src/main/java/com/example/dataflow/PubSubWriteWithAttributes.java
VeronicaWasson: That's idiomatic for Dataflow, to pass in arbitrary arguments to `PipelineOptionsFactory.fromArgs`, so the caller can set [pipeline options](https://beam.apache.org/documentation/programming-guide/#pipeline-options-cli), which can vary depending on the Beam runner being used.

(See examples of this in the [Beam docs](https://beam.apache.org/documentation/programming-guide/#creating-a-pcollection).)

I added a code comment to clarify this, similar to [here](https://github.com/GoogleCloudPlatform/java-docs-samples/blob/9ae60dc03af6c32ce90734c264461f7ab650b373/dataflow/snippets/src/main/java/com/example/dataflow/BigQueryWrite.java#L56).

dataflow/snippets/src/main/java/com/example/dataflow/PubSubWriteWithAttributes.java
VeronicaWasson: Thanks - I agree. I was trying to avoid creating a model class, but realistically the pipeline would receive structured data, so it makes sense to show that. I modified the sample to show structured data instead of just "message1" etc.

dataflow/snippets/src/test/java/com/example/dataflow/PubSubWriteIT.java
VeronicaWasson: Hm - it looks like most samples rely on GOOGLE_CLOUD_PROJECT but include an assertion in Setup. I've added that.l Let me know if I need to do something different here. Thanks!

VeronicaWasson:
I added the relevant doc bug to the PR description

VeronicaWasson:
@minherz Thanks for the review! I responded to the comments. Is there anything else that I need to do on this? thx!

minherz:
> I added the relevant doc bug to the PR description

The bug itself is not descriptive. Do you migrate code samples from https://github.com/GoogleCloudPlatform/dataflow-cookbook?

VeronicaWasson:
> The bug itself is not descriptive. Do you migrate code samples from https://github.com/GoogleCloudPlatform/dataflow-cookbook?

The bug is to port selected snippets from that repo, add integration tests and region tags, and then incorporate them into the documentation here: https://cloud.google.com/dataflow/docs/guides/io-connector-best-practices

I updated the bug with this information.

minherz (approved):
lgtm.

nit: I would recommend to keep the number of dependencies used in each Java code sample to the bare minimum (e.g. use of `org.apache.beam.sdk.coders` seems redudndant).

VeronicaWasson:
> lgtm.
> 
> nit: I would recommend to keep the number of dependencies used in each Java code sample to the bare minimum (e.g. use of `org.apache.beam.sdk.coders` seems redudndant).

thanks!

coders is needed but I did find an unused import in the test file. Removed.

VeronicaWasson:
@minherz Hi! If there aren't any other issues, can you merge this PR? (I don't have write access.) Thanks!

Issue #8935, opened 2023-12-13
Discontinue support for Java 8 runtime


minherz:
### Description

Google Cloud [continues][1] support of Java 8 runtime. However, the volume of Java applications using Java 8 runtime to call Google APIs is constantly decreasing and on December 2023 was at 16% out of all requests made by Java applications.

To decrease the maintenance load we want to discontinue support for code samples in Java 8. While we continue running tests using this runtime we will no longer mark the Java 8 runtime checks as required. The next step will be changing the cadence of Java 8 checks from being triggered on each push to the weekly scheduled.

### What you need to know

We do not plan to decommission / remove  code samples designed for Java 8. Code samples that configured to use Java 8 at build time will continue to be build "as is" but their tests will run in more modern runtimes.

If you use a code sample that is specifically designated for Java 8 (e.g. [appengine-java8]) we will no longer monitor its correctness. Please, consider using code samples designated to run in more modern runtimes such as Java 11.

### What you need to do

If you are a code sample contributor, please configure your code sample(s) to be build with Java 11 runtime.

[1]: https://cloud.google.com/java/docs/supported-java-versions#keeping_production_systems_current

minherz:
We will add a message to README file(s) to alert users about ceasing Java 8 runtime active support.

minherz:
README is updated with the description about the supported Java runtimes in #8957

PR #8938, opened 2023-12-15
chore(proc): prevent dependency upgrades for java8 code samples

minherz:
### Description

Remove Java8 badge and add section "Supported Java runtimes" to the root README.
Exclude code samples designated to run under Java 8 runtime from the renovate analysis.
This change is a part of the process described in #8935.
Excluding dependency updates for Java 8 code samples will allow to reduce a number of flaky tests in each renovate PR.

### DIFF
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -14,6 +14,8 @@
     "dependencyDashboardLabels": [
         "type: process",
     ],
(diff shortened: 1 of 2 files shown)


Sita04 (approved):
Thanks for the changes, Leonid.

PR #8960, opened 2023-12-28
chore(proc): make header check required

minherz:
### Description

Modify the repo settings to enforce header-check on every submitted change.


minherz:
Force merge the change without review.

PR #8962, opened 2024-01-02
bigtable (docs): add datasets for change stream tutorials in the docs

billyjacobson:
## Description

Fixes #<ISSUE-NUMBER>

## Checklist


Sita04:
@billyjacobson Considering the size of this dataset, can it be stored in a db and pulled when the tests need them?

billyjacobson:
> @billyjacobson Considering the size of this dataset, can it be stored in a db and pulled when the tests need them?

The data is for a tutorial in the docs and needs to be imported for users. There isn't any code that goes with it since the user would just import the data and then uses a dataflow template, bigquery and looker. There isn't really another way to get data into Bigtable easily, and I'd rather not create an entirely new repo for this or put it somewhere else because it ends up being more difficult to keep track of. Happy to discuss

minherz:
> > @billyjacobson Considering the size of this dataset, can it be stored in a db and pulled when the tests need them?
> 
> The data is for a tutorial in the docs and needs to be imported for users. There isn't any code that goes with it since the user would just import the data and then uses a dataflow template, bigquery and looker. There isn't really another way to get data into Bigtable easily, and I'd rather not create an entirely new repo for this or put it somewhere else because it ends up being more difficult to keep track of. Happy to discuss

Can you add arguments about the need for 10K of data records (in 2023 file)? I also don't think the solution of using this repo as a storage for tutorial data is the right solution. Please, consider using a designated storage bucket.

minherz (changes requested):
Please downgrade the size of the CSV files to few dozen of records. Addressing the [comment][1] regarding the use of this data for tutorial, the use of this repo for storing content other than code for tutorials is possible to the reasonable extend.
Storing CSV file with 10K records does not look like one. Please, consider using one of DevRel projects to store the data in the publicly available storage bucket.

[1]: https://github.com/GoogleCloudPlatform/java-docs-samples/pull/8962#issuecomment-1883073237

billyjacobson:
Added to a bucket with bigtable public datasets

PR #8983, opened 2024-01-25
feat(scc): add v2 samples - muteconfig and findings

Sita04:
Add new samples for Security Command Center v2 APIs.

Total: 12 samples (MuteConfig + Findings)


minherz:
QQ: we discussed to temporarily suspend adding new samples to the repo. What is necessity for these particular samples?

Sita04:
@minherz  Ack, but the samples are supporting a new launch. Hence, the priority in adding them.

security-command-center/snippets/src/main/java/drz/muteconfig/CreateMuteRule.java
```diff
@@ -0,0 +1,76 @@
...
+    // Available locations: "us", "eu", "global".
+    String location = "us";
+
+    // muteConfigId: Set a random id; max of 63 chars.
+    String muteConfigId = "random-mute-id-" + UUID.randomUUID();
+
+    createMuteRule(projectId, location, muteConfigId);
+  }
+
+  // Creates a DRZ compliant mute configuration in a project under a given location.
```
RajviM, line 45: remove

security-command-center/snippets/src/main/java/drz/muteconfig/UpdateMuteRule.java
```diff
@@ -0,0 +1,82 @@
...
+      UpdateMuteConfigRequest updateMuteConfigRequest =
+          UpdateMuteConfigRequest.newBuilder()
+              .setMuteConfig(updateMuteConfig)
+              // Set the update mask to specify which properties of the mute config should be
+              // updated.
+              // If empty, all mutable fields will be updated.
+              // Make sure that the mask fields match the properties changed in 'updateMuteConfig'.
+              // For more info on constructing update mask path, see the proto or:
+              // https://cloud.google.com/security-command-center/docs/reference/rest/v1/folders.muteConfigs/patch?hl=en#query-parameters
+              .setUpdateMask(FieldMask.newBuilder().addPaths("description").build())
```
RajviM, line 72: v2

security-command-center/snippets/src/main/java/drz/muteconfig/BulkMuteFindings.java
```diff
@@ -0,0 +1,74 @@
...
+
+    // muteRule: Expression that identifies findings that should be muted.
+    // Can also refer to an organization/ folder.
+    // eg: "resource.project_display_name=\"PROJECT_ID\""
+    String muteRule = "resource.project_display_name=\"" + projectId + "\"";
+
+    bulkMute(projectId, location, muteRule);
+  }
+
+  // Kicks off a DRZ compliant long-running operation (LRO) to bulk mute findings for a parent based on a filter.
```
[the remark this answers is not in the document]
Sita04: Removed the keyword "DRZ compliant".

security-command-center/snippets/src/main/java/drz/muteconfig/CreateMuteRule.java
Sita04: Removed the keyword "DRZ compliant".

security-command-center/snippets/pom.xml
```diff
@@ -69,8 +70,8 @@
     <dependency>
       <groupId>com.google.truth</groupId>
       <artifactId>truth</artifactId>
-      <version>1.4.0</version>
+      <version>1.2.0</version>
```
minherz, line 73: what is a reason for downgrading this package?

security-command-center/snippets/src/main/java/v2/muteconfig/BulkMuteFindings.java
```diff
@@ -0,0 +1,73 @@
...
+              // To create mute rules, see:
+              // https://cloud.google.com/security-command-center/docs/how-to-mute-findings#create_mute_rules
+              .setFilter(muteRule)
+              .build();
+
+      // ExecutionException is thrown if the below call fails.
+      BulkMuteFindingsResponse response =
+          client.bulkMuteFindingsAsync(bulkMuteFindingsRequest).get();
+      System.out.println("Bulk mute findings completed successfully! " + response);
+    } catch (IOException | InterruptedException | ExecutionException e) {
```
minherz, line 68: why this code sample catches "all" errors while others allow the exception to pass through?
additionally, consider how hard it is to test error scenarios when exceptions are handled this way.

security-command-center/snippets/src/main/java/v2/muteconfig/CreateMuteRule.java
```diff
@@ -0,0 +1,75 @@
...
+              .build();
+
+      // You can also create mute rules in an organization/ folder.
+      // Construct the parameters according to the parent resource.
+      //  * Organization -> client.createMuteConfig(OrganizationLocationName.of(...
+      //  * Folder -> client.createMuteConfig(FolderLocationName.of(...
+      MuteConfig response = client.createMuteConfig(
+          LocationName.of(projectId, location), muteConfig, muteConfigId);
+      System.out.println("Mute rule created successfully: " + response.getName());
+    } catch (IOException e) {
```
minherz, line 70: Similar BulkMuteFindings.java, what is a reason for catching all IO exceptions?

security-command-center/snippets/src/main/java/v2/muteconfig/DeleteMuteRule.java
```diff
@@ -0,0 +1,60 @@
...
+    // once, and can be reused for multiple requests.
+    try (SecurityCenterClient client = SecurityCenterClient.create()) {
+      // Use appropriate `MuteConfigName` methods depending on the parent type.
+      // folder -> MuteConfigName.ofFolderLocationMuteConfigName()
+      // organization -> MuteConfigName.ofOrganizationLocationMuteConfigName()
+      client.deleteMuteConfig(
+          MuteConfigName.ofProjectLocationMuteConfigName(projectId, location, muteConfigId));
+
+      System.out.println("Mute rule deleted successfully: " + muteConfigId);
+    } catch (IOException e) {
```
minherz, line 55: Same case for catching exceptions.

security-command-center/snippets/src/main/java/v2/muteconfig/GetMuteRule.java
```diff
@@ -0,0 +1,61 @@
...
+    try (SecurityCenterClient client = SecurityCenterClient.create()) {
+      // Use appropriate `MuteConfigName` methods depending on the parent type.
+      //  * organization -> MuteConfigName.ofOrganizationLocationMuteConfigName()
+      //  * folder -> MuteConfigName.ofFolderLocationMuteConfigName()
+
+      MuteConfigName muteConfigName = MuteConfigName.ofProjectLocationMuteConfigName(projectId,
+          location, muteConfigId);
+      MuteConfig muteConfig = client.getMuteConfig(muteConfigName);
+      System.out.println("Retrieved the mute config: " + muteConfig);
+    } catch (IOException e) {
```
minherz, line 56: All exceptions are hidden. Consider returning status and throwing unrelated to API call exceptions.

security-command-center/snippets/src/main/java/v2/muteconfig/SetUnmuteFinding.java
minherz, line 1: nit: Would it make more sense and save some code lines if the Mute and Unmute findings will be combined into a single snippet?

security-command-center/snippets/src/main/java/v2/muteconfig/UpdateMuteRule.java
```diff
@@ -0,0 +1,81 @@
...
+      UpdateMuteConfigRequest updateMuteConfigRequest =
+          UpdateMuteConfigRequest.newBuilder()
+              .setMuteConfig(updateMuteConfig)
+              // Set the update mask to specify which properties of the mute config should be
+              // updated.
+              // If empty, all mutable fields will be updated.
+              // Make sure that the mask fields match the properties changed in 'updateMuteConfig'.
+              // For more info on constructing update mask path, see the proto or:
+              // https://cloud.google.com/security-command-center/docs/reference/rest/v2/folders.muteConfigs/patch?hl=en#query-parameters
+              .setUpdateMask(FieldMask.newBuilder().addPaths("description").build())
```
minherz, line 71: Is this value random or it has to be "description"?  From the code snippet it isn't obvious that the call `  MuteConfig.newBuilder().setDescription` should correlate with the call  `UpdateMuteConfigRequest.newBuilder().setUpdateMask(FieldMask.newBuilder().addPaths("description").build())`.
Too many comments overload the code snippet.
Can this sample be made more simple?

minherz (commented):
* nit: it is generally recommended to narrow the scope of code snippets by placing the region tags around the actual code while omitting `main()` methods and other not relevant details. Mind that you can use more the region tag in more than place within the file so you can provide necessary `import` statements to ensure that copy-n-paste of the code snippet will compile.
* nit: The best practice is considered to return actual values for code snippets that get information. It allows to write more robust unit tests.
* nit: It would be a good practice either to catch the API related exception or explicitly declare them (using `throws` statement in the method declaration)
* Inconsistency in the error handling; there are several considerations regarding it:
   * Best practices for error handling in Java
   * Expose what kind of thrown exceptions are related to this particular API or user scenarion
   * Support writing unit tests that validate specific errors (thrown exceptions)

security-command-center/snippets/pom.xml
Sita04: It wasn't intentional. Thanks for the catch.

security-command-center/snippets/src/main/java/v2/muteconfig/SetUnmuteFinding.java
Sita04: Since we've been following one snippet in one class guideline, I would prefer to leave it as such. Plmk if you feel strongly about this.

security-command-center/snippets/src/main/java/v2/muteconfig/UpdateMuteRule.java
Sita04: Yes, it has to match the "updateMuteConfig" object defined above. The comment in line 68 here mentions here.

But like you suggested, I'll refactor the comments to make it more readable! Thanks

security-command-center/snippets/src/main/java/v2/muteconfig/BulkMuteFindings.java
Sita04: The code is a refactor of the existing samples for v1. Hence, didn't modify the way exceptions are caught.

The argument makes sense, Updating to remove all caught exceptions.

security-command-center/snippets/src/main/java/v2/muteconfig/CreateMuteRule.java
Sita04: ditto.

security-command-center/snippets/src/main/java/v2/muteconfig/GetMuteRule.java
Sita04: ditto

Sita04:
> * nit: it is generally recommended to narrow the scope of code snippets by placing the region tags around the actual code while omitting `main()` methods and other not relevant details. Mind that you can use more the region tag in more than place within the file so you can provide necessary `import` statements to ensure that copy-n-paste of the code snippet will compile.
> * nit: The best practice is considered to return actual values for code snippets that get information. It allows to write more robust unit tests.
> * nit: It would be a good practice either to catch the API related exception or explicitly declare them (using `throws` statement in the method declaration)
> * Inconsistency in the error handling; there are several considerations regarding it:
>   
>   * Best practices for error handling in Java
>   * Expose what kind of thrown exceptions are related to this particular API or user scenarion
>   * Support writing unit tests that validate specific errors (thrown exceptions)

Thanks for your review, Leonid!

1. This guideline is not standard and different languages have their own implementation. In Java, we have primarily resorted to following "one snippet per class" rule. Will follow up on this discussion internally to see if it's time to update this!

2. I've updated snippets that can benefit from returning results. Few snippets like listing and grouping findings, I've left the tests unchanged, primarily because this could mean passing unknown number of result records back to tests.

3. For exceptions, I've updated the snippets to follow the guideline of catching only API related exceptions.

security-command-center/snippets/src/test/java/v2/FindingsIT.java
```diff
@@ -0,0 +1,134 @@
...
+import v2.findings.GroupFindingsWithFilter;
+import v2.findings.ListAllFindings;
+import v2.findings.ListFindingsWithFilter;
+
+// Test v2 Findings samples.
+@RunWith(JUnit4.class)
+public class FindingsIT {
+
+  // TODO(Developer): Replace the below variables.
+  private static final String ORGANIZATION_ID = System.getenv("SCC_PROJECT_ORG_ID");
```
minherz, line 49: nit: it is a bit unconventional to use external environment variable for integration testing. Where it is defined? How is it ensured that this will work on every environment? If we use the "fixed" test project, can we just setup the "fixed" org id?

minherz:
> >  _..._
> Thanks for your review, Leonid!
> 
> 1. This guideline is not standard and different languages have their own implementation. In Java, we have primarily resorted to following "one snippet per class" rule. Will follow up on this discussion internally to see if it's time to update this!

I do not follow the reference about different languages here. The first bullet in the original comment explained that the code snippet (part of code enclosed by the region tag) should show only the relevant code to ensure that the documentation is not overloaded with extra content. The files in this PR use region tags in a way that code snippet includes the whole file with the exception of the copyright comment.
It is not a blocker, but including the main() method into code snippet does not contribute to the example of the API call usage but only helps to run the code sample if user decides to.

> 2. I've updated snippets that can benefit from returning results. Few snippets like listing and grouping findings, I've left the tests unchanged, primarily because this could mean passing unknown number of result records back to tests.

It looks like only one "get" code samples was modified.
It is not a blocker, but I would recommend considering to modify all "get" code samples. The delete/set/update code samples can communicate the status as not throwing exception ==> OK. If any of the delete/set/update code samples receive the status of the operation success on response, they can return that status as well.
There should be no concern regarding unknown number of results because the tests should validate the expected content. Currently they do not validate a number of results or all returned content either.

> 3. For exceptions, I've updated the snippets to follow the guideline of catching only API related exceptions.

I do not see any `catch` statements in the files. Is it supposed to be like this?

minherz (approved):
I think that using environment variables with IT tests is not good practice. In general we run IT tests vs. controlled environment. If a user does not have access to that environment, they can run the tests on push.
While it is not a blocker, leaving IT tests in this way is a source of errors for trying to run tests locally and because it requires special customization of the test setup to run this specific group of tests.
I strongly encourage to review it and replace the values with constant values that can run vs. java-docs-samples-testing environment.

security-command-center/snippets/src/test/java/v2/FindingsIT.java
Sita04: Thanks..

The org id is required here to account for the special testing permissions. These are fetched from Secret manager and are consistent in every run. More in the internal thread.

PR #8987, opened 2024-01-26
Quota project fix

aabmass:
## Description

Cloud Trace API requires a quota project. This change passes through `GOOGLE_CLOUD_QUOTA_PROJECT` environment variable, which Cloud Shell has automatically set. This is better than doing the resource project override and works better off Cloud Shell. I also renamed the file to reflect that it works with any credentials file, including SA keys.

Equivalent of https://github.com/GoogleCloudPlatform/golang-samples/pull/3760 from the Go repo

## Checklist


opentelemetry/spring-boot-instrumentation/docker-compose.creds.yaml
```diff
@@ -26,8 +28,8 @@ services:
   otelcol:
     # If the collector does not have permission to read the mounted volumes, set
     # USERID=$(id -u) to run the container as the current user
-    user: ${USERID}
+    user: $USERID
```
minherz, line 31: what is a reason for this change? To be syntactically correct the original line should be enclosed with quotes.

opentelemetry/spring-boot-instrumentation/docker-compose.creds.yaml
```diff
@@ -12,12 +12,14 @@
...
-# Use this compose file along with docker-compose.yaml to pass Application Default
-# Credentials from the host into the collector container:
+# Use this compose file along with docker-compose.yaml to pass a credentials file from the host
+# into the collector container:
 #
 # ```
-# export GOOGLE_APPLICATION_CREDENTIALS=$HOME/.config/gcloud/application_default_credentials.json
-# docker compose -f docker-compose.yaml -f docker-compose.adc.yaml up
+# GOOGLE_CLOUD_PROJECT="otel-quickstart-demos" \
+# GOOGLE_APPLICATION_CREDENTIALS="$HOME/.config/gcloud/application_default_credentials.json" \
```
minherz, line 20: what is a reason to advertise the anti-pattern of using the GSA key file for authentication?
the doc samples use the main use case which is to run the sample on Google Cloud.

opentelemetry/spring-boot-instrumentation/docker-compose.creds.yaml
```diff
@@ -26,8 +28,8 @@ services:
   otelcol:
     # If the collector does not have permission to read the mounted volumes, set
     # USERID=$(id -u) to run the container as the current user
-    user: ${USERID}
+    user: $USERID
     volumes:
-      - ${GOOGLE_APPLICATION_CREDENTIALS:-/dev/null}:/tmp/keys/gcp-credentials.json:ro
+      - ${GOOGLE_APPLICATION_CREDENTIALS?}:${GOOGLE_APPLICATION_CREDENTIALS}:ro
```
minherz, line 33: if I am not mistake this syntax results in failing if the `GOOGLE_APPLICATION_CREDENTIALS` env var is not set.
what is a reason for using this construct in the code sample? see my previous comment about using anti-patterns

minherz (changes requested):
this is contra

opentelemetry/spring-boot-instrumentation/docker-compose.creds.yaml
aabmass: The change is to be consistent with L35. Either style should be equivalent.

> To be syntactically correct the original line should be enclosed with quotes.

I'm not sure what you mean, what is wrong with this syntax?

opentelemetry/spring-boot-instrumentation/docker-compose.creds.yaml
aabmass: This is to allow users to run the example on their own machine. We are not recommending this pattern in the docs, just making it possible.

opentelemetry/spring-boot-instrumentation/docker-compose.creds.yaml
aabmass: Correct that is the purpose, to alert the user they need the environment variable with **this** docker compose file. If they want to get credentials from the environment (which doesn't work on a local machine), they can use the other compose file. The two options are explained in [the README](https://github.com/GoogleCloudPlatform/java-docs-samples/blob/2bd2953c40e178965f6f0216650c3e4a7e134877/opentelemetry/spring-boot-instrumentation/README.md#running-the-example) and in the devsite docs.

PR #9227, opened 2024-01-29
chore: update workflow badges

Sita04:
Versions:

Java 8 : badge removed as this version is in maintenance mode for existing samples.
 
Java 11 and 17: badge supported as periodic builds reflect health of the samples against these versions

Java 21: badge added to reflect nightly build results

### DIFF
--- a/README.md
+++ b/README.md
@@ -80,17 +80,9 @@ Apache header:
 
 [ADC]: https://developers.google.com/identity/protocols/application-default-credentials
 [auth_command]: https://cloud.google.com/sdk/gcloud/reference/beta/auth/application-default/login
-[java-8-badge]:
-https://storage.googleapis.com/cloud-devrel-kokoro-resources/java/badges/java-docs-samples-8.svg
-[java-8-link]:
-https://storage.googleapis.com/cloud-devrel-kokoro-resources/java/badges/java-docs-samples-8.html
-[java-11-badge]:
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -80,17 +80,9 @@ Apache header:
...
-[java-11-link]:
-https://storage.googleapis.com/cloud-devrel-kokoro-resources/java/badges/java-docs-samples-11.html
-[java-17-badge]:
-https://storage.googleapis.com/cloud-devrel-kokoro-resources/java/badges/java-docs-samples-17.svg
-[java-17-link]:
-https://storage.googleapis.com/cloud-devrel-kokoro-resources/java/badges/java-docs-samples-17.html
+[java-21-badge]:
+https://storage.googleapis.com/cloud-devrel-kokoro-resources/java/badges/java-docs-samples-21.svg
+[java-21-link]:
+https://storage.googleapis.com/cloud-devrel-kokoro-resources/java/badges/java-docs-samples-21.html
```
minherz, line 86: I understand the reason to remove Java8 - we don't support it.
Why do we remove Java11 and Java17? Shouldn't they reference the last execution (by one of PR checks)?

minherz (commented):
I suggest we will leave the Java11 and Java17 check results. We only need to make sure that they point to the "right" results.
If the current not-periodic Kokoro builds don't updates the badges, let's fix that instead of removing the badges.

README.md
Sita04: Updated it as discussed.

minherz (approved):
lgtm

PR #9239, opened 2024-02-07
chore: update header-checker-lint.yml to ignore year

Sita04:


minherz (changes requested):
I disagree. The year is the integral part of the license. By human mistake one can place a wrong year and it will be in the public access forever.
Please, correct the year in the #9238

Sita04:
This was meant as a temp fix for the PR. As per discussions, closing this PR.

PR #9272, opened 2024-03-27
feat: Text-embedding model tuning demo.

lee1premium:
## Description

Fixes: http://b/320696823

## Checklist


aiplatform/src/test/java/aiplatform/CreatePipelineJobEmbeddingModelTuningSampleTest.java
```diff
@@ -0,0 +1,149 @@
...
+        QUERIES,
+        CORPUS,
+        TRAIN_LABEL,
+        TEST_LABEL,
+        BATCH_SIZE,
+        ITERATIONS);
+
+    String got = bout.toString();
+    System.out.println("hylee:: " + got);
+    assertThat(got).contains(pipelineJobDisplayName);
```
minherz, line 146: Using STDOUT for test assertions is not recommended. Consider a case when the tested code changes printing string in a way that the asserted substring remains but totally changes the logic.
Please, refactor `createEmbeddingModelTuningPipelineJob()` to return status code that can be validated.

minherz (changes requested):

lee1premium:
@minherz Can we move on rather than addressing your concerns, after we create bugs for code quality problems? I cannot make a big difference over Vertex AI platform code samples.

I appreciate your comments/code quality concerns. I didn't come up with such bad coding practice. I just repeated the same, bad pratice from https://github.com/GoogleCloudPlatform/java-docs-samples/tree/main/aiplatform/src
* You can find a ton of samples using the same region tags.
* You can find a few other samples using the same catch clauses.
* You can find a few other samples using the same retry over deletion.
* You can find a ton of samples using STDOUT for testing.

aiplatform/src/main/java/aiplatform/CreatePipelineJobEmbeddingModelTuningSample.java
```diff
@@ -0,0 +1,128 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package aiplatform;
+
+// [START aiplatform_sdk_embedding_model_tuning]
```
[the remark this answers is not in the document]
yan283: As lee1premium mentioned there are quite a few aiplatform samples that include the complete class, including the main method

https://github.com/GoogleCloudPlatform/java-docs-samples/blob/main/aiplatform/src/main/java/aiplatform/BatchCreateFeaturesSample.java

https://github.com/GoogleCloudPlatform/java-docs-samples/blob/main/aiplatform/src/main/java/aiplatform/CreateBatchPredictionJobBigquerySample.java

I feel it is debatable which one is better to include in the documentation, but at least the current practice is consistent with the rest of the samples.    Do we have a new policy on this and plan to migrate all samples to it?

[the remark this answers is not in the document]
minherz: Yes, many legacy code samples enclose the whole file in region tags. It is incorrect. See https://googlecloudplatform.github.io/samples-style-guide/ for guidelines.
Following incorrect implementation is rarely an argument in new implementations. In this particular case the subject in question is to enclose in region tags two areas instead of one.

I am unsure what you consider as "debatable" here. The provided reference to guidelines should be comprehensive enough.

minherz:
> @minherz Can we move on rather than addressing your concerns, after we create bugs for code quality problems? I cannot make a big difference over Vertex AI platform code samples.
> 

The whole point of the code review is code quality. We don't need the review if we can skip this phase. I find the code hard for maintenance due to lack of resilience in test implementation and running times.
I pointed to a violation of guidelines and a redundant code lines that are easily fixable and improve the code quality.
Please remember, that this code represents Google engineering quality among other things. I do believe the investing some time worth the effort.

aiplatform/src/main/java/aiplatform/EmbeddingModelTuningSample.java
```diff
@@ -0,0 +1,129 @@
...
+      String queriesPath,
+      String corpusPath,
+      String trainLabelPath,
+      String testLabelPath,
+      int batchSize,
+      int iterations)
+      throws IOException {
+    Pattern apiEndpointPattern =
+        Pattern.compile("(?<Location>.+)(-autopush|-staging)?-aiplatform.+");
+    Matcher matcher = apiEndpointPattern.matcher(apiEndpoint);
```
minherz, line 84: nit: do we really need it in the code sample? For example the code does not validate project ID and other input parameters.

minherz (approved):
Thank you for your work. Consider addressing the minor comments.

aiplatform/src/main/java/aiplatform/EmbeddingModelTuningSample.java
```diff
@@ -0,0 +1,129 @@
...
+        taskType,
+        pipelineJobDisplayName,
+        outputDir,
+        queriesPath,
+        corpusPath,
+        trainLabelPath,
+        testLabelPath,
+        batchSize,
+        iterations);
+    // [END aiplatform_sdk_embedding_model_tuning]
```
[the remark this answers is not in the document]
lee1premium: Can you define a function with default values in Java? This pattern has been there for almost all the samples.

aiplatform/src/main/java/aiplatform/EmbeddingModelTuningSample.java
lee1premium: This is not optional, but rather required. This code extracts the required location from an API endpoint.

aiplatform/src/main/java/aiplatform/EmbeddingModelTuningSample.java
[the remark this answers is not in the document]
lee1premium: There are no reasonable defaults for batch size and iterations. You cannot know what machine learning problems and downstream tasks your future customers will train a machine learning model for.

gericdong (approved):

aiplatform/src/main/java/aiplatform/EmbeddingModelTuningSample.java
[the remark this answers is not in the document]
davidcavazos: Unfortunately I don't think Java supports default values. The only way I'm aware of would be through overloading, but that would make it even harder to read.

As for giving reasonable defaults, I think that can be done. I know they can change widely from case to case, but we can at least give a reasonable value for this specific example. What people are looking for is if the batch size should be something like 10, or 64, or 512, or 100,000. Even if it's not the most optimized, a reasonable default could be something that makes for reasonable enough performance while keeping memory usage low. Iterations can, again, be something that balances compute time with overall accuracy (however that's measured).

For the specific case of a code sample, we would usually incline towards using less resources and taking less time to complete so that users can play with the code without it being too expensive or resource intensive. For example, I usually try to keep my sample training jobs around 5 to 10 minutes, or at least less than 30 minutes. That would depend if, for example, after 5 minutes the results still look like noise, I would go higher, but always trying to keep it as fast as possible. I also like to give an example of what the results look like after training for longer periods like several hours so that they can see the results without having to go through all the hassle.

If it's something very important for this sample, there should be an explicit note both as comments in the code and in the README file.

aiplatform/src/main/java/aiplatform/EmbeddingModelTuningSample.java
```diff
@@ -0,0 +1,127 @@
...
+      String outputDir,
+      String queriesPath,
+      String corpusPath,
+      String trainLabelPath,
+      String testLabelPath,
+      int batchSize,
+      int iterations)
+      throws IOException {
+    Matcher matcher = Pattern.compile("^(?<Location>\\w+-\\w+)").matcher(apiEndpoint);
+    String location = matcher.matches() ? matcher.group("Location") : "us-central1";
```
davidcavazos, line 83: This is hard to understand what's going on. Instead of using a regular expression to extract the location, how about we simply pass the location as an extra parameter to the function? We can have an explicit comment that the location must match the endpoint used.

Otherwise, a simple split by `-` and getting the first item seems like a simpler solution than a regular expression.

lee1premium: I have already implemented the same logic in python, go, and nodejs as well.
* Found this simple regex "^(\w+-\w+)" works for almost all the cases.

In Python, given a funny api_endpoint = "us-central1-funny-staging-strage-prediction-aiplatform.sandy.boxer.googleapis.com"
* location = "-".join(api_endpoint.split(".")[0].split("-")[:2])  # I found it is hairy to do two splits and a join.

lee1premium: For consistency reasons, you would want this to be the same as other language samples.

davidcavazos (approved):

davidcavazos:
`err: signal: killed` after 27 minutes, looks like a timeout. Something's taking too long during the test.

lee1premium:
How can you ignore these irrelevant, flaky tests? I have no way to see the pipeline creation failure details at `projects/ucaip-sample-tests`.

### Java 11 - CI failed at aiplatform.CreateTrainingPipelineSampleTest and  aiplatform.CancelTrainingPipelineSampleTest

* Caused by: io.grpc.StatusRuntimeException: FAILED_PRECONDITION: The TrainingPipeline "projects/ucaip-sample-tests/locations/us-central1/trainingPipelines/6971726189157154816" is in state PIPELINE_STATE_FAILED and cannot be canceled.
* Caused by: io.grpc.StatusRuntimeException: FAILED_PRECONDITION: The TrainingPipeline "projects/ucaip-sample-tests/locations/us-central1/trainingPipelines/4361890205095952384" is in state PIPELINE_STATE_FAILED and cannot be canceled.

### Java 17 - CI failed at aiplatform.CancelTrainingPipelineSampleTest

* Caused by: io.grpc.StatusRuntimeException: FAILED_PRECONDITION: The TrainingPipeline "projects/ucaip-sample-tests/locations/us-central1/trainingPipelines/4551745076887289856" is in state PIPELINE_STATE_FAILED and cannot be canceled.

aiplatform/src/main/java/aiplatform/EmbeddingModelTuningSample.java
```diff
@@ -0,0 +1,127 @@
...
+              "queries_path", valueOf(queriesPath),
+              "corpus_path", valueOf(corpusPath),
+              "train_label_path", valueOf(trainLabelPath),
+              "test_label_path", valueOf(testLabelPath),
+              "batch_size", valueOf(batchSize),
+              "iterations", valueOf(iterations));
+      PipelineJob pipelineJob =
+          PipelineJob.newBuilder()
+              .setTemplateUri(
+                  "https://us-kfp.pkg.dev/ml-pipeline/llm-text-embedding/tune-text-embedding-model/v1.1.2")
```
kweinmeister, line 102: Consider pulling this out as a constant next to location at the beginning of the function.

kweinmeister (approved):

aiplatform/src/main/java/aiplatform/EmbeddingModelTuningSample.java
lee1premium: Done

PR #9294, opened 2024-04-18
fix: Adding missing grpc dependency

surbhigarg92:


dsimil (approved):

suztomo:
Would you fill the description about what it enables?

spanner/opencensus/pom.xml
```diff
@@ -59,6 +59,10 @@
       <artifactId>opencensus-contrib-grpc-metrics</artifactId>
       <version>${opencensus.version}</version>
     </dependency>
+    <dependency>
+      <groupId>io.grpc</groupId>
+      <artifactId>grpc-census</artifactId>
```
suztomo, line 64: Would you add source code comment when a user wants this?

Sita04 (approved):

spanner/opencensus/pom.xml
surbhigarg92: Added.

suztomo:
Would you fill the pull request description about what it enables?

(What was not working?)

PR #9314, opened 2024-05-08
Add a batch write flow control example for Bigtable

kongweihan:
## Description

Add a batch write flow control example for Bigtable

## Checklist

- [] These samples need a new **API enabled** in testing projects to pass (let us know which ones)
- [] These samples need a new/updated **env vars** in testing projects set to pass (let us know which ones)
- [] This sample adds a new sample directory, and I updated the [CODEOWNERS file](https://github.com/GoogleCloudPlatform/java-docs-samples/blob/main/.github/CODEOWNERS) with the codeowners for this sample
- [] This sample adds a new **Product API**, and I updated the [Blunderbuss issue/PR auto-assigner](https://github.com/GoogleCloudPlatform/java-docs-samples/blob/main/.github/blunderbuss.yml) with the codeowners for this sample 

### DIFF
--- a/bigtable/beam/batch-write-flow-control-example/pom.xml
+++ b/bigtable/beam/batch-write-flow-control-example/pom.xml
@@ -1,0 +1,120 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+    Copyright 2024 Google LLC
+
+    Licensed under the Apache License, Version 2.0 (the "License");
+    you may not use this file except in compliance with the License.
+    You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+    Unless required by applicable law or agreed to in writing, software
+    distributed under the License is distributed on an "AS IS" BASIS,
+    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+    See the License for the specific language governing permissions and
+    limitations under the License.
+-->
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+
+  <groupId>com.example.bigtable</groupId>
+  <artifactId>batch-write-flow-control-example</artifactId>
+  <version>1.0-SNAPSHOT</version>
+
+  <properties>
+    <maven.compiler.source>1.8</maven.compiler.source>
+    <maven.compiler.target>1.8</maven.compiler.target>
+    <apache_beam.version>2.56.0</apache_beam.version>
+  </properties>
+
+  <!--
+  The parent pom defines common style checks and testing strategies for our samples.
+  Removing or replacing it should not affect the execution of the samples in anyway.
+  -->
+  <parent>
+    <groupId>com.google.cloud.samples</groupId>
+    <artifactId>shared-configuration</artifactId>
+    <version>1.2.0</version>
+  </parent>
+
+  <dependencyManagement>
+    <dependencies>
+      <dependency>
+        <artifactId>libraries-bom</artifactId>
+        <groupId>com.google.cloud</groupId>
+        <scope>import</scope>
+        <type>pom</type>
+        <version>26.40.0</version>
+      </dependency>
+    </dependencies>
+  </dependencyManagement>
+  <dependencies>
+    <dependency>
+      <groupId>org.apache.beam</groupId>
+      <artifactId>beam-runners-google-cloud-dataflow-java</artifactId>
+      <version>${apache_beam.version}</version>
+    </dependency>
+    <dependency>
+      <groupId>org.apache.beam</groupId>
+      <artifactId>beam-runners-direct-java</artifactId>
+      <version>${apache_beam.version}</version>
+    </dependency>
+    <dependency>
+      <groupId>com.google.cloud.bigtable</groupId>
+      <artifactId>bigtable-hbase-beam</artifactId>
+      <version>2.14.0</version>
+    </dependency>
+
+    <dependency>
+      <groupId>com.google.cloud</groupId>
+      <artifactId>google-cloud-bigtable</artifactId>
+    </dependency>
+
+    <dependency>
+      <groupId>com.google.cloud.bigtable</groupId>
+      <artifactId>bigtable-client-core</artifactId>
+      <version>1.29.2</version>
+    </dependency>
+
+    <dependency>
+      <groupId>junit</groupId>
+      <artifactId>junit</artifactId>
+      <version>4.13.2</version>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>com.google.truth</groupId>
+      <artifactId>truth</artifactId>
+      <version>1.1.5</version>
+      <scope>test</scope>
+    </dependency>
+
+
+  </dependencies>
+
+  <repositories>
+    <repository>
+      <id>artifact-registry</id>
+      <url>artifactregistry://us-maven.pkg.dev/cloud-bigtable-ecosystem/debug-applovin</url>
+      <releases>
+        <enabled>true</enabled>
+      </releases>
+      <snapshots>
+        <enabled>true</enabled>
+      </snapshots>
+    </repository>
+  </repositories>
+
+  <build>
+    <extensions>
+      <extension>
+        <groupId>com.google.cloud.artifactregistry</groupId>
+        <artifactId>artifactregistry-maven-wagon</artifactId>
+        <version>2.2.0</version>
+      </extension>
+    </extensions>
+  </build>
+
+</project>
--- a/bigtable/beam/batch-write-flow-control-example/src/main/java/bigtable/BatchWriteFlowControlExample.java
+++ b/bigtable/beam/batch-write-flow-control-example/src/main/java/bigtable/BatchWriteFlowControlExample.java
@@ -1,0 +1,257 @@
+/*
+ * Copyright 2024 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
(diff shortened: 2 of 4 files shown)


bigtable/beam/batch-write-flow-control-example/pom.xml
```diff
@@ -0,0 +1,127 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+    Copyright 2021 Google LLC
```
billyjacobson, line 3: can you change to 2024

bigtable/beam/batch-write-flow-control-example/src/main/java/bigtable/BatchWriteFlowControlExample.java
```diff
@@ -0,0 +1,228 @@
...
+    String mutationLabel = String
+        .format("Create mutations that write %d columns of total %d bytes to each row",
+            options.getBigtableColsPerRow(), options.getBigtableColsPerRow()
+                * options.getBigtableBytesPerCol());
+
+    Pipeline p = Pipeline.create(options);
+
+    PCollection<Long> numbers = p.apply(generateLabel, GenerateSequence.from(0).to(numRows));
+
+    if (options.getUseCloudBigtableIo()) {
```
billyjacobson, line 83: do we need the code showing both ways for this? I know we recommend customers use one of these over the other but can never remember which one. Ideally we just show the recommended way. Or if the goal is to just show how to activate the flow control on the client then we could keep both but maybe the data being written can be simplified to not have as much information for customers to take in. 

For example, we could just write one row using each client and that can reduce the need to generate sequences and all the additional helper functions.

bigtable/beam/batch-write-flow-control-example/src/main/java/bigtable/BatchWriteFlowControlExample.java
```diff
@@ -0,0 +1,237 @@
...
+          .withTableId(options.getBigtableTableId())
+          .withFlowControl(true);  // This enables batch write flow control
+
+      mutations.apply(
+          String.format("Write data to table %s via BigtableIO", options.getBigtableTableId()),
+          write
+      );
+    }
+
+    p.run();
```
minherz, line 126: this block of code is hard to read. Since it is a code sample it should be easy to understand. Please, reformat it so it will look like steps each of the steps calling `apply` method of the pipeline. See the [dataflow-bigquery-read-tablerows](https://cloud.google.com/dataflow/docs/samples/dataflow-bigquery-read-tablerows) sample as a reference.

minherz, line 126: Does it intend for run asynchronously? Please, append `waitUntilFinish()` call to the result of the `run()`.

minherz (changes requested):
Hello,
Please address the following questions:

1. We ask to have a single code sample per file. This code looks like showing two code samples. What is this code sample demonstrates?
2. This code sample does not have regional tags. What documentation use it?
3. We do not host code samples without tests. What is a reason for lack of tests?

bigtable/beam/batch-write-flow-control-example/src/main/java/bigtable/BatchWriteFlowControlExample.java
kongweihan: Yes we recommend BigtableIO bug we need examples to show both connector as they're configured differently.

I'm hoping this example can demo the effect of flow control so that the job needs to write sufficient amount of data. Also user can play with the data shape with the parameters to understand flow control more if they want.

bigtable/beam/batch-write-flow-control-example/pom.xml
kongweihan: Done.

bigtable/beam/batch-write-flow-control-example/src/main/java/bigtable/BatchWriteFlowControlExample.java
kongweihan: This code sample shows how to enable the flow control feature. It has 2 parts because we have two different connectors and they're configured differently. I was unsure whether it'll worth duplicating the rest of the code to keep the sample simple. Please advice and I can split.

We're going to write the doc that will use this code sample.

Yes I'll add tests.

kongweihan: So far our example jobs all run async, is it a better practice to run sync? I'm happy to learn the reasoning of the practice.

minherz: > I was unsure whether it'll worth duplicating the rest of the code to keep the sample simple. Please advice and I can split.

It seems to me that both parts of the `if`/`else` block apply only two transformations where the first transformations are very similar and hard to distinguish in the current version.
I suggest either to look into an option to abstract these transformations (by inheriting from a proper Beam class) or, at least, to implement each of them in separate method with meaningful names.
Another option can be to implement each pipeline in a separate class from the start to the end.

> We're going to write the doc that will use this code sample.

Please either acquire a region tag or create a new tag using devrel/ site and add it to the code.

minherz: This is a code sample. It is a good practice to have the binary terminate _after_ the sampled behavior is complete.
If there is practical differences between implementing the batch flow control asynchronously and synchronously, consider creating multiple code samples that demonstrate the behaviors.

kongweihan: Split into 2 methods. Added region tag.

kongweihan: Done.

kongweihan:
> Hello, Please address the following questions:
> 
> 1. We ask to have a single code sample per file. This code looks like showing two code samples. What is this code sample demonstrates?
> 2. This code sample does not have regional tags. What documentation use it?
> 3. We do not host code samples without tests. What is a reason for lack of tests?

I think I've resolved the above comments. Any more thoughts on this? Thanks!

bigtable/beam/batch-write-flow-control-example/pom.xml
```diff
@@ -0,0 +1,127 @@
...
+    <artifactId>shared-configuration</artifactId>
+    <version>1.2.0</version>
+  </parent>
+
+  <dependencyManagement>
+    <dependencies>
+      <dependency>
+        <groupId>com.google.cloud</groupId>
+        <artifactId>google-cloud-bigtable-bom</artifactId>
+        <version>2.22.1-DEBUG-SNAPSHOT</version>
```
minherz, line 47: What is a reason to define a development version as a dependency? Additionally, this dependency overlaps with the `libraries-bom` dependency that includes (or should include) bigtable client library.

bigtable/beam/batch-write-flow-control-example/pom.xml
```diff
@@ -0,0 +1,127 @@
...
+        <type>pom</type>
+        <scope>import</scope>
+      </dependency>
+      <dependency>
+        <artifactId>libraries-bom</artifactId>
+        <groupId>com.google.cloud</groupId>
+        <scope>import</scope>
+        <type>pom</type>
+        <version>26.22.0</version>
```
minherz, line 56: what is a reason for using outdated BOM version? please, upgrade to the latest (AFAIK 26.40.0)

bigtable/beam/batch-write-flow-control-example/src/main/java/bigtable/BatchWriteFlowControlExample.java
```diff
@@ -0,0 +1,255 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package bigtable;
+
+import com.google.bigtable.v2.Mutation;
```
minherz, line 19: nit: consider adding START/END region tag comments to import statements as well in order to include the packages that are necessary for the code snippet to execute. It helps to documentation readers to identify the packages used for the code sample. See go/code-snippets-201#including-code-from-multiple-sections-of-a-file for technical details.

minherz (commented):
There are a couple of questions regarding pom.xml configuration. Apologies for missing them in the first review.
Please, have a look and check the versioning configuration as well as possible duplication of dependencies.

bigtable/beam/batch-write-flow-control-example/pom.xml
kongweihan: Removed.

bigtable/beam/batch-write-flow-control-example/pom.xml
kongweihan: Done.

bigtable/beam/batch-write-flow-control-example/src/main/java/bigtable/BatchWriteFlowControlExample.java
kongweihan: Added.

kongweihan:
@minherz Is there any other comments that I can address? Thank you!

minherz (approved):
lgtm. please review this regional tag in http://devrel if you have access

kongweihan:
> lgtm. please review this regional tag in http://devrel if you have access

What do you mean by reviewing it in http://devrel? IIUC after submit I'll notify our tech writer about the tags.

There're 2 pending reviewers in this PR, do you know how may I contact these reviewers? Thanks!

minherz:
> What do you mean by reviewing it in http://devrel? IIUC after submit I'll notify our tech writer about the tags.

I mean that the product (first word in the tag) is a registered product and there is no duplication or overlap with other code samples using the same region tag.

> There're 2 pending reviewers in this PR, do you know how may I contact these reviewers? Thanks!

I think that first matter of business should be to fix all failed Kokoro checks. You can ping @billyjacobson in the comment after that.

Issue #8950, opened 2023-12-21
InfiniteStreamRecognize - io.grpc.StatusRuntimeException: CANCELLED: Failed to read message when EnableVoiceActivityEvents is true


stefalbi:
## In which file did you encounter the issue?
https://github.com/GoogleCloudPlatform/java-docs-samples/blob/main/speech/src/main/java/com/example/speech/InfiniteStreamRecognize.java

## Did you change the file? If so, how?
Added enable voice activity event option and set audio channel count to 1

RecognitionConfig recognitionConfig =
          RecognitionConfig.newBuilder()
              .setEncoding(RecognitionConfig.AudioEncoding.LINEAR16)
              .setLanguageCode(languageCode)
              .setSampleRateHertz(16000)
              **.setAudioChannelCount(1)**
              .build();

StreamingRecognitionConfig streamingRecognitionConfig =
          StreamingRecognitionConfig.newBuilder()
              .setConfig(recognitionConfig)
              .setInterimResults(true)
              **.setEnableVoiceActivityEvents(true)
              .setVoiceActivityTimeout(VoiceActivityTimeout.newBuilder()
              		  							.setSpeechStartTimeout(Duration.newBuilder().setSeconds(15L))
            		  				  )**	
              .build();	

## Describe the issue
After I receive the event SPEECH_ACTIVITY_BEGIN I get the CancelledException:  Failed to read message.

minherz:
@stefalbi it will help us to triage the problem if you are able to answer to a few questions below:

1. Can you please confirm that the original example (without your changes) works as expected?
1. I assume that you added the links to [this code block][code1]. Can you run your code without setting VoiceActivityTimeout and report whether you get this exception?
1. I assume that the `CancelledException` is catched by the catch exception block at the [end of the sample][code2]. Can you please post the full content of the exception printed by the catch block?

Thank you

[code1]: https://github.com/GoogleCloudPlatform/java-docs-samples/blob/12e1b1ff3c933317f12e8bc24694140e50fdacf2/speech/src/main/java/com/example/speech/InfiniteStreamRecognize.java#L173-L184
[code2]: https://github.com/GoogleCloudPlatform/java-docs-samples/blob/12e1b1ff3c933317f12e8bc24694140e50fdacf2/speech/src/main/java/com/example/speech/InfiniteStreamRecognize.java#L296-L298

stefalbi:
1. yes, it works
[...]

PR #9336, opened 2024-05-20
Create test for a speech sample (InfiniteStreamRecognize)

### LINKED ISSUE #8261: Test needed for speech/src/main/java/com/example/speech/InfiniteStreamRecognize.java
This sample is missing a test.

ssvir:
## Description

Fixes #8261

## Checklist

### DIFF
--- a/speech/src/main/java/com/example/speech/InfiniteStreamRecognize.java
+++ b/speech/src/main/java/com/example/speech/InfiniteStreamRecognize.java
@@ -47,10 +47,6 @@ public class InfiniteStreamRecognize {
 
   private static final int STREAMING_LIMIT = 290000; // ~5 minutes
 
-  public static final String RED = "\033[0;31m";
-  public static final String GREEN = "\033[0;32m";
-  public static final String YELLOW = "\033[0;33m";
-
   // Creating shared object
   private static volatile BlockingQueue<byte[]> sharedQueue = new LinkedBlockingQueue<byte[]>();
   private static int BYTES_PER_BUFFER = 6400; // buffer size in bytes
@@ -68,7 +64,7 @@ public class InfiniteStreamRecognize {
   private static StreamController referenceToStreamController;
   private static ByteString tempByteString;
 
-  public static void main(String... args) {
+  public static void main(String... args) throws LineUnavailableException {
     InfiniteStreamRecognizeOptions options = InfiniteStreamRecognizeOptions.fromFlags(args);
     if (options == null) {
       // Could not parse.
@@ -76,8 +72,23 @@ public class InfiniteStreamRecognize {
       System.exit(1);
     }
 
+    // TODO(developer): Replace the variables before running the sample or use default
+    // the number of samples per second
+    int sampleRate = 16000;
+    // the number of bits in each sample
+    int sampleSizeInBits = 16;
+    // the number of channels (1 for mono, 2 for stereo, and so on)
+    int channels = 1;
+    // indicates whether the data is signed or unsigned
+    boolean signed = true;
+    // indicates whether the data for a single sample is stored in big-endian byte
+    // order (false means little-endian)
+    boolean bigEndian = false;
+
+    MicBuffer micBuffer = new MicBuffer(sampleRate, sampleSizeInBits, channels, signed, bigEndian);
     try {
-      infiniteStreamingRecognize(options.langCode);
+      infiniteStreamingRecognize(options.langCode, micBuffer, sampleRate,
+              RecognitionConfig.AudioEncoding.LINEAR16);
     } catch (Exception e) {
       System.out.println("Exception caught: " + e);
     }
@@ -95,26 +106,23 @@ public class InfiniteStreamRecognize {
                 - TimeUnit.MINUTES.toSeconds(TimeUnit.MILLISECONDS.toMinutes(millis))));
   }
 
-  public static void infiniteStreamingRecognize(String languageCode) throws Exception {
-    infiniteStreamingRecognize(languageCode, new MicBuffer(), getResponseObserver());
-  }
-
   /** Performs infinite streaming speech recognition */
-  public static void infiniteStreamingRecognize(String languageCode,
-                                     Runnable micBuffer,
-                                     ResponseObserver<StreamingRecognizeResponse> responseObserver)
-                                     throws Exception {
+  public static void infiniteStreamingRecognize(String languageCode, Runnable micBuffer,
+                                                int sampleRateHertz,
+                                                RecognitionConfig.AudioEncoding encoding)
+          throws Exception {
     // Creating microphone input buffer thread
     Thread micThread = new Thread(micBuffer);
     try (SpeechClient client = SpeechClient.create()) {
-      ClientStream<StreamingRecognizeRequest> clientStream;
-      clientStream = client.streamingRecognizeCallable().splitCall(responseObserver);
+      ResponseObserver<StreamingRecognizeResponse> responseObserver = getResponseObserver();
+      ClientStream<StreamingRecognizeRequest> clientStream = client.streamingRecognizeCallable()
+              .splitCall(responseObserver);
 
       RecognitionConfig recognitionConfig =
           RecognitionConfig.newBuilder()
-              .setEncoding(RecognitionConfig.AudioEncoding.LINEAR16)
+              .setEncoding(encoding)
               .setLanguageCode(languageCode)
-              .setSampleRateHertz(16000)
+              .setSampleRateHertz(sampleRateHertz)
               .build();
 
       StreamingRecognitionConfig streamingRecognitionConfig =
@@ -167,7 +175,6 @@ public class InfiniteStreamRecognize {
                     .setStreamingConfig(streamingRecognitionConfig)
                     .build();
 
-            System.out.println(YELLOW);
             System.out.printf("%d: RESTARTING REQUEST\n", restartCounter * STREAMING_LIMIT);
 
             startTime = System.currentTimeMillis();
@@ -215,7 +222,7 @@ public class InfiniteStreamRecognize {
 
             audioInput.add(tempByteString);
           }
-
+          Thread.sleep(100);
           clientStream.send(request);
         }
         clientStream.closeSend();
@@ -245,8 +252,7 @@ public class InfiniteStreamRecognize {
 
         SpeechRecognitionAlternative alternative = result.getAlternativesList().get(0);
         if (result.getIsFinal()) {
-          System.out.print(GREEN);
-          System.out.print("\033[2K\r");
+          System.out.print("\r");
           System.out.printf(
                   "%s: %s [confidence: %.2f]\n",
                   convertMillisToDate(correctedTime),
@@ -255,8 +261,7 @@ public class InfiniteStreamRecognize {
           isFinalEndTime = resultEndTimeInMS;
           lastTranscriptWasFinal = true;
         } else {
-          System.out.print(RED);
-          System.out.print("\033[2K\r");
+          System.out.print("\r");
           System.out.printf(
                   "%s: %s", convertMillisToDate(correctedTime), alternative.getTranscript());
           lastTranscriptWasFinal = false;
@@ -264,7 +269,9 @@ public class InfiniteStreamRecognize {
         checkStopRecognitionFlag(alternative.getTranscript().getBytes(StandardCharsets.UTF_8));
       }
 
-      public void onComplete() {}
+      public void onComplete() {
+        System.out.println("Recognition was stopped");
+      }
 
       public void onError(Throwable t) {}
     };
@@ -273,7 +280,7 @@ public class InfiniteStreamRecognize {
   private static void checkStopRecognitionFlag(byte[] flag) {
     if (flag.length < 10) {
       stopRecognition = new String(flag).trim().equalsIgnoreCase("exit");
-      putDataToSharedQueue("stop".getBytes(StandardCharsets.UTF_8));
+      putDataToSharedQueue("exit".getBytes(StandardCharsets.UTF_8));
     }
   }
 
@@ -281,10 +288,10 @@ public class InfiniteStreamRecognize {
   static class MicBuffer implements Runnable {
     TargetDataLine targetDataLine;
 
-    public MicBuffer() throws LineUnavailableException {
-      // SampleRate:16000Hz, SampleSizeInBits: 16, Number of channels: 1, Signed: true,
-      // bigEndian: false
-      AudioFormat audioFormat = new AudioFormat(16000, 16, 1, true, false);
+    public MicBuffer(int sampleRate, int sampleSizeInBits,
+                     int channels, boolean signed, boolean bigEndian)
+            throws LineUnavailableException {
+      AudioFormat audioFormat = new AudioFormat(sampleRate, sampleSizeInBits, channels, signed, bigEndian);
       DataLine.Info targetInfo =
               new Info(
                       TargetDataLine.class,
@@ -302,7 +309,6 @@ public class InfiniteStreamRecognize {
 
     @Override
     public void run() {
-      System.out.println(YELLOW);
       System.out.println("Start speaking...Say `exit` to stop");
       targetDataLine.start();
       byte[] data = new byte[BYTES_PER_BUFFER];
@@ -320,7 +326,8 @@ public class InfiniteStreamRecognize {
     try {
       sharedQueue.put(data.clone());
     } catch (InterruptedException e) {
-      System.out.println("Microphone input buffering interrupted : " + e.getMessage());
+      System.out.printf("Can't insert data to shared queue. Caused by : %s", e.getMessage());
+      throw new RuntimeException(e);
     }
   }
 }
--- a/speech/src/main/java/com/example/speech/InfiniteStreamRecognize.java
+++ b/speech/src/main/java/com/example/speech/InfiniteStreamRecognize.java
@@ -30,7 +30,6 @@ import com.google.cloud.speech.v1p1beta1.StreamingRecognizeRequest;
 import com.google.cloud.speech.v1p1beta1.StreamingRecognizeResponse;
 import com.google.protobuf.ByteString;
 import com.google.protobuf.Duration;
-
 import java.nio.charset.StandardCharsets;
 import java.text.DecimalFormat;
 import java.util.ArrayList;
@@ -101,8 +100,11 @@ public class InfiniteStreamRecognize {
   }
 
   /** Performs infinite streaming speech recognition */
-  public static void infiniteStreamingRecognize(String languageCode, Runnable micBuffer,
-                                                ResponseObserver<StreamingRecognizeResponse> responseObserver) throws Exception {
+  public static void infiniteStreamingRecognize
+                                    (String languageCode,
+                                     Runnable micBuffer,
+                                     ResponseObserver<StreamingRecognizeResponse> responseObserver)
+                                     throws Exception {
     // Creating microphone input buffer thread
     Thread micThread = new Thread(micBuffer);
     try (SpeechClient client = SpeechClient.create()) {
@@ -238,8 +240,7 @@ public class InfiniteStreamRecognize {
         StreamingRecognitionResult result = response.getResultsList().get(0);
         Duration resultEndTime = result.getResultEndTime();
         resultEndTimeInMS =
-                (int)
-                        ((resultEndTime.getSeconds() * 1000) + (resultEndTime.getNanos() / 1000000));
+                (int) ((resultEndTime.getSeconds() * 1000) + (resultEndTime.getNanos() / 1000000));
         double correctedTime =
                 resultEndTimeInMS - bridgingOffset + (STREAMING_LIMIT * restartCounter);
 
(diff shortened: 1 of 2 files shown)


speech/src/main/java/com/example/speech/InfiniteStreamRecognize.java
```diff
@@ -93,81 +95,19 @@ public static String convertMillisToDate(double milliSeconds) {
                 - TimeUnit.MINUTES.toSeconds(TimeUnit.MILLISECONDS.toMinutes(millis))));
   }
 
-  /** Performs infinite streaming speech recognition */
   public static void infiniteStreamingRecognize(String languageCode) throws Exception {
+    infiniteStreamingRecognize(languageCode, new MicBuffer(), getResponseObserver());
+  }
```
minherz, line 100: What is a value of using these two additional parameters instead of initiating them inside `infiniteStreamingRecognize`?
The current version makes the code sample more complicated to understand.
Please, refactor the code to remove unnecessary nested calls and initializations.
If your plan is to demonstrate both a creation of the response observer **AND** stream recognition implementation, consider creating a separate sample for the response observer in a stand alone file and reuse it in your code sample of stream recognition.

speech/src/test/java/com/example/speech/InfiniteStreamRecognizeTest.java
```diff
@@ -0,0 +1,72 @@
...
+  @Before
+  public void setUp() throws IOException, ExecutionException, InterruptedException,
+          TimeoutException {
+    bout = new ByteArrayOutputStream();
+    PrintStream out = new PrintStream(bout);
+    System.setOut(out);
+  }
+
+  @Test
+  public void infiniteStreamingRecognize() throws Exception {
```
minherz, line 51: the name of the test method should include "Test"

speech/src/test/java/com/example/speech/InfiniteStreamRecognizeTest.java
```diff
@@ -0,0 +1,72 @@
...
+    InfiniteStreamRecognize.putDataToSharedQueue(readAudioFile());
+    InfiniteStreamRecognize.putDataToSharedQueue("exit".getBytes(StandardCharsets.UTF_8));
```
minherz, line 53: Are these lines a part of the test or a part of the setup?

speech/src/main/java/com/example/speech/InfiniteStreamRecognize.java
```diff
@@ -293,10 +218,110 @@ public void onError(Throwable t) {}
...
+                  alternative.getTranscript(),
+                  alternative.getConfidence());
+          isFinalEndTime = resultEndTimeInMS;
+          lastTranscriptWasFinal = true;
+        } else {
+          System.out.print(RED);
+          System.out.print("\033[2K\r");
+          System.out.printf(
+                  "%s: %s", convertMillisToDate(correctedTime), alternative.getTranscript());
+          lastTranscriptWasFinal = false;
```
minherz, line 262: this is a bad practice for code samples. because

* it leads to flaky tests because it requires parsing captured `stdout` output which is not deterministic.
* it does not add to verbosity of code sample because not everyone who looks into the code will run it to check the actual output
* it creates overhead in understanding these lines while they do not contribute to the code sample

If these prints do not have the meaning in the scope of building the response observer, please remove it. If they serve some role, please refactor the code and make it more verbose and easy to understand. For example, consider using a format template and give it self-explaining name.

speech/src/main/java/com/example/speech/InfiniteStreamRecognize.java
```diff
@@ -293,10 +218,110 @@ public void onError(Throwable t) {}
...
+
+  // Microphone Input buffering
+  static class MicBuffer implements Runnable {
+    TargetDataLine targetDataLine;
+
+    public MicBuffer() throws LineUnavailableException {
+      // SampleRate:16000Hz, SampleSizeInBits: 16, Number of channels: 1, Signed: true,
+      // bigEndian: false
+      AudioFormat audioFormat = new AudioFormat(16000, 16, 1, true, false);
+      DataLine.Info targetInfo =
```
minherz, line 288: Using hard-coded parameters in the code sample leads to flaky tests because the support for parameters can change with the time. Unless there is a specific requirement for exactly these numbers, consider to convert them to arguments passed from the main method.
If they are constants, please define them as `final` constants with meaningful names.

speech/src/main/java/com/example/speech/InfiniteStreamRecognize.java
```diff
@@ -293,10 +218,110 @@ public void onError(Throwable t) {}
...
+                  "%s: %s", convertMillisToDate(correctedTime), alternative.getTranscript());
+          lastTranscriptWasFinal = false;
+        }
+        checkStopRecognitionFlag(alternative.getTranscript().getBytes(StandardCharsets.UTF_8));
+      }
+
+      public void onComplete() {}
+
+      public void onError(Throwable t) {}
+    };
```
minherz, line 270: Do you need to define these empty methods?
What is a recommended error handling technique that the sample can demonstrate?

speech/src/main/java/com/example/speech/InfiniteStreamRecognize.java
```diff
@@ -293,10 +218,110 @@ public void onError(Throwable t) {}
...
+        putDataToSharedQueue(data.clone());
+      }
+    }
+  }
+
+  public static void putDataToSharedQueue(byte[] data) {
+    try {
+      sharedQueue.put(data.clone());
+    } catch (InterruptedException e) {
+      System.out.println("Microphone input buffering interrupted : " + e.getMessage());
```
minherz, line 323: nit: Should the sample terminate in this case?

speech/src/test/java/com/example/speech/InfiniteStreamRecognizeTest.java
```diff
@@ -0,0 +1,72 @@
...
+    InfiniteStreamRecognize.infiniteStreamingRecognize("en-US",
+            () -> {}, InfiniteStreamRecognize.getResponseObserver());
+    // wait responses from server
+    Thread.sleep(5000);
+    assertThat(bout.toString().toLowerCase()).contains("hi i want to");
+  }
+
+  private byte[] readAudioFile() throws UnsupportedAudioFileException, IOException {
+    File file = new File("resources/commercial_mono.wav");
```
minherz, line 62: Please, move the path to the beginning of the file as a constant. It will make it more discoverable in event we are investigating resources used by tests.

speech/src/test/java/com/example/speech/InfiniteStreamRecognizeTest.java
```diff
@@ -0,0 +1,72 @@
...
+import javax.sound.sampled.AudioFormat;
+import javax.sound.sampled.AudioInputStream;
+import javax.sound.sampled.AudioSystem;
+import javax.sound.sampled.UnsupportedAudioFileException;
+import org.junit.Before;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.junit.runners.JUnit4;
+
+@RunWith(JUnit4.class)
```
minherz, line 38: Is it necessary "to pin" the specific version of JUnit runner? Placing it here means that when we migrate our test suite to JUnit5 this test will continue using JUnit4 runner instead. And we will have to manually change each test file having this argument. Unless this test uses some specific JUnit4 feature that is expected to be out of support in JUnit5, please remove this attribute.

minherz (changes requested):

minherz (changes requested):
Why this code sample does not have region tags?
This repository is intended for code samples that are used with documentation. Please, check for appropriate region tag.

speech/src/test/java/com/example/speech/InfiniteStreamRecognizeTest.java
ssvir: As a part of the test, readAudioFile(), we read the test file to recognise and in the next line, we command to stop recognition

speech/src/main/java/com/example/speech/InfiniteStreamRecognize.java
ssvir: I'm planning to investigate this part of an observer inside this issue https://github.com/GoogleCloudPlatform/java-docs-samples/issues/8950

speech/src/main/java/com/example/speech/InfiniteStreamRecognize.java
```diff
@@ -63,19 +60,35 @@ public class InfiniteStreamRecognize {
...
+    boolean signed = true;
+    // indicates whether the data for a single sample is stored in big-endian byte
+    // order (false means little-endian)
+    boolean bigEndian = false;
+
+    MicBuffer micBuffer = new MicBuffer(sampleRate, sampleSizeInBits, channels, signed, bigEndian);
     try {
-      infiniteStreamingRecognize(options.langCode);
+      infiniteStreamingRecognize(options.langCode, micBuffer, sampleRate,
+              RecognitionConfig.AudioEncoding.LINEAR16);
```
minherz, line 91: I understand that this method exits when a stream has a word "exit".
It would be useful to document it in the code. Consider renaming variable to be more verabit (eg "exitRecognized") or write a comment.
nit: Otherwise, if a user ends up killing the app, a graceful termination on SIGKILL might be useful if resources require release.

speech/src/main/java/com/example/speech/InfiniteStreamRecognize.java
```diff
@@ -285,18 +215,121 @@ public void onError(Throwable t) {}
...
+      public void onComplete() {
+        System.out.println("Recognition was stopped");
+      }
+
+      public void onError(Throwable t) {}
+    };
+  }
+
+  private static void checkStopRecognitionFlag(byte[] flag) {
+    if (flag.length < 10) {
```
minherz, line 281: nit: it would be nice to make more clear why `10` is selected as a threshold. consider using a constant with verbatim name or another method.

speech/src/main/java/com/example/speech/InfiniteStreamRecognize.java
```diff
@@ -285,18 +215,121 @@ public void onError(Throwable t) {}
...
+      public void onResponse(StreamingRecognizeResponse response) {
+        responses.add(response);
+        StreamingRecognitionResult result = response.getResultsList().get(0);
+        Duration resultEndTime = result.getResultEndTime();
+        resultEndTimeInMS =
+                (int) ((resultEndTime.getSeconds() * 1000) + (resultEndTime.getNanos() / 1000000));
+        double correctedTime =
+                resultEndTimeInMS - bridgingOffset + (STREAMING_LIMIT * restartCounter);
+
+        SpeechRecognitionAlternative alternative = result.getAlternativesList().get(0);
```
minherz, line 253: what happens if the list have more than 1 element

speech/src/main/java/com/example/speech/InfiniteStreamRecognize.java
```diff
@@ -285,18 +215,121 @@ public void onError(Throwable t) {}
...
+                  alternative.getConfidence());
+          isFinalEndTime = resultEndTimeInMS;
+          lastTranscriptWasFinal = true;
+        } else {
+          System.out.print("\r");
+          System.out.printf(
+                  "%s: %s", convertMillisToDate(correctedTime), alternative.getTranscript());
+          lastTranscriptWasFinal = false;
+        }
+        checkStopRecognitionFlag(alternative.getTranscript().getBytes(StandardCharsets.UTF_8));
```
minherz, line 269: this method will fail if the "exit" word is not in the first element. can such thing happen? can the code reflect that?

speech/src/test/java/com/example/speech/InfiniteStreamRecognizeTest.java
```diff
@@ -0,0 +1,72 @@
...
+  @Test
+  public void infiniteStreamingRecognizeTest() throws Exception {
+    InfiniteStreamRecognize.putDataToSharedQueue(readAudioFile());
+    InfiniteStreamRecognize.putDataToSharedQueue("exit".getBytes(StandardCharsets.UTF_8));
+    InfiniteStreamRecognize.infiniteStreamingRecognize("en-US",
+            () -> {}, 16000, RecognitionConfig.AudioEncoding.LINEAR16);
+    // wait responses from server
+    Thread.sleep(10000);
+    assertThat(bout.toString().toLowerCase()).contains("hi i want to");
+    assertThat(bout.toString().toLowerCase()).contains("recognition was stopped");
```
minherz, line 58: I listen to the commercial_mono.wav file and it does not contain a word "exit". Does it mean that `onComplete()` is called when the stream ends _OR_ stopped?
I recommend to consider distinguishing these two events. Currently the impression from the code sample is that onComplete() is called when the recognition is stopped.

minherz (commented):
Thank you for addressing majority of items.
I still see that testing uses stdout buffer but I recognize that in this particular code sample other exit statuses are harder to implement.

I found a potential bug when calling "checking stop recognition flag" in event the response contains multiple items. Please, look into that.

speech/src/main/java/com/example/speech/InfiniteStreamRecognize.java
ssvir: I added a comment on how to stop, and this will also be printed when the application runs.

speech/src/main/java/com/example/speech/InfiniteStreamRecognize.java
ssvir: May contain one or more recognition hypotheses (up to the
maximum specified in `max_alternatives`).
These alternatives are ordered in terms of accuracy, with the top (first)
alternative being the most probable, as ranked by the recognizer.

speech/src/main/java/com/example/speech/InfiniteStreamRecognize.java
ssvir: this code reflects only if the word "exit" will be present in a sentence , that's why we check element length before comparing

speech/src/test/java/com/example/speech/InfiniteStreamRecognizeTest.java
ssvir: As we have InfiniteStreamRecognize, I suppose we don't have some status-completed, so this code can call `onComplete()` only in one way when we stop the recognition

speech/src/main/java/com/example/speech/InfiniteStreamRecognize.java
minherz: We have guidelines how code sample should look like. This repository is intended to host code used mainly for documentation code snippets. See go/code-snippets-101#what-is-a-code-snippet for the definition. It is not intended for demo applications or other complex solutions.
I agree that _printing_ it makes more sense than place a comment. However, I see no code line that prints this instruction.

speech/src/main/java/com/example/speech/InfiniteStreamRecognize.java
```diff
@@ -285,18 +218,121 @@ public void onError(Throwable t) {}
...
 
+            checkStopRecognitionFlag(tempByteString.toByteArray());
+
             request =
                 StreamingRecognizeRequest.newBuilder().setAudioContent(tempByteString).build();
 
             audioInput.add(tempByteString);
           }
-
+          Thread.sleep(100);
```
minherz, line 228: What is a reason for this delay? Why the delay is 100 ms? Code samples have to be self-explanatory. It is hard to understand from the code whether this delay is required and why 100 ms value is selected.

speech/src/main/java/com/example/speech/InfiniteStreamRecognize.java
```diff
@@ -39,19 +40,16 @@
 import javax.sound.sampled.AudioSystem;
 import javax.sound.sampled.DataLine;
 import javax.sound.sampled.DataLine.Info;
+import javax.sound.sampled.LineUnavailableException;
 import javax.sound.sampled.TargetDataLine;
 
 public class InfiniteStreamRecognize {
 
   private static final int STREAMING_LIMIT = 290000; // ~5 minutes
```
minherz, line 48: nit: This code sample is called _infinite_ streaming. The name "streaming limit" can be potentially confusing. Consider providing more meaningful name.

speech/src/main/java/com/example/speech/InfiniteStreamRecognize.java
```diff
@@ -39,19 +40,16 @@
...
-
-  public static final String RED = "\033[0;31m";
-  public static final String GREEN = "\033[0;32m";
-  public static final String YELLOW = "\033[0;33m";
+  private static final String EXIT_WORD = "exit";
 
   // Creating shared object
   private static volatile BlockingQueue<byte[]> sharedQueue = new LinkedBlockingQueue<byte[]>();
-  private static TargetDataLine targetDataLine;
   private static int BYTES_PER_BUFFER = 6400; // buffer size in bytes
```
minherz, line 53: nit: Consider selecting multiple of 2 for the buffer size as a general guideline. Is it possible to explain how to select a certain value?

speech/src/main/java/com/example/speech/InfiniteStreamRecognize.java
```diff
@@ -63,19 +61,37 @@ public class InfiniteStreamRecognize {
   private static boolean newStream = true;
   private static double bridgingOffset = 0;
   private static boolean lastTranscriptWasFinal = false;
+  private static boolean stopRecognition = false;
   private static StreamController referenceToStreamController;
   private static ByteString tempByteString;
```
minherz, line 66: All constants in Java should use [UPPER_SNAKE_CASE](https://engdoc.corp.google.com/eng/doc/devguide/java/style/index.md?cl=head#s5.2.4-constant-names)

speech/src/main/java/com/example/speech/InfiniteStreamRecognize.java
minherz: My question was related to line 256. Is it possible that other, less confident, alternatives interpret "exit" while the first one does not?

minherz (commented):
Please provide region tags for this code snippet.
Please follow guidelines from go/code-snippets-201#including-code-from-multiple-sections-of-a-file to exclude the main method and all auxiliary methods from the code snippet.
If there is no intention to use this sample as a code snippet, please consider using a different repository to host this code.

speech/src/main/java/com/example/speech/InfiniteStreamRecognize.java
ssvir: I can't explain how to select a certain value.

speech/src/main/java/com/example/speech/InfiniteStreamRecognize.java
ssvir: I don't have google acc :(

ssvir: @minherz By constant, do you mean the variable marked as final or static is constant too?

speech/src/main/java/com/example/speech/InfiniteStreamRecognize.java
ssvir: I think we have some possibility, so in this case, we should repeat "exit".

speech/src/main/java/com/example/speech/InfiniteStreamRecognize.java
ssvir: Could you please send a direct link, or if this link needs google account, could you please copy it here a phrase

speech/src/main/java/com/example/speech/InfiniteStreamRecognize.java
minherz: > I don't have google acc :(

I looked into the case closer. It looks like you made changes to the code sample itself and not only added the test case. I will ask a Googler who opened the issue to review the changes you made to the code sample itself.


> @minherz By constant, do you mean the variable marked as final or static is constant too?

I mean "constants" in Java. It should be literals declared with `const` or objects defined with `final`. Static or not static should not play a role.

ssvir: Yeap, I refactored this sample to have a possibility to replace the voice recognition thread with a file or another source of an audio byte stream

speech/src/test/java/com/example/speech/InfiniteStreamRecognizeTest.java
minherz: Integration tests are supposed to be deterministic, terminated, run minimal time required to execute the test and do a proper setup and clean up action so they aren't dependent on the state of the testing environment and do not consume resources beyond required.
From your comment it sounds like the test might not terminate, making it a subject for flakiness.

Please, reconsider the testing strategy because we prefer not to have a test at all to having a flaky test.

speech/src/test/java/com/example/speech/InfiniteStreamRecognizeTest.java
minherz: Do you mean that the reading audio file and adding _text_ to the already read binary audio data is a part of the test? What do these lines test?

minherz (changes requested):
It is unclear whether the proposed testing strategy is deterministic. Loading audio file bytes and then adding text string to the same stream raises concerns about test performance.
With the current open questions and lack of access to the internal information regarding best practices for code samples I prefer to hold this change for now.
@Sita04 please have a look

telpirion:
It looks like this PR has been abandoned.  @ssvir please re-open if you intend to continue working on it. Thank you!

PR #9355, opened 2024-06-03
chore(proc): add periodic test config for TPC testing

minherz:
## Description

Adds kokoro configuration files (SCM side) to support testing a subset of the tests on PRPTST universe.
The new configuration is copied from java17 config files and modified to do the following:

* Reduce the timeout
* Use "local" bash script
* Execute testing using PRPTST project `java-docs-samples-testing` with the service account created in that project
* Retrieve the key from the "original" (on GDU) `java-docs-samples-testing` project
* Remove all unused commands and instructions
* Remove use of `btlr`

### DIFF
--- a/.kokoro/prptst/run_tests.sh
+++ b/.kokoro/prptst/run_tests.sh
@@ -1,0 +1,114 @@
+#!/bin/bash
+# Copyright 2024 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# `-e` enables the script to automatically fail when a command fails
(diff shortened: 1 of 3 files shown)


.kokoro/prptst/run_tests.sh
```diff
@@ -0,0 +1,114 @@
...
+# shellcheck disable=SC2076
+if ! [[ ",$JAVA_VERSION," =~ ",$POM_JAVA," ]]; then
+    echo -e "\n Skipping tests: Java version ($POM_JAVA) not required ($JAVA_VERSION)\n"
+    exit 0
+fi
+
+if [[ (",$JAVA_VERSION," =~ "17" || ",$JAVA_VERSION," =~ "21")  && ( "$file" == *"run/hello-broken"* || "$file" == *"flexible/java-11/pubsub"* || "$file" == *"flexible/java-11/cloudstorage"*|| "$file" == *"flexible/java-11/datastore"*) ]]; then
+    echo -e "\n Skipping tests: Sample ($file) tests do not work with Java runtimes 17 or greater\n"
+    exit 0
+fi
```
Sita04, line 91: Do you intend to retain this check to expand the test suite later on?
Or can this be removed as currently only compute is being tested in the universe.

Sita04 (approved):

.kokoro/prptst/run_tests.sh
minherz: I tried to keep a balance between a "reference solution" and a proof of concept. Most of the instructions related to the runtime can be removed. I left them "as is" partially to avoid possible conflicts with the configuration and partially because it will be useful if in (near) future someone will want to extend this script to additional runtimes.

PR #9366, opened 2024-06-06
chore: upgrade bom to latest

minherz:
## Description

Upgrade libraries-bom to the latest version in order to have universe domain support in Cloud SDK.


minherz:
After upgrading to the latest BOM tests for two classes start to fail. I discovered that both tests call **unitested** methods (code samples) in their setup routines (marked with `@BeforeAll` attribute). The test classes are temporary disabled until the tests are fixed. See issues #9373 and #9374 for additional information.

PR #9409, opened 2024-07-11
fix(spanner): Explicit declaration of dependencies for OpenTelemetry traces sample

nareshz:
This PR includes changes to add all the required dependencies for OpenTelemetry traces sample explicitly.

### DIFF
--- a/spanner/opentelemetry_traces/pom.xml
+++ b/spanner/opentelemetry_traces/pom.xml
@@ -29,7 +29,7 @@
       <dependency>
         <groupId>com.google.cloud</groupId>
         <artifactId>libraries-bom</artifactId>
-        <version>26.34.0</version>
+        <version>26.43.0</version>
         <type>pom</type>
         <scope>import</scope>
       </dependency>
@@ -51,7 +51,15 @@
     </dependency>
     <dependency>
       <groupId>io.opentelemetry</groupId>
+      <artifactId>opentelemetry-api</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>io.opentelemetry</groupId>
       <artifactId>opentelemetry-sdk</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>io.opentelemetry</groupId>
+      <artifactId>opentelemetry-sdk-common</artifactId>
     </dependency>
     <dependency>
       <groupId>io.opentelemetry</groupId>
@@ -70,7 +78,15 @@
     </dependency>
     <dependency>
       <groupId>io.opentelemetry</groupId>
+      <artifactId>opentelemetry-api</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>io.opentelemetry</groupId>
       <artifactId>opentelemetry-sdk</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>io.opentelemetry</groupId>
+      <artifactId>opentelemetry-sdk-common</artifactId>
     </dependency>
     <dependency>
       <groupId>io.opentelemetry</groupId>


Sita04:
@minherz PTAL at these changes.

nareshz:
@olavloite Can you please take a look at your convenience?

spanner/opentelemetry_traces/pom.xml
```diff
@@ -68,10 +76,18 @@
       <groupId>com.google.cloud</groupId>
       <artifactId>google-cloud-spanner</artifactId>
     </dependency>
+    <dependency>
+      <groupId>io.opentelemetry</groupId>
+      <artifactId>opentelemetry-api</artifactId>
+    </dependency>
```
minherz, line 82: why this dependency is included two times (see lines 52-55)?

minherz (commented):
what is a reason for adding all these dependencies?
is it the best practice to be followed?

spanner/opentelemetry_traces/pom.xml
nareshz: In documentation, we want to show dependencies required for using Cloud Trace exporter and OTLP exporter seperately. That's why two tags are defined in this pom.xml file - spanner_opentelemetry_traces_otlp_dependencies(otlp exporter dependencies) and spanner_opentelemetry_traces_cloudtrace_dependencies(cloud trace exporter dependencies). There maybe common dependency across both.

nareshz:
> what is a reason for adding all these dependencies? is it the best practice to be followed?

Yes, right. The reason is mainly to follow good practice of explicitly declaring all the dependencies required.

minherz:
> > what is a reason for adding all these dependencies? is it the best practice to be followed?
> 
> Yes, right. The reason is mainly to follow good practice of explicitly declaring all the dependencies required.

I am not aware of such good practice. Can you please reference me to Google documentation that defines it? If not, please provide another widely accepted guidance reference.

Thank you

spanner/opentelemetry_traces/pom.xml
```diff
@@ -49,10 +49,18 @@
       <groupId>com.google.cloud</groupId>
       <artifactId>google-cloud-spanner</artifactId>
     </dependency>
+    <dependency>
+      <groupId>io.opentelemetry</groupId>
+      <artifactId>opentelemetry-api</artifactId>
```
surbhigarg92, line 54: This dependency is already part of Spanner Java Client library. Why do we need to add this again here ?

spanner/opentelemetry_traces/pom.xml
```diff
@@ -49,10 +49,18 @@
...
+    </dependency>
     <dependency>
       <groupId>io.opentelemetry</groupId>
       <artifactId>opentelemetry-sdk</artifactId>
     </dependency>
+    <dependency>
+      <groupId>io.opentelemetry</groupId>
+      <artifactId>opentelemetry-sdk-common</artifactId>
```
surbhigarg92, line 62: Can you please help me remind and what was this needed for ?

nareshz: This is needed for the `io.opentelemetry.sdk.resources.Resource` class.

spanner/opentelemetry_traces/pom.xml
nareshz: As discussed offline with you @surbhigarg92 that we should only declare dependencies that needs to be added by customers, not making any changes to explicitly declare the dependencies.

spanner/opentelemetry_traces/pom.xml
minherz: The marked dependencies include the artifacts with the same groupId and artifactId. It does not matter how many components use it. They all have to use **the same** dependency. Creating multiple dependencies of the same artifact can lead to using several versions of the artifact. It is bad practice and should not be used.

PR #9419, opened 2024-07-24
docs(generativeai): add region tags for code samples

msampathkumar:
## Description

Adding `generativeaionvertexai_` region tags to replace old `aiplatform_`

In near future GenAI code samples, will only use `generativeaionvertexai_` region tag prefix.

Fixes: [b/350918509](b/350918509)

## Checklist

### DIFF
--- a/aiplatform/src/main/java/aiplatform/CreatePipelineJobModelTuningSample.java
+++ b/aiplatform/src/main/java/aiplatform/CreatePipelineJobModelTuningSample.java
@@ -17,6 +17,7 @@
 package aiplatform;
 
 // [START aiplatform_sdk_tuning]
+// [START generativeaionvertexai_sdk_tuning]
 import com.google.cloud.aiplatform.v1.CreatePipelineJobRequest;
 import com.google.cloud.aiplatform.v1.LocationName;
 import com.google.cloud.aiplatform.v1.PipelineJob;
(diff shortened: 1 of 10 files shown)


aiplatform/src/main/java/aiplatform/CreatePipelineJobModelTuningSample.java
```diff
@@ -116,3 +117,4 @@ static Value numberToValue(int n) {
 }
 
 // [END aiplatform_sdk_tuning]
+// [END generativeaionvertexai_sdk_tuning]
```
Sita04, line 120: IIUC, the region tags follow LIFO order. The last START region tag should be the first one to END?
Is this assumption right?

Sita04 (changes requested):

aiplatform/src/main/java/aiplatform/CreatePipelineJobModelTuningSample.java
msampathkumar: Hi @Sita04  Thank you. `LIFO order` is not necessary for region tags. https://github.com/GoogleCloudPlatform/python-docs-samples/blob/main/cdn/snippets.py is a good example. (SnippetBot also accepts this order)

In an upcoming request noted at [b/355457593](http://b/355457593), we will be clearing these unwanted region tags with prefix `aiplatform_`. So I request your understanding here.

msampathkumar: Marking this as resolved.

rsamborski (approved):

Sita04 (approved):

PR #9420, opened 2024-07-25
Remove execute/assessment from homepage load

faube:
Remove execute/assessment from homepage load.

This is no longer something we want to showcase in reCAPTCHA demosite.


Sita04:
@faube Is this change being made in the other languages as well? Has this been internally tested?

faube:
Yes, yes.

Sita04 (approved):

PR #9424, opened 2024-07-31
docs(samples): Add Kafka to Dataflow snippet

VeronicaWasson:
## Description

Add code snippets for reading from Kafka using the Dataflow managed I/O transform.

Doc bugs:

- b/338063763: Managed I/O documentation [includes Kafka support]
- b/342015050: Document best practices for using Kafka with Dataflow
- 
## Checklist

### DIFF
--- a/dataflow/snippets/pom.xml
+++ b/dataflow/snippets/pom.xml
@@ -156,6 +156,24 @@
       <version>${iceberg.version}</version>
     </dependency>
 
+    <!-- Kafka -->
+    <dependency>
+      <groupId>org.apache.beam</groupId>
+      <artifactId>beam-sdks-java-io-kafka</artifactId>
+      <version>${apache_beam.version}</version>
+    </dependency>
+    <dependency>
+      <groupId>org.apache.kafka</groupId>
+      <artifactId>kafka-clients</artifactId>
+      <version>3.8.0</version>
+    </dependency>
+    <dependency>
+      <groupId>org.testcontainers</groupId>
+      <artifactId>kafka</artifactId>
+      <version>1.20.0</version>
+      <scope>test</scope>
+    </dependency>
+
     <!-- Google Cloud -->
     <dependency>
       <groupId>com.google.cloud</groupId>
--- a/dataflow/snippets/pom.xml
+++ b/dataflow/snippets/pom.xml
@@ -163,11 +163,14 @@
       <version>${apache_beam.version}</version>
     </dependency>
     <dependency>
+      <!-- For sending Kafka messages in the integration test -->
       <groupId>org.apache.kafka</groupId>
       <artifactId>kafka-clients</artifactId>
       <version>3.8.0</version>
+      <scope>test</scope>
     </dependency>
     <dependency>
+      <!-- For running containerized Kafka instance in the integration test -->
       <groupId>org.testcontainers</groupId>
       <artifactId>kafka</artifactId>
       <version>1.20.0</version>
--- a/dataflow/snippets/src/main/java/com/example/dataflow/KafkaRead.java
+++ b/dataflow/snippets/src/main/java/com/example/dataflow/KafkaRead.java
@@ -31,7 +31,7 @@ import org.apache.beam.sdk.values.TypeDescriptors;
 
 public class KafkaRead {
 
-// [END dataflow_kafka_read]
+  // [END dataflow_kafka_read]
   public interface Options extends StreamingOptions {
     @Description("The Kafka bootstrap server. Example: localhost:9092")
     String getBootstrapServer();
@@ -93,8 +93,7 @@ public class KafkaRead {
             .write()
             .to(options.getOutputPath())
             .withSuffix(".txt")
-            .withNumShards(1)
-        );
+            .withNumShards(1));
     return pipeline;
   }
 }
--- a/dataflow/snippets/src/main/java/com/example/dataflow/KafkaRead.java
+++ b/dataflow/snippets/src/main/java/com/example/dataflow/KafkaRead.java
@@ -1,0 +1,91 @@
+/*
+ * Copyright 2024 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.example.dataflow;
+
+// [START dataflow_kafka_read]
+import com.google.common.collect.ImmutableMap;
+import java.io.UnsupportedEncodingException;
+import org.apache.beam.sdk.Pipeline;
+import org.apache.beam.sdk.io.TextIO;
+import org.apache.beam.sdk.managed.Managed;
+import org.apache.beam.sdk.options.Description;
+import org.apache.beam.sdk.options.PipelineOptionsFactory;
+import org.apache.beam.sdk.options.StreamingOptions;
+import org.apache.beam.sdk.transforms.MapElements;
+import org.apache.beam.sdk.values.TypeDescriptors;
+
+public class KafkaRead {
+
+  public interface Options extends StreamingOptions {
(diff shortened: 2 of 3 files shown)


dataflow/snippets/pom.xml
```diff
@@ -156,6 +156,24 @@
...
+    </dependency>
+    <dependency>
+      <groupId>org.apache.kafka</groupId>
+      <artifactId>kafka-clients</artifactId>
+      <version>3.8.0</version>
+    </dependency>
+    <dependency>
+      <groupId>org.testcontainers</groupId>
+      <artifactId>kafka</artifactId>
+      <version>1.20.0</version>
```
minherz, line 173: nit: is there a dependency between these two versions? should we set kafka client version as a variable?

dataflow/snippets/src/main/java/com/example/dataflow/KafkaRead.java
```diff
@@ -0,0 +1,91 @@
...
+        // Write the payload to a text file.
+        .apply(TextIO
+            .write()
+            .to(options.getOutputPath())
+            .withSuffix(".txt")
+            .withNumShards(1)
+    );
+    pipeline.run().waitUntilFinish();
+  }
+}
```
minherz, line 90: please move the code sample code to a stand-alone method.
call this method from the main pipeline.
implement a simple response in the method.
consider returning a response that can be easily validated in testing.

dataflow/snippets/src/main/java/com/example/dataflow/KafkaRead.java
```diff
@@ -0,0 +1,91 @@
...
+        .put("topic", options.getTopic())
+        .put("data_format", "RAW")
+        .put("max_read_time_seconds", 15)
+        .put("auto_offset_reset_config", "earliest")
+        .build();
+
+    // Build the pipeline.
+    pipeline
+        // Read messages from Kafka.
+        .apply(Managed.read(Managed.KAFKA).withConfig(config)).getSinglePCollection()
```
minherz, line 69: nit: I am not SME in using Kafka with Apache Beam (or Dataflow). What is a reason for using configuration two times: one as Options and another in the first `apply()` call?

minherz (commented):
I found no explanation regarding these code samples or the usage of the `dataflow_kafka_read` tag.
It seems that the use of the managed.managed.KAFKA isn't required to read from Kafka in Apache Beam or in Dataflow specifically. Can you provide more context for the decision to make the implementation "as is"?
Will it work the same way in Python and Go code samples?
I did not find the answer to these questions in the referenced issues.

dataflow/snippets/pom.xml
VeronicaWasson: As far as I know, there's no dependency. The kafka-clients module is for sending messages to Kafka, the testcontainers module is to spin up a Kafka container (these are both for the IT). So they only interact via Kafka which should be decoupled.

VeronicaWasson:
> I found no explanation regarding these code samples or the usage of the `dataflow_kafka_read` tag. It seems that the use of the managed.managed.KAFKA isn't required to read from Kafka in Apache Beam or in Dataflow specifically. Can you provide more context for the decision to make the implementation "as is"? Will it work the same way in Python and Go code samples? I did not find the answer to these questions in the referenced issues.

This isn't a Managed Kafka sample, it's showing reading Kafka from Dataflow. It should be use-able with Managed Kafka but that's not a dependency. 

The Java version specifically uses the [Managed I/O transform](/dataflow/docs/guides/managed-io) which is new. For now the Python version will use an older version of the Kafka I/O, as there is currently no Managed I/O transform for Python. (Same for Go although a Go version isn't currently planned.)

Relevant doc bugs: 

- b/338063763: Managed IO GA documentation [includes Kafka support]
- b/342015050: Document best practices for using Kafka with Dataflow

dataflow/snippets/src/main/java/com/example/dataflow/KafkaRead.java
VeronicaWasson: I could move the pipeline code into a separate method, but I need to include the Main method in the snippet. The Beam runner (which includes the Dataflow runner) invokes the Main method, so that's part of the sample.

Building the pipeline in Main is idiomatic for Beam: https://github.com/apache/beam/blob/master/examples/java/src/main/java/org/apache/beam/examples/MinimalWordCount.java

I'll update to return the pipeline result.

dataflow/snippets/pom.xml
minherz: Thank you. Since both these dependencies seem to be used for testing, please add the `test` scope to the `kafka-clients` dependency.

dataflow/snippets/src/main/java/com/example/dataflow/KafkaRead.java
minherz: I understand. We distinguish between the code sample and code snippet. The latter is displayed in the documentation while the former is a fully compilable and executable code.
Unless there is something specific about the code in the `main()` method of _any_ Java application that you want to demonstrate, the startup code (i.e. `main` method and other auxiliary lines) should be excluded from the snippet. This way the documentation shows only the lines that are relevant in the context and a reader get access the full code sample via a link to Github provided in the snippet widget.

dataflow/snippets/src/main/java/com/example/dataflow/KafkaRead.java
minherz: According to your clarifications, the use of `Options` is done for convenience so you can avoid coding a complex parsing logic. Given explanations provided in the [previous comment](https://github.com/GoogleCloudPlatform/java-docs-samples/pull/9424#discussion_r1714640234), please consider separating the parsing logic that effectively can be left at user's discretion, from the recommended method of reading from Kafka that requires the "property bag of configuration parameters". I would recommend to pass the parameters as arguments to the method that will be shown in the code snippet. In this form, the parsing of the arguments that you choose to use will not be part of the snippet and users will be able to implement it at their discretion.
This refactoring will also simplify testing the method since parameters can be explicitly set up in the testing method.

minherz:
Thank you for your clarifications. Please, follow up my feedback in the lines comments above.

dataflow/snippets/pom.xml
VeronicaWasson: done

dataflow/snippets/src/main/java/com/example/dataflow/KafkaRead.java
VeronicaWasson: I need to pass a PipelineOptions regardless, because that tells Beam which runner to use, along with any Dataflow runtime options, streaming options, etc. For testing I'm using the DirectRunner, and Beam parses that via `Pipeline.create(options)`. 

The standard way to pass custom options to a Dataflow pipeline is by extending PipelineOptions, as shown. Splitting them out from the pipeline options isn't a best practice.

dataflow/snippets/src/main/java/com/example/dataflow/KafkaRead.java
VeronicaWasson: I trimmed out main() by moving the logic for running the pipeline into the test, also the logic for instantiating the options class. That leaves the snippet focused on building the pipeline.

re: the options class, See my other comment. I do need to keep this class for the snippet.

dataflow/snippets/src/main/java/com/example/dataflow/KafkaRead.java
minherz, line 1: It looks like we had a misunderstanding. Let me reiterate on my comments to the provided code sample:

* The `main()` method is needed because we want to enable a user to run the code sample (and not only the tests) without writing code.
* The `main()` method should be excluded from the code snippet by using region tag to enclose import statements and relevant code blocks within the file. See go/code-snippets-201 for examples
* The code sample method, `createPipeline()` needs to receive three arguments that will be used to setup the property bag for the managed I/O transform. The use of the `Options` interface for this goal is opinionated approach and, as far as I could understand from [your comment](https://github.com/GoogleCloudPlatform/java-docs-samples/pull/9424#discussion_r1714217802), is not a recommended practice but used for convenience.

We want to avoid publishing opinionated code unless it follows best practices accepted in the industry or recommended by 1P or 3P providers.

Please, correct me if my understanding incorrect and passing arguments to the property bag using `Options` interface is the recommended practice. Otherwise, I strongly suggest to decouple them. We can use `Options` to parse arguments from the command line in the `main` method since the `main` method is not a part of the snippet.

VeronicaWasson: per discussion:

- Moved `main()` back into the snippet, and excluded it with region tags. 
- Kept the `Options` class, but excluded the class definition

minherz (approved):
lgtm

PR #9412, opened 2024-07-12
feat(videostitcher): add VOD config samples and tests

irataxy:
## Description

Fixes b:347726038

- [X] Please **merge** this PR for me once it is approved

PR #9438, opened 2024-08-06
feat(videostitcher): update other samples using feedback from #9412

irataxy:
## Description

* Add method comments
* Use try with resources
* Remove `videoStitcherServiceClient.close()`
* Return the response and use it in test cases for assertion
* Combine related tests into a single file rather than using individual files per test case
* Setup and tear down should be done once in a test lifecycle


Sita04 (approved):
Thanks for refactoring the existing samples. LGTM.
Please address the few nits before merging.

PR #9480, opened 2024-08-22
fix(compute): bug fixes HyperdisksIT test

TetyanaYahodska:
## Description

Fixes #<ISSUE-NUMBER>

## Checklist


Sita04:
Can you include description on the bugs this PR fixes.
Also, why do we need three different test classes? Can all the tests be grouped inside the same class as it belongs to a logical group?

TetyanaYahodska:
Having three tests in one class causes the PoolProvisionedIops value limit to be exceeded for one zone. 
Creating Hyperdisk Storage Pool with PoolProvisionedIops value less than 10000 cause error:
"message": "Provisioned IOPS 3000 is not a multiple of 10000."
Creating Hyperdisk Storage Pool with PoolProvisionedIops value more than 10000 cause error:
"Operation with name "operation-..." failed with status = HttpJsonStatusCode{statusCode=PERMISSION_DENIED} and message = FORBIDDEN"
Splitting these tests in separate file solved issue.

Sita04:
If same zone is the issue, would it be possible to declare different zones for each testcase, but within the same test file?

Sita04 (approved):

PR #9512, opened 2024-09-03
Update comments for IAM samples

melaniedejong:
## Description

Fixes the following issues with the code comments:

- Removes query string that forces documentation links to go to the English version of the page (?hl=en)
- Updates text to correctly indicate what each service account is for (several pages incorrectly said that the service accounts were to specify which key should be deleted)


iennae (approved):
Thanks for cleaning up the comments!

melaniedejong:
@iennae can you help me understand why the Kokoro CI - Java 11 test is failing? I don't think I changed anything in the code that would cause new failures.

PR #9513, opened 2024-09-04
feat(secretmanager): added samples for annotations

vipul7499:
## Description

Fixes #<ISSUE-NUMBER>

## Checklist

### DIFF
--- a/secretmanager/src/test/java/secretmanager/SnippetsIT.java
+++ b/secretmanager/src/test/java/secretmanager/SnippetsIT.java
@@ -150,7 +150,7 @@ public class SnippetsIT {
     ProjectName parent = ProjectName.of(PROJECT_ID);
 
     Secret secret;
-    if(addAnnotation){
+    if (addAnnotation) {
       secret = Secret.newBuilder()
       .setReplication(
         Replication.newBuilder()
@@ -159,7 +159,7 @@ public class SnippetsIT {
       .putLabels(LABEL_KEY, LABEL_VALUE)
       .putAnnotations(ANNOTATION_KEY, ANNOTATION_VALUE)
       .build();
-    }else{
+    } else {
       secret = Secret.newBuilder()
       .setReplication(
         Replication.newBuilder()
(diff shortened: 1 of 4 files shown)


secretmanager/src/test/java/secretmanager/SnippetsIT.java
```diff
@@ -154,6 +160,7 @@ private static Secret createSecret() throws IOException {
                             .setAutomatic(Replication.Automatic.newBuilder().build())
                             .build())
                     .putLabels(LABEL_KEY, LABEL_VALUE)
+                    .putAnnotations(ANNOTATION_KEY, ANNOTATION_VALUE)
```
Sita04, line 163: This will add the annotation to multiple secrets that call this method. Instead wydt about adding it only for TEST_SECRET?

secretmanager/src/test/java/secretmanager/SnippetsIT.java
```diff
@@ -432,6 +457,16 @@ public void testCreateUpdateSecretLabel() throws IOException {
...
   }
 
+  @Test
+  public void testEditSecretAnnotations() throws IOException {
+    SecretName name = SecretName.parse(TEST_SECRET.getName());
+    Secret updatedSecret = EditSecretAnnotations.editSecretAnnotations(
+        name.getProject(), name.getSecret(), UPDATED_ANNOTATION_KEY, UPDATED_ANNOTATION_VALUE);
+
+    assertThat(updatedSecret.getAnnotationsMap()).containsEntry(
+      UPDATED_ANNOTATION_KEY, UPDATED_ANNOTATION_VALUE);
```
Sita04, line 467: We can move this to BeforeClass annotation. That way the annotation is "added" to the existing TEST_SECRET.

Sita04 (changes requested):

secretmanager/src/test/java/secretmanager/SnippetsIT.java
vipul7499: I am not sure what you mean here, why would we want move a test to BeforeClass?

secretmanager/src/test/java/secretmanager/SnippetsIT.java
vipul7499: Fixed

secretmanager/src/test/java/secretmanager/SnippetsIT.java
Sita04: Few tests, like resource creation can be moved to BeforeClass so that the resource can be setup once and used by all the other tests in the class. This would avoid creating multiple resources for testing.

But I see you passed in the boolean to createSecret method. That works too.

Sita04 (approved):

PR #9547, opened 2024-10-01
feat(security-command-center): add sample for creating client with regional endpoint

hegemonic:
## Description

This sample shows how to create a Security Command Center client that uses a regional endpoint, along with a client that uses the default endpoint. That's because Security Command Center's regional endpoints support only a limited subset of Security Command Center resources that are subject to data residency controls: https://cloud.google.com/security-command-center/docs/data-residency-support#scc-regional-urls-rest

As a result, it's likely that customers will need to create two clients: one for their regional endpoint, to call the methods it supports, and another for the default endpoint, to call all other methods.

This sample will be used here: https://cloud.google.com/security-command-center/docs/data-residency-support#regional-urls

## Checklist


hegemonic:
The Kokoro CI failures don't appear to be related to this PR. Let me know if you disagree.

minherz:
Hello @hegemonic,
I have converted this PR to draft because it fails on multiple integration tests. Please address failing tests before submitting this PR for review.

hegemonic:
@minherz As I mentioned above, the Kokoro CI failures don't appear to be related to this PR. They involve a different set of tests that this PR doesn't update. Also, note that flaky-bot has already reported that these tests are failing at `HEAD`: #9542

If you don't agree, please let me know why this PR would cause these tests to fail.

hegemonic:
Yes, I'm familiar with Security Command Center, but I don't own any of these code samples—I'm just writing this one as a favor.

I'll ask one of the owners to fix the existing samples. For now, I'll move this back to a draft.

hegemonic:
#9580 was merged, so this PR is ready to review.

security-command-center/snippets/src/main/java/vtwo/client/CreateClientWithEndpoint.java
```diff
@@ -0,0 +1,49 @@
...
+  // regional endpoint.
+  public static void createClientWithEndpoint(String regionalEndpoint) throws java.io.IOException {
+    SecurityCenterSettings regionalSettings =
+        SecurityCenterSettings.newBuilder().setEndpoint(regionalEndpoint).build();
+
+    try (SecurityCenterClient client = SecurityCenterClient.create();
+        SecurityCenterClient regionalClient = SecurityCenterClient.create(regionalSettings)) {
+      System.out.println("Client initiated with endpoint: " + client.getSettings().getEndpoint());
+      System.out.println(
+          "Regional client initiated with endpoint: " + regionalClient.getSettings().getEndpoint());
```
minherz, line 44: nit: please follow [guidelines](https://github.com/GoogleCloudPlatform/java-docs-samples/blob/main/SAMPLE_FORMAT.md#method-structure:~:text=Snippet%20methods%20should%20return%20data) and return a value instead of printing to `stdout`.

security-command-center/snippets/src/main/java/vtwo/client/CreateClientWithEndpoint.java
```diff
@@ -0,0 +1,49 @@
...
+// [START securitycenter_set_client_endpoint_v2]
+
+import com.google.cloud.securitycenter.v2.SecurityCenterClient;
+import com.google.cloud.securitycenter.v2.SecurityCenterSettings;
+import java.io.IOException;
+
+public class CreateClientWithEndpoint {
+
+  public static void main(String[] args) throws IOException {
+    // TODO: Replace the value with your regional endpoint.
```
minherz, line 28: nit: consider using real value and explain in comments that region can be modified. AFAIK users can use an endpoint in _ANY_  region. It is not user's endpoint.

security-command-center/snippets/src/main/java/vtwo/client/CreateClientWithEndpoint.java
```diff
@@ -0,0 +1,49 @@
...
+  }
+
+  // Creates Security Command Center clients for the default endpoint and for a
+  // regional endpoint.
+  public static void createClientWithEndpoint(String regionalEndpoint) throws java.io.IOException {
+    SecurityCenterSettings regionalSettings =
+        SecurityCenterSettings.newBuilder().setEndpoint(regionalEndpoint).build();
+
+    try (SecurityCenterClient client = SecurityCenterClient.create();
+        SecurityCenterClient regionalClient = SecurityCenterClient.create(regionalSettings)) {
```
minherz, line 41: Why a user has to create two instances of `SecurityCenterClient`?

The code sample method is called "createClientWithEndpoint". Please, remove the client creation without regional settings.

security-command-center/snippets/src/main/java/vtwo/client/CreateClientWithEndpoint.java
```diff
@@ -0,0 +1,49 @@
...
+      System.out.println("Client initiated with endpoint: " + client.getSettings().getEndpoint());
+      System.out.println(
+          "Regional client initiated with endpoint: " + regionalClient.getSettings().getEndpoint());
+    }
+  }
+}
+
```
minherz, line 48: nit: In the sake of saving documentation space, consider removing empty lines that otherwise might be included into code snippet.

security-command-center/snippets/src/test/java/vtwo/CreateClientIT.java
minherz, line 1: After changing the method `createClientWithEndpoint` signature, update the test to assert on returned value instead of parsing `stdout`.

minherz (changes requested):

security-command-center/snippets/src/main/java/vtwo/client/CreateClientWithEndpoint.java
hegemonic: There are a couple of reasons why the sample needs to create both clients:

* Security Command Center's regional endpoint supports a limited subset of methods in the API. If you're using a regional endpoint, you'll also have to use the global endpoint.

    This behavior is pretty unusual—I'm not aware of any other Google Cloud service that requires customers to do this. Given how unusual this is, I'd like the code sample to make it very clear that customers must create two clients.

* I'd like the Java code sample to match the code samples for other languages.

Does that work for you?

security-command-center/snippets/src/main/java/vtwo/client/CreateClientWithEndpoint.java
hegemonic: Done.

FYI, many of the existing Security Command Center samples don't follow this guideline. For example: https://github.com/GoogleCloudPlatform/java-docs-samples/blob/81c4eaea6f3486df841d288f17942588b28dc0b1/security-command-center/snippets/src/test/java/vtwo/FindingsIT.java#L100-L118

security-command-center/snippets/src/main/java/vtwo/client/CreateClientWithEndpoint.java
hegemonic: Right now, the only regional endpoint that's available for Security Command Center is in `me-central2`; it's likely that only a very small percentage of our customers will need to use that endpoint, now and in the future. For that reason, I'd like to stick with the placeholder value. We could revisit this if and when Security Command Center adds more regional endpoints.

I did update and clarify the comment, though.

security-command-center/snippets/src/main/java/vtwo/client/CreateClientWithEndpoint.java
minherz: Yes, it is a known problem.

security-command-center/snippets/src/main/java/vtwo/client/CreateClientWithEndpoint.java
minherz: If there is such a limited usage of the endpoints, what is a reason to publish a sample?

security-command-center/snippets/src/main/java/vtwo/client/CreateClientWithEndpoint.java
minherz: I am unsure that I understand. Code example _is defined_ as "creating client with **regional** endpoint".

If the following code does not create **regional** endpoint why this code is included in the example?

```java
SecurityCenterClient client = SecurityCenterClient.create();
```

Your explanation does not sit well with me. Do you say that in order to use **regional** endpoint of SCC a developer has to create another (what it is called, global?) instance of `SecurityCenterClient`? What API endpoint does this "default" (or global) instance calls?

If this code sample confuses me, there is a good chance it will confuse an ordinary developer who has no experience about API endpoints and other specifics.

To your question:

> I'm not aware of any other Google Cloud service

AFAIK majority of services implement regional endpoints. Easy examples are GCS and PubSub. There are plenty more.
Additionally, many client libraries require endpoint (usually in the form of URL) parameter to initialize the object.

Regarding the following:

> I'd like the Java code sample to match the code samples for other languages.

This is a good goal when the code sample is clear and not ambiguous. Unfortunately, this is not the case with this particular code sample.

security-command-center/snippets/src/main/java/vtwo/client/CreateClientWithEndpoint.java
hegemonic: I can think of at least two reasons:

* Security Command Center has valued customers who use, or plan to use, the `me-central2` endpoint. We need to help those customers succeed.
* If and when Security Command Center adds more regional endpoints, we might have a very large number of customers who need this information.

security-command-center/snippets/src/main/java/vtwo/client/CreateClientWithEndpoint.java
hegemonic: > Your explanation does not sit well with me. Do you say that in order to use **regional** endpoint of SCC a developer has to create another (what it is called, global?) instance of `SecurityCenterClient`? What API endpoint does this "default" (or global) instance calls?
>
> ...
>
> AFAIK majority of services implement regional endpoints. Easy examples are GCS and PubSub. There are plenty more.

Let me see if I can explain this more clearly:

Yes, a number of Google Cloud services provide regional endpoints. Let's use [Cloud DLP](https://cloud.google.com/sensitive-data-protection/docs/reference/rest) as an example, because it's a pretty conventional service. (In contrast, Cloud Storage is an outlier in many different ways, and the only publicly visible endpoint for Cloud Pub/Sub [appears to be the global endpoint](https://cloud.google.com/pubsub/docs/reference/rest).)

As far as I can tell, Cloud DLP lets you use _any_ endpoint to call _any_ method in the API. In other words, if you use a regional service endpoint, and all of your data is in that region, you'll never have to call any other endpoint. If there are exceptions, then the [relevant documentation](https://cloud.google.com/sensitive-data-protection/docs/api-endpoints) doesn't mention them.

Security Command Center doesn't work that way:

* A [subset of Security Command Center resource types](https://cloud.google.com/security-command-center/docs/data-residency-support#scc-resources-subject-to-drz) are subject to data residency controls. If your data resides in a location with a regional endpoint (currently just `me-central2`), then you _must_ use the regional endpoint to manage these resource types.
* Other resource types _are not_ subject to data residency controls. As a result, you _cannot_ use a regional endpoint to manage other resource types. You must use the global endpoint.

As a result, if you enable data residency for Security Command Center, and your data resides in `me-central2`, then you _must_ create two separate clients:

* One for the regional endpoint, `https://securitycenter.me-central2.rep.googleapis.com`, to manage resources that reside in `me-central2`
* One for the global endpoint, `https://securitycenter.googleapis.com`, to manage resources that aren't subject to data residency controls

I'm not aware of any other Google Cloud services that require you to do this. That's why we need to provide a code sample—as you've pointed out, this requirement is surprising, and this isn't a pattern that our customers are already familiar with.

Does that make more sense? If not, what would you like me to change? (Keep in mind that customers are likely to see this sample in the context of the [data residency documentation](https://cloud.google.com/security-command-center/docs/data-residency-support), which already provides some of this context.)

> Additionally, many client libraries require endpoint (usually in the form of URL) parameter to initialize the object.

Out of curiosity, which ones? I don't think I've seen any that _require_ an endpoint parameter.

minherz: resolving this comment following our conversation. please, see two suggestions

security-command-center/snippets/src/main/java/vtwo/client/CreateClientWithEndpoint.java
minherz: resolving this comment following our conversation. please see two suggestions.

security-command-center/snippets/src/main/java/vtwo/client/CreateClientWithEndpoint.java
```diff
@@ -0,0 +1,58 @@
...
+import java.io.IOException;
+import java.util.HashMap;
+import java.util.Map;
+
+public class CreateClientWithEndpoint {
+
+  public static void main(String[] args) throws IOException {
+    // TODO: Replace the value with the endpoint for the region in which your
+    // Security Command Center data resides.
+    String regionalEndpoint = "securitycenter.REGION.rep.googleapis.com:443";
```
minherz, line 32: following our last conversation and because there is only one available region, please apply the suggested change
```suggestion
    String regionalEndpoint = "securitycenter.me-central2.rep.googleapis.com:443";
```

minherz: nit: consider opening an issue to follow up and add a link to documentation that enumerates available endpoints (or regions).

security-command-center/snippets/src/main/java/vtwo/client/CreateClientWithEndpoint.java
```diff
@@ -0,0 +1,58 @@
...
+  // regional endpoint.
+  public static Map<String, SecurityCenterClient> createClientWithEndpoint(String regionalEndpoint)
+      throws java.io.IOException {
+    Map<String, SecurityCenterClient> clients = new HashMap<>();
+    SecurityCenterSettings regionalSettings =
+        SecurityCenterSettings.newBuilder().setEndpoint(regionalEndpoint).build();
+    try (SecurityCenterClient client = SecurityCenterClient.create();
+        SecurityCenterClient regionalClient = SecurityCenterClient.create(regionalSettings)) {
+      clients.put("client", client);
+      clients.put("regionalClient", regionalClient);
```
minherz, line 52: following our last conversation and because there is only one available region, please apply the suggested change
```suggestion
    try (SecurityCenterClient regionalClient = SecurityCenterClient.create(regionalSettings)) {
      clients.put("regionalClient", regionalClient);
```

minherz (changes requested):
as discussed, i've provided the suggested changes following last GVC

minherz (approved):
lgtm. thank you for productive discussion.

PR #9562, opened 2024-10-08
chore(docs): change guidelines for snippet return values

minherz:
## Description

Modify snippet method's return value guidelines to enable more robust testing.
Fix markdown lint warning for the modified section of the document.

### DIFF
--- a/SAMPLE_FORMAT.md
+++ b/SAMPLE_FORMAT.md
@@ -348,33 +348,39 @@ public static void exampleSnippet(String projectId, String filePath) {
     // Snippet content ...
 }
 ```
+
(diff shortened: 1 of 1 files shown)


SAMPLE_FORMAT.md
```diff
@@ -348,33 +348,39 @@ public static void exampleSnippet(String projectId, String filePath) {
     // Snippet content ...
```
Sita04, line 348: nit: maybe update this to: `public static String exampleSnippet(...` to show a non-null return type?

Sita04 (approved):
Thanks for updating the guidelines!

PR #9586, opened 2024-10-15
feat: Java SDK Api Key authorization sample

ldetmer:
## Description

Java Cloud SDK introduced a new API to allow users to authenticate via an API key.  This provides a sample usage of the new setApiKey function.

## Checklist


auth/pom.xml
```diff
@@ -104,11 +108,6 @@ limitations under the License.
         </executions>
         <configuration>
           <executable>java</executable>
-          <arguments>
-            <argument>-classpath</argument>
-            <classpath />
```
ldetmer, line 109: please note I had to remove this specification as it was causing all AuthExamples tests to fail.  I have updated the readme so DownscopingExample can run without it.

auth/README.md
```diff
@@ -35,6 +35,11 @@ You can then run a given `ClassName` via:
     mvn exec:java -Dexec.mainClass=com.google.cloud.auth.samples.AuthExample
         -Dexec.args="compute"
 
+### Authenticating using an api key
+
+    mvn exec:java -Dexec.mainClass=com.google.cloud.auth.samples.AuthExample
+        -Dexec.args="apikey [api key]"
```
suztomo, line 41: Can you add a reference where one can get an API key?

Also would you make it clear that "[api key]" should be replaced with the value of an API key?

auth/README.md
```diff
@@ -35,6 +35,11 @@ You can then run a given `ClassName` via:
     mvn exec:java -Dexec.mainClass=com.google.cloud.auth.samples.AuthExample
         -Dexec.args="compute"
 
+### Authenticating using an api key
```
suztomo, line 38: Would you mention what it does? Reference the title of the JSON file example above.

minherz (changes requested):
Hello, thank you for submitting this code sample.

Please do the following modifications:

* split code samples one per class (and file)
* resolve code snippet mismatching region tags
* add test(s) to the new code sample

auth/src/main/java/com/google/cloud/auth/samples/AuthExample.java
```diff
@@ -89,6 +93,20 @@ static void authAppEngineStandard() throws IOException {
...
+    LanguageServiceClient client = LanguageServiceClient.create(settings);
+    Document document =
+        Document.newBuilder().setContent("Hello World!").setType(Document.Type.PLAIN_TEXT).build();
+
+    AnalyzeSentimentResponse actualResponse = client.analyzeSentiment(document);
+
+    System.out.println(actualResponse.getDocumentSentiment().toString());
+  }
+  // [END auth_cloud_api_keys]
+
```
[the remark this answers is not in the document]
ldetmer: done

auth/src/main/java/com/google/cloud/auth/samples/AuthExample.java
```diff
@@ -20,6 +20,10 @@
 import com.google.auth.appengine.AppEngineCredentials;
 import com.google.auth.oauth2.ComputeEngineCredentials;
 import com.google.auth.oauth2.GoogleCredentials;
+import com.google.cloud.language.v2.AnalyzeSentimentResponse;
+import com.google.cloud.language.v2.Document;
+import com.google.cloud.language.v2.LanguageServiceClient;
+import com.google.cloud.language.v2.LanguageServiceSettings;
```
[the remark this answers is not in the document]
ldetmer: done

auth/src/main/java/com/google/cloud/auth/samples/ApiKeyAuthExample.java
```diff
@@ -0,0 +1,56 @@
...
+    client.close();
+  }
+  // [END auth_cloud_api_keys]
+
+  public static void main(String[] args) throws IOException {
+    if (args.length >= 1) {
+      authApiKey(args[0]);
+    } else {
+      throw new IllegalArgumentException("Api key is required for this test");
+    }
```
minherz, line 54: nit: We discourage use arguments for code sample methods.

Just define a variable or a constant that will be passed as API key to the `authApiKey()` method.

Look at other methods for additional examples.

auth/src/main/java/com/google/cloud/auth/samples/ApiKeyAuthExample.java
```diff
@@ -0,0 +1,56 @@
...
+  static void authApiKey(String apiKey) throws IOException {
+    LanguageServiceSettings settings =
+        LanguageServiceSettings.newBuilder().setApiKey(apiKey).build();
+    LanguageServiceClient client = LanguageServiceClient.create(settings);
+    Document document =
+        Document.newBuilder().setContent("Hello World!").setType(Document.Type.PLAIN_TEXT).build();
+
+    AnalyzeSentimentResponse actualResponse = client.analyzeSentiment(document);
+
+    System.out.println(actualResponse.getDocumentSentiment().toString());
```
minherz, line 43: nit: please follow the recommend practice to return value instead of printing it. And modify the test to assert on that value. Using captured stdout proved to be error prone and lead to flaky tests.

auth/src/test/java/com/google/cloud/auth/samples/ApiKeyAuthExampleIT.java
minherz, line 1: nit: all tests can be combined in a single test class. the requirement for "class per sample" is for the source code itself. feel free to move the test to the shared class if you find it more convenient.

auth/src/main/java/com/google/cloud/auth/samples/ApiKeyAuthExample.java
```diff
@@ -0,0 +1,56 @@
...
+
+  // [START auth_cloud_api_key]
+  static void authApiKey(String apiKey) throws IOException {
+    LanguageServiceSettings settings =
+        LanguageServiceSettings.newBuilder().setApiKey(apiKey).build();
+    LanguageServiceClient client = LanguageServiceClient.create(settings);
+    Document document =
+        Document.newBuilder().setContent("Hello World!").setType(Document.Type.PLAIN_TEXT).build();
+
+    AnalyzeSentimentResponse actualResponse = client.analyzeSentiment(document);
```
minherz, line 41: Please use [try-with-resources](https://docs.oracle.com/javase/tutorial/essential/exceptions/tryResourceClose.html) construct to close the client at the end of the use instead of explicitly calling `close()`

minherz (changes requested):
Thank you for addressing the feedback quickly. There are a couple of additional changes related to using the standard language constructs as best practices and some guidance regarding code sample parameters and return types.

If you find time, would you mind to split other code samples in the same folder to stand-alone classes? It is not big deal if you don't but we are trying to use each opportunity to improve our code when we touch it.

Thank you.

auth/src/main/java/com/google/cloud/auth/samples/ApiKeyAuthExample.java
ldetmer: done

auth/src/main/java/com/google/cloud/auth/samples/ApiKeyAuthExample.java
ldetmer: updated

auth/src/test/java/com/google/cloud/auth/samples/ApiKeyAuthExampleIT.java
ldetmer: moved to AuthExampleIT

auth/src/main/java/com/google/cloud/auth/samples/ApiKeyAuthExample.java
ldetmer: done

auth/src/test/java/com/google/cloud/auth/samples/AuthExampleIT.java
```diff
@@ -57,4 +59,19 @@ public void testAuthExplicitNoPath() throws IOException {
...
+    String keyDisplayName = "Test API Key";
+    String service = "language.googleapis.com";
+    String method = "google.cloud.language.v2.LanguageService.AnalyzeSentiment";
+    Key apiKey = AuthTestUtils.createTestApiKey(projectId, keyDisplayName, service, method);
+
+    String output = ApiKeyAuthExample.authenticateUsingApiKey(apiKey.getKeyString());
+
+    assertTrue(output.contains("magnitude:"));
+
+    AuthTestUtils.deleteTestApiKey(apiKey.getName());
```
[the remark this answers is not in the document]
ldetmer: good catch thanks, done!

auth/src/test/java/com/google/cloud/auth/samples/AuthTestUtils.java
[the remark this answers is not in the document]
ldetmer: unfortunately this logic only exists in sample code that I would have to refactor to support my use case and I think potentially would make the sample less readable.

minherz (approved):
thank you for the quick fixes

PR #9587, opened 2024-10-16
feat(dataplex): add code samples for Entry

jacspa96:
## Description

Fixes b/287183421

## Checklist

### DIFF
--- a/dataplex/snippets/src/main/java/dataplex/CreateEntry.java
+++ b/dataplex/snippets/src/main/java/dataplex/CreateEntry.java
@@ -39,6 +39,7 @@ public class CreateEntry {
     System.out.println("Successfully created entry: " + createdEntry.getName());
   }
 
+  // Method to create Entry
   public static Entry createEntry(
       String projectId, String location, String entryGroupId, String entryId) throws Exception {
     EntryGroupName entryGroupName = EntryGroupName.of(projectId, location, entryGroupId);
@@ -59,8 +60,8 @@ public class CreateEntry {
                             "projects/dataplex-types/locations/global/aspectTypes/generic")
                         .setData(
                             Struct.newBuilder()
-                                // "Generic" Aspect Type have fields called "type" and "system,
-                                // it is just illustration how to fill in the fields of the Aspect.
+                                // "Generic" Aspect Type have fields called "type" and "system.
+                                // The values below are a sample of possible options.
                                 .putFields(
                                     "type",
                                     Value.newBuilder().setStringValue("example value").build())
@@ -72,9 +73,7 @@ public class CreateEntry {
             .build();
 
     // Initialize client that will be used to send requests. This client only needs to be created
-    // once, and can be reused for multiple requests. After completing all of your requests, call
-    // the "close" method on the client to safely clean up any remaining background resources,
-    // or use "try-with-close" statement to do this automatically.
+    // once, and can be reused for multiple requests.
     try (CatalogServiceClient client = CatalogServiceClient.create()) {
       return client.createEntry(entryGroupName, entry, entryId);
     }
--- a/dataplex/snippets/src/main/java/dataplex/CreateEntry.java
+++ b/dataplex/snippets/src/main/java/dataplex/CreateEntry.java
@@ -42,39 +42,38 @@ public class CreateEntry {
   // Method to create Entry
   public static Entry createEntry(
       String projectId, String location, String entryGroupId, String entryId) throws Exception {
-    EntryGroupName entryGroupName = EntryGroupName.of(projectId, location, entryGroupId);
-    Entry entry =
-        Entry.newBuilder()
-            // Example of system Entry Type.
-            // It is also possible to specify custom Entry Type.
-            .setEntryType("projects/dataplex-types/locations/global/entryTypes/generic")
-            .setEntrySource(
-                EntrySource.newBuilder().setDescription("description of the entry").build())
-            .putAllAspects(
-                Map.of(
-                    "dataplex-types.global.generic",
-                    Aspect.newBuilder()
-                        // This is required Aspect Type for "generic" Entry Type.
-                        // For custom Aspect Type required Entry Type would be different.
-                        .setAspectType(
-                            "projects/dataplex-types/locations/global/aspectTypes/generic")
-                        .setData(
-                            Struct.newBuilder()
-                                // "Generic" Aspect Type have fields called "type" and "system.
-                                // The values below are a sample of possible options.
-                                .putFields(
-                                    "type",
-                                    Value.newBuilder().setStringValue("example value").build())
-                                .putFields(
-                                    "system",
-                                    Value.newBuilder().setStringValue("example system").build())
-                                .build())
-                        .build()))
-            .build();
-
     // Initialize client that will be used to send requests. This client only needs to be created
     // once, and can be reused for multiple requests.
     try (CatalogServiceClient client = CatalogServiceClient.create()) {
+      EntryGroupName entryGroupName = EntryGroupName.of(projectId, location, entryGroupId);
+      Entry entry =
+          Entry.newBuilder()
+              // Example of system Entry Type.
+              // It is also possible to specify custom Entry Type.
+              .setEntryType("projects/dataplex-types/locations/global/entryTypes/generic")
+              .setEntrySource(
+                  EntrySource.newBuilder().setDescription("description of the entry").build())
+              .putAllAspects(
+                  Map.of(
+                      "dataplex-types.global.generic",
+                      Aspect.newBuilder()
+                          // This is required Aspect Type for "generic" Entry Type.
+                          // For custom Aspect Type required Entry Type would be different.
+                          .setAspectType(
+                              "projects/dataplex-types/locations/global/aspectTypes/generic")
+                          .setData(
+                              Struct.newBuilder()
+                                  // "Generic" Aspect Type have fields called "type" and "system.
+                                  // The values below are a sample of possible options.
+                                  .putFields(
+                                      "type",
+                                      Value.newBuilder().setStringValue("example value").build())
+                                  .putFields(
+                                      "system",
+                                      Value.newBuilder().setStringValue("example system").build())
+                                  .build())
+                          .build()))
+              .build();
       return client.createEntry(entryGroupName, entry, entryId);
     }
   }
--- a/dataplex/snippets/src/main/java/dataplex/CreateEntry.java
+++ b/dataplex/snippets/src/main/java/dataplex/CreateEntry.java
@@ -1,0 +1,83 @@
+/*
+ * Copyright 2024 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package dataplex;
+
+// [START dataplex_create_entry]
+import com.google.cloud.dataplex.v1.Aspect;
+import com.google.cloud.dataplex.v1.CatalogServiceClient;
+import com.google.cloud.dataplex.v1.Entry;
+import com.google.cloud.dataplex.v1.EntryGroupName;
+import com.google.cloud.dataplex.v1.EntrySource;
+import com.google.protobuf.Struct;
+import com.google.protobuf.Value;
+import java.util.Map;
+
+public class CreateEntry {
+
+  public static void main(String[] args) throws Exception {
+    // TODO(developer): Replace these variables before running the sample.
+    String projectId = "MY_PROJECT_ID";
+    String location = "MY_LOCATION";
+    String entryGroupId = "MY_ENTRY_GROUP_ID";
+    String entryId = "MY_ENTRY_ID";
+
+    Entry createdEntry = createEntry(projectId, location, entryGroupId, entryId);
+    System.out.println("Successfully created entry: " + createdEntry.getName());
+  }
+
+  public static Entry createEntry(
+      String projectId, String location, String entryGroupId, String entryId) throws Exception {
+    EntryGroupName entryGroupName = EntryGroupName.of(projectId, location, entryGroupId);
+    Entry entry =
+        Entry.newBuilder()
+            // Example of system Entry Type.
+            // It is also possible to specify custom Entry Type.
+            .setEntryType("projects/dataplex-types/locations/global/entryTypes/generic")
+            .setEntrySource(
+                EntrySource.newBuilder().setDescription("description of the entry").build())
+            .putAllAspects(
+                Map.of(
+                    "dataplex-types.global.generic",
+                    Aspect.newBuilder()
+                        // This is required Aspect Type for "generic" Entry Type.
+                        // For custom Aspect Type required Entry Type would be different.
+                        .setAspectType(
+                            "projects/dataplex-types/locations/global/aspectTypes/generic")
+                        .setData(
+                            Struct.newBuilder()
+                                // "Generic" Aspect Type have fields called "type" and "system,
+                                // it is just illustration how to fill in the fields of the Aspect.
+                                .putFields(
+                                    "type",
+                                    Value.newBuilder().setStringValue("example value").build())
+                                .putFields(
+                                    "system",
+                                    Value.newBuilder().setStringValue("example system").build())
+                                .build())
+                        .build()))
+            .build();
+
+    // Initialize client that will be used to send requests. This client only needs to be created
+    // once, and can be reused for multiple requests. After completing all of your requests, call
+    // the "close" method on the client to safely clean up any remaining background resources,
+    // or use "try-with-close" statement to do this automatically.
+    try (CatalogServiceClient client = CatalogServiceClient.create()) {
+      return client.createEntry(entryGroupName, entry, entryId);
+    }
+  }
+}
+// [END dataplex_create_entry]
--- a/dataplex/snippets/src/main/java/dataplex/CreateEntry.java
+++ b/dataplex/snippets/src/main/java/dataplex/CreateEntry.java
@@ -39,7 +39,7 @@ public class CreateEntry {
     System.out.println("Successfully created entry: " + createdEntry.getName());
   }
 
-  // Method to create Entry
+  // Method to create Entry located in projectId, location, entryGroupId and with entryId
   public static Entry createEntry(
       String projectId, String location, String entryGroupId, String entryId) throws Exception {
     // Initialize client that will be used to send requests. This client only needs to be created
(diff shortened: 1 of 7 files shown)


dataplex/snippets/src/main/java/dataplex/GetEntry.java
```diff
@@ -0,0 +1,78 @@
...
+    GetEntryRequest getEntryRequest =
+        GetEntryRequest.newBuilder()
+            .setName(EntryName.of(projectId, location, entryGroupId, entryId).toString())
+            // View determines which Aspects are returned with the Entry, available options:
+            // not set - Defaults to FULL
+            // BASIC - Returns entry only, without aspects
+            // FULL - Returns all required aspects as well as the keys of all non-required aspects
+            // CUSTOM - Returns Aspects matching custom fields in GetEntryRequest (max 100)
+            // ALL - Returns all aspects (max 100)
+            .setView(EntryView.FULL)
```
cerberus97, line 61: Let's set view to CUSTOM in the example, since it doesn't matter too much and at least this way the types and paths make a bit more sense.

dataplex/snippets/src/main/java/dataplex/GetEntry.java
```diff
@@ -0,0 +1,78 @@
...
+            // Following 2 lines will be ignored, because "View" is set to FULL.
+            // Their purpose is to demonstrate how to filter the Aspects returned for Entry
+            // when "View" is set to CUSTOM.
+            .addAspectTypes("projects/dataplex-types/locations/global/aspectTypes/generic")
+            .addPaths("generic")
```
cerberus97, line 66: nit: let's call it something else like "my_path", because generic is also the name fo the type

dataplex/snippets/src/main/java/dataplex/ListEntries.java
```diff
@@ -0,0 +1,56 @@
...
+public class ListEntries {
+
+  public static void main(String[] args) throws IOException {
+    // TODO(developer): Replace these variables before running the sample.
+    String projectId = "MY_PROJECT_ID";
+    String location = "MY_LOCATION";
+    String entryGroupId = "MY_ENTRY_GROUP_ID";
+
+    List<Entry> entries = listEntries(projectId, location, entryGroupId);
+    entries.forEach(aspectType -> System.out.println("Entry name: " + aspectType.getName()));
```
cerberus97, line 37: Would it be feasible to show how to set the filter (cs/google/cloud/dataplex/v1/catalog.proto;rcl=686230339;l=2991) here?

dataplex/snippets/src/main/java/dataplex/LookupEntry.java
```diff
@@ -0,0 +1,80 @@
+/*
```
cerberus97, line 1: Same comments as GetEntry for this

dataplex/snippets/src/main/java/dataplex/UpdateEntry.java
```diff
@@ -0,0 +1,92 @@
...
+    Entry entry =
+        Entry.newBuilder()
+            .setName(EntryName.of(projectId, location, entryGroupId, entryId).toString())
+            .setEntrySource(
+                EntrySource.newBuilder().setDescription("updated description of the entry").build())
+            .putAllAspects(
+                Map.of(
+                    "dataplex-types.global.generic",
+                    Aspect.newBuilder()
+                        // This is required Aspect Type for "generic" Entry Type.
```
cerberus97, line 54: I don't think this comment is necessary for Update, could be slightly misleading because it could be interpreted as "You must specify required type during update" which isn't true

dataplex/snippets/src/main/java/dataplex/GetEntry.java
cerberus97: Discussed offline, resolving. Sample doesn't work with CUSTOM view.

cerberus97 (approved):

dataplex/snippets/src/main/java/dataplex/CreateEntry.java
```diff
@@ -0,0 +1,83 @@
...
+    Entry entry =
+        Entry.newBuilder()
+            // Example of system Entry Type.
+            // It is also possible to specify custom Entry Type.
+            .setEntryType("projects/dataplex-types/locations/global/entryTypes/generic")
+            .setEntrySource(
+                EntrySource.newBuilder().setDescription("description of the entry").build())
+            .putAllAspects(
+                Map.of(
+                    "dataplex-types.global.generic",
```
Sita04, line 54: Same question here. Is this a pre-defined constant or will it need to be updated before running the sample?

dataplex/snippets/src/main/java/dataplex/CreateEntry.java
```diff
@@ -0,0 +1,83 @@
...
+                    Aspect.newBuilder()
+                        // This is required Aspect Type for "generic" Entry Type.
+                        // For custom Aspect Type required Entry Type would be different.
+                        .setAspectType(
+                            "projects/dataplex-types/locations/global/aspectTypes/generic")
```
Sita04, line 59: ditto.

dataplex/snippets/src/main/java/dataplex/CreateEntry.java
```diff
@@ -0,0 +1,83 @@
...
+                        .setData(
+                            Struct.newBuilder()
+                                // "Generic" Aspect Type have fields called "type" and "system,
+                                // it is just illustration how to fill in the fields of the Aspect.
```
Sita04, line 63:
```suggestion
                                // "Generic" Aspect Type have fields called "type" and "system.
                                // The values below are a sample of possible options.
```

dataplex/snippets/src/main/java/dataplex/CreateEntry.java
```diff
@@ -0,0 +1,83 @@
...
+                                    "system",
+                                    Value.newBuilder().setStringValue("example system").build())
+                                .build())
+                        .build()))
+            .build();
+
+    // Initialize client that will be used to send requests. This client only needs to be created
+    // once, and can be reused for multiple requests. After completing all of your requests, call
+    // the "close" method on the client to safely clean up any remaining background resources,
+    // or use "try-with-close" statement to do this automatically.
```
Sita04, line 77:
```suggestion
    // once, and can be reused for multiple requests.
```
Please remove the last line in the comment (from all samples) as this is not applicable to the samples anymore. The try with resource block auto-closes the client.

dataplex/snippets/src/main/java/dataplex/CreateEntry.java
```diff
@@ -0,0 +1,83 @@
...
+    try (CatalogServiceClient client = CatalogServiceClient.create()) {
```
Sita04, line 78: All code inside this method should be enclosed in this try block. Please move this line and the comment above to line 44.

Applies to all samples.

dataplex/snippets/src/main/java/dataplex/GetEntry.java
```diff
@@ -0,0 +1,78 @@
...
+
+// [START dataplex_get_entry]
+import com.google.cloud.dataplex.v1.CatalogServiceClient;
+import com.google.cloud.dataplex.v1.Entry;
+import com.google.cloud.dataplex.v1.EntryName;
+import com.google.cloud.dataplex.v1.EntryView;
+import com.google.cloud.dataplex.v1.GetEntryRequest;
+import java.io.IOException;
+
+// Sample to get Entry
```
Sita04, line 27:
```suggestion
```

dataplex/snippets/src/main/java/dataplex/GetEntry.java
```diff
@@ -0,0 +1,78 @@
...
+    System.out.println("Entry retrieved successfully: " + entry.getName());
+    entry
+        .getAspectsMap()
+        .keySet()
+        .forEach(aspectKey -> System.out.println("Retrieved aspect for entry: " + aspectKey));
+  }
+
+  // When Entry is created in Dataplex for example for BigQuery table,
+  // access permissions might differ between Dataplex and source system.
+  // Get method checks permissions in Dataplex.
```
Sita04, line 47:
```suggestion
  // "Get" method, checks permissions in Dataplex.
```

dataplex/snippets/src/main/java/dataplex/GetEntry.java
```diff
@@ -0,0 +1,78 @@
...
+
+    GetEntryRequest getEntryRequest =
+        GetEntryRequest.newBuilder()
+            .setName(EntryName.of(projectId, location, entryGroupId, entryId).toString())
+            // View determines which Aspects are returned with the Entry, available options:
+            // not set - Defaults to FULL
+            // BASIC - Returns entry only, without aspects
+            // FULL - Returns all required aspects as well as the keys of all non-required aspects
+            // CUSTOM - Returns Aspects matching custom fields in GetEntryRequest (max 100)
+            // ALL - Returns all aspects (max 100)
```
Sita04, line 60: Having more comment lines is not preferred in most cases, unless it highlights an aspect of the code itself. Please consider pointing to the docs link instead, like:

" // View determines which Aspects are returned with the Entry. For all available options, see <docs-link>"

dataplex/snippets/src/main/java/dataplex/GetEntry.java
```diff
@@ -0,0 +1,78 @@
...
+            .setView(EntryView.FULL)
+            // Following 2 lines will be ignored, because "View" is set to FULL.
+            // Their purpose is to demonstrate how to filter the Aspects returned for Entry
+            // when "View" is set to CUSTOM.
```
Sita04, line 64: In this case, could we update the view to EntryView.CUSTOM in line 61?

dataplex/snippets/src/main/java/dataplex/ListEntries.java
```diff
@@ -0,0 +1,69 @@
...
+// [START dataplex_list_entries]
+import com.google.cloud.dataplex.v1.CatalogServiceClient;
+import com.google.cloud.dataplex.v1.Entry;
+import com.google.cloud.dataplex.v1.EntryGroupName;
+import com.google.cloud.dataplex.v1.ListEntriesRequest;
+import com.google.common.collect.ImmutableList;
+import java.io.IOException;
+import java.util.List;
+
+// Sample to list Entries
```
Sita04, line 28:
```suggestion
```

dataplex/snippets/src/main/java/dataplex/ListEntries.java
```diff
@@ -0,0 +1,69 @@
...
+    ListEntriesRequest listEntriesRequest =
+        ListEntriesRequest.newBuilder()
+            .setParent(EntryGroupName.of(projectId, location, entryGroupId).toString())
+            // A filter on the entries to return. Filters are case-sensitive.
+            // You can filter the request by the following fields:
+            // * entry_type
+            // * entry_source.display_name
+            // The comparison operators are =, !=, <, >, <=, >=, with lexical order.
+            // You can use the logical operators AND, OR, NOT in the filter.
+            // You can use wildcard "*", but for entry_type full project id or number is needed
```
Sita04, line 52: Consider removing these comments and instead adding a docs link for more info.

dataplex/snippets/src/main/java/dataplex/LookupEntry.java
```diff
@@ -0,0 +1,80 @@
...
+
+// [START dataplex_lookup_entry]
+import com.google.cloud.dataplex.v1.CatalogServiceClient;
+import com.google.cloud.dataplex.v1.Entry;
+import com.google.cloud.dataplex.v1.EntryName;
+import com.google.cloud.dataplex.v1.EntryView;
+import com.google.cloud.dataplex.v1.LookupEntryRequest;
+import java.io.IOException;
+
+// Sample to lookup Entry
```
Sita04, line 27:
```suggestion
```

dataplex/snippets/src/main/java/dataplex/LookupEntry.java
```diff
@@ -0,0 +1,80 @@
...
+    System.out.println("Entry retrieved successfully: " + entry.getName());
+    entry
+        .getAspectsMap()
+        .keySet()
+        .forEach(aspectKey -> System.out.println("Retrieved aspect for entry: " + aspectKey));
+  }
+
+  // When Entry is created in Dataplex for example for BigQuery table,
+  // access permissions might differ between Dataplex and source system.
+  // Lookup method checks permissions in source system.
```
Sita04, line 47:
```suggestion
  // "Lookup" method checks permissions in source system.
```

dataplex/snippets/src/main/java/dataplex/LookupEntry.java
```diff
@@ -0,0 +1,80 @@
...
+  // Please also refer to GetEntry code sample, which checks permissions in Dataplex.
```
Sita04, line 48: When this sample is included in the docs, this instruction might become obscure. Please update the comment to something like: "// Please also refer how to get an Entry, which checks permissions in Dataplex."

dataplex/snippets/src/main/java/dataplex/LookupEntry.java
```diff
@@ -0,0 +1,80 @@
...
+              // The project to which the request should be attributed
+              .setName(projectLocation)
+              // The resource name of the Entry
+              .setEntry(EntryName.of(projectId, location, entryGroupId, entryId).toString())
+              // View determines which Aspects are returned with the Entry, available options:
+              // not set - Defaults to FULL
+              // BASIC - Returns entry only, without aspects
+              // FULL - Returns all required aspects as well as the keys of all non-required aspects
+              // CUSTOM - Returns Aspects matching custom fields in GetEntryRequest (max 100)
+              // ALL - Returns all aspects (max 100)
```
Sita04, line 68: Kindly replace comments with docs link.

dataplex/snippets/src/main/java/dataplex/LookupEntry.java
```diff
@@ -0,0 +1,80 @@
...
+              .setView(EntryView.FULL)
+              // Following 2 lines will be ignored, because "View" is set to FULL.
+              // Their purpose is to demonstrate how to filter the Aspects returned for Entry
+              // when "View" is set to CUSTOM.
```
Sita04, line 72: ditto. Is it possible to replace "EntryView.FULL" to "EntryView.CUSTOM"?

dataplex/snippets/src/main/java/dataplex/UpdateEntry.java
```diff
@@ -0,0 +1,90 @@
...
+                                        .setStringValue("updated example system")
+                                        .build())
+                                .build())
+                        .build()))
+            .build();
+
+    // Update mask specifies which fields will be updated.
+    // If empty mask is given, all modifiable fields from the request will be used for update.
+    // If update mask is specified as "*" it is treated as full update,
+    // that means fields not present in the request will be emptied.
```
Sita04, line 77: Consider replacing this with "For more information on update masks, see: <docs-link>"

dataplex/snippets/src/main/java/dataplex/CreateEntry.java
Sita04, line 1: All comments in this file also apply to UpdateEntry.

dataplex/snippets/src/test/java/dataplex/EntryIT.java
```diff
@@ -0,0 +1,120 @@
...
+  private static String requireProjectIdEnvVar() {
+    String value = System.getenv("GOOGLE_CLOUD_PROJECT");
+    assertNotNull(
+        "Environment variable GOOGLE_CLOUD_PROJECT is required to perform these tests.", value);
+    return value;
+  }
+
+  @BeforeClass
+  public static void checkRequirements() {
+    requireProjectIdEnvVar();
```
Sita04, line 48: Also verify if GOOGLE_APPLICATION_CREDENTIALS is present.

Sita04 (changes requested):
Thanks for the samples!

Most of the comments are nits. Please ping me for further clarifications.

dataplex/snippets/src/main/java/dataplex/CreateEntry.java
```diff
@@ -0,0 +1,83 @@
...
+  }
+
+  public static Entry createEntry(
+      String projectId, String location, String entryGroupId, String entryId) throws Exception {
+    EntryGroupName entryGroupName = EntryGroupName.of(projectId, location, entryGroupId);
+    Entry entry =
+        Entry.newBuilder()
+            // Example of system Entry Type.
+            // It is also possible to specify custom Entry Type.
+            .setEntryType("projects/dataplex-types/locations/global/entryTypes/generic")
```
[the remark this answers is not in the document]
jacspa96: It is path to system resource that should always be present. That way user does not need to point to their custom resource, that they might not have. That makes sample self contained and I'd treat it as constant.

dataplex/snippets/src/main/java/dataplex/CreateEntry.java
jacspa96: It is path to system resource that should always be present. That way user does not need to point to their custom resource, that they might not have. That makes sample self contained and I'd treat it as constant.

dataplex/snippets/src/main/java/dataplex/CreateEntry.java
jacspa96: It is path to system resource that should always be present. That way user does not need to point to their custom resource, that they might not have. That makes sample self contained and I'd treat it as constant.

dataplex/snippets/src/main/java/dataplex/GetEntry.java
jacspa96: It was conscious decision to set it to FULL. We don't want to limit returned results, but only show how to set these filters if someone wished to do that. We want to present how to access aspects attached to returned entry and setting view to CUSTOM might result in no aspect being returned.

dataplex/snippets/src/main/java/dataplex/LookupEntry.java
jacspa96: ditto. It is technically possible, but in my view not desired.

dataplex/snippets/src/test/java/dataplex/EntryIT.java
jacspa96: Running tests locally fails because of that check (but work without), and it seems checks performed in this repo are working either way. Also, I found this check only in 6 files in this repo, so it doesn't seem to be a standard. Therefore I will not add this.

dataplex/snippets/src/main/java/dataplex/CreateEntry.java
```diff
@@ -0,0 +1,83 @@
...
+    String projectId = "MY_PROJECT_ID";
+    String location = "MY_LOCATION";
+    String entryGroupId = "MY_ENTRY_GROUP_ID";
+    String entryId = "MY_ENTRY_ID";
+
+    Entry createdEntry = createEntry(projectId, location, entryGroupId, entryId);
+    System.out.println("Successfully created entry: " + createdEntry.getName());
+  }
+
+  public static Entry createEntry(
```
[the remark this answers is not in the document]
Sita04: The comment style we follow is something like:

`// Creates an entry to/in/with/ etc...`

dataplex/snippets/src/test/java/dataplex/EntryIT.java
Sita04: This is a best practice that most of the samples created in the last couple years, follow. This would ensure that the samples and tests can be run anywhere (in Cloud Shell/ workstations) where this config is auto. 

It is also useful to our customers to understand how to authenticate before running the sample. I'd highly encourage to add this in the test.

Sita04 (approved):
Couple nits. Kindly address those before merging.

PR #9596, opened 2024-10-19
feat(compute): add compute consume reservation samples

TetyanaYahodska:
## Description
Sample in NodeJs - https://github.com/GoogleCloudPlatform/nodejs-docs-samples/pull/3877/files

Fixes #<ISSUE-NUMBER>

## Checklist


minherz (changes requested):
please modify the code samples to maintain **a single** code sample per file/class.

TetyanaYahodska:
These samples demonstrate how to create instances with different reservation consumption options. So, changes are required for both the instance and reservation. 
Documentation for these three samples:
https://cloud.google.com/compute/docs/instances/reservations-consume#consuming_a_specific_shared_reservation, https://cloud.google.com/compute/docs/instances/reservations-consume#consuming_a_specific_single-project_reservation, https://cloud.google.com/compute/docs/instances/reservations-consume#consuming_instances_from_any_matching_reservation

compute/cloud-client/src/main/java/compute/reservation/ConsumeAnyMatchingReservation.java
```diff
@@ -0,0 +1,184 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package compute.reservation;
+
+// [START compute_consume_any_matching_reservation]
+
```
minherz, line 20: remove empty line

minherz (changes requested):
As I explained earlier we have a policy of having a single code sample per file. Having two methods violates this requirement. If you want to demonstrate a workflow that involves executing multiple methods and different client libraries you can do it in the single method that examples the workflow.
If you have further questions, please schedule a meeting and argument your case.

compute/cloud-client/src/main/java/compute/reservation/ConsumeAnyMatchingReservation.java
TetyanaYahodska: Done

TetyanaYahodska:
In this PR, I fixed the empty lines before the tags for the samples created earlier, removed the cleanup methods from the setup methods, and fixed other issues based on the recommendations I received. The idea behind this was to keep our code consistent.

compute/cloud-client/src/test/java/compute/reservation/ConsumeReservationsIT.java
```diff
@@ -0,0 +1,164 @@
...
+@Timeout(value = 6, unit = TimeUnit.MINUTES)
+public class ConsumeReservationsIT {
+
+  private static final String PROJECT_ID = System.getenv("GOOGLE_CLOUD_PROJECT");
+  private static final String ZONE = "us-central1-a";
+  private static final String  RESERVATION_NAME = "test-reservaton-" + UUID.randomUUID();
+  private static final String INSTANCE_FOR_SPR = "test-instance-for-spr-" + UUID.randomUUID();
+  private static final String INSTANCE_FOR_ANY_MATCHING = "test-instance-" + UUID.randomUUID();
+  private static final String SPECIFIC_SHARED_INSTANCE =
+      "test-instance-shared-" + UUID.randomUUID();
```
minherz, line 56: nit: recommend to use the same random ID; it helps to troubleshoot failed tests

compute/cloud-client/src/test/java/compute/reservation/CrudOperationsReservationIT.java
```diff
@@ -111,7 +83,6 @@ public void testGetReservation()
   }
 
   @Test
-  @Order(4)
   public void testListReservation() throws IOException {
```
minherz, line 86: assert on `RESERVATION_NAME` in line 91 (I cannot comment on untouched code lines). Asserting on prefix will pass the test in event there is reservation from a parallel tests running there.

compute/cloud-client/src/test/java/compute/reservation/ReservationIT.java
minherz, line 1: nit: please review the use of the test order by this class.

minherz (commented):
Thorough work. Thank you. Everything looks good.  I've left a few of nit comments for your considerations.

Besides the comments, before merging these changes, please check the following two items:

1. These tests look heavy to me. They create and delete a lot of VM instances and templates. Please, consider opening a feature request and suggesting more lightweight method of testing these code samples. Keep it mind that we do not need to test that client libraries used in these samples work as expected. We only need to test that our code samples use these client libraries as expected.

1. It looks like three different code samples (ConsumeAnyMatchingReservation, ConsumeSingleProjectReservation and ConsumeSpecificSharedReservation) differ only by the value the object `ReservationAffinity` that is used when instance resource is created. Maintaining all of the is a clear overhead. Consider replacing these code samples with the code that showing how each of the different reservation affinity can be created.

compute/cloud-client/src/test/java/compute/reservation/ConsumeReservationsIT.java
TetyanaYahodska: Thank you! Fixed

compute/cloud-client/src/test/java/compute/reservation/CrudOperationsReservationIT.java
TetyanaYahodska: You are right, We can't predict reservations' order. That's why I'm testing prefix. Because the goal of the listReservations() method is to get list with reservations. Actually, we can delete this assertion and left only isNotNull() assertion. Would you recommend to delete this assertion?

minherz: I suggest to use a known reservation name that is supposed to be there.
If we cannot do it then I prefer the validation that you suggest because both of them can be false positive but your suggestion let's us lose enforcing the order which makes testing run faster.

minherz (approved):
Before merging, please have a look at my reply to [this comment thread](https://github.com/GoogleCloudPlatform/java-docs-samples/pull/9596#discussion_r1826306998). No actions required but consider making changes discussed in the thread.

compute/cloud-client/src/test/java/compute/reservation/CrudOperationsReservationIT.java
TetyanaYahodska: As the purpose of this test is list the reservations and not give as some defined reservation, I prefer to left it as is.

PR #9585, opened 2024-10-15
feat(tpu): add create/delete/get samples 

TetyanaYahodska:
## Description
Sample in Python [feat: TPU *Create / *get / *create_with_script / *delete Samples by Thoughtseize1 · Pull Request #12690 · GoogleCloudPlatform/python-docs-samples](https://github.com/GoogleCloudPlatform/python-docs-samples/pull/12690)
Documentation - https://cloud.google.com/tpu/docs/managing-tpus-tpu-vm
Fixes #<ISSUE-NUMBER>

Note: Before submitting a pull request, please open an issue for discussion if you are not associated with Google.

## Checklist

- [x] I have followed [Sample Format Guide](https://github.com/GoogleCloudPlatform/java-docs-samples/blob/main/SAMPLE_FORMAT.md)
- [ ] `pom.xml` parent set to latest `shared-configuration`
- [ ] Appropriate changes to README are included in PR
- [ ] These samples need a new **API enabled** in testing projects to pass (let us know which ones)
- [ ] These samples need a new/updated **env vars** in testing projects set to pass (let us know which ones)
- [x] **Tests** pass:   `mvn clean verify` **required**
- [x] **Lint**  passes: `mvn -P lint checkstyle:check` **required**
- [ ] **Static Analysis**:  `mvn -P lint clean compile pmd:cpd-check spotbugs:check` **advisory only**
- [x] This sample adds a new sample directory, and I updated the [CODEOWNERS file](https://github.com/GoogleCloudPlatform/java-docs-samples/blob/main/.github/CODEOWNERS) with the codeowners for this sample
- [ ] This sample adds a new **Product API**, and I updated the [Blunderbuss issue/PR auto-assigner](https://github.com/GoogleCloudPlatform/java-docs-samples/blob/main/.github/blunderbuss.yml) with the codeowners for this sample 
- [ ] Please **merge** this PR for me once it is approved

PR #9600, opened 2024-10-23
feat(tpu): add  tpu queued resources create/delete/force delete/get/list samples

TetyanaYahodska:
## Description
Issue with Deleting operation in google.cloud.tpu_v2alpha1
1. When running on macOS, the user will encounter an IllegalStateException, which is better handled at the library level and not delegated to the user, as it does not affect the delete method, since it is successful.
Screenshot 2024-10-23 at 13 08 56 (2)

2.We are using the v2alpha1 library to delete Queued Resources and we faced an issue with the unpack operation even though it still succeeds. I have handled this in a catch block, but I believe that this is not a very good way to pass this exception to the end user.
Screenshot 2024-10-23 at 13 15 37 (2)

Log file -  [log.txt](https://github.com/user-attachments/files/17491663/log.txt)

Fixes #<ISSUE-NUMBER>

## Checklist


TetyanaYahodska:
Will be merged after https://github.com/GoogleCloudPlatform/java-docs-samples/pull/9585

PR #9613, opened 2024-10-29
feat(tpu): add tpu queued resources create/get/delete  samples

TetyanaYahodska:
## Description

Issue with Deleting operation in google.cloud.tpu_v2alpha1

When running on macOS, the user will encounter an IllegalStateException, which is better handled at the library level and not delegated to the user, as it does not affect the delete method, since it is successful.
Screenshot 2024-10-23 at 13 08 56 (2)

2.We are using the v2alpha1 library to delete Queued Resources and we faced an issue with the unpack operation even though it still succeeds. I have handled this in a catch block, but I believe that this is not a very good way to pass this exception to the end user.
Screenshot 2024-10-23 at 13 15 37 (2)

Log file - [log.txt](https://github.com/user-attachments/files/17491663/log.txt)
Sample in Python - https://github.com/GoogleCloudPlatform/python-docs-samples/pull/12716/files
Fixes #<ISSUE-NUMBER>

## Checklist


tpu/src/main/java/tpu/CreateQueuedResource.java
```diff
@@ -0,0 +1,121 @@
...
+
+      CreateQueuedResourceRequest request =
+          CreateQueuedResourceRequest.newBuilder()
+              .setParent(parent)
+              .setQueuedResourceId(queuedResourceId)
+              .setQueuedResource(queuedResource)
+              .build();
+
+      // You can wait until TPU Node is READY,
+      // and check its status using getTpuVm() from "tpu_vm_get" sample.
```
minherz, line 115: what is a meaning for this comment? either implement the waiting (mind that you will need to mock this in the tests) or consider showing it in a separate code sample.

tpu/src/main/java/tpu/DeleteQueuedResource.java
```diff
@@ -0,0 +1,88 @@
...
+import com.google.api.gax.retrying.RetrySettings;
+import com.google.cloud.tpu.v2alpha1.DeleteQueuedResourceRequest;
+import com.google.cloud.tpu.v2alpha1.TpuClient;
+import com.google.cloud.tpu.v2alpha1.TpuSettings;
+import java.io.IOException;
+import java.util.concurrent.ExecutionException;
+import org.threeten.bp.Duration;
+// Uncomment this imports for getting node name
+//import com.google.cloud.tpu.v2alpha1.GetQueuedResourceRequest;
+//import com.google.cloud.tpu.v2alpha1.QueuedResource;
```
minherz, line 29: these comments are irrelevant for this code sample.
delete the comment and the commented lines.

tpu/src/main/java/tpu/DeleteQueuedResource.java
```diff
@@ -0,0 +1,88 @@
...
+    clientSettings
+        .deleteQueuedResourceSettings()
+        .setRetrySettings(
+            RetrySettings.newBuilder()
+                .setInitialRetryDelay(Duration.ofMillis(5000L))
+                .setRetryDelayMultiplier(2.0)
+                .setInitialRpcTimeout(Duration.ZERO)
+                .setRpcTimeoutMultiplier(1.0)
+                .setMaxRetryDelay(Duration.ofMillis(45000L))
+                .setTotalTimeout(Duration.ofHours(24L))
```
minherz, line 63: same question as in the previous example. why to customize the retrying logic here and not in many other code samples? why these specific custom values?

tpu/src/main/java/tpu/DeleteQueuedResource.java
```diff
@@ -0,0 +1,88 @@
...
+    // once, and can be reused for multiple requests.
+    try (TpuClient tpuClient = TpuClient.create(clientSettings.build())) {
+      // Before deleting the queued resource it is required to delete the TPU VM.
+      // Uncomment these lines to retrieve node name
+      //GetQueuedResourceRequest getRequest =
+      //    GetQueuedResourceRequest.newBuilder().setName(name).build();
+      //QueuedResource queuedResource = tpuClient.getQueuedResource(getRequest);
+      //String nodeName = queuedResource.getTpu().getNodeSpec(0).getNode().getName();
+      // For more information about deleting TPU
+      // see https://cloud.google.com/tpu/docs/managing-tpus-tpu-vm
```
minherz, line 75: the comments are unclear. Does the sample not showing how to delete a queued resource? If it does, please delete the comments and commented lines. If it does not, modify the code sample to fully demonstrate the delete process. I recommend to discuss the problem of additional operation or what resource would be best used for queued resource manipulations with a technical writer who will use these code samples.

tpu/src/main/java/tpu/DeleteQueuedResource.java
```diff
@@ -0,0 +1,88 @@
...
+
+      DeleteQueuedResourceRequest request =
+              DeleteQueuedResourceRequest.newBuilder().setName(name).build();
+
+      // Waiting for updates in the library. Until then, the operation will complete successfully,
+      // but the user will receive an error message with UnknownException and IllegalStateException.
```
minherz, line 81: what is a meaning for this comment? if operation completes successfully there should be no exception. if this is a problem of the library, resolve the problem in the library _BEFORE_  implementing a code sample.

tpu/src/main/java/tpu/GetQueuedResource.java
```diff
@@ -50,4 +50,4 @@ public static QueuedResource getQueuedResource(
     }
   }
 }
-//[END tpu_queued_resources_get]
+//[END tpu_queued_resources_get]
```
minherz, line 53: nit: undo this change since it seems like doing nothing. reduced changed lines

minherz (commented):
please update the description of the PR because it does not seem adequately reflect the submitted changes. for example, adding new code samples does not make sense in respect to the "Issue with Deleting operation...".

Review and address lines containing comments and instructions to users and provide justification for using custom retrying configurations that aren't used anywhere else.

tpu/src/main/java/tpu/CreateQueuedResource.java
```diff
@@ -0,0 +1,121 @@
...
+    clientSettings
+        .createQueuedResourceSettings()
+        .setRetrySettings(
+            RetrySettings.newBuilder()
+                .setInitialRetryDelay(Duration.ofMillis(5000L))
+                .setRetryDelayMultiplier(2.0)
+                .setInitialRpcTimeout(Duration.ZERO)
+                .setRpcTimeoutMultiplier(1.0)
+                .setMaxRetryDelay(Duration.ofMillis(45000L))
+                .setTotalTimeout(Duration.ofHours(24L))
```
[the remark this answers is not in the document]
TetyanaYahodska: As I remember, for delete operation and sometimes for create/update operation we have CancellationException error due to timeout.We have these settings in the library and I want to show users how we can fix such errors. This code appears only once for queued resource creation to not overload others samples.

tpu/src/main/java/tpu/CreateQueuedResource.java
```diff
@@ -0,0 +1,121 @@
...
+                      .addNodeSpec(
+                          QueuedResource.Tpu.NodeSpec.newBuilder()
+                              .setParent(parent)
+                              .setNode(node)
+                              .setNodeId(nodeName)
+                              .build())
+                      .build())
+              // You can request a queued resource using a reservation by specifying it in code
+              //.setReservationName(
+              // "projects/YOUR_PROJECT_ID/locations/YOUR_ZONE/reservations/YOUR_RESERVATION_NAME")
```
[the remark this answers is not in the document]
TetyanaYahodska: It was asked to add this line - "Add a commented piece that shows how to specify reservation. We won't have a TPU reservation though, so no way to test it."

tpu/src/main/java/tpu/CreateQueuedResource.java
TetyanaYahodska: Thank you! Deleted

tpu/src/main/java/tpu/DeleteQueuedResource.java
TetyanaYahodska: As I remember, for delete operation and sometimes for create/update operation we have CancellationException error due to timeout.We have these settings in the library and I want to show users how we can fix such errors.

tpu/src/main/java/tpu/DeleteQueuedResource.java
TetyanaYahodska: Thank you! Deleted code.Let's keep this comment for user.

tpu/src/main/java/tpu/DeleteQueuedResource.java
TetyanaYahodska: Yes, there is a problem in the library. Should I delete this sample?

tpu/src/main/java/tpu/GetQueuedResource.java
TetyanaYahodska: Excuse me, can you explain your point? What should I change?

tpu/src/main/java/tpu/CreateQueuedResource.java
```diff
@@ -0,0 +1,121 @@
...
+      String parent = String.format("projects/%s/locations/%s", projectId, zone);
+      Node node =
+          Node.newBuilder()
+              .setName(nodeName)
+              .setAcceleratorType(tpuType)
+              .setRuntimeVersion(tpuSoftwareVersion)
+              .setQueuedResource(
+                  String.format(
+                      "projects/%s/locations/%s/queuedResources/%s",
+                      projectId, zone, queuedResourceId))
```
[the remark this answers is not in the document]
TetyanaYahodska: Thank you! Fixed

tpu/src/main/java/tpu/DeleteQueuedResource.java
TetyanaYahodska: Thank you! Deleted

tpu/src/main/java/tpu/GetQueuedResource.java
minherz: I do not understand why this line appears as changed. Apparently there was some change in the text of the line.
In general, we try to avoid such modifications because they do not carry any meaning but increase a complexity of the commit by increasing the number of lines that are reported as changed.
If you cannot figure out what is the cause of the change, you can ignore this.

tpu/src/main/java/tpu/CreateQueuedResource.java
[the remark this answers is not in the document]
minherz: It is a great idea. Such code sample can be useful and this particular use case can serve as a practical example. Please create an issue to describe it. And notify Maciej about it.
Having said that, the example that you described is different from one that is submitted. This example (as the region tag implies) demonstrates the use of API to create a queued resource. This is different example and adding this type of improvement introduces a confusing message to documentation reader, especially compared to code snippets in other languages or code snippets in Java for different operations.

Please, remove customized retry settings in this and other code samples in this PR.

tpu/src/main/java/tpu/CreateQueuedResource.java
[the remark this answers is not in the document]
minherz: Where it was asked and by whom? The PR description does not contain any information.

Code snippets are intended to demonstrate coding practices and not document possible behaviors.
Adding long comments increase the number of lines that a reader has to go through in order to find the code.

If you like to do a code sample that sets up reservation name, pass it as argument to the method and configure it. You can comment on the line saying it is optional.

tpu/src/main/java/tpu/DeleteQueuedResource.java
minherz: See similar comment in the Create*.java file

tpu/src/main/java/tpu/DeleteQueuedResource.java
minherz: This comment does not seem to contribute to the code example. Either the following call (line 74) completes or throws exception and exits. If there is an exception that is being thrown still means that operation complete correctly, process it in this code.
If in your opinion there is a problem in the library, report this problem by opening an issue in the library's Github repo.

tpu/src/main/java/tpu/DeleteQueuedResource.java
```diff
@@ -0,0 +1,80 @@
...
+      tpuClient.deleteQueuedResourceAsync(request).get();
+
+      System.out.printf("Deleted Queued Resource: %s\n", name);
```
minherz, line 76: Delete these lines. They are redundant. If you feel it is required, return the status.

minherz (commented):
thank you for explaining your reasons. i have provided further clarifications about requested changes.

tpu/src/main/java/tpu/CreateQueuedResource.java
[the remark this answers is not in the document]
TetyanaYahodska: It was asked by Maciej Strzelczyk. But you are right, We don't have this information in hte documentation. I delete this comments.

tpu/src/main/java/tpu/CreateQueuedResource.java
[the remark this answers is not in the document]
TetyanaYahodska: Thank you. Deleted

tpu/src/main/java/tpu/DeleteQueuedResource.java
TetyanaYahodska: Thank you. Deleted

tpu/src/main/java/tpu/DeleteQueuedResource.java
TetyanaYahodska: Thank you. Deleted

tpu/src/main/java/tpu/DeleteQueuedResource.java
TetyanaYahodska: Thank you. Deleted

tpu/src/main/java/tpu/GetQueuedResource.java
TetyanaYahodska: This looks strange, but I don't see any difference between these two lines.

minherz (approved):
lgtm. good work!

PR #9684, opened 2024-11-19
feat(managedkafka): increase LRO timeout to 1 hour in client samples

rohithrajan-ai:
* feat(managedkafka): increase timeouts to 1 hour for Create/Update/Delete cluster 
* feat(managedkafka): added code sample for polling logs in CreateCluster

### DIFF
--- a/managedkafka/examples/pom.xml
+++ b/managedkafka/examples/pom.xml
@@ -29,7 +29,7 @@
             <dependency>
                 <groupId>com.google.cloud</groupId>
                 <artifactId>libraries-bom</artifactId>
-                <version>26.40.0</version>
+                <version>26.50.0</version>
                 <type>pom</type>
                 <scope>import</scope>
             </dependency>
@@ -40,13 +40,7 @@
     <dependency>
         <groupId>com.google.cloud</groupId>
         <artifactId>google-cloud-managedkafka</artifactId>
-        <version>0.1.0</version>
     </dependency>
-        <dependency>
-            <groupId>com.google.api.grpc</groupId>
-            <artifactId>proto-google-cloud-managedkafka-v1</artifactId>
-            <version>0.1.0</version>
-        </dependency>
         <dependency>
             <groupId>junit</groupId>
             <artifactId>junit</artifactId>
--- a/managedkafka/examples/src/main/java/examples/CreateCluster.java
+++ b/managedkafka/examples/src/main/java/examples/CreateCluster.java
@@ -76,13 +76,7 @@ public class CreateCluster {
     ManagedKafkaSettings.Builder settingsBuilder = ManagedKafkaSettings.newBuilder();
     TimedRetryAlgorithm timedRetryAlgorithm = OperationTimedPollAlgorithm.create(
         RetrySettings.newBuilder()
-            .setInitialRetryDelayDuration(Duration.ofSeconds(5L))
-            .setRetryDelayMultiplier(1.5)
-            .setMaxRetryDelayDuration(Duration.ofSeconds(30L))
-            .setInitialRpcTimeoutDuration(Duration.ZERO) // ignored for LROs
-            .setRpcTimeoutMultiplier(1.0) // ignored for LROs
-            .setMaxRpcTimeoutDuration(Duration.ZERO) // ignored for LROs
-            .setTotalTimeoutDuration(Duration.ofHours(1L)) // set polling timeout to 1 hour
+            .setTotalTimeoutDuration(Duration.ofHours(1L))
             .build());
     settingsBuilder.createClusterOperationSettings()
         .setPollingAlgorithm(timedRetryAlgorithm);
--- a/managedkafka/examples/src/main/java/examples/CreateCluster.java
+++ b/managedkafka/examples/src/main/java/examples/CreateCluster.java
@@ -17,7 +17,13 @@
 package examples;
 
 // [START managedkafka_create_cluster]
+
 import com.google.api.gax.longrunning.OperationFuture;
+import com.google.api.gax.longrunning.OperationSnapshot;
+import com.google.api.gax.longrunning.OperationTimedPollAlgorithm;
+import com.google.api.gax.retrying.RetrySettings;
+import com.google.api.gax.retrying.RetryingFuture;
+import com.google.api.gax.retrying.TimedRetryAlgorithm;
 import com.google.cloud.managedkafka.v1.AccessConfig;
 import com.google.cloud.managedkafka.v1.CapacityConfig;
 import com.google.cloud.managedkafka.v1.Cluster;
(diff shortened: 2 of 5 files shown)


managedkafka/examples/src/main/java/examples/CreateCluster.java
```diff
@@ -64,17 +72,59 @@ public static void createCluster(
...
+    TimedRetryAlgorithm timedRetryAlgorithm = OperationTimedPollAlgorithm.create(
+        RetrySettings.newBuilder()
+            .setInitialRetryDelayDuration(Duration.ofSeconds(5L))
+            .setRetryDelayMultiplier(1.5)
+            .setMaxRetryDelayDuration(Duration.ofSeconds(30L))
+            .setInitialRpcTimeoutDuration(Duration.ZERO) // ignored
+            .setRpcTimeoutMultiplier(1.0) // ignored
+            .setMaxRpcTimeoutDuration(Duration.ZERO) // ignored
+            .setTotalTimeoutDuration(Duration.ofHours(1L)) // set polling timeout to 1 hour
+            .build());
```
acocuzzo, line 86: Can we get the default settings as a builder + only change the value we need? Here and elsewhere. Otherwise it makes the samples very verbose.

managedkafka/examples/src/main/java/examples/CreateCluster.java
```diff
@@ -64,17 +72,59 @@ public static void createCluster(
...
           managedKafkaClient.createClusterOperationCallable().futureCall(request);
+
+      // Get the initial LRO and print details.
+      OperationSnapshot operation = future.getInitialFuture().get();
+      System.out.printf("Cluster creation started. Operation name: %s\nDone: %s\nMetadata: %s\n",
+          operation.getName(),
+          operation.isDone(),
+          future.getMetadata().get().toString());
+
+      // Polling logs to see operation progress
```
acocuzzo, line 117: I don't think we need this comment, and it's confusing (should be just Polling for operation progress).

managedkafka/examples/src/main/java/examples/CreateCluster.java
```diff
@@ -64,17 +72,59 @@ public static void createCluster(
...
+        .setPollingAlgorithm(timedRetryAlgorithm);
+
+    try (ManagedKafkaClient managedKafkaClient = ManagedKafkaClient.create(
+        settingsBuilder.build())) {
+      // Optionally print OperationCallSettings to investigate LRO timeout errors
+      // due to configuration set on the client side
+      // System.out.println("Create Cluster Settings (OperationCallSettings): " +
+      //     managedKafkaClient.getSettings().createClusterOperationSettings().getPollingAlgorithm()
+      //         .toString());
+
```
acocuzzo, line 97: I think we can remove this.

managedkafka/examples/pom.xml
```diff
@@ -40,13 +40,7 @@
     <dependency>
         <groupId>com.google.cloud</groupId>
         <artifactId>google-cloud-managedkafka</artifactId>
-        <version>0.1.0</version>
```
acocuzzo, line 43: Why was this dependency removed?

managedkafka/examples/src/main/java/examples/CreateCluster.java
rohithrajan-ai: I'm not quite sure I follow what you mean by "get the default settings as a builder"? I'm already doing that by creating the builder and I'm only setting the `createClusterOperationSettings` which is what we need to change?

I intentionally left the RPC timeout values with the ignored comment so users know that these settings will be ignored for LROs (I should probably state explicitly that it is ignored for LROs, oops). I can remove these if you want though but left them in there for now.

managedkafka/examples/src/main/java/examples/CreateCluster.java
rohithrajan-ai: Removed!

The reason I said "polling logs .." is because future.get() also polls under the hood so this while loop is doing the same thing but so we can see progress by looking at the log statements being printed out.

managedkafka/examples/src/main/java/examples/CreateCluster.java
rohithrajan-ai: Removed!

managedkafka/examples/pom.xml
rohithrajan-ai: I was following the guidance that was provided here: https://cloud.google.com/java/docs/reference/google-cloud-managedkafka/latest/overview#maven 

There is no version provided for this dependency and removing the version caused errors importing that specific version of grpc. Moreover, I'm not sure why it was added and removing it didn't affect being able to run the client library.

Talked to Adam and couldn't exactly figure out the reason this was added unless I'm missing something?

managedkafka/examples/src/main/java/examples/CreateCluster.java
acocuzzo: We should only set RetrySettings.Builder values where we need to change the default value.

See docs:
https://cloud.google.com/java/docs/reference/gax/latest/com.google.api.gax.retrying.RetrySettings.Builder

1. `setInitialRetryDelayDuration` defaults to 5 seconds, so there is no need to set it. 
2.  `setRetryDelayMultiplier` This is the default value, there is no need to set it.
3. For `setMaxRetryDelayDuration` the default is (45 seconds), so it's fine to set to 30 seconds, but might not really be necessary.
4. `setInitialRpcTimeoutDuration`, `setRpcTimeoutMultiplier`, and `setMaxRpcTimeoutDuration` are not necessary since they are ignored for LROs and you are just setting them to the default values.
I think we should only set RetrySettings.newBuilder()..setTotalTimeoutDuration(Duration.ofHours(1L)).build());

managedkafka/examples/src/main/java/examples/DeleteCluster.java
```diff
@@ -38,13 +45,38 @@ public static void main(String[] args) throws Exception {
 
   public static void deleteCluster(String projectId, String region, String clusterId)
       throws Exception {
-    try (ManagedKafkaClient managedKafkaClient = ManagedKafkaClient.create()) {
+
+    // Create the settings to configure the timeout for polling operations
```
acocuzzo, line 49: Same here, we should only set .setTotalTimeoutDuration(Duration.ofHours(1L))

managedkafka/examples/src/main/java/examples/CreateCluster.java
rohithrajan-ai: Ahhh I see what you meant - I was looking at the defaults set in the `RetrySettings` builder (which doesn't have the LRO defaults) but the LRO-specific values are set in our `ManagedKafkaStubSettings` builder.

Fixed!

managedkafka/examples/src/main/java/examples/DeleteCluster.java
rohithrajan-ai: Fixed!

managedkafka/examples/src/main/java/examples/UpdateCluster.java
```diff
@@ -48,11 +55,35 @@ public static void updateCluster(
...
+    TimedRetryAlgorithm timedRetryAlgorithm = OperationTimedPollAlgorithm.create(
+        RetrySettings.newBuilder()
+            .setInitialRetryDelayDuration(Duration.ofSeconds(5L))
+            .setRetryDelayMultiplier(1.5)
+            .setMaxRetryDelayDuration(Duration.ofSeconds(30L))
+            .setInitialRpcTimeoutDuration(Duration.ZERO) // ignored for LROs
+            .setRpcTimeoutMultiplier(1.0) // ignored for LROs
+            .setMaxRpcTimeoutDuration(Duration.ZERO) // ignored for LROs
+            .setTotalTimeoutDuration(Duration.ofHours(1L))  // set polling timeout to 1 hour
+            .build());
```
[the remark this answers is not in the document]
rohithrajan-ai: Fixed!

acocuzzo (approved):

Mukamik (approved):
Lgtm

Mukamik (approved):
LGTM

PR #9689, opened 2024-11-21
feat: add basic proxy

igorbernstein2:
Stacked PRs:
 * #9695
 * #9692
 * #9691
 * #9690
 * __->__#9689
 * #9688

--- --- ---

PR #9690, opened 2024-11-21
feat: move credential handling responsibility to the proxy

igorbernstein2:
Stacked PRs:
 * #9695
 * #9692
 * #9691
 * __->__#9690
 * #9689
 * #9688

--- --- ---

### DIFF
--- a/bigtable/bigtable-proxy/src/main/java/com/google/cloud/bigtable/examples/proxy/core/ProxyHandler.java
+++ b/bigtable/bigtable-proxy/src/main/java/com/google/cloud/bigtable/examples/proxy/core/ProxyHandler.java
@@ -16,6 +16,7 @@
 
 package com.google.cloud.bigtable.examples.proxy.core;
 
+import io.grpc.CallCredentials;
 import io.grpc.CallOptions;
 import io.grpc.Channel;
 import io.grpc.ClientCall;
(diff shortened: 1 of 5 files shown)


bigtable/bigtable-proxy/src/main/java/com/google/cloud/bigtable/examples/proxy/core/ProxyHandler.java
```diff
@@ -25,15 +26,23 @@
...
+  public ProxyHandler(Channel channel, CallCredentials callCredentials) {
     this.channel = channel;
+    this.callCredentials = callCredentials;
   }
 
   @Override
   public ServerCall.Listener<ReqT> startCall(ServerCall<ReqT, RespT> serverCall, Metadata headers) {
-    CallOptions callOptions = CallOptions.DEFAULT;
+    // Strip incoming credentials
+    headers.removeAll(AUTHORIZATION_KEY);
```
mutianf, line 43: So if someone runs their client with a credential they also need to start the proxy handler with that credential? why not just using the incoming credentials?

mutianf (commented):
Do we need to add auth library to the dependency? The test failed with compilation errors.

bigtable/bigtable-proxy/src/main/java/com/google/cloud/bigtable/examples/proxy/core/ProxyHandler.java
igorbernstein2: Thats not quite right. The proxy listening port uses an insecure channel (plaintext) and the customer's client is expected to use emulator mode to connect to the proxy, which disables credentials. The only usecase that I want to support here is the proxy owning the creds

mutianf (approved):

bhshkh (approved):

PR #9691, opened 2024-11-21
feat: add metrics

igorbernstein2:
Stacked PRs:
 * #9695
 * #9692
 * __->__#9691
 * #9690
 * #9689
 * #9688

--- --- ---

### DIFF
--- a/bigtable/bigtable-proxy/src/main/java/com/google/cloud/bigtable/examples/proxy/metrics/MetricsImpl.java
+++ b/bigtable/bigtable-proxy/src/main/java/com/google/cloud/bigtable/examples/proxy/metrics/MetricsImpl.java
@@ -1,0 +1,230 @@
+/*
+ * Copyright 2024 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.cloud.bigtable.examples.proxy.metrics;
+
+import com.google.api.gax.core.FixedCredentialsProvider;
+import com.google.api.gax.grpc.GrpcTransportChannel;
+import com.google.api.gax.rpc.FixedTransportChannelProvider;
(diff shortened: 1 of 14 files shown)


bigtable/bigtable-proxy/src/main/java/com/google/cloud/bigtable/examples/proxy/metrics/MetricsImpl.java
```diff
@@ -0,0 +1,227 @@
...
+  public MetricsImpl(Credentials credentials, String projectId) throws IOException {
+    meterProvider = createMeterProvider(credentials, projectId);
+    Meter meter = meterProvider.meterBuilder("bigtableproxy").build();
+
+    serverCallsStarted =
+        meter
+            .counterBuilder(METRIC_PREFIX + "server.call.started")
+            .setDescription(
+                "The total number of RPCs started, including those that have not completed.")
+            .setUnit("{call}")
```
mutianf, line 73: should the unit be 1?

bigtable/bigtable-proxy/src/main/java/com/google/cloud/bigtable/examples/proxy/metrics/MetricsImpl.java
```diff
@@ -0,0 +1,227 @@
...
+        .build();
+
+    MetricConfiguration config =
+        MetricConfiguration.builder()
+            .setProjectId(projectId)
+            .setMetricServiceSettings(metricServiceSettingsBuilder.build())
+            .setInstrumentationLibraryLabelsEnabled(false)
+            .build();
+
+    MetricExporter exporter = GoogleCloudMetricExporter.createWithConfiguration(config);
```
mutianf, line 171: Did you test exporting the metrics? Wondering if we need to create the metrics descriptor?

mutianf: And also which monitoed resource it's exported to by default?

igorbernstein2: I think thats out of scope....I'm using the MetricExporter for exporting metrics, so I'm relying on its correctness. The cost of doing a full end to end test is quite high (waiting for 5 minutes for publication) and the benefit is quite low.

It will use the autodetected monitored resource that GoogleCloudMetricExporter uses by default....GCE VM if on gce, gke pod if on gke, generic node if on prem, etc

mutianf: If that's the case can we have the option to not export metrics?

bhshkh (approved):

PR #9743, opened 2024-11-28
feat(securitycenter): Add Resource SCC Management API Org ETD Custom Module code samples (Create, Delete, List, Get)

lovenishs04:
## Description

Fixes # b/347346890, b/347346971, b/347346892, b/347346464
Reference to b/347346890, b/347346971, b/347346892, b/347346464
This PR adds SCC Managament API Org Event Threat Detection Custom Module Code Samples for Create, Delete, List & Get.

## Checklist

### DIFF
--- a/security-command-center/snippets/src/main/java/management/api/CreateEventThreatDetectionCustomModule.java
+++ b/security-command-center/snippets/src/main/java/management/api/CreateEventThreatDetectionCustomModule.java
@@ -1,0 +1,103 @@
+/*
+ * Copyright 2024 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package management.api;
+
+// [START securitycenter_create_event_threat_detection_custom_module]
+import com.google.cloud.securitycentermanagement.v1.CreateEventThreatDetectionCustomModuleRequest;
+import com.google.cloud.securitycentermanagement.v1.EventThreatDetectionCustomModule;
+import com.google.cloud.securitycentermanagement.v1.EventThreatDetectionCustomModule.EnablementState;
+import com.google.cloud.securitycentermanagement.v1.SecurityCenterManagementClient;
+import com.google.protobuf.ListValue;
+import com.google.protobuf.Struct;
+import com.google.protobuf.Value;
+import java.io.IOException;
+
+public class CreateEventThreatDetectionCustomModule {
+
+  public static void main(String[] args) throws IOException {
+    // https://cloud.google.com/security-command-center/docs/reference/security-center-management/rest/v1/organizations.locations.eventThreatDetectionCustomModules/create
+    // TODO: Developer should replace project_id with a real project ID before running this code
+    String projectId = "project_id";
+
+    String customModuleDisplayName = "custom_module_display_name";
+
+    createEventThreatDetectionCustomModule(projectId, customModuleDisplayName);
+  }
+
+  public static EventThreatDetectionCustomModule createEventThreatDetectionCustomModule(
+      String projectId, String customModuleDisplayName) throws IOException {
+
+    // Initialize client that will be used to send requests. This client only needs
+    // to be created
+    // once, and can be reused for multiple requests.
+    try (SecurityCenterManagementClient client = SecurityCenterManagementClient.create()) {
+
+      String name =
+          String.format(
+              "projects/%s/locations/global/eventThreatDetectionCustomModules/%s",
+              projectId, "custom_module");
+
+      // define the metadata and other config parameters severity, description,
+      // recommendation below
+      Struct metadataStruct =
+          Struct.newBuilder()
+              .putFields("severity", Value.newBuilder().setStringValue("MEDIUM").build())
+              .putFields(
+                  "description",
+                  Value.newBuilder().setStringValue("add your description here").build())
+              .putFields(
+                  "recommendation",
+                  Value.newBuilder().setStringValue("add your recommendation here").build())
+              .build();
+      Struct configStruct =
+          Struct.newBuilder()
+              .putFields("metadata", Value.newBuilder().setStructValue(metadataStruct).build())
+              .putFields(
+                  "ips",
+                  Value.newBuilder()
+                      .setListValue(
+                          ListValue.newBuilder()
+                              .addValues(Value.newBuilder().setStringValue("0.0.0.0").build())
+                              .build())
+                      .build())
+              .build();
+
+      // define the Event Threat Detection custom module configuration, update the EnablementState
+      // below
+      EventThreatDetectionCustomModule eventThreatDetectionCustomModule =
+          EventThreatDetectionCustomModule.newBuilder()
+              .setName(name)
+              .setConfig(configStruct)
+              .setDisplayName(customModuleDisplayName)
+              .setEnablementState(EnablementState.ENABLED)
+              .setType("CONFIGURABLE_BAD_IP")
+              .build();
+
+      CreateEventThreatDetectionCustomModuleRequest request =
+          CreateEventThreatDetectionCustomModuleRequest.newBuilder()
+              .setParent(String.format("projects/%s/locations/global", projectId))
+              .setEventThreatDetectionCustomModule(eventThreatDetectionCustomModule)
+              .build();
+
+      EventThreatDetectionCustomModule response =
+          client.createEventThreatDetectionCustomModule(request);
+
+      return response;
+    }
+  }
+}
+// [END securitycenter_create_event_threat_detection_custom_module]
--- a/security-command-center/snippets/src/main/java/management/api/CreateEventThreatDetectionCustomModule.java
+++ b/security-command-center/snippets/src/main/java/management/api/CreateEventThreatDetectionCustomModule.java
@@ -25,6 +25,8 @@ import com.google.protobuf.ListValue;
 import com.google.protobuf.Struct;
 import com.google.protobuf.Value;
 import java.io.IOException;
+import java.util.HashMap;
+import java.util.Map;
 
 public class CreateEventThreatDetectionCustomModule {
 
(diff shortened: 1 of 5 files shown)


security-command-center/snippets/src/main/java/management/api/CreateEventThreatDetectionCustomModule.java
```diff
@@ -0,0 +1,103 @@
...
+              .putFields("metadata", Value.newBuilder().setStructValue(metadataStruct).build())
+              .putFields(
+                  "ips",
+                  Value.newBuilder()
+                      .setListValue(
+                          ListValue.newBuilder()
+                              .addValues(Value.newBuilder().setStringValue("0.0.0.0").build())
+                              .build())
+                      .build())
+              .build();
```
minherz, line 77: nit: Do you mean there is no protobuf record for the module's configuration? This long configuration setting looks very confusing and hard to read. Please, confirm with the API/ client library team that there is no other way. If there is no other way, please refactor the code to make it readable. Consider defining the properties in a map and to use protobuf method to convert the map to the protobuf structure.

security-command-center/snippets/src/main/java/management/api/CreateEventThreatDetectionCustomModule.java
```diff
@@ -0,0 +1,103 @@
...
+              .setName(name)
+              .setConfig(configStruct)
+              .setDisplayName(customModuleDisplayName)
+              .setEnablementState(EnablementState.ENABLED)
+              .setType("CONFIGURABLE_BAD_IP")
+              .build();
+
+      CreateEventThreatDetectionCustomModuleRequest request =
+          CreateEventThreatDetectionCustomModuleRequest.newBuilder()
+              .setParent(String.format("projects/%s/locations/global", projectId))
```
minherz, line 92: nit: is it possible that parent is not required since it duplicates information placed in the name? is it possible to define the name of the resource to use one project ID and location but to set parent to another project and location?

security-command-center/snippets/src/main/java/management/api/CreateEventThreatDetectionCustomModule.java
```diff
@@ -0,0 +1,103 @@
...
+    // https://cloud.google.com/security-command-center/docs/reference/security-center-management/rest/v1/organizations.locations.eventThreatDetectionCustomModules/create
+    // TODO: Developer should replace project_id with a real project ID before running this code
+    String projectId = "project_id";
+
+    String customModuleDisplayName = "custom_module_display_name";
+
+    createEventThreatDetectionCustomModule(projectId, customModuleDisplayName);
+  }
+
+  public static EventThreatDetectionCustomModule createEventThreatDetectionCustomModule(
```
minherz, line 41: nit: I guess the code sample demonstrates creating a thread detection for specific kind of threats: filter out traffic from a specific IP. Should the name of the code sample reflect this?

security-command-center/snippets/src/test/java/management/api/EventThreatDetectionCustomModuleTest.java
```diff
@@ -0,0 +1,166 @@
...
+          client.listEventThreatDetectionCustomModules(request);
+      // Iterate over the response and delete custom module one by one which start with
+      // java_sample_custom_module
+      for (EventThreatDetectionCustomModule module : response.iterateAll()) {
+        if (module.getDisplayName().startsWith("java_sample_custom_module")) {
+          String customModuleId = extractCustomModuleId(module.getName());
+          deleteCustomModule(PROJECT_ID, customModuleId);
+        }
+      }
+    }
```
minherz, line 89: nit: consider silencing exception. it is better to leave leftovers with and log the problem than fail a lot of tests.

security-command-center/snippets/src/test/java/management/api/EventThreatDetectionCustomModuleTest.java
```diff
@@ -0,0 +1,166 @@
...
+
+  // cleanupExistingCustomModules clean up all the existing custom module
+  private static void cleanupExistingCustomModules() throws IOException {
+    try (SecurityCenterManagementClient client = SecurityCenterManagementClient.create()) {
+      ListEventThreatDetectionCustomModulesRequest request =
+          ListEventThreatDetectionCustomModulesRequest.newBuilder()
+              .setParent(String.format("projects/%s/locations/global", PROJECT_ID))
+              .build();
+      ListEventThreatDetectionCustomModulesPagedResponse response =
+          client.listEventThreatDetectionCustomModules(request);
```
minherz, line 80: please refactor to clean up only modules created by this collection of tests. do not delete "other" modules.

security-command-center/snippets/src/test/java/management/api/EventThreatDetectionCustomModuleTest.java
```diff
@@ -0,0 +1,166 @@
...
+      for (EventThreatDetectionCustomModule module : response.iterateAll()) {
+        if (module.getDisplayName().startsWith("java_sample_custom_module")) {
+          String customModuleId = extractCustomModuleId(module.getName());
+          deleteCustomModule(PROJECT_ID, customModuleId);
+        }
+      }
+    }
+  }
+
+  // extractCustomModuleID extracts the custom module Id from the full name
```
minherz, line 92: nit: it is a good practice to explain briefly regex. add explanation that it parses suffix after the last slash character.

security-command-center/snippets/src/main/java/management/api/CreateEventThreatDetectionCustomModule.java
```diff
@@ -0,0 +1,103 @@
...
+
+    // Initialize client that will be used to send requests. This client only needs
+    // to be created
+    // once, and can be reused for multiple requests.
+    try (SecurityCenterManagementClient client = SecurityCenterManagementClient.create()) {
+
+      String name =
+          String.format(
+              "projects/%s/locations/global/eventThreatDetectionCustomModules/%s",
+              projectId, "custom_module");
```
minherz, line 52: using a constant name for ID is OK for the code sample, but will be a problem for testing. consider providing the module ID as a parameter a use a literal for display name.

security-command-center/snippets/src/test/java/management/api/EventThreatDetectionCustomModuleTest.java
```diff
@@ -0,0 +1,166 @@
...
+  private static EventThreatDetectionCustomModule createCustomModule(
+      String projectId, String customModuleDisplayName) throws IOException {
+    if (!Strings.isNullOrEmpty(projectId) && !Strings.isNullOrEmpty(customModuleDisplayName)) {
+      EventThreatDetectionCustomModule response =
+          CreateEventThreatDetectionCustomModule.createEventThreatDetectionCustomModule(
+              projectId, customModuleDisplayName);
+      return response;
+    }
+    return null;
+  }
```
minherz, line 114: this method is redundant. it is an exact replica of createEventThreatDetectionCustomModule() except for testing for empty string arguments which look unnecessary for private method inside test class.

minherz (changes requested):

security-command-center/snippets/src/main/java/management/api/CreateEventThreatDetectionCustomModule.java
lovenishs04: Actually I have used the generic name for all the code samples (because in this sample also we are trying to create the ETD custom module) and in other existing samples also. And all IP related information is available in the documentation link which I have added in the comment. And if you still want me to change the name then can you please suggest me the name?

security-command-center/snippets/src/main/java/management/api/CreateEventThreatDetectionCustomModule.java
lovenishs04: Addressed

security-command-center/snippets/src/main/java/management/api/CreateEventThreatDetectionCustomModule.java
lovenishs04: Even I have also noticed that we are passing duplicate information in both parent and name. Actually I tried to remove the parent since we are passing the duplicate information but I am getting an exception, seems it will be a required field to pass in the request

security-command-center/snippets/src/test/java/management/api/EventThreatDetectionCustomModuleTest.java
lovenishs04: Yes this condition I have already added in the code (if (module.getDisplayName().startsWith("java_sample_custom_module"))). It will be going to delete the modules created by this collection of tests and not the other modules

security-command-center/snippets/src/test/java/management/api/EventThreatDetectionCustomModuleTest.java
lovenishs04: Addressed

security-command-center/snippets/src/main/java/management/api/CreateEventThreatDetectionCustomModule.java
lovenishs04: Actually this sample code is for creating the custom module and we don't have any module ID in place to pass in the parameter, so the module ID will be generated after the creation of custom modules. I have used this constant just as a placeholder. If you want me to change the name then I can use any other name?

security-command-center/snippets/src/test/java/management/api/EventThreatDetectionCustomModuleTest.java
lovenishs04: Actually I have created this method because in every test case I am creating the custom module first and then performing the test case. So in order to use the same code I have created this method and calling it in every test case and also you might have noticed I have called the same method createEventThreatDetectionCustomModule() inside this method.

security-command-center/snippets/src/main/java/management/api/CreateEventThreatDetectionCustomModule.java
lovenishs04: Addressed, now only parent is used

security-command-center/snippets/src/main/java/management/api/CreateEventThreatDetectionCustomModule.java
lovenishs04: Addressed, as already mentioned removed this variable name and now using the parent only

minherz (approved):
please note that this PR is approved while having a potential test flakiness due to the code that deletes modules based on the display name which is the same in all executions of the tests.

please, use random prefix or suffix for the value of the display name to ensure that clean up operation deletes only resources used in the current testing session.

consider other remarks.

security-command-center/snippets/src/test/java/management/api/EventThreatDetectionCustomModuleTest.java
lovenishs04: Addressed, added etd prefix to the display name and it stands for event threat detection

security-command-center/snippets/src/test/java/management/api/EventThreatDetectionCustomModuleTest.java
lovenishs04: Addressed, removed this separate method, now making the direct call in every testcase

security-command-center/snippets/src/main/java/management/api/CreateEventThreatDetectionCustomModule.java
```diff
@@ -0,0 +1,98 @@
...
+              .putFields("metadata", Value.newBuilder().setStructValue(metadataStruct).build())
+              .putFields(
+                  "ips",
+                  Value.newBuilder()
+                      .setListValue(
+                          ListValue.newBuilder()
+                              .addValues(Value.newBuilder().setStringValue("0.0.0.0").build())
+                              .build())
+                      .build())
+              .build();
```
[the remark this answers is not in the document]
lovenishs04: Addressed

security-command-center/snippets/src/test/java/management/api/EventThreatDetectionCustomModuleTest.java
minherz: It does not help in the case this test is executed in parallel by multiple processes. For example, consider when this test is run in parallel in two environments with different JDKs.

Please using `java.util.UUID`. For the reference you can see how it is used in the [compute tests](https://github.com/GoogleCloudPlatform/java-docs-samples/blob/05b16a2d97aad93cdac6fac4a27d8b85794758f1/compute/cloud-client/src/test/java/compute/InstanceOperationsIT.java#L79).

minherz (approved):

security-command-center/snippets/src/test/java/management/api/EventThreatDetectionCustomModuleTest.java
lovenishs04: Addressed, as suggested, now added the random UUID as a suffix to the display name

PR #9750, opened 2024-11-29
chore: misc non-functional cleanup

igorbernstein2:
NOTE: this depends on and includes https://github.com/GoogleCloudPlatform/java-docs-samples/pull/9749

Just some random nonfunctional cleanup - comments, logging, etc

PR #9751, opened 2024-11-29
feat: add jitter to anti idle task

igorbernstein2:
NOTE: this depends on and includes https://github.com/GoogleCloudPlatform/java-docs-samples/pull/9750


bigtable/bigtable-proxy/src/main/java/com/google/cloud/bigtable/examples/proxy/channelpool/DataChannel.java
```diff
@@ -96,16 +106,30 @@ public DataChannel(
...
+  }
+
+  private void warmTask() {
     try {
       warm();
     } catch (RuntimeException e) {
       LOGGER.warn("anti idle ping failed, forcing reconnect", e);
       inner.enterIdle();
+    } finally {
+      synchronized (scheduleLock) {
```
mutianf, line 126: hmm is this gonna be a problem? why are we calling warm task recursively?

mutianf (approved):

PR #9881, opened 2024-12-16
Merge from Main to ludo-java21

ludoch:


iennae:
@ludoch can you refresh your branch from main and only submit changes for review for the changes you are initiating? This looks like you might have been working from a stale branch.

PR #9987, opened 2025-01-22
docs(memorystore): added valkey caching guide

dackers86:
## Description

Fixes #<ISSUE-NUMBER>

## Checklist

### DIFF
--- a/memorystore/valkey/caching/GUIDE.md
+++ b/memorystore/valkey/caching/GUIDE.md
@@ -1,0 +1,189 @@
+# Building a Caching Service on Google Cloud using Valkey, Spring Boot, and PostgreSQL
+
+Modern applications need to deliver fast, responsive user experiences at scale.
+
+In this guide, there are architectural concepts and deployment steps for creating a high-performance caching service on Google Cloud. Using a combination of Java, Spring Boot, PostgreSQL, and Valkey, you can reduce latency while also reducing the load on your database.
+
+## Why Caching Matters
+
+- **Speed & Latency:** Storing frequently requested data in memory avoids repeated round-trip queries to databases, while reducing response times.
+- **Scalability:** By reducing the workload on your database, applications can serve data directly from memory, increasing the capacity for requests.
+
+## What You’ll Build
+
+You’ll set up a caching service that:
+
+1. **Works with a PostgreSQL database** to store long-lived, persistent records.
+2. **Incorporates Valkey** as a high-speed, in-memory cache, fronting the PostgreSQL database.
+3. **Uses Spring Boot** to expose REST endpoints, providing a simple interface for reading, writing, and invalidating cached data.
+4. **A solution that can be deployed to Google Cloud Platform (GCP)** for production, leveraging services like Cloud Run, Cloud SQL, and Memorystore.
+
+By following this guide, you’ll have a reference architecture ready to adapt, test, and deploy to meet the performance needs of your application.
+
+## Architecture Overview
+
+- **Spring Boot Application:** Serves as the middle tier for responding to API calls. When a request is received, the API checks Valkey for cached results; if no entries are found, then the API will retrieve data from the PostgreSQL database and update the cache.
+- **Valkey (In-Memory Cache):** A Redis-like memory store that keeps hot data ready to be served instantly.
+- **PostgreSQL Database:** Your source of truth for all data. The cache reduces how often the app queries this database.
+- **Google Cloud Infrastructure:** Deployed using Terraform, you can host the application on Cloud Run, store data in Cloud SQL for PostgreSQL, and leverage Memorystore for Valkey.
+
+## Step-by-step Guide
+
+To begin, we are will generate an API with the following routes:
+
+_create_: For creating new items.
+_delete_: For creating deleting items.
+_create_: For creating new items.
+_create_: For creating new items.
+
+### Creating a new application
+
+The first step is to initialize a brand new Spring Boot application. The [offical guide](https://spring.io/guides/gs/spring-boot) demonstrates how to generate a new project using [Spring Initializer](https://start.spring.io/).
+
+1. We chosen `Maven` as the project type for this demonstration..
+2. Select Sprint Boot version 3.4.1
+3. Complete the appropriate metadata.
+4. Choose your preffered Packing for downloading
+5. Select `Java 17` for your Java version.
+6. Finally, generate and extract the files.
+
+### Installing additional dependencies
+
+Next, ensure the following dependencies have been added to your POM.xml file.
+
+#### Jedis
+
+Add the folowing snippet toconnect directly to the Memorystore for Valkey instance.
+
+```xml
+        <!-- Jedis: Redis Java Client -->
+        <dependency>
+            <groupId>redis.clients</groupId>
+            <artifactId>jedis</artifactId>
+            <version>4.3.0</version> <!-- Use the latest version -->
+        </dependency>
+```
+
+#### Jakarta
+
+To ensure that our api routes are correctly validated. Add the following dependency:
+
+```xml
+   <!-- Add Validation support-->
+   <dependency>
+      <groupId>jakarta.validation</groupId>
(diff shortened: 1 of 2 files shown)


memorystore/valkey/caching/GUIDE.md
```diff
@@ -0,0 +1,189 @@
...
+        <dependency>
+            <groupId>redis.clients</groupId>
+            <artifactId>jedis</artifactId>
+            <version>4.3.0</version> <!-- Use the latest version -->
+        </dependency>
+```
+
+#### Jakarta
+
+To ensure that our api routes are correctly validated. Add the following dependency:
```
fewtrell, line 69: can we elaborate a little bit here? It's not clear to me what this means.

memorystore/valkey/caching/GUIDE.md
```diff
@@ -0,0 +1,189 @@
...
+   Item createdItem = new Item(
+   itemId,
+   item.getName(),
+   item.getDescription(),
+   item.getPrice()
+   );
+
+   /** Cache the data in Memorystore for valkey with the Time-to-live value **/
+   String idString = Long.toString(itemId);
+   jedis.set(idString, createdItem.toJSONObject().toString());
```
fewtrell, line 108: expire can be included in set() as a SetParam https://github.com/redis/jedis/blob/3e537f948d17a5511db3b8f57cd4e093407852b0/src/main/java/redis/clients/jedis/commands/StringBinaryCommands.java#L14

This is a more recommended pattern as it cuts QPS in half versues calling set() followed by expire() every time.

memorystore/valkey/caching/GUIDE.md
```diff
@@ -0,0 +1,189 @@
...
+### Reading from the Cache (Retrieving Values)
+
+This method demonstrates how to efficiently retrieve data using a caching layer (Memorystore) to improve performance. The function first checks the cache for the requested item and handles its time-to-live (TTL) appropriately. If the item is not found in the cache, it fetches the data from the database, caches it, and then returns the result.
+
+```java
+/** Import the Jedis library */
+import redis.clients.jedis.Jedis;
+
+public Item get(long id) {
+   /** Ensure that the item id is a string for retirval from Memorystore */
```
fewtrell, line 125: retrieval (spelling)

memorystore/valkey/caching/GUIDE.md
```diff
@@ -0,0 +1,189 @@
...
+    String idString = Long.toString(id);
+
+    /* Check if the item exists in  the cache */
+    if (jedis.exists(idString)) {
+      // If the data exists in the cache extend the TTL
+      jedis.expire(idString, DEFAULT_TTL);
```
fewtrell, line 131: i'm curious what the purpose of this is. If the purpose is to do some sort of least-recently-used eviction algorithm why not just use redis built in allkeys-lru maxmemory-policy?

The reason I ask is now we're converting every read into 2 reads and a write which is going to triple QPS.

also TTL often indicates some sort of allowed staleness. this technique will allow a frequently read item to stay in cache indefinitely - allowing it to get very stale.

memorystore/valkey/caching/GUIDE.md
fewtrell: recommend use this throughout.

memorystore/valkey/caching/GUIDE.md
```diff
@@ -0,0 +1,189 @@
...
+```java
+/** Import the Jedis library */
+import redis.clients.jedis.Jedis;
+
+public Item get(long id) {
+   /** Ensure that the item id is a string for retirval from Memorystore */
+    String idString = Long.toString(id);
+
+    /* Check if the item exists in  the cache */
+    if (jedis.exists(idString)) {
```
fewtrell, line 129: i wonder if we should have 'try catch/finally' around jedis commands so that if they fail the db.get() still runs? that way a broken cache doesn't break the app?

memorystore/valkey/caching/GUIDE.md
```diff
@@ -0,0 +1,189 @@
...
+- **PostgreSQL Database:** Your source of truth for all data. The cache reduces how often the app queries this database.
+- **Google Cloud Infrastructure:** Deployed using Terraform, you can host the application on Cloud Run, store data in Cloud SQL for PostgreSQL, and leverage Memorystore for Valkey.
+
+## Step-by-step Guide
+
+To begin, we are will generate an API with the following routes:
+
+_create_: For creating new items.
+_delete_: For creating deleting items.
+_create_: For creating new items.
```
fewtrell, line 36: should it be create,read, update, delete?
or create,read, delete?

same question for following files.

memorystore/valkey/caching/GUIDE2.md
```diff
@@ -0,0 +1,188 @@
+# Building a Caching Service on Google Cloud using Valkey, Spring Boot, and PostgreSQL
```
fewtrell, line 1: this file is very similar to GUIDE.md - what's the difference supposed to be here?

dackers86: This was an alternative introduction to ensure that the article covered the correct tone and content.

I can remove this if the first article matches expectations.

memorystore/valkey/caching/GUIDE.md
dackers86: This example has been updated, with new instructions prior explaining the meaning behind the code example.

memorystore/valkey/caching/GUIDE.md
```diff
@@ -56,120 +55,151 @@ Next, ensure the following dependencies have been added to your POM.xml file.
...
-   /** Search for the item in the database */
-   Optional<Item> item = itemsRepository.get(id);
+    // No cached item exists, search for the item in the database
+    Optional < Item > item = itemsRepository.get(id);
 
-    /** Check if a record has been found, If the data doesn't exist in the database, return null */
+    // Check if a record has been found, If the data doesn't exist in the database, return null
     if (item.isEmpty()) {
-      return null;
+        return null;
```
fewtrell, line 164: curious if it's possible to cache a null value.

There's pros & cons to this. in production if you're used to a high cache hit ratio & client keeps retrying on a missing item you'll see a lot of traffic go to the database b/c the cache is not protecting.

sometimes folks will put the null in cache but with a shorter TTL.

memorystore/valkey/caching/GUIDE.md
```diff
@@ -0,0 +1,219 @@
...
+    }
+
+    // Get the database item, and convert it into a string value
+    Item dbItem = item.get();
+    String itemString = dbItem.toJSONObject().toString();
+
+    // An item has been found in the database. Cache it with the ID, value, and TTL.
+    try {
+        // Update the cache with the id, TTL value and string object
+        jedis.setex(idString, DEFAULT_TTL, itemString);
```
fewtrell, line 174: shouldn't we be checking return code here? (for every jedis call in fact)?

CorieW: The API states that `setex` returns "OK". 

Notably, when the connection drops and a request is made using Jedis, an exception is thrown then, but the current try/catching that exists allows for fallback onto the database. If I were to get rid of the try/catching, unhandled exceptions would cause status code 500 whenever a problem with Jedis occurs (such as when there is no connection with the Valkey server).

fewtrell: ok i think this case makes sense. Let's make sure we are checking error codes wherever appropriate.

memorystore/valkey/caching/GUIDE.md
dackers86: I don't believe it possible to cache null values.

An example error message would be 

```bash
Setex with null value failed as expected: Cannot read the array length because "raw" is null
```

A potential alternative would be set an empty string or a placeholder value `<Empty>`. An update on any front-end applications would be required to know how to handle the returned value.

fewtrell: sgtm thanks for looking.

fewtrell (approved):

PR #10002, opened 2025-01-31
docs(memorystore): added valkey session snippets

dackers86:
## Description

Fixes #<ISSUE-NUMBER>

## Checklist

### DIFF
--- a/memorystore/valkey/session/snippets/src/main/java/samples/MemorystoreClearBasket.java
+++ b/memorystore/valkey/session/snippets/src/main/java/samples/MemorystoreClearBasket.java
@@ -1,0 +1,60 @@
+/*
+ * Copyright 2025 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR IMPLIED WARRANTIES OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import redis.clients.jedis.Jedis;
+import redis.clients.jedis.JedisPool;
+
+public final class MemorystoreClearBasket {
+
+    /** Replace the Memorystore instance ID. */
+    private static final String INSTANCE_ID = "INSTANCE_ID";
+
+    /** Replace the Memorystore port, if not the default port. */
+    private static final int PORT = 6379;
+
+    /** User ID for managing the basket */
+    private static final String USER_ID = "USER_ID";
+
+    private MemorystoreClearBasket() {
+        // No-op; won't be called
+    }
+
+    /**
+     * Clears a user's basket in Memorystore.
+     *
+     * @param args command-line arguments
+     */
+    public static void main(final String[] args) {
+        // Connect to the Memorystore instance
+        JedisPool pool = new JedisPool(INSTANCE_ID, PORT);
+
+        try (Jedis jedis = pool.getResource()) {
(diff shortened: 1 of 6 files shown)


memorystore/valkey/session/snippets/src/main/java/samples/MemorystoreClearBasket.java
```diff
@@ -0,0 +1,60 @@
...
+    }
+
+    /**
+     * Clears a user's basket in Memorystore.
+     *
+     * @param args command-line arguments
+     */
+    public static void main(final String[] args) {
+        // Connect to the Memorystore instance
+        JedisPool pool = new JedisPool(INSTANCE_ID, PORT);
```
fewtrell, line 42: in above try block you both pool & jedis in the 'try' clause. now just 'jedis' shouldn't they be consistent?

memorystore/valkey/session/snippets/src/main/java/samples/MemorystoreLogoutUser.java
```diff
@@ -0,0 +1,58 @@
...
+     *
+     * @param args command-line arguments
+     */
+    public static void main(final String[] args) {
+        // Connect to the Memorystore instance
+        try (JedisPool pool = new JedisPool(INSTANCE_ID, PORT);
+                Jedis jedis = pool.getResource()) {
+
+            // Check if the session exists
+            if (!jedis.exists(TOKEN)) {
```
fewtrell, line 46: can you drop this by checking return code of del? If del=>0 then the key doesn't exist. now we only doe 1 call instead of 2 to redis. at scale this could matter.

memorystore/valkey/session/snippets/src/main/java/samples/MemorystoreRemoveItemFromBasket.java
```diff
@@ -0,0 +1,71 @@
...
+            // Remove the item from the user's basket
+            long newQty = jedis.hincrBy(basketKey, ITEM_ID, REMOVE_QUANTITY);
+
+            // Print the item removed
+            System.out.printf("Removed %d items from basket: %s%n", REMOVE_QUANTITY, ITEM_ID);
+
+            // Remove the item if the quanitity is less than or equal to 0
+            if (newQty <= 0) {
+                // Remove the item from the basket
+                jedis.hdel(basketKey, ITEM_ID);
```
fewtrell, line 62: in general we should be checking all return codes from jedis.

minherz (changes requested):
Your code samples do not follow the [guidelines](https://github.com/GoogleCloudPlatform/java-docs-samples/blob/main/SAMPLE_FORMAT.md).
I am switching this PR to draft.
Please, adjust your code samples and resubmit the PR.

On the side note, what is a purpose of these code samples?
They show how to use the Redis client library to access Redis instance which is Memorystore under the hood. How are they different from code samples in the official Redis documentation?

memorystore/valkey/session/snippets/src/main/java/samples/MemorystoreRemoveItemFromBasket.java
CorieW: `hdel` will never return 0 because `hincrBy` executes first. Even if `basketKey` or `ITEM_ID` do not exist, `hincrBy` seems to automatically create an item in the basket (will create the basket too if it doesn't exist) with the value of `REMOVE_QUANTITY`. This guarentees that an item with the provided ID `ITEM_ID` will always exist in `basketKey` before executing `hdel`.

Perhaps a check before `hincrBy` could be added to see if an the item exists, but this would increase the calls to Valkey by 1?

fewtrell (approved):

minherz (changes requested):
please add tests for each of your code samples.
review guidelines about testing framework and methods.
ensure that execution time for the tests is bound and short.
provision necessary objects and then clean up after the test is complete.
any sequential testing should be justified.
currently we implement two methods e2e tests that run vs testing environment and mocked tests. review other code samples for the reference.

telpirion:
See comments on #9986

PR #10022, opened 2025-02-11
feat(cloud sql): add lazy refresh property for postgres, sql server and mysql

kgala2:
Fixes #388034268

Enabled lazy refresh for the Java Cloud SQL connector (Postgres, MySQL, and SQL Server).

### DIFF
--- a/cloud-sql/mysql/servlet/src/main/java/com/example/cloudsql/ConnectorConnectionPoolFactory.java
+++ b/cloud-sql/mysql/servlet/src/main/java/com/example/cloudsql/ConnectorConnectionPoolFactory.java
@@ -73,9 +73,9 @@ public class ConnectorConnectionPoolFactory extends ConnectionPoolFactory {
     config.addDataSourceProperty("ipTypes", "PUBLIC,PRIVATE");
     // [START cloud_sql_mysql_servlet_connect_unix]
 
-    // setting the refresh strategy to LAZY
-    // to refresh the tokens when they are needed, rather than on a regular interval
-    // this is recommended for serverless environments to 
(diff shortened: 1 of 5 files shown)


iennae:
make sure that this update matches the other language samples guidance as per @jackwotherspoon

hessjcg (approved):

cloud-sql/mysql/servlet/src/main/java/com/example/cloudsql/ConnectorConnectionPoolFactory.java
```diff
@@ -73,6 +73,12 @@ public static DataSource createConnectionPool() {
     config.addDataSourceProperty("ipTypes", "PUBLIC,PRIVATE");
     // [START cloud_sql_mysql_servlet_connect_unix]
 
+    // Setting cloudSqlRefreshStrategy to lazy
+    // to refresh the tokens when they are needed, rather than on a scheduled interval.
```
jackwotherspoon, line 77: "refresh the tokens" is vague language and can cause confusion. Let's avoid this kind of language in our docs.

A "refresh" refreshes not only an OAuth2 token but also the ephemeral certificate.

```java
// cloudSqlRefreshStrategy set to "lazy" is used to perform a
// refresh when needed, rather than on a scheduled interval.
// This is recommended for serverless environments to
// avoid background refreshes from throttling CPU.
```

jackwotherspoon (changes requested):
Let's align the comments more closely with what we used in Go

jackwotherspoon (approved):
LGTM 👍 ✅

PR #10065, opened 2025-04-10
feat(modelarmor): Added code samples for screening user prompt, model response and a PDF file

tirthrajsinh-zala-crest:
## Description

Created samples for user prompt and model response sanitation using model armor APIs.

## Checklist

- [x] These samples need a new **API enabled** in testing projects to pass (let us know which ones)
- [x] These samples need a new/updated **env vars** in testing projects set to pass (let us know which ones)


harshnasitcrest:
> This PR has the same global issues as #10068 .
> 
> See internal discussion about content.

@telpirion Thanks for your review! All global issues have been resolved in this PR along with sensitive content replaced based on https://github.com/GoogleCloudPlatform/python-docs-samples/pull/13305

I will include additional tests, along with a sample of the Advanced SDP template, in a separate PR. This will prevent the increased diff from making it difficult for review.

modelarmor/src/test/java/modelarmor/SnippetsIT.java
```diff
@@ -0,0 +1,650 @@
...
+    System.setOut(null);
+  }
+
+  private static String randomId() {
+    Random random = new Random();
+    return "java-ma-" + random.nextLong();
+  }
+
+  // Create Model Armor templates required for tests.
+  private static Template createMaliciousUriTemplate() throws IOException {
```
utsav1810, line 170: Add a similar method to createRaiTemplate() with all subtypes.

modelarmor/src/test/java/modelarmor/SnippetsIT.java
```diff
@@ -0,0 +1,650 @@
...
+  }
+
+  @Test
+  public void testSanitizeModelResponseWithRaiTemplate() throws IOException {
+    String modelResponse = "To make cheesecake without oven, you'll need to follow these steps...";
+
+    SanitizeModelResponseResponse response = SanitizeModelResponse.sanitizeModelResponse(PROJECT_ID,
+        LOCATION_ID, TEST_RAI_TEMPLATE_ID, modelResponse);
+
+    assertEquals(FilterMatchState.NO_MATCH_FOUND,
```
utsav1810, line 514: We also want to explicitly test here that all RaiFilterTypes have FilterMatchState.NO_MATCH_FOUND. Add assert statements for all 4 subtypes.

utsav1810 (approved):

modelarmor/src/test/java/modelarmor/SnippetsIT.java
harshnasitcrest: I've utilized CreateTemplate.java that creates a template with all the RAI filters.

utsav1810: Understood.

modelarmor/src/test/java/modelarmor/SnippetsIT.java
harshnasitcrest: Isn't it implied that if filter match state is NO_MATCH_FOUND, filter match state for all the sub-filters will also be NO_MATCH_FOUND? Ref: https://cloud.google.com/security-command-center/docs/reference/model-armor/rest/v1/SanitizationResult

utsav1810: The implication is valid. The test case is to make sure that the value in the field is also correct. If due to any reason, the response structure (for RAI result) changes in the proto but the new client library is not built along with those changes, these tests will fail (as they should) and we can rectify it before the changes are public.

harshnasitcrest:
@telpirion Merge conflicts are resolved on this branch. Tests and lints are passing locally. Could you proceed with merging this PR?
Screenshot 2025-05-13 at 9 24 04 PM

telpirion (approved):

PR #10101, opened 2025-06-05
chore: add back region tags for sample browser

iennae:
## Description

### DIFF
--- a/retail/interactive-tutorials/src/main/java/product/AddFulfillmentPlaces.java
+++ b/retail/interactive-tutorials/src/main/java/product/AddFulfillmentPlaces.java
@@ -15,6 +15,7 @@
  */
 
 // [START retail_add_fulfillment_places]
+
 package product;
 
 import static setup.SetupCleanup.createProduct;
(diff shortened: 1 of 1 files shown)


retail/interactive-tutorials/src/main/java/product/AddFulfillmentPlaces.java
```diff
@@ -14,6 +14,8 @@
  * limitations under the License.
  */
 
+// [START retail_add_fulfillment_places]
+
```
glasnt, line 18: nit: excess space after region tag

iennae: that's a required empty line according to lint (I don't see an extra space at the end of the line?)

iennae: we can always do a followup edit to remove as I'm not sure what the extra space is, but really need this to exist so I can show folks how to get sample browser working. Thanks for the review!

PR #10106, opened 2025-06-13
chore(modelarmor): Added floor settings tests

mihirvala-crestdata:
## Description

Added Floor settings tests in model armor as issue b/424365799 is fixed.

## Checklist

### DIFF
--- a/modelarmor/src/test/java/modelarmor/SnippetsIT.java
+++ b/modelarmor/src/test/java/modelarmor/SnippetsIT.java
@@ -28,6 +28,8 @@ import com.google.cloud.modelarmor.v1.DetectionConfidenceLevel;
 import com.google.cloud.modelarmor.v1.FilterConfig;
 import com.google.cloud.modelarmor.v1.FilterMatchState;
 import com.google.cloud.modelarmor.v1.FilterResult;
+import com.google.cloud.modelarmor.v1.FloorSetting;
+import com.google.cloud.modelarmor.v1.FloorSettingName;
 import com.google.cloud.modelarmor.v1.LocationName;
 import com.google.cloud.modelarmor.v1.MaliciousUriFilterSettings;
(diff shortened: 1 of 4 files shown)


modelarmor/src/test/java/modelarmor/SnippetsIT.java
```diff
@@ -380,6 +404,67 @@ private static void deleteTemplate(String templateId) throws IOException {
...
+
+        client.updateFloorSetting(request);
+      }
+    }
+  }
+
+  // Tests for Folder setting snippets.
+  @Test
+  public void testGetOrganizationFloorSetting() throws IOException {
+    GetOrganizationFloorSetting.getOrganizationFloorSetting(ORGANIZATION_ID);
```
bhardwaj-prerna, line 433: Can we have assertions for the actual value that was fetched by the `getOrganizationFloorSetting` method? Similarly for other tests.

harshnasitcrest:
**NOTE:** Please set following environment variables required for running tests:
MA_FOLDER_ID: `695279264361`
MA_ORG_ID: `951890214235`

modelarmor/src/test/java/modelarmor/SnippetsIT.java
mihirvala08: We followed a similar pattern to other snippet tests (see [Secret Manager tests](https://github.com/GoogleCloudPlatform/java-docs-samples/blob/main/secretmanager/src/test/java/secretmanager/SnippetsIT.java#L551)). The goal of these tests is to verify that the snippets run successfully by asserting the presence of a success log.

glasnt:
(summary of internal discussions) Only the get floor test is failing, not the update. This shouldn't be a permission issue, in this case. The error message is also incorrect. This may be an issue with the sample implementation, rather than the test.

harshnasitcrest:
@glasnt Fixed the issue with help from Model Armor team! All tests ran successful.

glasnt (approved):
🥳

PR #10160, opened 2025-09-02
feat(genai): add thinking samples

jdomingr:
## Description

Add new thinking samples:

- Thinking with text
- Thinking budget with text
- Thinking including thoughts

## Checklist

### DIFF
--- a/genai/snippets/src/main/java/genai/thinking/ThinkingIncludeThoughtsWithText.java
+++ b/genai/snippets/src/main/java/genai/thinking/ThinkingIncludeThoughtsWithText.java
@@ -1,0 +1,106 @@
+/*
+ * Copyright 2025 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package genai.thinking;
+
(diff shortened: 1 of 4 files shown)


genai/snippets/src/main/java/genai/thinking/ThinkingWithText.java
```diff
@@ -0,0 +1,99 @@
...
+    // once, and can be reused for multiple requests.
+    try (Client client =
+        Client.builder()
+            .location("global")
+            .vertexAI(true)
+            .httpOptions(HttpOptions.builder().apiVersion("v1").build())
+            .build()) {
+
+      GenerateContentResponse response =
+          client.models.generateContent(modelId, "solve x^2 + 4x + 4 = 0", null);
```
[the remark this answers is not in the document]
jdomingr: That's not true. So far, in the java SDK all overloads for generateContent take the config, which I'm sending as null.

genai/snippets/src/main/java/genai/thinking/ThinkingIncludeThoughtsWithText.java
```diff
@@ -0,0 +1,107 @@
...
+import com.google.genai.types.GenerateContentConfig;
+import com.google.genai.types.GenerateContentResponse;
+import com.google.genai.types.HttpOptions;
+import com.google.genai.types.ThinkingConfig;
+
+public class ThinkingIncludeThoughtsWithText {
+
+  public static void main(String[] args) {
+    // TODO(developer): Replace these variables before running the sample.
+    String modelId = "gemini-2.5-flash";
```
msampathkumar: Please re-check model name!

jdomingr: @msampathkumar I've just fixed it in both samples that use the pro instead of the flash model. Thanks!

jdomingr:
@ddobrin Thanks for suggestions! I applied them. The only exception was changing the naming used for region tags, since I can't change that. However, I changed the filenames to use the same naming used for the region tags.

ddobrin (approved):
Suggested changes addressed. Thank you

PR #10173, opened 2025-09-25
feat(genai): add image generation samples (2)

jdomingr:
## Description
Add new image generation samples using imagen-3.0-capability-001 model:

- Image generation with subject reference
- Image generation with style reference
- Image generation with raw reference
- Image generation with scribble control reference
- Image generation with canny control reference

## Checklist

### DIFF
--- a/genai/snippets/src/main/java/genai/imagegeneration/ImageGenCannyCtrlTypeWithTextAndImage.java
+++ b/genai/snippets/src/main/java/genai/imagegeneration/ImageGenCannyCtrlTypeWithTextAndImage.java
@@ -37,18 +37,7 @@ public class ImageGenCannyCtrlTypeWithTextAndImage {
     cannyEdgeCustomization(modelId, outputGcsUri);
   }
 
-  /**
-   * Generates an image with a Canny edge image and text prompt.
-   *
-   * <p>This function demonstrates controlled customization. It uses a "Canny edge" image, which is
-   * a black-and-white line drawing that outlines the shapes of objects. The model is instructed to
-   * follow this structural outline to generate an image based the text prompt.
-   *
-   * @param modelId The GenAI model to use for generating the image.
-   * @param outputGcsUri A GCS URI where the generated image will be saved. Example:
-   *     "gs://your-bucket/your-prefix"
-   * @return An Optional containing the GCS URI of the generated image if successful.
-   */
+  // Generates an image using controlled customization with a Canny Edge image and a text prompt.
   public static Optional<String> cannyEdgeCustomization(String modelId, String outputGcsUri) {
     // Client Initialization. Once created, it can be reused for multiple requests.
     try (Client client = Client.builder().location("global").vertexAI(true).build()) {
--- a/genai/snippets/src/main/java/genai/imagegeneration/ImageGenSubjectReferenceWithTextAndImage.java
+++ b/genai/snippets/src/main/java/genai/imagegeneration/ImageGenSubjectReferenceWithTextAndImage.java
@@ -39,19 +39,8 @@ public class ImageGenSubjectReferenceWithTextAndImage {
     subjectCustomization(modelId, outputGcsUri);
   }
 
-  /**
-   * Demonstrates subject customization.
-   *
-   * <p>This function customizes a subject image using a subject reference image of type person,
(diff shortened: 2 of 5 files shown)


genai/snippets/src/main/java/genai/imagegeneration/ImageGenSubjectReferenceWithTextAndImage.java
```diff
@@ -0,0 +1,106 @@
...
+                  ControlReferenceConfig.builder().controlType("CONTROL_TYPE_FACE_MESH").build())
+              .build();
+
+      EditImageResponse imageResponse =
+          client.models.editImage(
+              modelId,
+              "a portrait of a woman[1] in the pose of the control image[2]in a watercolor style by"
+                  + " a professional artist, light and low-contrast stokes, bright pastel colors,"
+                  + " a warm atmosphere, clean background, grainy paper, bold visible brushstrokes,"
+                  + " patchy details",
```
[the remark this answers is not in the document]
jdomingr: I cannot simplify this into a single line because it would be greater than the max line length

genai/snippets/src/main/java/genai/imagegeneration/ImageGenCannyCtrlTypeWithTextAndImage.java
```diff
@@ -0,0 +1,84 @@
...
+public class ImageGenCannyCtrlTypeWithTextAndImage {
+
+  public static void main(String[] args) {
+    // TODO(developer): Replace these variables before running the sample.
+    String modelId = "imagen-3.0-capability-001";
+    String outputGcsUri = "gs://your-bucket/your-prefix";
+    generateImageWithCannyCtrl(modelId, outputGcsUri);
+  }
+
+  // Generates an image with a canny control reference image and text input
```
iennae: would recommend a more extensive header explaining the samples goals suggestion: 
```java
/**
   * Generates an image using a Canny edge image to control the output structure.
   *
   * <p>This function demonstrates controlled editing. It uses a "Canny edge" image, which is a
   * black-and-white line drawing that outlines the shapes of objects. The model is instructed to
   * follow this structural outline while creatively filling in the details based on the text
   * prompt.
   *
   * @param projectId The Google Cloud project ID.
   * @param location The Google Cloud location to run the request.
   * @param outputGcsUri A GCS URI where the generated image will be saved.
   *     Example: "gs://your-bucket/your-prefix"
   * @return An Optional containing the GCS URI of the generated image if successful.
   */
   ```

File: genai/snippets/src/main/java/genai/imagegeneration/ImageGenCannyCtrlTypeWithTextAndImage.java [hunk 3]
```diff
@@ -0,0 +1,84 @@
...
+      ControlReferenceImage controlReferenceImage =
+          ControlReferenceImage.builder()
+              .referenceId(1)
+              .referenceImage(
+                  Image.builder()
+                      .gcsUri("gs://cloud-samples-data/generative-ai/image/car_canny.png")
+                      .build())
+              .config(ControlReferenceConfig.builder().controlType("CONTROL_TYPE_CANNY").build())
+              .build();
+
```
iennae, line 55: missing some context that might be helpful 

```java
      // The `[1]` in the prompt text refers to the `referenceId` set above,
      // instructing the model to apply the Canny edge control to the "red car".
 ```

jdomingr: Thanks for your suggestion!

genai/snippets/src/main/java/genai/imagegeneration/ImageGenCannyCtrlTypeWithTextAndImage.java
jdomingr: @iennae Thanks for the suggestion. I added a more extensive header to each sample. Please, let me know if I should improve something else about the header.

jdomingr: I had to change the comments again, since according to the java sample format we should include a short comment and avoid using the javadoc format.

PR #10197, opened 2025-11-03
feat: update functions samples

maigovannon:
## Description
Creating Java25 doc samples for Cloud Run Functions and updating some of the existing ones to Java21. These are forked changes from https://github.com/GoogleCloudPlatform/java-docs-samples/pull/10193

## Checklist

### DIFF
--- a/functions/v2/pubsub/pom.xml
+++ b/functions/v2/pubsub/pom.xml
@@ -101,6 +101,25 @@
         </configuration>
       </plugin>
       <plugin>
+        <groupId>org.jacoco</groupId>
+        <artifactId>jacoco-maven-plugin</artifactId>
+        <version>0.8.11</version>
+        <executions>
+          <execution>
+            <goals>
+              <goal>prepare-agent</goal>
+            </goals>
+          </execution>
+          <execution>
+            <id>report</id>
+            <phase>prepare-package</phase>
+            <goals>
+              <goal>report</goal>
+            </goals>
+          </execution>
+        </executions>
+      </plugin>
+      <plugin>
         <groupId>org.apache.maven.plugins</groupId>
         <artifactId>maven-surefire-plugin</artifactId>
         <!-- version 3.0.0-M4 does not load JUnit5 correctly -->


functions/v2/pubsub/pom.xml
```diff
@@ -100,6 +100,25 @@
           <functionTarget>functions.SubscribeToTopic</functionTarget>
         </configuration>
       </plugin>
+      <plugin>
```
iennae, line 103: I get why you would add this but I'm not sure we should be adding this to the sample code. We want to illustrate using the Google Cloud Run function with Pub/Sub with a minimal example. This feels like it's adding complexity.

functions/helloworld/helloworld/pom.xml
```diff
@@ -50,8 +50,8 @@
   <!-- [START functions_example_pom] -->
   <!-- [START functions_example_pom_dependencies] -->
   <properties>
-    <maven.compiler.target>11</maven.compiler.target>
-    <maven.compiler.source>11</maven.compiler.source>
+    <maven.compiler.target>21</maven.compiler.target>
```
iennae, line 53: We generally do not have samples for every run time in Functions. Java 11 is the minimum required runtime https://docs.cloud.google.com/run/docs/runtimes/java and that's why this is 11 currently. If y'all are thinking to move to latest supported (non-preview) that's fine, but it would be helpful to know that's the intent here. We should not have duplicate samples showing the basic usecase.

functions/v2/pubsub/pom.xml
maigovannon: This was actually needed to make `mvn clean verify` work properly. Without this, it fails.

functions/helloworld/helloworld/pom.xml
maigovannon: Java11 is actually no longer supoprted and was decommissioned on 10/31: https://docs.cloud.google.com/run/docs/runtime-support#java

ludoch (approved):

PR #10251, opened 2026-04-20
feat(spanner): Update OpenTelemetry OTLP trace sample to use Google C…

surbhigarg92:
…loud OTLP endpoint

### DIFF
--- a/spanner/opentelemetry_traces/src/main/java/com/example/spanner/OpenTelemetryUsage.java
+++ b/spanner/opentelemetry_traces/src/main/java/com/example/spanner/OpenTelemetryUsage.java
@@ -27,6 +27,7 @@ import com.google.cloud.spanner.SpannerOptions;
 import com.google.cloud.spanner.Statement;
 import io.opentelemetry.api.OpenTelemetry;
 import io.opentelemetry.exporter.otlp.trace.OtlpGrpcSpanExporter;
+import io.opentelemetry.exporter.otlp.trace.OtlpGrpcSpanExporterBuilder;
 import io.opentelemetry.sdk.OpenTelemetrySdk;
 import io.opentelemetry.sdk.resources.Resource;
 import io.opentelemetry.sdk.trace.SdkTracerProvider;
@@ -34,8 +35,8 @@ import io.opentelemetry.sdk.trace.export.BatchSpanProcessor;
 import io.opentelemetry.sdk.trace.export.SpanExporter;
 import io.opentelemetry.sdk.trace.samplers.Sampler;
 import java.io.IOException;
+import java.net.URI;
 import java.util.Collections;
-import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 
@@ -50,9 +51,21 @@ public class OpenTelemetryUsage {
   static String instanceId = "my-instance";
   static String databaseId = "my-database";
 
-  static boolean useCloudTraceExporter = false; // Replace to true for Cloud Trace exporter
-  static String otlpEndpoint =
-      "https://telemetry.googleapis.com"; // Replace with your OTLP endpoint
+  // Use the OTLP exporter by default. Set to true to use the Cloud Trace exporter instead.
+  static boolean useCloudTraceExporter = false;
+
+  // The OTLP endpoint to send traces to.
+  // It is recommended to use environment variables (OTEL_EXPORTER_OTLP_ENDPOINT) for configuration.
+  static String otlpEndpoint = System.getenv("OTEL_EXPORTER_OTLP_TRACES_ENDPOINT");
+
+  static {
+    if (otlpEndpoint == null) {
+      otlpEndpoint = System.getenv("OTEL_EXPORTER_OTLP_ENDPOINT");
+    }
+    if (otlpEndpoint == null) {
+      otlpEndpoint = "https://telemetry.googleapis.com"; // Google OTLP gRPC endpoint
+    }
+  }
 
   public static void main(String[] args) throws IOException {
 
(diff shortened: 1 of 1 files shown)


spanner/opentelemetry_traces/src/main/java/com/example/spanner/OpenTelemetryUsage.java
```diff
@@ -46,20 +50,20 @@ public class OpenTelemetryUsage {
   static String instanceId = "my-instance";
   static String databaseId = "my-database";
 
-  // Replace these variables to use OTLP Exporter
-  static boolean useCloudTraceExporter = true; // Replace to false for OTLP
-  static String otlpEndpoint = "http://localhost:4317"; // Replace with your OTLP endpoint
+  static boolean useCloudTraceExporter = false; // Replace to true for Cloud Trace exporter
+  static String otlpEndpoint =
```
michaelsafyan: The customer's OTLP configuration should be respected (see relevant OTel environment variables configuring the exporter). However, auth and project ID should get automatically injected/added if the endpoint is the Telemetry API or one of its regional variants.

spanner/opentelemetry_traces/src/main/java/com/example/spanner/OpenTelemetryUsage.java
```diff
@@ -74,15 +78,37 @@ public static void main(String[] args) {
...
+  public static Spanner getSpannerWithOtlpExporter() throws IOException {
     // [START spanner_opentelemetry_traces_otlp_usage]
-    Resource resource = Resource
-        .getDefault().merge(Resource.builder().put("service.name", "My App").build());
-
+    Resource resource =
+        Resource.getDefault()
+            .merge(
+                Resource.builder()
+                    .put("service.name", "My App")
```
michaelsafyan: Why "My App"? The service name really ought to be supplied by the program, either through an explicit parameter or through environment variables. 

Note that there are standard OTel environment variables that are typically used to supply this.

spanner/opentelemetry_traces/src/main/java/com/example/spanner/OpenTelemetryUsage.java
```diff
@@ -74,15 +78,37 @@ public static void main(String[] args) {
...
+
+    GoogleCredentials credentials =
+        GoogleCredentials.getApplicationDefault()
+            .createScoped(Collections.singleton("https://www.googleapis.com/auth/trace.append"));
     OtlpGrpcSpanExporter otlpGrpcSpanExporter =
-        OtlpGrpcSpanExporter
-            .builder()
-            .setEndpoint(otlpEndpoint) // Replace with your OTLP endpoint
+        OtlpGrpcSpanExporter.builder()
+            .setEndpoint(otlpEndpoint)
```
michaelsafyan: The credentials and header logic should be conditioned on the endpoint matching the GCP Telemetry API. But that endpoint should not be hardcoded. It should be easy for customers to use the library with a collector of their choice.

spanner/opentelemetry_traces/src/main/java/com/example/spanner/OpenTelemetryUsage.java
surbhigarg92: Sure changed it to fetch from OTEL_EXPORTER_OTLP_ENDPOINT/OTEL_EXPORTER_OTLP_TRACES_ENDPOINT 

Have used google telemetry endpoint as default.

spanner/opentelemetry_traces/src/main/java/com/example/spanner/OpenTelemetryUsage.java
surbhigarg92: Removed

spanner/opentelemetry_traces/src/main/java/com/example/spanner/OpenTelemetryUsage.java
surbhigarg92: Added this as a conditional logic for "telemetry.googleapis.com"

spanner/opentelemetry_traces/src/main/java/com/example/spanner/OpenTelemetryUsage.java
```diff
@@ -74,84 +91,80 @@ public static void main(String[] args) {
...
     // Enable OpenTelemetry traces before Injecting OpenTelemetry
     SpannerOptions.enableOpenTelemetryTraces();
 
     // Inject OpenTelemetry object via Spanner options or register it as global object.
     // To register as the global OpenTelemetry object,
     // use "OpenTelemetrySdk.builder()....buildAndRegisterGlobal()".
-    SpannerOptions options = SpannerOptions.newBuilder()
-        .setOpenTelemetry(openTelemetry)
-        .build();
+    SpannerOptions options = SpannerOptions.newBuilder().setOpenTelemetry(openTelemetry).build();
```
sakthivelmanii, line 167: Can we remove this change? This looks more cleaner

sakthivelmanii: similarly in other places.

spanner/opentelemetry_traces/src/main/java/com/example/spanner/OpenTelemetryUsage.java
```diff
@@ -50,9 +51,21 @@ public class OpenTelemetryUsage {
...
+
+  // The OTLP endpoint to send traces to.
+  // It is recommended to use environment variables (OTEL_EXPORTER_OTLP_ENDPOINT) for configuration.
+  static String otlpEndpoint = System.getenv("OTEL_EXPORTER_OTLP_TRACES_ENDPOINT");
+
+  static {
+    if (otlpEndpoint == null) {
+      otlpEndpoint = System.getenv("OTEL_EXPORTER_OTLP_ENDPOINT");
+    }
+    if (otlpEndpoint == null) {
```
michaelsafyan, line 65: I think if "OTEL_EXPORTER" or "OTEL_TRACES_EXPORTER" are set to "gcp" or if environment variable "GOOGLE_CLOUD_TRACING_ENABLED" is set to "true", then you can fallback like this.

Otherwise, I think you should fallback to the default behavior of Open Telemetry. If "OTEL_EXPORTER" or "OTEL_TRACES_EXPORTER" is set to "otlp" and neither "OTEL_EXPORTER_OTLP_ENDPOINT" nor "OTEL_OTLP_EXPORTER_TRACES_ENDPOINT" are set, then the default is to use "http://localhost:4317" for gRPC and "http://localhost:4318" for HTTP. See:

https://opentelemetry.io/docs/languages/sdk-configuration/otlp-exporter/

spanner/opentelemetry_traces/src/main/java/com/example/spanner/OpenTelemetryUsage.java
```diff
@@ -81,52 +94,31 @@ public static void main(String[] args) throws IOException {
...
-    BatchSpanProcessor otlpGrpcSpanProcessor =
-        BatchSpanProcessor.builder(otlpGrpcSpanExporter).build();
+        Resource.getDefault().merge(Resource.builder().put("gcp.project_id", projectId).build());
+
+    OtlpGrpcSpanExporterBuilder exporterBuilder =
+        OtlpGrpcSpanExporter.builder().setEndpoint(otlpEndpoint);
+
+    // Add authentication only if the endpoint is the Google Cloud Telemetry API.
+    // The standard endpoint is telemetry.googleapis.com
+    if (otlpEndpoint.contains("telemetry.googleapis.com")) {
```
michaelsafyan, line 104: Note that you also need to match regionalized addresses:

   telemetry.googleapis.com    (global)
   telemetry.us-central1.rep.googleapis.com  (regionalized in cloud region us-central1)
   telemetry.us-east1.rep.googleapis.com       (regionalized in us-east1)
   ...

Prefixed with "https://" or suffixed with ":443" (optionally with "dns:// " as a prefix). Ex:

   telemetry.googleapis.com
    https://telemetry.googleapis.com
    telemetry.googleapis.com:443
    dns://telemetry.googleapis.com:443
    https://telemetry.us-central1.rep.googleapis.com
    telemetry.us-central1.rep.googleapis.com:443
    ...

However, you also want this match to apply only to the domain (it would be a mistake to match "otherdomain.com/proxy/telemetry.googleapis.com").

spanner/opentelemetry_traces/src/main/java/com/example/spanner/OpenTelemetryUsage.java
```diff
@@ -81,52 +94,31 @@ public static void main(String[] args) throws IOException {
...
+    }
 
-    // Create a new tracer provider
     sdkTracerProvider =
         SdkTracerProvider.builder()
-            // Use Otlp exporter or any other exporter of your choice.
-            .addSpanProcessor(otlpGrpcSpanProcessor)
+            .addSpanProcessor(BatchSpanProcessor.builder(exporterBuilder.build()).build())
             .setResource(resource)
             .setSampler(Sampler.traceIdRatioBased(0.1))
```
michaelsafyan, line 119: Use the `OTEL_TRACES_SAMPLER` flag to control sampling behavior?

https://opentelemetry.io/docs/languages/sdk-configuration/general/#otel_traces_sampler

PR #10336, opened 2026-08-24
feat(genai): Multimodal Image Reference Samples Migration to `gemini-2.5-flash-image`

Kef131:
## Description

Fixes #
b/549909475

- Migrate 5 controlled/reference image generation samples from deprecated Imagen `client.models.editImage` to `gemini-2.5-flash-image` using `client.models.generateContent` with multimodal `Part` inputs:
  - `ImageGenCannyCtrlTypeWithTextAndImage.java`
  - `ImageGenRawReferenceWithTextAndImage.java`
  - `ImageGenScribbleCtrlTypeWithTextAndImage.java`
  - `ImageGenStyleReferenceWithTextAndImage.java`
  - `ImageGenSubjectReferenceWithTextAndImage.java`
- Upload generated image bytes directly to the target destination GCS bucket URI using Cloud Storage SDK.
- Enable and un-ignore all tests in `ImageGenerationIT.java`, verifying full test suite passes against live Vertex AI.

### Testing

- [x] **I have tested this change on a live environment and verified it works as intended.**

### Compliance & Style

### Post-Approval Actions


angelcaamal (approved):
LGTM!

genai/snippets/src/main/java/genai/boundingbox/BoundingBoxWithTxtImg.java
```diff
@@ -90,8 +90,11 @@ public static void plotBoundingBoxes(String imageUrl, List<BoundingBox> bounding
             Color.ORANGE);
 
     for (int i = 0; i < boundingBoxes.size(); i++) {
-      BoundingBox boundingBox = boundingBoxes.get((i));
+      BoundingBox boundingBox = boundingBoxes.get(i);
       List<Integer> box2d = boundingBox.getBox2d();
+      if (box2d == null || box2d.size() < 4) {
```
amcolin, line 96: question: What does the `4` means on this context?

Kef131: The 4 represents the four coordinates to define a 2D bounding box -> https://openjfx.io/javadoc/23/javafx.graphics/javafx/geometry/BoundingBox.html#%3Cinit%3E(double,double,double,double) 

Due to repetitive fails in Kokoro, it led me to prevents `IndexOutOfBoundsException`, that's the reason for `box2d.size() < 4` 

If you see this changes out of context, it because [this kokoro fails](https://btx.cloud.google.com/invocations/cce092f9-77d2-40aa-9d94-bdf52d5560ae/targets/cloud-devrel%2Fjava%2Fjava-docs-samples%2Fjava17%2Fpresubmit;config=default/log), and until this commit https://github.com/GoogleCloudPlatform/java-docs-samples/pull/10336/commits/3d7aba9cbaaddd52ef73f418420e79f6027a656d it passed everything.

amcolin: Great! Would it make sense if we stor it in a variable like `IndexThreshold` or a comment that helps understanding the purpose?

Kef131: I don't see [a comment in other languages examples](https://docs.cloud.google.com/gemini-enterprise-agent-platform/models/bounding-box-detection.md#googlegenaisdk_boundingbox_with_txt_img-java_genai_sdk), although I guess the error would not happen in other contexts, also this change is out of the scope of the gemini-2.5 flash migration, I will prefer to do this change in a different ticket/PR but this currently affects this one.
 I will also doubt about the reliability of the whole `BoundingBoxWithTxtImg.java` due `boundingBoxes.get((i));` is an incorrect sintax.
 In any case, an extra comment doesn't affect and is a pretty quick change 👍

amcolin (approved):

PR #10345, opened 2026-09-07
refactor: Update micronaut and postgresql libraries

Kef131:
## Description

Fixes #
b/511911721
b/511714659
b/511714136
b/511093634

### Testing

- [x] **I have tested this change on a live environment and verified it works as intended.**

### Compliance & Style

### Post-Approval Actions


flexible/java-17/micronaut-helloworld/pom.xml
```diff
@@ -33,7 +33,7 @@
     <exec.mainClass>com.example.appengine.Application</exec.mainClass>
     <maven.compiler.target>11</maven.compiler.target>
     <maven.compiler.source>11</maven.compiler.source>
-    <micronaut.version>3.10.3</micronaut.version>
+    <micronaut.version>4.10.22</micronaut.version>
```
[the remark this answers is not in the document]
Kef131: Besides in the next PR doesn't show a issue upgrading https://github.com/GoogleCloudPlatform/java-docs-samples/pull/10264, for this case, I will keep 3.10

angelcaamal (approved):
LGTM!

dataflow/snippets/src/test/java/com/example/dataflow/ApacheIcebergIT.java
```diff
@@ -155,21 +174,38 @@ public void setUp() throws IOException {
...
+        // Ignore bucket cleanup errors in test teardown.
+      }
+      bucketName = null;
     }
   }
 
   @Test
   public void testApacheIcebergRestCatalog() throws IOException, InterruptedException {
+    Assume.assumeTrue(
+        "Skipping test: GOOGLE_CLOUD_PROJECT must be set",
```
XrossFox, line 203: Wouldn't be better to fail this test if the project wasn't set? I'd think you are supposed to set up all env vars before execution, and not doing so, be a valid failure.

Kef131: The problem is that Kokoro tests doesn't take into account where the var is set. Therefore, it is going to continue failing Kokoro and not letting continue with the processes. Indeed this will require a new Kokoro ticket which is out of the scope of this PR.
One alternative is to add a comment that the var should be set by the user.

XrossFox (approved):
LGTM

amcolin (approved):
LGTM

